PR #9, opened 2018-03-14
WIP : added the data and the example packages skeleton

pierreglaser:
the submodule datasets contains a method to fetch data. 
the folder examples contains a simple script for now, that can be enhanced to get more insight at what the encoders are doing


dirty_cat/datasets/datasets.py
```diff
@@ -0,0 +1,24 @@
...
+def fetch_data():
+    """
+    function that fetches example data
+    - either from the internet
+    - or locally
+    because the base example dataset (midwest survey) chosen is very small,
+    it is locally stored in data/midwest_survey.csv
+    in the future, for bigger dataset, this function will get the data
+    from a specific url, and store it (optional) before returning it.
+    :return:
```
GaelVaroquaux: You should abide the numpy doc standard in the docstrings (see http://www.sphinx-doc.org/en/master/ext/napoleon.html#google-vs-numpy ).

dirty_cat/datasets/datasets.py
```diff
@@ -0,0 +1,24 @@
+import pandas as pd
```
GaelVaroquaux: Did you add pandas as a dependencies? In particular on the continuous integration?

dirty_cat/datasets/datasets.py
```diff
@@ -0,0 +1,24 @@
...
+import os
+
+# TODO: speficy pandas in dependencies
+
+FOLDER_PATH= os.path.dirname(os.path.realpath(__file__))
+print(FOLDER_PATH)
+DATA_NAME = 'midwest_survey.csv'
+
+
+def fetch_data():
```
GaelVaroquaux: Maybe a slightly more specific function name: "fetch_midwest_survey"?

dirty_cat/datasets/datasets.py
```diff
@@ -0,0 +1,24 @@
...
+    """
+    function that fetches example data
+    - either from the internet
+    - or locally
+    because the base example dataset (midwest survey) chosen is very small,
+    it is locally stored in data/midwest_survey.csv
+    in the future, for bigger dataset, this function will get the data
```
GaelVaroquaux: I'd rather have one function per dataset in the future (like in scikit-learn).

examples/midwest_survey.py
```diff
@@ -0,0 +1,14 @@
+"""
+base example script. the midwest survey datasets will be included in the
+package because of its small size (500kb)
+"""
+
+from dirty_cat.datasets import fetch_data
+from dirty_cat.similarity_encoder import SimilarityEncoder
+
+if __name__ == '__main__':
```
GaelVaroquaux: For examples, don't bother with checking for main.

dirty_cat/datasets/datasets.py
GaelVaroquaux: Actually, I think that I would rather return the filename and use pandas in the example to load. It's more didactic.

GaelVaroquaux:
This is super useful! Great!

I made a few comments. Once they are addressed, the next step is to add sphinx-gallery to the doc generation (don't forget to install it on circle-ci).

PR #10, opened 2018-03-15
fix setup

jeromedockes:
use scikit-learn way of getting version number in setup


jeromedockes:
@GaelVaroquaux could say a word on why we do it like this, rather than e.g. reading it from a text file (which is what is done for the long description)?

jeromedockes:
> @GaelVaroquaux could say a word on why we do it like this, rather than e.g. reading it from a text file (which is what is done for the long description)?

close in favor of #11

PR #28, opened 2018-03-20
DOC: formatting of the similarity options

jorisvandenbossche:
The first option is not formatted correctly: https://dirty-cat.github.io/stable/generated/dirty_cat.SimilarityEncoder.html#dirty_cat.SimilarityEncoder

This patch is one way to solve it (although it looks a bit ugly in the source code)


GaelVaroquaux:
Thanks Joris! Merging

PR #31, opened 2018-06-06
Add reference to documentation

pcerda:


dirty_cat/similarity_encoder.py
```diff
@@ -62,6 +62,12 @@ class SimilarityEncoder(BaseEstimator, TransformerMixin):
     This encoding is an alternative to OneHotEncoder in the case of
     dirty categorical variables.
 
+    For a detailed description of the method, see
+    `Similarity encoding for learning with dirty categorical variables
+    <https://hal.inria.fr/hal-01806175>`_ by Cerda, Varoquaux, Kégl. 2018
+    (accepted for publication at: Machine Learning journal, Springer).
```
GaelVaroquaux: This should be in the "Notes" section of the docstring.

dirty_cat/target_encoder.py
```diff
@@ -20,6 +20,10 @@ class TargetEncoder(BaseEstimator, TransformerMixin):
     probability of y conditional on this category.
     In addition it takes an empirical Bayes approach to shrink the estimate.
 
+    For more details, see Micci-Barreca, 2001: A preprocessing scheme for
```
GaelVaroquaux: This should be in the "Notes" section of the docstring.

dirty_cat/similarity_encoder.py
GaelVaroquaux: Error: the "References" section

dirty_cat/target_encoder.py
GaelVaroquaux: Error: the "References" section

GaelVaroquaux:
OK, merging because people are starting to look at the package. I'll fix this myself.

PR #47, opened 2018-10-18
Update fetching.py

hamedonline:
Capital letter fix.


pierreglaser:
@hamedonline thanks for those fixes. I think that would be better if you could merge all the PRs that you created into a single one. pinging @pcerda

PR #48, opened 2018-10-24
Modified scikit versions to 0.20

mcuny:


build_tools/circle/build_doc.sh
```diff
@@ -127,6 +127,8 @@ pip install python-Levenshtein
 #cloning and installing from the ColumnTransformer branch
 git clone https://github.com/scikit-learn/scikit-learn.git
 cd scikit-learn
+git fetch --all
```
pierreglaser: We used to install scikit-learn from master because we wanted to have the ColumnTransformer object before 0.20 was released and on PyPi. Now that 0.20 is released, we can go back to the previous `pip install scikit-learn==0.20`

PR #60, opened 2018-11-08
Subclass SimilarityEncoder of OneHotEncoder

dsleo:
Replace _BaseEncoder by OneHotEncoder

### DIFF
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -149,12 +149,6 @@ class SimilarityEncoder(OneHotEncoder):
         if categories == 'auto' and n_prototypes is not None:
             warnings.warn('n_prototypes parameter ignored with category type \'auto\'')
 
-        # To delete unused OneHotEncoder attributes
-        del self.n_values
-        del self._handle_deprecations
-        del self.active_features_
-        del self.feature_indices_
-
     def get_most_frequent(self, prototypes):
         """ Get the most frequent category prototypes
         Parameters
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -2,9 +2,7 @@ import numpy as np
 from scipy import sparse
 
 from sklearn.feature_extraction.text import CountVectorizer, HashingVectorizer
-from sklearn.preprocessing._encoders import _BaseEncoder
-from sklearn.utils import check_array
-
+from sklearn.preprocessing._encoders import OneHotEncoder
 
 from dirty_cat import string_distances
 
@@ -52,7 +50,7 @@ _VECTORIZED_EDIT_DISTANCES = {
 }
 
 
-class SimilarityEncoder(_BaseEncoder):
+class SimilarityEncoder(OneHotEncoder):
     """Encode string categorical features as a numeric array.
 
     The input to this transformer should be an array-like of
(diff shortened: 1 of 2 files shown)


dirty_cat/similarity_encoder.py
```diff
@@ -2,9 +2,7 @@
 from scipy import sparse
 
 from sklearn.feature_extraction.text import CountVectorizer, HashingVectorizer
-from sklearn.preprocessing._encoders import _BaseEncoder
-from sklearn.utils import check_array
-
+from sklearn.preprocessing._encoders import OneHotEncoder
```
GaelVaroquaux: Can you import from sklearn.preprocessing rather than the private module.

dsleo: sure !

GaelVaroquaux:
Sorry, conflicts to resolve.

vinzeebreak:
There has been some problems with the first merge, we will look into it the sooner we can.

dirty_cat/similarity_encoder.py
```diff
@@ -3,7 +3,7 @@
 import numpy as np
 from scipy import sparse
 from sklearn.feature_extraction.text import CountVectorizer, HashingVectorizer
-from sklearn.preprocessing._encoders import _BaseEncoder
+from sklearn.preprocessing import OneHotEncoder
```
mcuny: @dsleo can I fix this the merge conflict and commit it to your forks master? Like that we can merge this and close the PR :)

dsleo: @mcuny weird, I did that change as suggested by Gael yesterday but sure go ahead.

dsleo:
Thinking more about this, and trying not to break LSP, I see two options:  
1. subclass `OneHotEncoder` and raise a `NotImplementedError` on the methods not used in the child class.  
2.  subclass `BaseEnconder` as before and copy the `get_feature_names` from `OneHotEncoder`. After all, it's such small copy/paste.   

@GaelVaroquaux @pcerda: what do you think ?

GaelVaroquaux:
> Thinking more about this, and trying not to break LSP,

I wouldn't worry too much about purity, here.

>  1. subclass OneHotEncoder and raise a NotImplementedError on the methods not used in the child class.

That's not very user friendly.

>  2. subclass BaseEnconder as before and copy the get_feature_names from OneHotEncoder. After all, it's such small copy/paste.

I still prefer subclassing OneHotEncoder, because we inherit
automatically the details of its interface. For instance, if scikit-learn
changes the interface, our code breaks, which is good (because then we
can fix it).

I liked your solution :)

dsleo:
Okay fair enough. Only thing is that the del methods only work for removing parameters not attributes of the class and that's why the checks fail ! 
I'm probably missing a trivial solution but I don't see how to do that easily, unless we create an abstract parent class of OneHotEncoder with only the methods we are interested in here and make SimilarityEncoder subclass of that...

GaelVaroquaux:
Well, can give up on removing those attributes. They are quite harmless, no? 

vinzeebreak:
I did not add the `NotImplementedError` but if necessary, I can add them in a new commit.

GaelVaroquaux:
I resolved the conflict and will merge.

GaelVaroquaux:
Tests passed. Docs look good. Merging.

PR #67, opened 2018-11-15
DOC: better midwest example

GaelVaroquaux:
I was explaining this example to a friend, and found that it lacked printing the values, to explain what this is about.


GaelVaroquaux:
Merging, because it is a bit trivial.

Issue #64, opened 2018-11-12
Benchmarking K-means with different hashing dims and sample sizes


mcuny:
Add option for sample weight parameter in kmeans
Benchmark K-means in similarity encoding with: 

hash dimensions: 64, 128
max samples at: 20k, 50k, 100k and nunique

GaelVaroquaux:
Closing as fixed

PR #68, opened 2018-11-16
#64 Benchmark with plots + scalability example

mcuny:
I added a parameter to the constructor of similarity to enable weighting the categories when a k-means strategy is used. The weights are the frequency of apparition of the categories.

The example is not fully finished, I'm gonna add some plotting. Just need some guidelines as how long the example can be and how much we want to stress the similarity encoder (like should I just take the benchmark use cases and make a really long file describing all the situations?)

The benchmark plots for different situations:
- 100 prototypes and 10 cross validations
- number of rows: 10k, nuniques, 20k, 50k and 100k
- activates weighting on k-mean strategy or not
- vectorizer: 
    - count
    - hash: 2 ** 14, 2 ** 16, 2 ** 18, 2 ** 20
- strategies:
    - most-frequent
    - k-means

We plot the scores by vectoriser and strategy and the time taken by vectoriser and strategy.
@pcerda @GaelVaroquaux

### DIFF
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -64,7 +64,7 @@ def get_prototype_frequencies(prototypes):
 
 
 def get_kmeans_protoypes(X, n_prototypes, hashing_dim=128,
-                         ngram_range=(3, 3), sparse=False):
+                         ngram_range=(3, 3), sparse=False, sample_weight=None):
     """
     Computes prototypes based on:
       - dimensionality reduction (via hashing n-grams)
@@ -79,7 +79,10 @@ def get_kmeans_protoypes(X, n_prototypes, hashing_dim=128,
     if not sparse:
         projected = projected.toarray()
     kmeans = KMeans(n_clusters=n_prototypes)
-    kmeans.fit(projected)
+    if sample_weight is not None:
+        kmeans.fit(projected, sample_weight=sample_weight)
+    else:
+        kmeans.fit(projected)
     centers = kmeans.cluster_centers_
     neighbors = NearestNeighbors()
     neighbors.fit(projected)
@@ -166,7 +169,7 @@ class SimilarityEncoder(_BaseEncoder):
 
     def __init__(self, similarity='ngram', ngram_range=(3, 3),
                  categories='auto', dtype=np.float64,
-                 handle_unknown='ignore', hashing_dim=None, n_prototypes=None):
+                 handle_unknown='ignore', hashing_dim=None, n_prototypes=None, sample_weight=False):
 
         self.categories = categories
         self.dtype = dtype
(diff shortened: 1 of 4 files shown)


dirty_cat/similarity_encoder.py
```diff
@@ -166,7 +169,7 @@ class SimilarityEncoder(_BaseEncoder):
 
     def __init__(self, similarity='ngram', ngram_range=(3, 3),
                  categories='auto', dtype=np.float64,
-                 handle_unknown='ignore', hashing_dim=None, n_prototypes=None):
+                 handle_unknown='ignore', hashing_dim=None, n_prototypes=None, sample_weight=False):
```
GaelVaroquaux: In terms of usability, I am not very enthusiastic at the idea that we are adding an additional parameter to the similarity encoder. We need to find a good default for this parameter, and maybe just hard-code the value.

dirty_cat/test/benchmark_sprint.py
```diff
@@ -0,0 +1,162 @@
+import warnings
```
GaelVaroquaux: This file should be moved to a new directory, called "benchmarks", at the root of the project. Also, we should find a better name for it :). And add a small docstring to the file to explain what it does.

dirty_cat/test/benchmark_sprint.py
```diff
@@ -0,0 +1,162 @@
...
+    df = pd.DataFrame(columns=['vectorizer', 'strategy', 'score'])
+    df['vectorizer'] = vectorizer
+    df['strategy'] = strategy
+    df['score'] = scores
+
+    sns.set(style='ticks', palette='muted')
+    sns.boxplot(x='vectorizer', y='score', hue='strategy', data=df)
+    plt.ylabel("Mean score on 10 cross validations")
+    plt.xlabel('Vectorizer used')
+    plt.legend(bbox_to_anchor=(1.05, 1), loc=2, borderaxespad=0.)
```
GaelVaroquaux: I am not seeing the legend on my plot. Not sure why. It makes interpreting the plot harder.
screenshot

dirty_cat/test/benchmark_sprint.py
```diff
@@ -0,0 +1,162 @@
...
+    plt.title(title)
+    plt.tight_layout()
+    plt.show()
```
GaelVaroquaux: Do not put a "show" in the middle of the script: it can be a blocking call, and hence interrupt the script. Put it at the end.

GaelVaroquaux:
Here is the example, as rendered by CircleCI:
https://304-124964482-gh.circle-artifacts.com/0/doc/auto_examples/04_scalability.html#sphx-glr-auto-examples-04-scalability-py

GaelVaroquaux:
I am +1 for merge, but after removing hashing from the example: it is useless and consumes a lot of time on that example for no good reason.

dirty_cat/similarity_encoder.py
GaelVaroquaux: Let's just hard-code the best option for sample_weight here, unless you see a big issue with it.

PR #70, opened 2018-11-16
OPTIM: better ngram-similarity

GaelVaroquaux:
Improve memory usage and improve speed.


GaelVaroquaux:
Merging, as this fixes failures on master (warnings in new numpy).

Issue #69, opened 2018-11-16
Expose the seed of the kmeans


GaelVaroquaux:
We need to expose the seed of the kmeans in the similarity encoder, to make the pipelines fully reproducible. For this, we need to add a "random_state" option to the similarity encoder, as in many scikit-learn models, and pass it to kmeans.

mcuny:
Is the passed option an integer used as a seed to build a numpy RandomState instance or do we just pass the integer to k-means? (asking because both are possible in the documentation)

GaelVaroquaux:
I think that:

* in fit we use check_random_state to create self.random_state_ from self.random_state
* we pass self.random_state_ to KMeans

⁣Sent from my phone. Please forgive typos and briefness.​

GaelVaroquaux:
Closing as fixed

PR #72, opened 2018-11-17
#69 Exposing random state in similarity encoder

mcuny:


dirty_cat/similarity_encoder.py
```diff
@@ -64,7 +64,7 @@ def get_prototype_frequencies(prototypes):
 
 
 def get_kmeans_protoypes(X, n_prototypes, hashing_dim=128,
```
GaelVaroquaux: Unrelated to this PR, but there is a typo in the function name

dirty_cat/similarity_encoder.py
```diff
@@ -176,9 +178,14 @@ def __init__(self, similarity='ngram', ngram_range=(3, 3),
         self.hashing_dim = hashing_dim
         self.n_prototypes = n_prototypes
 
+        if random_state and categories != 'k-means':
+            raise ValueError('Random state can only be used with a k-means strategy')
```
GaelVaroquaux: Remove this check: it's fine to pass a random state anyhow

GaelVaroquaux:
Could you add a test that fitting twice in the same data gives the same result. 

Aside from this, LGTM

dirty_cat/test/test_reproducibility.py
```diff
@@ -0,0 +1,23 @@
+import numpy as np
```
GaelVaroquaux: The convention is usually that tests go in a file that has the same name as the file containing the functionality that they test, but with "test_" before.

dirty_cat/test/test_reproducibility.py
```diff
@@ -0,0 +1,23 @@
...
+
+def test_reproducibility():
+    res1 = get_result(2454)
+    res2 = get_result(2454)
+    assert (np.array_equal(res1, res2))
+
+    random_state = np.random.RandomState(32456)
+    res1 = get_result(random_state)
+    res2 = get_result(random_state)
+    assert (np.array_equal(res1, res2))
```
GaelVaroquaux: Did these tests fail before? I am asking because the second part does not seem to test what it should: given that a stream is passed, different random streams are drawn in each execution. Hence, if the test passes, it's probably because it is not very stringent.

GaelVaroquaux: Hint: to get a failure without seeding, you probably need a clustering problem that is not well posed, ie with string that all have the same distances in n-gram space. Try something like X = np.array([chr(i) for i in range(32, 127)]).reshape((-1, 1)). And I wouldn't bother separating X and X_test, I would just try sim_enc.fit_transform(X).

GaelVaroquaux: Actually, I just tried that example, and it seems that it creates a very bad behavior in sim_enc, so we definitely need to test for it, and figure out what is wrong:
<pre>
X = np.array([chr(i) for i in range(32, 127)]).reshape((-1, 1))
sim_enc = SimilarityEncoder(categories='k-means', n_prototypes=10)
</pre>

GaelVaroquaux: OK, I understood the problem with the above: the strings are so short that they all have an n-gram similarity of zero. Hence nothing woks.

Here is a suggested test, that fails without your fix:
<pre>
X = np.array([' %s ' % chr(i) for i in range(32, 127)]).reshape((-1, 1))
prototypes = sim_enc.fit(X).categories_
for i in range(10):
   assert (prototypes == sim_enc.fit(X).categories_[0]).all()
</pre>

mcuny: @GaelVaroquaux I pushed a fix. How ever I'm not really sure of what I did.
The only way I was able to get a positiv result on the case where we pass a RandomState is by doing a deep copy of it at every fit call, otherwise calling check_random_state when the type of random_state is int or None.

GaelVaroquaux: Indeed, passing a random_state does not guarantee reproducibility, hence it should probably be removed from the test.

dirty_cat/similarity_encoder.py
```diff
@@ -228,7 +227,10 @@ def fit(self, X, y=None):
 
         n_samples, n_features = X.shape
         self.categories_ = list()
-        self.random_state_ = check_random_state(self.random_state)
+        if isinstance(self.random_state, np.random.RandomState):
+            self.random_state_ = copy.deepcopy(self.random_state)
```
GaelVaroquaux: You should not do that: the semantics are that passing an initialized RandomState does not protect against side effects. Only passing a seed does.

GaelVaroquaux: PS: the fact that your test is failing when you are passing an initialized RandomState shows that it is a good test :)

dirty_cat/test/test_similarity_encoder.py
```diff
@@ -90,11 +90,32 @@ def test_similarity_encoder():
...
+    prototypes = sim_enc.fit(X).categories_[0]
+    for i in range(10):
+        assert (not np.array_equal(prototypes, sim_enc.fit(X).categories_[0]))
+
+    sim_enc = similarity_encoder.SimilarityEncoder(categories='k-means', n_prototypes=10, random_state=435)
+    X = np.array([' %s ' % chr(i) for i in range(32, 127)]).reshape((-1, 1))
+    prototypes = sim_enc.fit(X).categories_[0]
+    for i in range(10):
+        assert (np.array_equal(prototypes, sim_enc.fit(X).categories_[0]))
+
```
GaelVaroquaux: I would just remove the lines below: the test is not meant to pass in such situation. The lines above are enough.

mcuny: Yup I asked CDS team to explain RandomState, sorry for the miscomprehension. Finishing something real fast and re-pushing this.

build_tools/circle/build_doc.sh
```diff
@@ -123,6 +123,7 @@ conda create -n $CONDA_ENV_NAME --yes --quiet python="${PYTHON_VERSION:-*}" \
 source activate testenv
 pip install sphinx-gallery
 pip install python-Levenshtein
+pip install memory_profiler
```
GaelVaroquaux: This does not belong to this PR

dirty_cat/similarity_encoder.py
```diff
@@ -157,7 +158,9 @@ class SimilarityEncoder(OneHotEncoder):
         HashingVectorizer with a number of features equal to `hashing_dim`.
     n_prototypes: number of prototype we want to use.
         Useful when `most_frequent` or `k-means` is used.
-        Must be a positive non null integer.
+        Must be a positiv non null integer.
```
GaelVaroquaux: Typo

mcuny: @GaelVaroquaux I think I got all the merge errors. If it's good for you we can merge it, then merge this into my other PR and rerun the benchmark and tests.

dirty_cat/similarity_encoder.py
```diff
@@ -237,16 +240,16 @@ def fit(self, X, y=None):
 
         n_samples, n_features = X.shape
         self.categories_ = list()
+        self.random_state_ = check_random_state(self.random_state) if self.random_state is not None else None
```
GaelVaroquaux: You should just use check_random_state here, nothing more.

dirty_cat/test/test_similarity_encoder.py
```diff
@@ -90,11 +90,25 @@ def test_similarity_encoder():
...
+    proto = get_kmeans_prototypes(X_test, 3)
     assert np.array_equal(np.sort(proto), np.sort(X_test))
+
+
+def test_reproducibility():
+    sim_enc = similarity_encoder.SimilarityEncoder(categories='k-means', n_prototypes=10)
+    X = np.array([' %s ' % chr(i) for i in range(32, 127)]).reshape((-1, 1))
+    prototypes = sim_enc.fit(X).categories_[0]
+    for i in range(10):
+        assert (not np.array_equal(prototypes, sim_enc.fit(X).categories_[0]))
```
GaelVaroquaux: Don't do this part of the test: you might find centers that match by chance. Just leave the second part.

GaelVaroquaux:
LGTM. Merging. Thank you!

PR #74, opened 2018-11-19
Benchmark test scalability + weights ink mean

mcuny:


benchmarks/similarity_scores_time_benchmark.py
```diff
@@ -0,0 +1,147 @@
...
+import pandas as pd
+import seaborn as sns
+from sklearn import pipeline, linear_model, model_selection
+from sklearn.compose import ColumnTransformer
+from sklearn.preprocessing import OneHotEncoder
+
+from dirty_cat import SimilarityEncoder
+from dirty_cat.datasets import fetch_traffic_violations
+
+warnings.simplefilter(action='ignore', category=FutureWarning)
```
GaelVaroquaux: Which FutureWarning is showing up?

mcuny: This one: "Default solver will be changed to 'lbfgs' in 0.22. Specify a solver to silence this warning."

examples/04_scalability.py
```diff
@@ -0,0 +1,179 @@
...
+print('\nSimilarity encoding with a most frequent strategy')
+
+sim_enc = SimilarityEncoder(similarity='ngram', categories='most_frequent', n_prototypes=100)
+
+column_trans = ColumnTransformer(
+    # adding the dirty column
+    transformers=transformers + [('sim_enc', sim_enc, ['Description'])],
+    remainder='drop')
+
+from time import time
```
GaelVaroquaux: I think that we don't need this import: it was done above, no?

examples/04_scalability.py
```diff
@@ -0,0 +1,179 @@
...
+
+################################################################################
+# SimilarityEncoder with k-means strategy
+# ---------------------------------------
+#
+# K-means strategy is also a dimensionality reduction technique. But we apply
+# a K-means and nearest neighbors algorithm to find the prototypes. The number
+# of prototypes is set manually.
+
+print('\nSimilarity encoding with a k-means strategy')
```
GaelVaroquaux: We should probably remove this print: on the webpage it's not useful given the section title.

examples/04_scalability.py
```diff
@@ -0,0 +1,179 @@
...
+
+################################################################################
+# SimilarityEncoder with most frequent strategy
+# ---------------------------------------------
+#
+# The most frequent strategy selects the n most frequent values in a dirty
+# categorical variable to reduce the dimensionality of the problem and thus
+# speed things up. We select manually the number of prototypes we want to use.
+
+print('\nSimilarity encoding with a most frequent strategy')
```
GaelVaroquaux: We should probably remove this print: the section title is enough.

examples/04_scalability.py
```diff
@@ -0,0 +1,179 @@
...
+# SimilarityEncoder with default options
+# --------------------------------------
+#
+# Let us build our vectorizer, using a ColumnTransformer to combine
+# one-hot encoding and similarity encoding
+from sklearn.preprocessing import OneHotEncoder
+from sklearn.compose import ColumnTransformer
+from dirty_cat import SimilarityEncoder
+
+print('\nBasic similarity encoding')
```
GaelVaroquaux: We should probably remove this print: the section title is enough.

benchmarks/similarity_scores_time_benchmark.py
GaelVaroquaux: OK, how about we suppress it by using "linear_model.LogisticRegression(solver='lbfgs')"?

mcuny: When I do that it asks to increase the number of iterations (it fails to converge), how much should I put?

GaelVaroquaux:
> +warnings.simplefilter(action='ignore', category=FutureWarning)

> When I do that it asks to increase the number of iterations (it fails to
> converge), how much should I put?

OK, let's keep your solution. Just add a comment above the warning. Also,
move the simplefilter to next to were you import the warnings module.

GaelVaroquaux:
LGTM. Merging. Thank you!

PR #76, opened 2018-11-29
Added example with dtype float32

mcuny:
Shows difference in memory consumption between a float64 and float32

### DIFF
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -267,6 +267,9 @@ class SimilarityEncoder(OneHotEncoder):
 
     def transform(self, X):
         """Transform X using specified encoding scheme.
+            For an n-gram similarity, the output dtype
+            will be the instance dtype, but the input dtype is
+            always float 32.
 
         Parameters
         ----------
@@ -319,7 +322,7 @@ class SimilarityEncoder(OneHotEncoder):
                                                                  ngram_range=(min_n, max_n),
                                                                  hashing_dim=self.hashing_dim,
                                                                  dtype=np.float32)
-                last = len(cats)
+                last += len(cats)
             return out
         else:
             raise ValueError("Unknown similarity: '%s'" % self.similarity)
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -267,9 +267,6 @@ class SimilarityEncoder(OneHotEncoder):
 
     def transform(self, X):
         """Transform X using specified encoding scheme.
-            For an n-gram similarity, the output dtype
-            will be the instance dtype, but the input dtype is
-            always float 32.
 
         Parameters
         ----------
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -59,7 +59,7 @@ def ngram_similarity(X, cats, ngram_range, hashing_dim, dtype=np.float64):
     for x, out_row in zip(X, out):
         out_row[:] = SE_dict[x]
 
-    return np.nan_to_num(out)
+    return np.nan_to_num(out, copy=False)
 
 
 def get_prototype_frequencies(prototypes):
@@ -310,13 +310,16 @@ class SimilarityEncoder(OneHotEncoder):
 
         elif self.similarity == 'ngram':
             min_n, max_n = self.ngram_range
-            out = []
+
+            total_length = sum(len(x) for x in self.categories_)
+            out = np.empty((len(X), total_length), dtype=self.dtype)
+            last = 0
             for j, cats in enumerate(self.categories_):
-                encoder = ngram_similarity(X[:, j], cats,
-                                           ngram_range=(min_n, max_n),
-                                           hashing_dim=self.hashing_dim,
-                                           dtype=self.dtype)
-                out.append(encoder)
-            return np.hstack(out)
+                out[:, last:last + len(cats)] = ngram_similarity(X[:, j], cats,
+                                                                 ngram_range=(min_n, max_n),
+                                                                 hashing_dim=self.hashing_dim,
+                                                                 dtype=np.float32)
+                last = len(cats)
+            return out
         else:
             raise ValueError("Unknown similarity: '%s'" % self.similarity)
--- a/examples/04_scalability.py
+++ b/examples/04_scalability.py
@@ -12,8 +12,6 @@ either chosen as the most frequent categories, or with kmeans clustering.
 # in the solver
 import warnings
 
-import numpy as np
-
 warnings.simplefilter(action='ignore', category=FutureWarning)
 
 ################################################################################
@@ -124,7 +122,7 @@ print('Type of X:', X.dtype)
 # We can run a cross-validation
 from sklearn import linear_model, pipeline, model_selection
 
-log_reg = linear_model.LogisticRegression(solver='newton-cg')
+log_reg = linear_model.LogisticRegression()
 
 model = pipeline.make_pipeline(column_trans, log_reg)
 results = resource_used(model_selection.cross_validate)(model, df, y, )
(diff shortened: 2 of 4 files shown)


examples/04_scalability.py
```diff
@@ -197,3 +200,87 @@ def wrapped_func(*args, **kwargs):
...
+print('Time to vectorize: %s' % (t1 - t0))
+################################################################################
+# We can run a cross-validation
+model = pipeline.make_pipeline(column_trans, log_ref)
+results = resource_used(model_selection.cross_validate)(model, df, y)
+print("Cross-validation score: %s" % results['test_score'])
+
+################################################################################
+# Most frequent strategy to define prototypes used with 32 bits float dtype
+# -------------------------------------------------------------------------
```
GaelVaroquaux: I would not duplicate the rest of the example (ie the most frequent and the kmeans)

examples/04_scalability.py
```diff
@@ -197,3 +200,87 @@ def wrapped_func(*args, **kwargs):
...
+
+column_trans = ColumnTransformer(
+    # adding the dirty column
+    transformers=transformers + [('sim_enc', sim_enc, ['Description'])],
+    remainder='drop')
+
+t0 = time()
+X = column_trans.fit_transform(df)
+t1 = time()
+print('Time to vectorize: %s' % (t1 - t0))
```
GaelVaroquaux: Can you show here that X is of dtype float32, by printing X.dtype (in a new cell)?

examples/04_scalability.py
```diff
@@ -197,3 +200,87 @@ def wrapped_func(*args, **kwargs):
...
+################################################################################
+# We can run a cross-validation
+model = pipeline.make_pipeline(column_trans, log_ref)
+results = resource_used(model_selection.cross_validate)(model, df, y)
+print("Cross-validation score: %s" % results['test_score'])
```
GaelVaroquaux: Here I would add a small paragraph mentioning that not all scikit-learn models work efficiently with float32, and hence it is worth checking that their indeed is an improvement in resources used.

examples/04_scalability.py
```diff
@@ -197,3 +200,55 @@ def wrapped_func(*args, **kwargs):
...
+
+################################################################################
+# Here we see that the memory consumption is reduced compared to the
+# case where 64 bits float are used.
+# Which is not the case with all the sci-kit learn models, where some
```
GaelVaroquaux: It's spelled "scikit-learn" :).

As you can see from the example:
https://385-124964482-gh.circle-artifacts.com/0/doc/auto_examples/04_scalability.html
there is little gain in memory used during the cross-validation. This is because the LBFG solver does not work with float32 and casts in float64.

Can you try changing the whole example to use the "saga" solver. I think that it should be faster and work with 32 bit.

examples/04_scalability.py
```diff
@@ -197,3 +200,55 @@ def wrapped_func(*args, **kwargs):
...
+
+column_trans = ColumnTransformer(
+    # adding the dirty column
+    transformers=transformers + [('sim_enc', sim_enc, ['Description'])],
+    remainder='drop')
+
+t0 = time()
+X = column_trans.fit_transform(df)
+t1 = time()
+print('Type of vectorized data:', X.dtype)
```
GaelVaroquaux: Can you do the print in a different "cell". What I call a cell is a comment that starts with a line of "#": it is displayed specially by sphinx-gallery, which renders the example.

Also, can you do it after printing the time to vectorize.

examples/04_scalability.py
```diff
@@ -147,14 +151,13 @@ def wrapped_func(*args, **kwargs):
 # Check now that prediction is still as good
 model = pipeline.make_pipeline(column_trans, log_ref)
```
GaelVaroquaux: Unrelated to your PR, but I just realized that there is a typo in the whole example: "log_ref" should be changed to "log_reg".

examples/04_scalability.py
mcuny: I did and the score go way down (they are all around 0.7 against 0.78, 0.79 with LBFG). Should I stay with saga?

GaelVaroquaux:
> I did and the score go way down (they are all around 0.7 against 0.78, 0.79 with LBFG). Should I stay with saga?

OK, try 'newton-cg'.

GaelVaroquaux:
Oh well, newton-cg' saves memory, but to the cost of time.

Let's give up, go back to LBFGS.

How about this: can you demo it in the example with most frequent, and explain that the log_reg will internally cast to 64 bit (as LBFGS does not support 32 bit), hence the memory saving here are only those during the similarity encoding. However, with a small number of prototypes, the most expensive operation memory-wise is during encoding, as the subsequent representation is low dimension.

You can then name this section "Saving memory by using float32".

I think that this is a useful message to convey.

GaelVaroquaux:
With the internal use of float32, I'm observing a bit of speed up on CircleCI: https://390-124964482-gh.circle-artifacts.com/0/doc/auto_examples/04_scalability.html
However, indeed, the peak memory isn't really improved. It must be dominated by a different part of the code.

dirty_cat/similarity_encoder.py
```diff
@@ -310,13 +310,16 @@ def transform(self, X):
...
-                                           ngram_range=(min_n, max_n),
-                                           hashing_dim=self.hashing_dim,
-                                           dtype=self.dtype)
-                out.append(encoder)
-            return np.hstack(out)
+                out[:, last:last + len(cats)] = ngram_similarity(X[:, j], cats,
+                                                                 ngram_range=(min_n, max_n),
+                                                                 hashing_dim=self.hashing_dim,
+                                                                 dtype=np.float32)
+                last = len(cats)
```
GaelVaroquaux: I think that here you have a bug, no? Shouldn't it be "last += len(cats)"?

examples/04_scalability.py
```diff
@@ -44,7 +47,7 @@ def wrapped_func(*args, **kwargs):
 # --------------------------------
 #
 # We first download the dataset:
-from dirty_cat.datasets import fetch_traffic_violations
+from dirty_cat.datasets.fetching import fetch_traffic_violations
```
GaelVaroquaux: We prefer to avoid importing from subpackages; here I think that it's better to directly import from datasets.

examples/04_scalability.py
```diff
@@ -106,19 +109,24 @@ def wrapped_func(*args, **kwargs):
...
-
 t0 = time()
 X = column_trans.fit_transform(df)
 t1 = time()
 print('Time to vectorize: %s' % (t1 - t0))
+
+################################################################################
+# Check the dtype of the vectorized data
+print('Type of X:', X.dtype)
+
```
GaelVaroquaux: I don't think that this is necessary for the example.

examples/04_scalability.py
```diff
@@ -106,19 +109,24 @@ def wrapped_func(*args, **kwargs):
...
-log_ref = linear_model.LogisticRegression()
 
-model = pipeline.make_pipeline(column_trans, log_ref)
-results = resource_used(model_selection.cross_validate)(model, df, y)
-print("Cross-validation score: %s" % results['test_score'])
+log_reg = linear_model.LogisticRegression()
+
+model = pipeline.make_pipeline(column_trans, log_reg)
+results = resource_used(model_selection.cross_validate)(model, df, y, )
+print("Normal Cross-validation score: %s" % results['test_score'])
```
GaelVaroquaux: Rather than "Normal" I would use "Default setting"

GaelVaroquaux: Specifically: "Cross-validation score with default settings"

examples/04_scalability.py
```diff
@@ -145,16 +152,15 @@ def wrapped_func(*args, **kwargs):
 
 ################################################################################
 # Check now that prediction is still as good
-model = pipeline.make_pipeline(column_trans, log_ref)
+model = pipeline.make_pipeline(column_trans, log_reg)
 results = resource_used(model_selection.cross_validate)(model, df, y)
-print("Cross-validation score: %s" % results['test_score'])
+print("Most frequent Cross-validation score: %s" % results['test_score'])
```
GaelVaroquaux: "Cross-validation score with most frequent"

examples/04_scalability.py
```diff
@@ -172,21 +178,21 @@ def wrapped_func(*args, **kwargs):
 
 ################################################################################
 # Check now that prediction is still as good
-model = pipeline.make_pipeline(column_trans, log_ref)
+model = pipeline.make_pipeline(column_trans, log_reg)
 results = resource_used(model_selection.cross_validate)(model, df, y)
-print("Cross-validation score: %s" % results['test_score'])
+print("Kmeans Cross-validation score: %s" % results['test_score'])
```
GaelVaroquaux: "Cross-validation score with k-means"

dirty_cat/similarity_encoder.py
```diff
@@ -267,6 +267,9 @@ def fit(self, X, y=None):
 
     def transform(self, X):
         """Transform X using specified encoding scheme.
+            For an n-gram similarity, the output dtype
+            will be the instance dtype, but the input dtype is
+            always float 32.
```
GaelVaroquaux: I don't think that this is necessary: it's confusing for a user who does not look at the internals.

examples/04_scalability.py
```diff
@@ -8,9 +8,12 @@
 
 """
 
+import warnings
```
GaelVaroquaux: This should be positioned below the relevant comment (ie where it was before :D )

examples/04_scalability.py
```diff
@@ -197,3 +201,49 @@ def wrapped_func(*args, **kwargs):
 [t.set(size=16) for t in ax2.get_yticklabels()]
 plt.tight_layout()
 
+################################################################################
+# Similarity encoder with most frequent strategy and float32 dtype
+# ----------------------------------------------------------------
+#
+# The most frequent strategy selects the n most frequent values in a dirty
+# categorical variable to reduce the dimensionality of the problem and thus
+# speed things up. We select manually the number of prototypes we want to use.
```
GaelVaroquaux: The sentences right above seem redundant with what was said before. I am not sure that they are necessary.

examples/04_scalability.py
```diff
@@ -197,3 +201,49 @@ def wrapped_func(*args, **kwargs):
 [t.set(size=16) for t in ax2.get_yticklabels()]
 plt.tight_layout()
 
+################################################################################
+# Similarity encoder with most frequent strategy and float32 dtype
+# ----------------------------------------------------------------
```
GaelVaroquaux: I would use as a title "Reduce memory usage during encoding using float32".

I would add a sentence stressing that the subsequent scikit-learn model may upcast to float64, but that it leads to memory saving during the encoding, as can be seen below.

GaelVaroquaux:
Also, you need to add a line to the CHANGES.rst file (create an entry for the next release, 0.0.6).

GaelVaroquaux:
LGTM. Merging.

Issue #77, opened 2018-11-30
Change default of "handle_unknown" in SimilarEncoder


GaelVaroquaux:
The default should be to "ignore", rather than "error", as the SimilarityEncoder encoders very well unknown categories.

We should change the default, adapt examples (ie remove the option from the examples that set it, and always use the default), and add a line in the CHANGES.rst.

LilianBoulard:
Just checked, and it's already done !

PR #78, opened 2018-11-30
#77 Removed unknown parameters as ignore is default, corrected documentation

mcuny:


mcuny:
I did not change anything in the Change.rst file as we did not actually change the similarity encoder class.

GaelVaroquaux:
Indeed!

LGTM, and the tests pass. Merging. Thank you!

PR #81, opened 2018-12-11
Filling with 0s when we don't have enough prototypes

mcuny:
Not sure if this is the correct way to implement it, maybe using NumPy functions would be better?

### DIFF
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -97,7 +97,7 @@ def get_kmeans_prototypes(X, n_prototypes, hashing_dim=128,
     if indexes_prototypes.shape[0] < n_prototypes:
         indexes_prototypes = \
             np.append(indexes_prototypes,
-                      [[0] * indexes_prototypes.shape[1]] * (n_prototypes - indexes_prototypes.shape[0]),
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
Can you add a test please (no functionality without tests).

dirty_cat/similarity_encoder.py
```diff
@@ -95,6 +95,10 @@ def get_kmeans_prototypes(X, n_prototypes, hashing_dim=128,
     neighbors.fit(projected)
     indexes_prototypes = np.unique(neighbors.kneighbors(centers, 1)[-1])
     if indexes_prototypes.shape[0] < n_prototypes:
+        indexes_prototypes = \
+            np.append(indexes_prototypes,
+                      [0] * (n_prototypes - indexes_prototypes.shape[0]),
```
GaelVaroquaux: You should use np.empty here.

GaelVaroquaux:
You need to add a test for this functionality.

GaelVaroquaux:
I think that we can close this PR, because it's been 4 years and that functionality does not seem crucial.

Still, thanks for everything you've done with us, @mcuny

PR #87, opened 2018-12-13
Fix error when list of categories

pierreglaser:
Also increased the coverage a bit by adding a test about this feature (giving a list of categories to `SimilarityEncoder`).
The test relies on the `auto` behavior though, which may change, so @GaelVaroquaux if you are uncomforatable with it I can also simply check a hardcoded similarity matrix.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,9 @@
 Release 0.0.6
 =============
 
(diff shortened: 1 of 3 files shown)


GaelVaroquaux:
Thank you very much. Merging!

PR #88, opened 2018-12-14
DOC, CLN fix bullet list indentation in docstring

pierreglaser:
`sphinx` outputs warnings when building the docs because bullet-lists in the docstring were not properly indented. Fixed that. Will merge soon if no-one objects.


pierreglaser:
Sounds good 👍

PR #90, opened 2018-12-14
DOC first version of the non-linear online example

pierreglaser:
Example about non-linear starts stabilizing IMO, so I can start taking feedback. 

In short, the example shows a classification problem for the `traffic_violations` dataset.

5 types of models are implemented:
* `SimilarityEncoder` + `SVC`
* `SimilarityEncoder` + `RBFSampler` + `SVC`
* `SimilarityEncoder` + `RBFSampler` + `SGD`
* `SimilarityEncoder` fitted on a subset of the data + `RBFSampler` + `SGD`
* `SimilarityEncoder` fitted on a subset of the data  + `RBFSampler` + online `SGD`

Training set sizes still dummy to make CI pass fast. For now, the scripts takes around 15 sec to run.
`RBFSampler` can easily be switched in favor of `Nystroem`.

### DIFF
--- a/examples/05_scaling_non_linear_models.py
+++ b/examples/05_scaling_non_linear_models.py
@@ -1,6 +1,7 @@
 r"""
 Fitting scalable, non-linear models on data with dirty categories
 =================================================================
+
 Non-linear models form a very large model class. Among others, this class
 includes:
 
@@ -15,23 +16,24 @@ grasp more complex relationships between the input and the output of a machine
 learning problem. However, using a non-linear model model often comes at a
 cost:
 
-* for neural networks, this cost is an extended model tuning time, in order to
+* For neural networks, this cost is an extended model tuning time, in order to
   achieve good optimization and network architecture.
-* gradient boosting machines do not tend to scale extremely well with
+* Gradient Boosting Machines do not tend to scale extremely well with
   increasing sample size, as all the data needs to be loaded into the main
   memory.
-* for kernel methods, parameter fitting requires the inversion of a gram matrix
-  of size :math:`n \times m` (:math:`n` being the number of samples), yiedling
+* For kernel methods, parameter fitting requires the inversion of a gram matrix
+  of size :math:`n \times n` (:math:`n` being the number of samples), yiedling
   a quadratic dependency (with n) in the final compmutation time.
 
 
-All is not lost though. For kernel methods, there exists approximation
-algorithms, that drops the quadratic dependency with the sample size while
+All is not lost though. For kernel methods, there exist approximation
+algorithms that drop the quadratic dependency with the sample size while
 ensuring almost the same model capacity.
 
 In this example, you will learn how to:
     1. build a ML pipeline that uses a kernel method.
-    2. make this pipeline scalable.
+    2. make this pipeline scalable, by using online algorithms and dimension
+       reduction methods.
 
 
 .. note::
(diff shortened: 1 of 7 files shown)


GaelVaroquaux:
Thanks for the example. I just had a quick look and here are my immediate reactions:

* It's very didactic. You did a good job inserting text and mark up to make it easy to follow
* It's a bit long, we should find things to remove. We really need to keep the online part. Maybe we should remove things before
* The example limits itself to n=3000. I would like to run it on the full dataset, which is 50 times bigger. This means that non online methods will not be scalable. We can compare non online methods on a subset of the data (3000 for instance) to online methods on the full data.

pierreglaser:
> It's a bit long

Agreed. We can remove one model. `SimilarityEncoder fitted on a subset of the data + RBFSampler + SGD` (4) seems like a good candidate.

>The example limits itself to n=3000

Actually, there are 2 variables defining the training set size:

* `train_set_size`, that defines the training set size for non-online models. Currently set to 3000
* `online_train_set_size` that defines the training set size for the online model. Currently set to 10000.

We can scale it up to x4 (12000 and 40000), which I did on my laptop. Afterwards, we reach our few minutes time budget for the example. Note that the test set is the same for every (online and non-online) models.

GaelVaroquaux:
In the list, I would probably remove the following:

* SimilarityEncoder + RBFSampler + SVC
* SimilarityEncoder + RBFSampler + SGD
* SimilarityEncoder fitted on a subset of the data + RBFSampler + SGD

I would only keep SimilarityEncoder + SVC rbf and the online solution. I
would explain that the RBFSampler is necessary to approximate the
non-linear model fit by a non-linear feature transformation, that can
then be fit online.

> We can scale it up to x4 (12000 and 40000), which I did on my laptop.
> Afterwards, we reach our few minutes time budget for the example.

OK. But does the time cost come from the online part, or the other one. I
would be in favor of keeping the non online small. I think that it is
didactic to show that only enables basically unbounded fits.

Also, in terms of speed, it may be interesting to reduce a bit the
embedding dimension of the SimilarityEncoder using tricks in
https://dirty-cat.github.io/stable/auto_examples/04_dimension_reduction_and_performance.html
maybe with 200 protoptypes and the most_frequent strategy. This should
make everything faster, and I would hope that it doesn't change much in
terms of generalization.

> Note that the test set is the same for every (online and non-online) models.

Good to have the same test set!

pierreglaser:
>  it may be interesting to reduce a bit the embedding dimension of the SimilarityEncoder
I already use only 100 categories (the first 100 rows of the training set). So dimension is already pretty low. But I can switch to using the most frequent strategy.

GaelVaroquaux:
>     I already use only 100 categories (the first 100 rows of the training set).
>     So dimension is already pretty low. But I can switch to using the most
>     frequent strategy.

Most frequent on a large subset might be a better recommendation.

pierreglaser:
Just updated the example taking the previous remarks into consideration.

pierreglaser:
(There is some noise in the diff - `conf.py` -  that comes from an unfortunate rebase, don't worry about it)

GaelVaroquaux:
A few quick comments:

* It would be good to have a plot: we try to have all our example quite visual. Maybe comparing batch SVM RBF on 2000 data points to the RBF sampler + online on 50000 data points? Or a convergence plot as a function of the number of iterations.
* The RBF embedding dimension should probably be significantly higher. Indeed, the price to pay to approximate kernels is to have a dimensionality much higher than the input dimensionality.
* I think that I would prefer running on the for data, rather than limiting to 50000.

GaelVaroquaux:
Can you merge with master, to fix the conflict and have the CI run

GaelVaroquaux:
Hum, failure on circleci

pierreglaser:
Weird, I cannot reproduce locally. May have to do with the package versions?

GaelVaroquaux:
Detail: would you mind removing the side bar. It doesn't fit in my screen (I blow up the fonts a lot, admitted).

examples/05_scaling_non_linear_models.py
```diff
@@ -0,0 +1,537 @@
...
+models reach their expressivity limit: whatever the number of samples the
+training set may have, past some point, it's precision won't get any better.
+
+Non-linear models, however, tend to *scale* better with sample size: they are
+able to digest the information in the additional samples to get a better
+estimate of the link between the input and the output. This idea is illustrated
+in this plot:
+
+.. image:: ../_static/model_capacity.png
+   :align: center
```
GaelVaroquaux: I would remove this image, and rather point to a relevant page.

In general, I would try to summarize a bit the whole discussion here, to be more concise. However, if you want, I can do it myself after the merge.

examples/05_scaling_non_linear_models.py
```diff
@@ -0,0 +1,537 @@
...
+from sklearn.metrics import average_precision_score
+# define the different train set sizes on which to evaluate the model
+train_set_sizes = [train_set_size // 10, train_set_size // 3, train_set_size]
+
+def evaluate_model(model, X_train, y_train, X_test, y_test, train_set_sizes,
+                   one_hot_encoder):
+    train_times, test_scores = [], []
+
+    for n in train_set_sizes:
+        train_time, _ = profile(model.fit)(X_train[:n], y_train[:n])
```
GaelVaroquaux: Given that "profile" is used only once, I think that the example would be simpler by doing the time accounting manually:
```
t0 = time.time()
model.fit(...)
train_time = time.time() - t0
```

examples/05_scaling_non_linear_models.py
```diff
@@ -0,0 +1,537 @@
...
+        y_pred_onehot = one_hot_encoder.transform(y_pred.reshape(-1, 1))
+        y_test_onehot = one_hot_encoder.transform(y_test.reshape(-1, 1))
+
+        test_score = average_precision_score(y_test_onehot, y_pred_onehot)
+
+        train_times.append(train_time)
+        test_scores.append(test_score)
+    return train_times, test_scores
+
+train_times_svc, test_scores_svc = evaluate_model(model, X_train, y_train,
```
GaelVaroquaux: Given that "evaluate_model" is used only once, I would not define a function, and I would simply inline the code, to make the example simpler.

examples/05_scaling_non_linear_models.py
```diff
@@ -0,0 +1,537 @@
...
+
+###############################################################################
+# Increasing training size cleary improves model accuracy. However, the
+# training time and the input size increase quadratically with the training set
+# size.  Indeed, kernel methods need to process an entire :math:`n \times n`
+# matrix at once. In order for this matrix to be loaded into memory, :math:`n`
+# has to remain low: Using 30000 observations, the input is a :math:`30000
+# \times 30000` matrix.  If composed of 32-bit floats, its total size is around
+# :math:`30000^2 \times 32 = 2.8 \times 10^{10} \text{bits} = 4\text{GB}`,
+# which is the caracteristic size of many personal laptops.
```
GaelVaroquaux: I think that you can remove "which is the caracteristic size of many personal laptops". In general, it is useful to say less. It's a classic mantra of communication: "less is more", in other words, the less you say, the more what you say in impactful (I should learn that IRL :D ).

examples/05_scaling_non_linear_models.py
```diff
@@ -0,0 +1,537 @@
...
+    random_state=42, ngram_range=(2, 4))
+
+# Fit the rbf_sampler with the similarity matrix.
+column_transformer = make_column_transformer(
+    (similarity_encoder, ['NONPROPRIETARYNAME']),
+    (OneHotEncoder(handle_unknown='ignore'), ['DOSAGEFORMNAME', 'ROUTENAME']),
+    sparse_threshold=1)
+
+transformed_categories = column_transformer.fit_transform(X_encoder)
+
```
GaelVaroquaux: Maybe say here (in a small comment on one or two lines) that the gamma is an important hyperparameter that needs to be set on the data.

examples/05_scaling_non_linear_models.py
```diff
@@ -0,0 +1,537 @@
...
+from sklearn.linear_model.stochastic_gradient import SGDClassifier
+
+online_train_set_size = 100000
+# Filter warning on max_iter and tol
+warnings.filterwarnings('ignore', module='sklearn.linear_model')
+sgd_classifier = SGDClassifier(
+    max_iter=1, tol=None, random_state=42, average=10)
+
+###############################################################################
+# We can now start the training. There are two intricated loops, one iterating
```
GaelVaroquaux: I would use "nested" instead of "intricated".

GaelVaroquaux:
I think that I would prefer merging the 2 figures in a single one with 2 subplots one on top of the other:

* The test accuracy as a function of time for both approaches (batch and online)
* The number of samples visited as a function of time (also for both approaches.

GaelVaroquaux:
Ok, I had a full look at the example. Overall, I think that it shows very well what we wanted to show: scalable non-linear models that will benefit from many samples.

I made a bunch of comments, and overall I would summarize them by saying that now we need to find a way of simplifying this example to make it easier to follow for beginners. There are also probably too many considerations that are a bit theoretical or mathematical that we can remove.

Thanks @pierreglaser !

examples/05_scaling_non_linear_models.py
pierreglaser: I removed the image. Feel free to change the text after the merge, it will probably be faster that me iterating several times upon your feedback.

GaelVaroquaux:
Oh no, it looks like Circle has failed :(.

pierreglaser:
yes, stupid mistake: I changed the logger frequency, which caused the downsampling of `online_train_set_sizes`, hence `matplotlib` complaining about the input shape.

GaelVaroquaux:
Very nice and very didactic!

I am definitely in favor of merging this, and I might address phrasing details later.

One important detail: looking at the curves, it seems that it is possible to illustrate the benefits of online with only half of the computing time. That might be nice, to reduce the load on the CI servers.

GaelVaroquaux:
> By setting n_epoch to 1 I assume. I was about to do that, but then I thought it
> might confuse the reader to see an outer loop with only one iteration. I do not
> have a strong opinion on that subject though.

I think that I would remove the outer loop and suggest in a comment that
it would probably be a good idea to sacrifice a bit more computing time
and do multiple passes on the data, unless it is very big.

GaelVaroquaux:
Very nice! Thank you. Merging!

PR #93, opened 2018-12-17
DOC change note cell background

### LINKED ISSUE #92: DOC note block and python code block have the same background
Title says it call. 
I suggest using light grey for the note cells.

pierreglaser:
Closes #92 

What do you think?
ref: https://www.color-hex.com/color/e9e9e9

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -137,6 +137,7 @@ html_theme_options = {
     'link': "#076B00",
 #    'gray_3': "#090",
     'fixed_sidebar': 'true',
+    'note_bg': "#E9E9E9"
 }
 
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
Can you make it a bit lighter. For instance "rgb(246, 248, 250);" (classic grey-box color, used eg on github).

doc/conf.py
```diff
@@ -138,7 +138,7 @@
 #    'gray_3': "#090",
     'fixed_sidebar': 'true',
     'note_bg': "#E9E9E9",
```
GaelVaroquaux: Can you change it on note too?

pierreglaser: Sorry about that.

PR #95, opened 2018-12-17
DOC 'scikit-learn'->'sklearn' sphinx_gallery_conf

### LINKED ISSUE #94: intersphinx does not link to scikit-learn for inline code
we should use `sklearn` and not `scikit-learn` for the key of `sphinx_gallery_conf`, in order to enable inline code links to the `scikit-learn` documentation.

pierreglaser:
closes #94

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -204,11 +204,11 @@ texinfo_documents = [
 
 # Configuration for intersphinx
 intersphinx_mapping = {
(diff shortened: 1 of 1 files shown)


pierreglaser:
while we're at it, might as well use the new format for the intersphinx_mapping config.

GaelVaroquaux:
LGTM. Merging.

PR #96, opened 2018-12-17
DOC enable toc in example sidebars

pierreglaser:


GaelVaroquaux:
I had disabled it because I liked the lighter design :). I'm open to
discussion, but it is hard to keep a fixed sidebar (as opposed to
scrolling) and the toc: it tends to create a too high sidebar.

pierreglaser:
Yes I figured it was voluntary. The example in #90 is pretty long, that is why I wanted to enable it. 
I add a toc in the main body instead.

GaelVaroquaux:
Adding a table in the body seems right. I was about to suggest it, but wanted to hear your thoughts. 

GaelVaroquaux:
This should be closed, shouldn't it?

pierreglaser:
yes. closing

PR #99, opened 2018-12-21
ENH accelerate ngram_similarity

pierreglaser:
Accelerate the computation in `SimilarityEncoder.transform` by:

* Parallelizing the similarity computations using `joblib`
* Computing the count vectors of the vocabulary at fitting time and not at transform time.

### DIFF
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -1,6 +1,7 @@
 import warnings
 
 import numpy as np
+from joblib import Parallel, delayed
 from scipy import sparse
 from sklearn.cluster import KMeans
 from sklearn.feature_extraction.text import CountVectorizer, HashingVectorizer
@@ -9,6 +10,46 @@ from sklearn.preprocessing import OneHotEncoder
 from sklearn.utils import check_random_state
 
 from . import string_distances
+from .string_distances import get_ngram_count, preprocess
+
+
+def _ngram_similarity_one_sample(
+        x_count_vector, vocabulary_count_matrix, str_x,
+        vocabulary_ngram_counts, se_dict, unq_X, i, ngram_range):
+    """Compute the similarity between one new sample and a vocabulary.
+
+
+    Parameters
+    ----------
+    x_count_vector: np.array
+        count vector of the sample based on the ngrams of the vocabulary
+    vocabulary_count_matrix: np.array
+        count vector of the vocabulary based on its ngrams
+    str_x: str
+        the actual sample string
+    vocabulary_ngram_counts: np.array
+        number of ngrams for each unique element of the vocabulary
+    se_dict: dict
+        dictionary containing the similarities for each x in unq_X
+    unq_X: np.array
+        the arrayes of all unique samples
+    i: str
+        the index of x_count_vector in the csr count matrix
+    ngram_range: tuple
+
+    """
+    nonzero_idx = x_count_vector.indices
+    nonzero_vals = x_count_vector.data
+
+    samegrams = np.asarray(
+        (vocabulary_count_matrix[:, nonzero_idx].minimum(nonzero_vals)).sum(
+            axis=1))
+
+    allgrams = get_ngram_count(
+        str_x, ngram_range) + vocabulary_ngram_counts - samegrams
+    similarity = np.divide(samegrams, allgrams)
+    se_dict[unq_X[i]] = similarity.reshape(-1)
+
 
 
 def ngram_similarity(X, cats, ngram_range, hashing_dim, dtype=np.float64):
@@ -161,6 +202,9 @@ class SimilarityEncoder(OneHotEncoder):
         Must be a positive non null integer.
     random_state: either an int used as a seed, a RandomState instance or None.
         Useful when `k-means` strategy is used.
+    n_jobs: int, optional
+        maximum number of processes used to compute similarity matrices. Used
+        only if ``fast=True`` in ``SimilarityEncoder.transform``
 
     Attributes
     ----------
@@ -181,7 +225,8 @@ class SimilarityEncoder(OneHotEncoder):
 
     def __init__(self, similarity='ngram', ngram_range=(2, 4),
                  categories='auto', dtype=np.float64,
-                 handle_unknown='ignore', hashing_dim=None, n_prototypes=None, random_state=None):
+                 handle_unknown='ignore', hashing_dim=None, n_prototypes=None,
+                 random_state=None, n_jobs=None):
         super().__init__()
         self.categories = categories
         self.dtype = dtype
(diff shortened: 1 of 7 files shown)


dirty_cat/similarity_encoder.py
```diff
@@ -9,6 +10,46 @@
 from sklearn.utils import check_random_state
 
 from . import string_distances
+from .string_distances import get_ngram_count, preprocess
+
+
+def _ngram_similarity_one_sample(
```
GaelVaroquaux: Maybe add "inplace" to the name, and clearly indicate that you are modifying inplace on variables (super important to understand the code).

pierreglaser:
(forgot to add joblib in the CI scripts)

dirty_cat/similarity_encoder.py
pierreglaser: indeed!

GaelVaroquaux:
Comparing https://82-125209614-gh.circle-artifacts.com/0/doc/auto_examples/04_dimension_reduction_and_performance.html#sphx-glr-auto-examples-04-dimension-reduction-and-performance-py to https://dirty-cat.github.io/stable/auto_examples/04_dimension_reduction_and_performance.html#sphx-glr-auto-examples-04-dimension-reduction-and-performance-py it seems like the fit_transform is significantly slower. Maybe it's just an artifact of the CI (non repeatable infrastructure), but it would be worth checking.

dirty_cat/similarity_encoder.py
```diff
@@ -316,11 +394,61 @@ def transform(self, X):
             out = np.empty((len(X), total_length), dtype=self.dtype)
             last = 0
             for j, cats in enumerate(self.categories_):
-                out[:, last:last + len(cats)] = ngram_similarity(X[:, j], cats,
-                                                                 ngram_range=(min_n, max_n),
-                                                                 hashing_dim=self.hashing_dim,
-                                                                 dtype=np.float32)
+                if fast:
+                    encoded_Xj = self._ngram_similarity_fast(X[:, j], j)
```
GaelVaroquaux: Do we have any reason not to want "fast"? Given that we are already paying the price of fitting the vectorizers during fit.

dirty_cat/similarity_encoder.py
```diff
@@ -316,11 +394,61 @@ def transform(self, X):
...
+        X: np.array, list
+            observations being transformed.
+        col_idx: int
+            the column index of X in the original feature matrix.
+        """
+        min_n, max_n = self.ngram_range
+        vectorizer = self._vectorizers[col_idx]
+
+        unq_X = np.unique(X)
+        unq_X_ = np.array(list(map(preprocess, unq_X)))
```
GaelVaroquaux: I wonder whether this line doesn't go faster if you indicate the dtype to np.array. Though it might be negligeable, and thus not worth doing it.

dirty_cat/string_distances.py
```diff
@@ -19,6 +20,52 @@
...
+    min_n, max_n = ngram_range
+    ngram_count = 0
+
+    for i in range(min_n, max_n + 1):
+        ngram_count += len(X) - i + 1
+
+    return ngram_count
+
+
+def preprocess(x):
```
GaelVaroquaux: I think that we need a more explicit name. Maybe "count_vectorizer_preprocess".

dirty_cat/similarity_encoder.py
pierreglaser: It allowed an easy comparison between the two methods during development. Indeed I think this can be removed before merging, but before that I can show the output a minimal script showing time comparisons for the two methods.

dirty_cat/string_distances.py
```diff
@@ -19,6 +20,52 @@
...
+    prior to the CountVectorizer.
+    """
+
+    # preprocessing step done in ngram_similarity
+    x = ' %s ' % x
+
+    # preprocessing step done in the CountVectorizer
+    _white_spaces = re.compile(r"\s\s+")
+
+    return _white_spaces.sub(' ', x)
```
GaelVaroquaux: Can we turn off this preprocessing in count vectorizer? We don't really want it, and it probably slows things down.

pierreglaser: nope!

GaelVaroquaux: :(

dirty_cat/string_distances.py
pierreglaser: We can choose a more explicit name, but `count_vectorizer_preprocess` may be misleading as some of the preprocessing done was done by `ngram_similarity` (adding spaces at the extremities of the string)

dirty_cat/similarity_encoder.py
pierreglaser: In terms of dtype, what do you have in mind? `object`? a fixed-size unicode character length?

GaelVaroquaux:
> In terms of dtype, what do you have in mind? object? a fixed-size unicode character length?

Good point: it cannot be determined without scanning the full list, hence
it cannot be done faster. Stupid suggestion!

pierreglaser:
> it seems like the fit_transform is significantly slower

running this script:
```python
import pandas as pd
import numpy as np
import time
from dirty_cat.datasets import fetch_traffic_violations
from dirty_cat.similarity_encoder import SimilarityEncoder


def profile(func):
    def profiled_func(*args, **kwargs):
        t0 = time.perf_counter()
        res = func(*args, **kwargs)
        total_time = time.perf_counter() - t0
        return total_time, res
    return profiled_func


info = fetch_traffic_violations()

df = pd.read_csv(info['path'], nrows=2000, usecols=['Description'])

categories = df[['Description']].values

similarity_encoder = SimilarityEncoder(
    similarity='ngram', ngram_range=(2, 4),
    dtype=np.float32)

similarity_encoder.fit(categories)

time_classic, similarity_matrix = profile(similarity_encoder.transform)(
    categories, fast=False)
time_fast, similarity_matrix_fast = profile(similarity_encoder.transform)(
    categories, fast=True)

assert np.allclose(similarity_matrix, similarity_matrix_fast)
print('classic method: {:.3f}'.format(time_classic))
print('accelerated method: {:.3f}'.format(time_fast))
```
with `n_jobs=None` (so 1) consistently outputs results around this range:
```
classic method: 0.674
accelerated method: 0.498
```

Let me know if you want a more robust experiment.

GaelVaroquaux:
Same script with 50000 rows? 

pierreglaser:
```
classic method: 10.774
accelerated method: 9.759
```
It seems like as the number of rows increase, the gain becomes smaller.

dirty_cat/similarity_encoder.py
```diff
@@ -199,6 +244,10 @@ def __init__(self, similarity='ngram', ngram_range=(2, 4),
         if categories == 'auto' and n_prototypes is not None:
             warnings.warn('n_prototypes parameter ignored with category type \'auto\'')
 
+        self._vectorizers = None
+        self._vocabulary_count_matrices = None
+        self._vocabulary_ngram_counts = None
```
GaelVaroquaux: These variable names should end with an underscore to indicate that they are derived from data (scikit-learn conventions)

GaelVaroquaux: Also, in scikit-learn, we tend not to instantiate variables in init, but in fit

GaelVaroquaux:
Thank you. Merging!

Issue #103, opened 2019-05-07
similarity encoder broken by recent change in scikit-learn _BaseEncoder


jeromedockes:
_BaseEncoder._check_X now returns X_columns, n_samples, n_features
(from git blame:

b2344a4e3f7924d92ff02b6e9dc476919da606da (maikia                2019-03-01 11:02:57 +0100  70)         return X_columns, n_samples, n_features

)

willettk:
Echoed that this breaks on the current version of scikit-learn. I'm able to run the [example script](https://dirty-cat.github.io/stable/auto_examples/02_fit_predict_plot_employee_salaries.html#sphx-glr-download-auto-examples-02-fit-predict-plot-employee-salaries-py) using scikit-learn 0.20.2, but fitting a pipeline with the SimilarityEncoder() returns an error when using scikit-learn 0.21.1. 

This means that most of the example scripts will currently only work for one minor version, since ColumnTransformer() (which is used in almost all examples) was only introduced in 0.20.0.

amueller:
@willettk "minor" is the only kind of version there is for sklearn - apart from bug-fix ;)

GaelVaroquaux:
This is fixed, I believe. Closing.

PR #104, opened 2019-06-17
[BUG] _check_X from _BaseEncoder returns a tuple in new scikit-learn version

pcerda:
#103


dirty_cat/similarity_encoder.py
```diff
@@ -266,7 +266,9 @@ def fit(self, X, y=None):
         -------
         self
         """
-        X = self._check_X(X)
+
+        Xlist, n_samples, n_features = self._check_X(X)
```
GaelVaroquaux: Hum, I think that we should try to work on different versions of scikit-learn. Hence we probably need a test on the version of scikit-learn.

You can find examples of such test here:
https://github.com/nilearn/nilearn/search?q=looseversion&unscoped_q=looseversion

dirty_cat/similarity_encoder.py
```diff
@@ -266,7 +272,13 @@ def fit(self, X, y=None):
         -------
         self
         """
-        X = self._check_X(X)
+
+        if LooseVersion(sklearn.__version__) > LooseVersion('0.21'):
+            Xlist, n_samples, n_features = self._check_X(X)
+        else:
+            Xlist = self._check_X(X)
```
jeromedockes: should be transposed

TwsThomas:
Can be close thanks to https://github.com/dirty-cat/dirty_cat/pull/108

PR #105, opened 2019-10-11
Added minhash encoder and the Gamma-Poisson factorization. Also added…

alexis-cvetkov:
… a reformat_input method for similarity, minhash and gamma-poisson encoders so that input lists, arrays and pandas objects can be used in fit_transform.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,6 @@
 language: python
 env:
-    - CONDA_DEPENDENCIES='python-Levenshtein distance sklearn joblib numpy scipy requests pytest-cov codecov'
+    - CONDA_DEPENDENCIES='python-Levenshtein distance sklearn joblib numpy scipy requests pytest-cov codecov coverage'
 python:
     - 3.6
 install:


dirty_cat/minhash_encoder.py
```diff
@@ -0,0 +1,148 @@
...
+                   "minmax_hash not implemented with murmur"
+            return self.minhash(
+                    string, n_components=self.n_components,
+                    ngram_range=self.ngram_range)
+        else:
+            raise ValueError("hashing function must be 'fast' or"
+                             "'murmur', got '{}'"
+                             "".format(self.hashing))
+
+    def reformat_input(self, X):
```
GaelVaroquaux: This blank line shouldn't be there.

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,478 @@
+import numpy as np
+
+import pandas as pd
```
jeromedockes: if you need pandas add it to the `install_requires` list in setup.py to fix the CI error

GaelVaroquaux:
> if you need pandas add it to the install_requires list in setup.py to fix the CI error

I don't think that we should depend on pandas. Is it easy to do without?

jeromedockes:
Gael Varoquaux <notifications@github.com> writes:

> I don't think that we should depend on pandas. Is it easy to do without?

as far as I can tell it is only used to check input type, to accept
dataframes and series. I guess we can remove it and only accept arrays,
or instead check the presence of an attribute `to_numpy`

GaelVaroquaux:
> or instead check the presence of an attribute `to_numpy`

The later.

Issue #115, opened 2020-04-14
dirty_cat jaro distance is different from python-Levenshtein package


alexis-cvetkov:
The function test_compare_implementations of test_string_distances.py compares dirty_cat implementation of jaro and jaro_winkler metrics with the python-Levenshtein package.
In practice these 2 implementations are different, but I believe it has never been signaled by Travis since the comparison is done on a small number (N=10) of random strings for which both method returns 0 with very high probability.

Here is a simple code to expose the difference:

```
from dirty_cat import string_distances
import Levenshtein

for s1, s2 in[('acc', 'babbb'), ('acc', 'bbabb'), ('acc', 'bbbab')]:
    print(f'dirty_cat {s1, s2} = {string_distances._jaro_winkler(s1, s2, winkler=False)}')
    print(f'Levenshtein {s1, s2} = {Levenshtein.jaro(s1, s2)}')
    print('-'*50)

>>> dirty_cat ('acc', 'babbb') = 0.5111111111111111
>>> Levenshtein ('acc', 'babbb') = 0.5111111111111111
>>> --------------------------------------------------
>>> dirty_cat ('acc', 'bbabb') = 0.0
>>> Levenshtein ('acc', 'bbabb') = 0.5111111111111111
>>> --------------------------------------------------
>>> dirty_cat ('acc', 'bbbab') = 0.0
>>> Levenshtein ('acc', 'bbbab') = 0.0
>>> --------------------------------------------------
```
From this examples and other tests I've made I believe that the difference lies in the definition of matching characters. The dirty_cat implementation uses the [wikipedia](https://en.wikipedia.org/wiki/Jaro%E2%80%93Winkler_distance) definition: two characters from s1 and s2 are considered matching only if they are the same and not farther than
`int(max(|s1|,|s2|) / 2) - 1` = 1 in my example.

Instead, the Levensthein implementation seems to use `int( (min(|s1|,|s2|) + 1) / 2)` = 2.

GaelVaroquaux:
Fixed in #116. Thanks!

PR #116, opened 2020-04-20
Fixed dirty_cat jaro_winkler implementation to match Levensthein.

alexis-cvetkov:
Should resolve issue #115. There was 2 problems with our implementation of the Jaro and Jaro-Winkler distance.
- `search_range`, i.e. the maximum range when matching characters was different: `max_len // 2 -1` against `(min_len - 1) // 2`.
- In some cases like `min_len <= 3`, the Jaro-Winkler distance wasn't computed even though we set `winkler=True`.

I also added more thorough tests in `test_compare_implementations`. It should now discriminate implementations with different `search_range`.

### DIFF
--- a/dirty_cat/test/test_string_distances.py
+++ b/dirty_cat/test/test_string_distances.py
@@ -1,6 +1,7 @@
 import unittest
 
 import numpy as np
+import random
 
 try:
     import Levenshtein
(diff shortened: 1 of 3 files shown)


dirty_cat/test/test_string_distances.py
```diff
@@ -22,6 +23,19 @@ def _random_string_pairs(n_pairs=50):
         pairs.append((s1, s2))
     return pairs
 
+def _random_common_char_pairs(n_pairs=50):
+    """
+    Return string pairs with a common char at random positions. This should
+    discriminate different thresholds for matching chararacters in Jaro distance.
+    """
+    # Make strings with random length and common char at index 0
```
GaelVaroquaux, line 31: You should not use the global random number generator, as this leads to tests that are not perfectly reproducible. Rather, create a local random number generator.
```suggestion
    # Make strings with random length and common char at index 0
    rng = np.random.RandomState(seed=seed)
```

dirty_cat/test/test_string_distances.py
```diff
@@ -22,6 +23,19 @@ def _random_string_pairs(n_pairs=50):
         pairs.append((s1, s2))
     return pairs
 
+def _random_common_char_pairs(n_pairs=50):
+    """
+    Return string pairs with a common char at random positions. This should
+    discriminate different thresholds for matching chararacters in Jaro distance.
+    """
```
GaelVaroquaux, line 30: The code in this function does not follow the PEP8 coding convention: you need spaces after the commas, and before and after "+", "*".

alexis-cvetkov:
I have added your suggestions, I think it is good for merging now!

GaelVaroquaux:
Looks good to me. However, a small detail is that we need to add a note that we changed this, for users to see. It should go here: https://github.com/dirty-cat/dirty_cat/blob/master/CHANGES.rst

PR #119, opened 2020-05-14
[WIP] fix bug in examples

TwsThomas:
Change the download source from official link to openML. 
For employees_salaries

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,7 @@
+Release 0.0.7
+=============
+* **datasets.fetch_employee_salaries**: change the origin of download for employee_salaries.
+
 Release 0.0.6
 =============
 * **SimilarityEncoder**: Fixed a bug when using the Jaro-Winkler distance as a
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -2,7 +2,7 @@ Release 0.0.7
 =============
 * **datasets.fetch_employee_salaries**: change the origin of download for employee_salaries.
   - The function now return a bunch with a dataframe under the field "data",
-   and not the path to the csv file. 
+    and not the path to the csv file. 
   - The field "description" has been renamed to "DESCR".
 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,6 +1,10 @@
 Release 0.0.7
 =============
 * **datasets.fetch_employee_salaries**: change the origin of download for employee_salaries.
+  - The function now return a bunch with a dataframe under the field "data",
+   and not the path to the csv file. 
+  - The field "description" has been renamed to "DESCR".
+
 
 Release 0.0.6
 =============
--- a/dirty_cat/datasets/fetching.py
+++ b/dirty_cat/datasets/fetching.py
@@ -13,6 +13,7 @@ import urllib
 from collections import namedtuple
 import contextlib
 import warnings
+from sklearn.datasets import fetch_openml
 
 from ..datasets.utils import md5_hash, _check_if_exists, \
     _uncompress_file, \
@@ -357,7 +358,6 @@ def fetch_employee_salaries():
 
     """
 
-    from sklearn.datasets import fetch_openml
     data = fetch_openml(data_id=42125, as_frame=True)
     return data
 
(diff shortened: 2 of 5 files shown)


GaelVaroquaux:
Darn, still failing!

Is it the download that is failing? If so, we should change where we download from. I had in mind that we had uploaded this data on openML, but I am not sure.

TwsThomas:
It seems to me it is the download indeed.
I will work on that with openML.

GaelVaroquaux:
> I will work on that with openML.

Did we already upload it there? I have in mind that we had started
uploading the datasets.

For the download, you can use
https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_openml.html

TwsThomas:
> Did we already upload it there? I have in mind that we had started
uploading the datasets.

Most of them are already in openML (cf the PR https://github.com/dirty-cat/datasets/pull/4)

> For the download, you can use
https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_openml.html

I will do it with that ok.

GaelVaroquaux:
> Most of them are already in openML (cf the PR dirty-cat/datasets#4)

Well, that work is paying out today!!! That's good news.

dirty_cat/datasets/fetching.py
```diff
@@ -346,18 +346,23 @@ def fetch_employee_salaries():
...
+            - the target (under the ``target`` key)
 
     References
     ----------
     https://catalog.data.gov/dataset/employee-salaries-2016
 
     """
 
-    return fetch_dataset(EMPLOYEE_SALARIES_CONFIG, show_progress=False)
+    from sklearn.datasets import fetch_openml
```
GaelVaroquaux, line 360: I think that I would prefer if this import was moved to the top of the module, with the other imports.

examples/02_fit_predict_plot_employee_salaries.py
```diff
@@ -21,16 +21,17 @@
...
 employee_salaries = fetch_employee_salaries()
-print(employee_salaries['description'])
+print(employee_salaries['DESC'])
+
 
 ################################################################################
 # Then we load it:
 import pandas as pd
-df = pd.read_csv(employee_salaries['path']).astype(str)
+df = data = employee_salaries['data']
```
GaelVaroquaux, line 30:
```suggestion
df = employee_salaries['data']
```

examples/02_fit_predict_plot_employee_salaries.py
```diff
@@ -21,16 +21,17 @@
...
 
 # Test if load was unsuccesful
 if '"code" : "authentication_required"' in str(df.iloc[0]):
-    print('Error while loading the data') #raise IOError
+    raise IOError
```
GaelVaroquaux, line 34: Should we remove the if clause: we shouldn't need it anymore, right?

GaelVaroquaux:
CI is still failing because other examples need to be updated. But this is looking good!

Can you also add an entry in CHANGES.rst that documents the changes of API in the fetcher?

CHANGES.rst
```diff
@@ -1,3 +1,7 @@
+Release 0.0.7
+=============
+* **datasets.fetch_employee_salaries**: change the origin of download for employee_salaries.
```
GaelVaroquaux, line 3: Please also mention that this changed the function signature: the function now return a Bunch with a dataframe under the field "data", and not the path to the csv file. Also, the field "description" has been renamed to "DESCR".

Mentioning this is important, because these changes can break user code, and they need to quickly identify them from the changelog.

CHANGES.rst
```diff
@@ -1,6 +1,10 @@
 Release 0.0.7
 =============
 * **datasets.fetch_employee_salaries**: change the origin of download for employee_salaries.
+  - The function now return a bunch with a dataframe under the field "data",
+   and not the path to the csv file.
```
GaelVaroquaux, line 5:
```suggestion
    and not the path to the csv file. 
```
The indentation above is not valid restructured text, I believe.

GaelVaroquaux:
Hurray, merging!

PR #123, opened 2020-05-15
Fix examples_02

TwsThomas:
On my laptop, example 02 raise ValueError because of Nan in "gender". 
Here is a possible fix. (which slightly rise the variance in scores)


alexis-cvetkov:
Looks good for merging to me!

GaelVaroquaux:
Awesome!

We should add support for missing values in our encoders later

PR #125, opened 2020-05-15
Fix example 05 to check for empty chunks

alexis-cvetkov:
There was some issue with example 05 that prevented CircleCI to build the doc with sphinx.
When performing online training on chunks of a dataframe, one of the chunks is empty after the preprocessing step and thus cannot be used for training. I simply added a test to check for empty chunks and skip iterations if needed.


GaelVaroquaux:
What a strange problem. Thanks a lot. This does restore things. Merging.

Issue #124, opened 2020-05-15
Add support for missing values in the encoders


GaelVaroquaux:
Encoding a missing value as a vector of zeros is a reasonable thing. Our theoretical study (https://arxiv.org/abs/1902.06931) shows that the most important thing is to encode them in a special value that can be later picked up by the supervised step.

Our encoders should have an option that controls whether missing values are encoded as zeros or an error is raised (following scikit-learn encoders).

amy12xx:
Is this like the parameter handle_unknown{‘error’, ‘ignore’}, default=’error’ in sklearn's one hot encoder?

GaelVaroquaux:
Thanks for asking!

Not quite: handle_unknown governs everything that happens for categories not seen in the train set. However, the encoders that we have in dirty-cat can deal with those, as long as they are strings.

Rather, it would be to deal with NaNs, which encode missingness.

amy12xx:
Can I take a look at it? I maybe a little slow on fixing it though. 

I have yet to read your study, but is it adequate to encode all missing values with the same encoding (0) and assume they will be handled downstream or better to encode them as unique categories?

I looked briefly at the code, and the sklearn function _check_X() throws the error when there are NaNs. Would a reasonable solution be to impute the NaNs with a special value (for e.g. __nan__) before this check, and then handle later on to encode to 0?

GaelVaroquaux:
It would be great if you'd work on this!!!

I indeed think that the simplest way to embed missing values would be to cast them to a vector of zeros.

Now, with regards to how to implement this, we are hitting a few problems.

I had a quick look, and it seems that, for the SimilarityEncoder at least, replacing nans by the empty string '' works (though it spits out nasty warnings that we should capture at the right place, ie as deep as possible).

```
import pandas as pd
import numpy as np

a = [['a',    'b'],
     ['b',    'a'],
     ['b',    np.nan],
     ['a',    'c'],
     [np.nan, 'a'],
    ]

from dirty_cat import SimilarityEncoder
se = SimilarityEncoder()

X = se.fit_transform(pd.DataFrame(a).fillna(value=''))
```

We could start with this strategy, and do a first PR for the similarity? Note that I don't think that the solution that I coded (using pd's fillna) should be implemented on all types of input. We might have to have a different treatment if the input is a list, and yet another if it is a numpy array.

[...]

PR #128, opened 2020-06-14
[MRG] handle missing values for similarity encoder

amy12xx:

Changes: Added missing values support for similarity encoder

### DIFF
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -407,7 +407,8 @@ class SimilarityEncoder(OneHotEncoder):
             raise ValueError(template % self.handle_missing)
         if hasattr(X, 'iloc') and X.isna().values.any():
             if self.handle_missing == 'error':
-                msg = ("Found missing values in input data.")
+                msg = ("Found missing values in input data; set "
+                       "handle_missing='' to encode with missing values")
                 raise ValueError(msg)
             if self.handle_missing != 'error':
                 X = X.fillna(self.handle_missing)
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -403,10 +403,6 @@ class SimilarityEncoder(OneHotEncoder):
 
         """
 
-        if self.handle_missing not in ['error', '']:
-            template = ("handle_missing should be either 'error' or "
-                        "'', got %s")
-            raise ValueError(template % self.handle_missing)
         if hasattr(X, 'iloc') and X.isna().values.any():
             if self.handle_missing == 'error':
                 msg = ("Found missing values in input data; set "
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -415,16 +415,13 @@ class SimilarityEncoder(OneHotEncoder):
             X = np.asarray(X, dtype=object)
 
         if hasattr(X, 'dtype'):
-            if np.issubdtype(X.dtype, np.str_):
-                X = np.asarray(X, dtype=object)
             mask = _object_dtype_isnan(X)
             if X.dtype.kind == 'O' and mask.any():
                 if self.handle_missing == 'error':
                     msg = ("Found missing values in input data.")
                     raise ValueError(msg)
                 if self.handle_missing != 'error':
-                    X = np.ma.array(X, mask=mask)
-                    X = np.ma.filled(X, self.handle_missing)
+                    X[mask] = self.handle_missing
 
         if LooseVersion(sklearn.__version__) > LooseVersion('0.21'):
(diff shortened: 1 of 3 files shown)


dirty_cat/similarity_encoder.py
```diff
@@ -369,6 +401,31 @@ def transform(self, X, fast=True):
...
+                msg = ("Found missing values in input data.")
+                raise ValueError(msg)
+            if self.handle_missing != 'error':
+                X = X.fillna(self.handle_missing)
+        elif not hasattr(X, 'dtype') and isinstance(X, list):
+            X = np.asarray(X, dtype=object)
+
+        if hasattr(X, 'dtype'):
+            if np.issubdtype(X.dtype, np.str_):
+                X = np.asarray(X, dtype=object)
```
GaelVaroquaux, line 419: If the dtype is a string, we don't have to worry about missing values: there cannot be any. So we should just skip ahead, rather than converting.

amy12xx: Ah yes, I removed it from one function but I guess I missed the other.

GaelVaroquaux:
Great!

Next step is to add a few tests. Probably with lists, pandas, and numpy arrays of dtype object.

amy12xx:
@GaelVaroquaux the build failed because I used pandas library in the test script.

amy12xx:
Just checking if you have an update for me regarding the pandas test? I'm a little eager to close this PR :-)

GaelVaroquaux:
Sorry, I had indeed lost track of this: very busy work week.

So, we need the test and the package to work without pandas installed, so you should delay the import and skip this specific bit of test when pandas is not importable.

However, given that we also want to test with pandas (as you rightfully point out), I've created a test matrix: jobs with and without pandas on travis https://github.com/dirty-cat/dirty_cat/pull/129

Can you rebase (or merge) on master, to benefit from this.

Thanks a lot!!

dirty_cat/similarity_encoder.py
```diff
@@ -273,6 +283,28 @@ def fit(self, X, y=None):
         self
         """
 
+        if self.handle_missing not in ['error', '']:
+            template = ("handle_missing should be either \'error\' or "
```
GaelVaroquaux, line 287: Slight detail: you don't need to protect the ' when the string is with doublequotes: "a 'b' c" is valid.
```suggestion
            template = ("handle_missing should be either 'error' or "
```

dirty_cat/similarity_encoder.py
```diff
@@ -273,6 +283,28 @@ def fit(self, X, y=None):
...
+                        "\'\', got %s")
```
GaelVaroquaux, line 288:
```suggestion
                        "'', got %s")
```

dirty_cat/similarity_encoder.py
```diff
@@ -369,6 +401,28 @@ def transform(self, X, fast=True):
 
         """
 
+        if self.handle_missing not in ['error', '']:
+            template = ("handle_missing should be either 'error' or "
+                        "'', got %s")
+            raise ValueError(template % self.handle_missing)
+        if hasattr(X, 'iloc') and X.isna().values.any():
+            if self.handle_missing == 'error':
+                msg = ("Found missing values in input data.")
```
GaelVaroquaux, line 410:
```suggestion
                msg = ("Found missing values in input data; set handling_missing='' "
                            "to encode with missing values")
```

dirty_cat/test/test_similarity_encoder.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import pandas as pd
```
GaelVaroquaux, line 2: This import has to be protected, so that the tests run even without pandas.

dirty_cat/test/test_similarity_encoder.py
```diff
@@ -45,6 +46,38 @@ def test_fast_ngram_similarity():
...
+            assert e.__str__() == 'Found missing values in input data.'
+            return
+    elif missing == '':
+        ans = sim_enc.fit_transform(observations)
+        assert np.allclose(encoded, ans)
+    else:
+        try:
+            sim_enc = similarity_encoder.SimilarityEncoder(handle_missing=missing)
+        except ValueError as e:
+            assert e.__str__() == "handle_missing should be either \'error\' or \'\', got %s" % missing
```
GaelVaroquaux, line 77:
```suggestion
            assert e.__str__() == "handle_missing should be either 'error' or '', got %s" % missing
```

amy12xx:
Thanks for all the guidance. I can maybe look at the other encoders next if this can be replicated there.

GaelVaroquaux:
> I can maybe look at the other encoders next if this can be replicated there.

Yes! This would be absolutely great!!

Thanks a lot!

PR #130, opened 2020-06-29
[MRG] handle missing values for target encoder

amy12xx:

Changes: Added missing values support for target encoder

### DIFF
--- a/dirty_cat/target_encoder.py
+++ b/dirty_cat/target_encoder.py
@@ -4,6 +4,7 @@ import numpy as np
 from sklearn.preprocessing import LabelEncoder
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils import check_array
+from sklearn.utils.fixes import _object_dtype_isnan
 
 
 def lambda_(x, n):
@@ -20,7 +21,7 @@ class TargetEncoder(BaseEstimator, TransformerMixin):
     probability of y conditional on this category.
     In addition it takes an empirical Bayes approach to shrink the estimate.
 
- 
+
     Parameters
     ----------
     categories : 'auto' or a list of lists/arrays of values.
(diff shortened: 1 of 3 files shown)


amy12xx:
2 observations while fixing this:

1. Target encoder does not have an inverse_transform.
2. If there are 2 categories with the same counts, and y distribution, they will get mapped to the same encoding (I guess it can be interpreted as having the same impact?)

Are both intended behaviour?

dirty_cat/target_encoder.py
```diff
@@ -44,8 +45,14 @@ class TargetEncoder(BaseEstimator, TransformerMixin):
         present during transform (default is to raise). When this parameter
         is set to 'ignore' and an unknown category is encountered during
         transform, the resulting one-hot encoded columns for this feature
-        will be all zeros. In the inverse transform, an unknown category
-        will be denoted as None.
+        will be all zeros.
```
GaelVaroquaux, line 48: I think that the sentence that you removed is actually not about missing values, but about a category that has not been seen at train. I should probably be put back, maybe rephrasing so that it is more clear, something like:
```suggestion
        will be all zeros. In the inverse transform, an unknown category
        (not seen during training) will be denoted as None.
```

amy12xx: I removed it because TargetEncoder does not have an inverse_transform() function. 

```
>>> te.inverse_transform(xx)
AttributeError: 'TargetEncoder' object has no attribute 'inverse_transform'
```

GaelVaroquaux:
I merged. Thanks a lot for your contribution!!!

GaelVaroquaux:
Sorry, I am catching up, with some delay.

>     1. Target encoder does not have an inverse_transform.

Indeed. We could implement one, but it is challenging to make it controlled statistically.

>     2. If there are 2 categories with the same counts, and y distribution, they will get mapped to the same encoding (I guess it can be interpreted as having the same impact?)

Yes. This is intended. It is consistent with the model of TargetEncoder.
 
> Also, if I use a list for the target y, it throws an error. (Works fine with numpy and pandas array)

This is something where we could improve.

amy12xx:
> Also, if I use a list for the target y, it throws an error. (Works fine with numpy and pandas array)
I can fix this when I take a look at MinhashEncoder.

> 1. Target encoder does not have an inverse_transform.
> Indeed. We could implement one, but it is challenging to make it controlled statistically.
Yes, I think for now we can maybe just create an issue, and pick it up later.

PR #132, opened 2020-07-31
[MRG] minhash handle missing values

amy12xx:

Changes: Added missing values support for minhash encoder

### DIFF
--- a/dirty_cat/minhash_encoder.py
+++ b/dirty_cat/minhash_encoder.py
@@ -199,8 +199,7 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
                 if isinstance(x, float):
                     nan_idx.append(i)
                 elif x not in self.hash_dict:
-                    self.hash_dict[x] = self.get_fast_hash(x)
-                    X_out[i, :] = self.hash_dict[x]
+                    X_out[i, :] = self.hash_dict[x] = self.get_fast_hash(x)
                 else:
                     X_out[i, :] = self.hash_dict[x]
 
@@ -209,11 +208,10 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
                 if isinstance(x, float):
                     nan_idx.append(i)
                 elif x not in self.hash_dict:
-                    self.hash_dict[x] = self.minhash(
+                    X_out[i, :] = self.hash_dict[x] = self.minhash(
                         x,
                         n_components=self.n_components,
                         ngram_range=self.ngram_range)
-                    X_out[i, :] = self.hash_dict[x]
                 else:
                     X_out[i, :] = self.hash_dict[x]
 
--- a/dirty_cat/minhash_encoder.py
+++ b/dirty_cat/minhash_encoder.py
@@ -24,10 +24,11 @@ from sklearn.cluster import KMeans
 from sklearn.feature_extraction.text import CountVectorizer, HashingVectorizer
 from sklearn.neighbors import NearestNeighbors
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.utils import check_random_state, murmurhash3_32 
+from sklearn.utils import check_random_state, murmurhash3_32
 
 from .fast_hash import ngram_min_hash
 from .utils import LRUDict
+
 
 class MinHashEncoder(BaseEstimator, TransformerMixin):
     """
@@ -48,21 +49,29 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
         might have some concern with its entropy.
     minmax_hash : bool, default=False
         if True, return min hash and max hash concatenated.
+        handle_missing : 'error' or '' (default)
+        Whether to raise an error or impute with blank string '' if missing
+        values (NaN) are present during fit (default is to impute).
+        When this parameter is set to '', and a missing value is encountered
+        during fit_transform, the resulting encoded columns for this feature
+        will be all zeros.
     References
     ----------
     For a detailed description of the method, see
     `Encoding high-cardinality string categorical variables
     <https://hal.inria.fr/hal-02171256v4>`_ by Cerda, Varoquaux (2019).
-    
+
     """
 
     def __init__(self, n_components, ngram_range=(2, 4),
-                 hashing='fast', minmax_hash=False):
+                 hashing='fast', minmax_hash=False,
+                 handle_missing=''):
         self.ngram_range = ngram_range
         self.n_components = n_components
         self.hashing = hashing
         self.minmax_hash = minmax_hash
         self.count = 0
+        self.handle_missing = handle_missing
 
     def get_unique_ngrams(self, string, ngram_range):
         """ Return the set of unique n-grams of a string.
@@ -113,11 +122,10 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
             min_hashes = np.minimum(min_hashes, hash_array)
         return min_hashes/(2**32-1)
 
-    def get_hash(self, string):
+    def get_fast_hash(self, string):
         """
-        Encode a string with murmur or fast hashing function.
-        fast hashing supports both min_hash and minmax_hash encoding,
-        whereas murmur only supports min_hash encoding.
+        Encode a string with fast hashing function.
+        fast hashing supports both min_hash and minmax_hash encoding.
         Parameters
         ----------
         string : str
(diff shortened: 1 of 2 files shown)


amy12xx:
Hi @GaelVaroquaux, here's an attempt at it. there's probably room for improvement. I used the profile_encoder() and main() in test to check the perf.

BEFORE:
------------------------------------------------------------
start test
test passed
time profile_encoder(MinHashEncoder, hashing=fast)
0.6618 seconds >> 0.6987 seconds >> 0.7474 seconds
time profile_encoder(MinHashEncoder, hashing=fast)
0.6814 seconds >> 0.9635 seconds
time profile_encoder(MinHashEncoder, hashing=fast)
0.6376 seconds >> 0.7556 seconds
time profile_encoder(MinHashEncoder, hashing=fast) with minmax
0.5987 seconds >> 0.5887 seconds
time profile_encoder(MinHashEncoder, hashing=fast) with minmax
0.5654 seconds >> 0.7014 seconds
time profile_encoder(MinHashEncoder, hashing=fast) with minmax
0.5932 seconds >> 0.6894 seconds
time profile_encoder(MinHashEncoder, hashing=murmur)
1.427 seconds >> 1.954
Done


AFTER:
------------------------------------------------------------
start test
test passed
time profile_encoder(MinHashEncoder, hashing=fast)
0.723 seconds
time profile_encoder(MinHashEncoder, hashing=fast)
0.691 seconds
time profile_encoder(MinHashEncoder, hashing=fast)
0.709 seconds
time profile_encoder(MinHashEncoder, hashing=fast) with minmax
0.643 seconds
time profile_encoder(MinHashEncoder, hashing=fast) with minmax
0.607 seconds
time profile_encoder(MinHashEncoder, hashing=fast) with minmax
0.6 seconds
time profile_encoder(MinHashEncoder, hashing=murmur)
1.506 seconds
Done

dirty_cat/minhash_encoder.py
```diff
@@ -127,27 +135,30 @@ def get_hash(self, string):
...
+        Parameters
+        ----------
+        string : str
+            The string to encode.
+        Returns
+        -------
+        array, shape (n_components, )
+            The encoded string, using specified encoding scheme.
+        """
+        return self.minhash(
```
GaelVaroquaux, line 159: When I see a function that small, I find that we have one level of indirection too much. Can we just inline the "minhash" method in here and remove it?

dirty_cat/minhash_encoder.py
```diff
@@ -184,14 +195,48 @@ def transform(self, X):
...
+                    "n_components should be even when minmax_hash=True"
+        if self.hashing == 'murmur':
+            assert not(self.minmax_hash),\
+                   "minmax_hash not implemented with murmur"
+        if self.handle_missing not in ['error', '']:
+            template = ("handle_missing should be either 'error' or "
+                        "'', got %s")
+            raise ValueError(template % self.handle_missing)
         X_out = np.zeros((len(X), self.n_components))
+        X = X.ravel()
```
GaelVaroquaux, line 209: This ravel is new, isn't it?

Why do we need it? I wonder if it won't have side effects.

dirty_cat/minhash_encoder.py
```diff
@@ -184,14 +195,48 @@ def transform(self, X):
...
-            if x not in self.hash_dict:
-                self.hash_dict[x] = self.get_hash(x)
+        nan_idx = []
+
+        if self.hashing == 'fast':
+            for i, x in enumerate(X):
+                if isinstance(x, float):
+                    nan_idx.append(i)
+                elif x not in self.hash_dict:
+                    self.hash_dict[x] = self.get_fast_hash(x)
```
GaelVaroquaux, line 219: I think that we need to populate X_out in this loop and remove the one below. The reason being that we can exceed the cache of hash_dict (a bug in the current code).

dirty_cat/test/test_minhash_encoder.py
```diff
@@ -56,25 +58,76 @@ def test_get_unique_ngrams():
...
+            else:
+                assert np.array_equal(y[1], z)
+                assert np.array_equal(y[-1], z)
+        else:
+            msg = "handle_missing should be either 'error' or '', got %s" % missing
+            assert_raise_message(ValueError, msg, encoder.fit_transform, X)
+            return
+
+
+def test_missing_values():
```
GaelVaroquaux, line 123: Maybe here we should be using pytest.parametrize:

https://docs.pytest.org/en/stable/parametrize.html

dirty_cat/minhash_encoder.py
amy12xx: The subsequent for loop check breaks if we have an array shape of (N, 1). Ravel turns it into a 1-D array. I can use np.resize instead.

dirty_cat/test/test_minhash_encoder.py
amy12xx: fixed

dirty_cat/minhash_encoder.py
amy12xx: fixed, had to add one more else condition.

dirty_cat/minhash_encoder.py
amy12xx: fixed

dirty_cat/minhash_encoder.py
GaelVaroquaux: OK. I worry about data being put in with a shape of (N, P), where the ravel could have "interesting" consequences. Maybe this cannot happen, and we can worry about it later, if we hit the problem.

dirty_cat/minhash_encoder.py
```diff
@@ -184,14 +178,53 @@ def transform(self, X):
...
+                    X_out[i, :] = self.hash_dict[x]
```
GaelVaroquaux, line 203: I think that the following would work:
```suggestion
                    X_out[i, :] = self.hash_dict[x] = self.get_fast_hash(x)
```
It should be faster, as it would avoid a value lookup in X_out, which is probably not completely cheap.

dirty_cat/minhash_encoder.py
```diff
@@ -184,14 +178,53 @@ def transform(self, X):
...
+        elif self.hashing == 'murmur':
+            for i, x in enumerate(X):
+                if isinstance(x, float):
+                    nan_idx.append(i)
+                elif x not in self.hash_dict:
+                    self.hash_dict[x] = self.minhash(
+                        x,
+                        n_components=self.n_components,
+                        ngram_range=self.ngram_range)
+                    X_out[i, :] = self.hash_dict[x]
```
GaelVaroquaux, line 216: Same comment about avoiding a look up into hash_dict

dirty_cat/test/test_minhash_encoder.py
```diff
@@ -1,5 +1,7 @@
 import numpy as np
 import time
+import pytest
+from sklearn.utils._testing import assert_raise_message
```
GaelVaroquaux, line 4: I worry that we are importing from a private submodule of sklearn, and this may break without warning.

I think that I would prefer that we use the pattern listed here:
https://docs.pytest.org/en/stable/assert.html#assertions-about-expected-exceptions

dirty_cat/test/test_minhash_encoder.py
```diff
@@ -56,25 +58,70 @@ def test_get_unique_ngrams():
...
     encoder.fit(X)
     y = encoder.transform(X)
     assert y.shape == (len(X), 50)
     eta = time.time() - t0
     return eta
 
 
+@pytest.mark.parametrize("input_type", ['list', 'numpy', 'pandas'])
+@pytest.mark.parametrize("missing", ['aaa', 'error', ''])
+@pytest.mark.parametrize("hashing", ["fast", "murmur"])
```
GaelVaroquaux, line 84: Sorry to nitpick, but the tensor-product of options here leads to a expensive combinatorial: this test is taking more time than it should. Could would downsample here? Some think like:
```
@pytest.mark.parametrize("input_type,missing,hashing", [
                                               ['list', 'aaa', 'fast'],
                                               ...
                                           ])
```
for a few well-chosen values

dirty_cat/minhash_encoder.py
amy12xx: 👍

dirty_cat/minhash_encoder.py
amy12xx: So, we have this on line 175 to assert a shape of (N, ) or (N, 1). At this time a resize later on, can assume this shape, in which case I think it should not be a problem? I used resize() instead in case ravel() is newer. 

```
assert X.ndim == 1 or (X.ndim == 2 and X.shape[1] == 1), f"ERROR:\
        shape {X.shape} of input array is not supported."
```

GaelVaroquaux:
This is great! Merging! Thank you!!

PR #133, opened 2020-08-18
Fix: Minhash encoder does not perform well on the dirty-cat examples

### LINKED ISSUE #127: Minhash encoder does not perform well on the dirty-cat examples
To solve this, the parameter `n_components` of the MinHashEncoder should be set to 30 at least.

TwsThomas:
Fix #127  

In example 02, puting `n_components` = 300 gives results as good as one_hot_encoder (.85) but it take more time.
I put  `n_components` = 100 to have decent results.

### DIFF
--- a/dirty_cat/minhash_encoder.py
+++ b/dirty_cat/minhash_encoder.py
@@ -38,7 +38,8 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
     Parameters
     ----------
     n_components : int
-        The number of dimension of encoded strings.
+        The number of dimension of encoded strings. Numbers around 300 tend to
+        lead to good prediction performance, but with more computational cost.
     ngram_range : tuple (min_n, max_n), default=(2, 4)
         The lower and upper boundary of the range of n-values for different
         n-grams to be extracted. All values of n such that min_n <= n <= max_n.
--- a/dirty_cat/minhash_encoder.py
+++ b/dirty_cat/minhash_encoder.py
@@ -57,7 +57,7 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
     
     """
 
-    def __init__(self, n_components, ngram_range=(2, 4),
+    def __init__(self, n_components=30, ngram_range=(2, 4),
                  hashing='fast', minmax_hash=False):
         self.ngram_range = ngram_range
         self.n_components = n_components
--- a/examples/02_fit_predict_plot_employee_salaries.py
+++ b/examples/02_fit_predict_plot_employee_salaries.py
@@ -73,7 +73,7 @@ encoders_dict = {
     'one-hot': OneHotEncoder(handle_unknown='ignore', sparse=False),
     'similarity': SimilarityEncoder(similarity='ngram'),
     'target': TargetEncoder(handle_unknown='ignore'),
-    'minhash': MinHashEncoder(n_components=10, ngram_range=(2, 4),
+    'minhash': MinHashEncoder(n_components=100, ngram_range=(2, 4),
                               hashing='fast', minmax_hash=False),
     'numerical': FunctionTransformer(None)}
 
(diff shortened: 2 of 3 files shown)


examples/03_fit_predict_plot_midwest_survey.py
```diff
@@ -69,7 +69,7 @@
 encoder_dict = {
     'one-hot': OneHotEncoder(handle_unknown='ignore', sparse=False),
     'similarity': SimilarityEncoder(similarity='ngram'),
-    'minhash': MinHashEncoder(n_components=10, ngram_range=(2, 4),
+    'minhash': MinHashEncoder(n_components=30, ngram_range=(2, 4),
```
pcerda, line 72: Why not n_components=100 here too ?

TwsThomas: Well, going to 100 only improves from 0.676 to 0.678.
Both are similar to the similarity encoding score (0.674).

pcerda: ok, thanks for the info

GaelVaroquaux:
Sounds good!!

Maybe we should put a discussion somewhere that the number of components
is a tradeoff: numbers around 300 tend to lead to good prediction
performance, but with more computational cost.

This could go at least in the docstring for the corresponding parameter,
but maybe also in the example where we use a large number of components.

examples/02_fit_predict_plot_employee_salaries.py
```diff
@@ -73,7 +73,7 @@
     'one-hot': OneHotEncoder(handle_unknown='ignore', sparse=False),
     'similarity': SimilarityEncoder(similarity='ngram'),
     'target': TargetEncoder(handle_unknown='ignore'),
-    'minhash': MinHashEncoder(n_components=10, ngram_range=(2, 4),
+    'minhash': MinHashEncoder(n_components=100, ngram_range=(2, 4),
```
GaelVaroquaux, line 76:
```suggestion
    'minhash': MinHashEncoder(n_components=100,
```
If I am not wrong, this is the default parameter, so we can remove it (as well as the ones after), with the goal to make the example simpler.

examples/02_fit_predict_plot_employee_salaries.py
```diff
@@ -130,11 +130,11 @@ def make_pipeline(encoding_method):
...
 # Finally, we plot the scores on a boxplot:
-# We notice that the MinHashEncoder performs poorly compared to other encoding
-# methods. There are two reasons for that: the MinHashEncoder performs better
-# with tree-based models than linear models (see example 03), and the
-# low-dimensionality of encodings (increasing n_components improves
-# performances.
+# We notice that the MinHashEncoder does not performs as well compared to 
+# other encoding methods.
+# There are two reasons for that: the MinHashEncoder performs better
+# with tree-based models than linear models (see example 03), and also
```
GaelVaroquaux, line 136: We could use a `:ref:` syntax in to link to example 03.

examples/03_fit_predict_plot_midwest_survey.py
GaelVaroquaux, line 72: Same comment on the parameters that take their default values.

Also, in the light of this PR, I suggest to put a default value for n_components, and set it at 30.

dirty_cat/minhash_encoder.py
```diff
@@ -38,7 +38,8 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
     Parameters
     ----------
     n_components : int
```
GaelVaroquaux, line 40:
```suggestion
    n_components : int, default=30
```

examples/02_fit_predict_plot_employee_salaries.py
```diff
@@ -130,11 +129,15 @@ def make_pipeline(encoding_method):
...
-# We notice that the MinHashEncoder performs poorly compared to other encoding
-# methods. There are two reasons for that: the MinHashEncoder performs better
-# with tree-based models than linear models (see example 03), and the
-# low-dimensionality of encodings (increasing n_components improves
-# performances.
+# We notice that the MinHashEncoder does not performs as well compared to 
+# other encoding methods.
+# There are two reasons for that: the MinHashEncoder performs better
+# with tree-based models than linear models (
+# :ref:`see example 03<sphx_glr_auto_examples_03_fit_predict_plot_midwest_survey.html#sphx-glr-auto-examples-03-fit-predict-plot-midwest-survey-py>`)
```
GaelVaroquaux, line 136: The pointer to the ref is actually not the name of the file. You need to look at the source code of the generated file to figure it out. I believe that it is `sphx_glr_auto_examples_03_fit_predict_plot_midwest_survey.py`

TwsThomas: I did not find any occurrence of "sphx_glr" in the source code :/ 

It look likes the :ref: in exemple 05 is also broken 
https://github.com/dirty-cat/dirty_cat/blob/47c5b7650c92f292234dd79cbdc30997fe31dc70/examples/05_scaling_non_linear_models.py#L55

GaelVaroquaux:
Great! Merging. Thanks!!

PR #135, opened 2020-09-01
replace sklearn assert function with pytest and changes.rst

amy12xx:
Changes:
- Replaced sklearn's assert_raise_message() with pytest raises in target and similarity encoder.
- update to changes.rst for minhash handle_missing functionality.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,9 @@
+Release 0.0.8
+=============
+
+* **MinHashEncoder**: Added a "handle_missing" attribute to allow encoding
+  with missing values.
+
 Release 0.0.7
 =============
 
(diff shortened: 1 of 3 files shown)


CHANGES.rst
```diff
@@ -1,3 +1,9 @@
+Release 0.0.8
+=============
+
+* **MinHashEncoder**: Added a "handle_missing" attribute to allow encoding
+  with missing values.
+
```
GaelVaroquaux, line 6: I think that I released 0.0.7 after you added this, hence it should be added to the 0.0.7.

GaelVaroquaux (commented):
Thanks a lot. A tiny comment, but aside from that it's a great addition and it's ready to merge.

CHANGES.rst
amy12xx: Yes, I wasn't too sure about that. I've fixed it.

GaelVaroquaux (approved):
LGTM. Merging. Thanks a lot!!

Issue #140, opened 2020-10-04
Warning messages in notebook 04_dimension_reduction_and_performance


amy12xx:
Too many ConvergenceWarning messages in the current notebook make it hard to see the actual output. 
https://dirty-cat.github.io/stable/auto_examples/04_dimension_reduction_and_performance.html#sphx-glr-auto-examples-04-dimension-reduction-and-performance-py

Phaide:
Hello @amy12xx,

This is indeed annoying.
One workaround is to use the "max_iter" argument during the LogisticRegression instantiation.
I've submitted PR #141, applying that method. 
Best regards

GaelVaroquaux:
Fixed in #141

PR #141, opened 2020-10-07
Minor modification to avoid ConvergenceWarnings in Notebook 04.

Phaide:
Added max_iter argument to the LogisticRegression instantiation in Notebook 04.
Solves "Warning messages in notebook 04_dimension_reduction_and_performance #140".

Source: https://stackoverflow.com/a/62113936
By the way, as mentioned in the answer:
> You may also standardize your data as the warning said, with `sklearn.preprocessing.scale()`.

### DIFF
--- a/examples/04_dimension_reduction_and_performance.py
+++ b/examples/04_dimension_reduction_and_performance.py
@@ -116,7 +116,7 @@ print('Time to vectorize: %s' % (t1 - t0))
 # We can run a cross-validation
 from sklearn import linear_model, pipeline, model_selection
 
-# We add max_iter to avoid ConvergenceWarning
+# We specify max_iter to avoid convergence warnings
 log_reg = linear_model.LogisticRegression(max_iter=10000)
 
 model = pipeline.make_pipeline(column_trans, log_reg)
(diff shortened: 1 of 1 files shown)


examples/04_dimension_reduction_and_performance.py
```diff
@@ -116,7 +116,8 @@ def wrapped_func(*args, **kwargs):
 # We can run a cross-validation
 from sklearn import linear_model, pipeline, model_selection
 
-log_reg = linear_model.LogisticRegression()
+# We add max_iter to avoid ConvergenceWarning
```
GaelVaroquaux, line 119:
```suggestion
# We specify max_iter to avoid convergence warnings
```
I'd rather write "convergence warnings" then "ConvergenceWarning" to facilitate the understanding by non technical users.

amy12xx:
aside from comment that @GaelVaroquaux left, LGTM. Thanks @Phaide

amy12xx:
@Phaide were you able to download the dataset, i.e. execute this line:
```
data = fetch_traffic_violations()
```
It seem to time out for me, and I see that the example notebook also displays the same warning message.

Phaide:
Yes, the first time I launched this line it issued a warning and remained stuck for about a minute, then proceeded.
Since then I can load it on the fly.

FYI:

image

The function is in dirty_cat/datasets/fetching.py, and it does download the file from the web, and stores it under [Python_packages_path]/dirty_cat/data/traffic_violations/rows.csv

Deleting the file triggers the download again when calling the function.

amy12xx:
Thanks, I was able to download it, just took a while.

GaelVaroquaux:
OK, the CI is green. Merging!!

PR #144, opened 2020-10-11
[MRG] remove race and gender attributes from example 04_dimension_reduction_and_performance

### LINKED ISSUE #143: Example 04 uses race and gender attributes
I know the purpose of the example is on performance, but we are currently illustrating it on the traffic violations dataset, and using race and gender attributes while building the model. It may be better to remove these two attributes (race, and gender) from the modeling.

amy12xx:
Fixes #143


GaelVaroquaux:
Looks good to me. I'm merging.

Unrelated: it seems that the example is no longer showing what we would like to show: the SimilarityEncoder is fast even with the default options. I don't know why this is the case... We should probably have a deeper look when we have time.

PR #147, opened 2020-10-23
Refactor of ``fetching.py`` to use exclusively OpenML.org's datasets

Phaide:
Hello,
This is a major refactor of the file ``fetching.py``, that is in charge of fetching datasets off the web.
Historically, they were fetched from various websites.

Our aim with this update is to only use OpenML.org's API, through Scikit-learn's [``fetch_openml()``](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_openml.html) function.
This allows us to have a much more reliable and unified interface, and avoids losing access to datasets due to deletion, renaming, etc.

For instance, with the current system, 4 on the 7 datasets are unavailable (403 Access Denied, website down, etc.).

The user-interface stays the same with the functions ``fetch_*()`` (e.g ``fetch_open_payments()``), still returning a similar dictionary.

As I'm writing this, not all the datasets have been uploaded to OpenML ; I will take care of this in the coming week(s).

Now unecessary code has been removed from ``utils.py`` files.

*Note on the testing part: this is my first time experimenting with mocking !*


dirty_cat/datasets/fetching.py
```diff
@@ -56,24 +55,24 @@ def fetch_openml_dataset(dataset_id: int, data_directory: str = get_data_dir())
 
     Parameters
     ----------
-    dataset_id
+    dataset_id: int
         The ID of the dataset to fetch.
-    data_directory
-        Optional. A directory to save the data to.
```
GaelVaroquaux: It's still optional, isn't it?

dirty_cat/datasets/fetching.py
```diff
@@ -292,36 +299,36 @@ def _features_to_csv_format(features: Features) -> str:
...
 
 
-def fetch_traffic_violations():
+def fetch_traffic_violations() -> dict:
     """Fetches the traffic violations dataset."""
     return fetch_openml_dataset(dataset_id=TRAFFIC_VIOLATIONS_ID)
 
 
-def fetch_drug_directory():
+def fetch_drug_directory() -> dict:
```
GaelVaroquaux, line 332: Small comment: refrain from doing changes not related to the purpose of a PR. They make the PR harder to review.

You can make those changes in another PR (these changes are indeed useful).

dirty_cat/datasets/fetching.py
```diff
@@ -128,9 +127,13 @@ def _download_and_write_openml_dataset(dataset_id: int, data_directory: str) ->
     ----------
     dataset_id: int
         The ID of the dataset to download.
+    data_directory: str
+        The directory in which the data will be saved.
 
     """
-    # The ``fetch_openml()`` function returns a scikit-learn Bunch object,
+    from sklearn.datasets import fetch_openml
```
GaelVaroquaux, line 134: Is there a reason why you moved the import locally?

Typically, local imports are frowned upon. If there is a good reason, indicate it here with a comment.

GaelVaroquaux (commented):
This looks great! I'll let you improve a bit the coverage before considering merging.

One thing that you could do, is add an example that downloads all the data, to act as a test (circleCI, which runs the examples, has access to internet). The example could load each of the CSVs and call ".head()" to display it.

dirty_cat/datasets/fetching.py
Phaide: You're absolutely right, however, importing the function ``fetch_openml`` locally is the only way I found to be able to mock it in the test file.

When importing the function in the file header (like I did in the first commit), the test ``test__download_and_write_openml_dataset()`` raises an AssertionError at ``mock_fetch_openml.assert_called_once_with(data_id=1, data_home=test_data_dir)`` (line 149).

I moved the import back in the header and launching the test gives this screen :

Screenshot from 2020-10-29 10-38-37

Perhaps there is another mocking method fitting the case.

PS: the comment ``# The following returns "Called 0 times" although I'm pretty sure it does get called...`` line 148 is outdated.

dirty_cat/datasets/fetching.py
Phaide: Yes, my bad.

Phaide:
Hello,

The code coverage for the folder is [maxed out](https://codecov.io/gh/dirty-cat/dirty_cat/compare/484f1eb499a87f0cdc9986e31deadf2ae0ae9f23...1b6c9af91b9f469db20d015e6474921faecdd97a/tree/dirty_cat/datasets).
I suspect that considering more lines were deleted than added, the overall coverage for the project lowered.

I'm waiting for #149 to be resolved.

Phaide:
All datasets were uploaded to OpenML, except the full open_payments: it is about 6 GB, but the maximum size of a dataset on OpenML is 2 GB.

They plan on upgrading this limit later on. Meanwhile, we will have to work with this one.

I'll upload a new notebook presenting the new interface asap, and we'll be able to merge.

Phaide:
Regarding the CI error, it is linked to scikit-learn's ``fetch_openml()`` function, which needs pandas to work (and therefore fails on the worker without pandas).

[From the doc](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_openml.html):
image

I don't see an imediate workaround or solution to this problem.

One solution that springs to mind would be to catch the import error, and require pandas being installed for retreiving datasets from OpenML, essentially making pandas a requirement for the example datasets fetching part.

What do you think ?

GaelVaroquaux:
Open a bug on scikit-learn

Phaide:
Actually, I don't think it is a problem coming from their end: when using ``as_frame=True`` it is implied that pandas is required (as it returns a pandas dataframe) ; this is the reason of the CI error.

``as_frame=False`` on the other end will return an error when trying to fetch (so not related to pandas):

Full traceback:
```
C:\Users\Phaide\Dropbox\Github\dirty_cat\dirty_cat\datasets\fetching.py:86: UserWarning: Could not find the dataset locally. Downloading it from OpenML... This might take a while.If the process is interrupted, some files will be invalid/incomplete.To fix this problem, delete the CSV file if it exists. The system will recreate it on the next run.
  warnings.warn(
  File "C:/Users/Phaide/Dropbox/Github/dirty_cat/dirty_cat/datasets/fetching.py", line 17, in <module>
    from dirty_cat.datasets.utils import get_data_dir
  File "C:\Users\Phaide\Dropbox\Github\dirty_cat\dirty_cat\datasets\__init__.py", line 5, in <module>
    from .fetching import fetch_midwest_survey, fetch_employee_salaries, \
  File "C:\Users\Phaide\Dropbox\Github\dirty_cat\dirty_cat\datasets\fetching.py", line 333, in <module>
    df_info = fetch_drug_directory()
  File "C:\Users\Phaide\Dropbox\Github\dirty_cat\dirty_cat\datasets\fetching.py", line 330, in fetch_drug_directory
    return fetch_openml_dataset(dataset_id=DRUG_DIRECTORY_ID)
  File "C:\Users\Phaide\Dropbox\Github\dirty_cat\dirty_cat\datasets\fetching.py", line 91, in fetch_openml_dataset
    _download_and_write_openml_dataset(dataset_id=dataset_id, data_directory=data_directory)
  File "C:\Users\Phaide\Dropbox\Github\dirty_cat\dirty_cat\datasets\fetching.py", line 143, in _download_and_write_openml_dataset
    fetch_openml(data_id=dataset_id, data_home=data_directory, as_frame=False)
  File "C:\Users\Phaide\AppData\Local\Programs\Python\Python38\lib\site-packages\sklearn\utils\validation.py", line 72, in inner_f
    return f(**kwargs)
  File "C:\Users\Phaide\AppData\Local\Programs\Python\Python38\lib\site-packages\sklearn\datasets\_openml.py", line 782, in fetch_openml
    raise ValueError('STRING attributes are not supported for '
ValueError: STRING attributes are not supported for array representation. Try as_frame=True
```

but I think it's rather a problem from the dataset we are trying to fetch (in this case the [drug directory](https://www.openml.org/d/42787)) than the API that's broken.

Finally, leaving ``as_frame`` unspecified is a mix of both, and doesn't work either (as illustrated by the CI error).

Phaide:
Hi

I messed up my fork, so I created a new PR: #155 

See you there :)

Issue #146, opened 2020-10-19
Apply minhash_encoder to more than 1024 categories returns -1


jp-varela:
Hy all,
I am trying to apply minhash_encoder to a somewhat large dataset of strings (~200k distinct).
I was testing my code with 10 strings, and it was running fine.
But when I tested using all dataset, most of the strings were represented as all '-1' vectors.
I took a look at the source code and find this line inside 'minhash_encoder.py', that maybe is causing the problem:
`self.hash_dict = LRUDict(capacity=2**10)`
Not sure why this is used, but I checked with 1025 strings, and only the first one returns -1.
This encoder should work with a lot more variables, right?

Code to replicate:
```
from dirty_cat import MinHashEncoder
import random
import string

def get_random_string(length):
    letters = string.ascii_lowercase
    result_str = ''.join(random.choice(letters) for i in range(length))
    return result_str

# 1024 categories -> all ok
raw_data = [get_random_string(10) for x in range(1024)]
hash_encoder = MinHashEncoder(n_components=10)
transformed_values = hash_encoder.fit_transform(raw_data)
print(transformed_values)

# 1025 categories -> first represented as -1's
raw_data = [get_random_string(10) for x in range(1025)]
hash_encoder = MinHashEncoder(n_components=10)
transformed_values = hash_encoder.fit_transform(raw_data)
print(transformed_values)
```

GaelVaroquaux:
It seems that we indeed have a bug. I thought that the latest changes to
the MinHashEncoder made it robust to cache overflow, but it seems not.
We'll look at this ASAP.

Can you put here the code that you used to demo the bug, it will save us
time. Thank you!

amy12xx:
I tried to replicate using the employee dataset, and using a small cache capacity (=10), but wasn't able to replicate it. An example would indeed be helpful.

jp-varela:
Just edited the issue with code to replicate it.

jp-varela:
@amy12xx I am using version 0.0.7

amy12xx:
Was not able to replicate this on the latest, I am using windows. @Phaide / @GaelVaroquaux may have some luck?

Phaide:
Hi all,
I tried on both Windows 10 and Ubuntu 20.04, and indeed, I get the first line as only -1s using the 1025 strings.
I'm using version 0.0.7 as well.

amy12xx:
[...]

PR #148, opened 2020-10-26
regression test for cache overflow

amy12xx:
Regression test for cache overflow problem (rows are encoded with -1s). 
Fixes: #146

### DIFF
--- a/dirty_cat/minhash_encoder.py
+++ b/dirty_cat/minhash_encoder.py
@@ -73,7 +73,7 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
         self.minmax_hash = minmax_hash
         self.count = 0
         self.handle_missing = handle_missing
-        self.capacity = 2**10
+        self._capacity = 2**10
 
     def get_unique_ngrams(self, string, ngram_range):
         """ Return the set of unique n-grams of a string.
@@ -159,7 +159,7 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
         self
             The fitted MinHashEncoder instance.
         """
-        self.hash_dict = LRUDict(capacity=self.capacity)
+        self.hash_dict = LRUDict(capacity=self._capacity)
         return self
 
     def transform(self, X):
--- a/dirty_cat/minhash_encoder.py
+++ b/dirty_cat/minhash_encoder.py
@@ -73,6 +73,7 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
         self.minmax_hash = minmax_hash
         self.count = 0
         self.handle_missing = handle_missing
+        self.capacity = 2**10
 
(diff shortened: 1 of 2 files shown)


Phaide (approved):
Looks good to me !

dirty_cat/minhash_encoder.py
```diff
@@ -73,6 +73,7 @@ def __init__(self, n_components=30, ngram_range=(2, 4),
         self.minmax_hash = minmax_hash
         self.count = 0
         self.handle_missing = handle_missing
+        self.capacity = 2**10
```
GaelVaroquaux, line 76: Given that this parameter is not in the init and not documented, I wonder whether we should not make it private: "_capacity". WDYT?

dirty_cat/test/test_minhash_encoder.py
```diff
@@ -124,6 +126,21 @@ def test_missing_values(input_type, missing, hashing):
...
+def test_cache_overflow():
+    # Regression test for cache overflow resulting in -1s in encoding
+    def get_random_string(length):
+        letters = ascii_lowercase
+        result_str = ''.join(random.choice(letters) for i in range(length))
+        return result_str
+
+    encoder = MinHashEncoder(n_components=3)
+    capacity = encoder.capacity
+    raw_data = [get_random_string(10) for x in range(capacity)]
```
GaelVaroquaux, line 138: Should it be "capacity", or "capacity + 1"? A simple way of figuring this would be to see if this code fails with the old version of dirty_cat, where there was a bug.

GaelVaroquaux (commented):
Overall looks great. I do have two questions (as comments in the code).

Thanks!!

dirty_cat/test/test_minhash_encoder.py
Phaide: On release 0.0.7, I confirm using ``capacity + 1`` returns some ``-1``s.

image

The same assertions launched from master returns no ``-1``s.

image

So I assume we should add this second assertion (with ``== 0`` instead of ``== 1``).

dirty_cat/minhash_encoder.py
amy12xx: That makes sense. Changed it to _capacity.

dirty_cat/test/test_minhash_encoder.py
amy12xx: Good catch, thanks. Had tested on old version, but missed in the test. Should be fixed now.

amy12xx:
@GaelVaroquaux @Phaide  thanks for your comments. Have made the changes. Let me know if you find anything else amiss.

Phaide (approved):
LGTM ! Thanks @amy12xx !

GaelVaroquaux:
Thanks!! Merging.

PR #151, opened 2020-11-19
Fix multiprocessing Runtime Exception

### LINKED ISSUE #145: 04_dimension_reduction_and_performance throws a multiprocessing exception on windows
This is due to the resource_used function used in this line:
results = resource_used(model_selection.cross_validate)(model, df, y, )

This can be fixed by having this inside the if __name__ == '__main__': block. I tested this on windows, and can confirm that doing this fixes the problem.

See here: https://stackoverflow.com/questions/18204782/runtimeerror-on-windows-trying-python-multiprocessing 

Details of exception raised:
File "C:\Users\Amanda\Miniconda3\envs\dirtycat\lib\multiprocessing\spawn.py", line 134, in _check_not_importing_main
    raise RuntimeError('''
RuntimeError: 
        An attempt has been made to start a new process before the
        current process has finished its bootstrapping phase.

        This probably means that you are not using fork to start your
        child processes and you have forgotten to use the proper idiom
        in the main module:

            if __name__ == '__main__':
                freeze_support()
                ...

        The "freeze_support()" line can be omitted if the program
        is not going to be frozen to produce an executable.

Phaide:
Wrapped everything inside ``if __name__ == "__main__":`` ; this should do the trick.
Fixes #145.


GaelVaroquaux:
I think that you have mixed commits in this branch coming from another PR. You should create a new branch and cherry-pick the relevant commit.

PR #152, opened 2020-11-19
Fix multiprocessing Runtime Error

### LINKED ISSUE #145: 04_dimension_reduction_and_performance throws a multiprocessing exception on windows

Phaide:
Wrapped everything in ``if __name__ == "__main__":`` ; this should do the trick.
Fixes #145


GaelVaroquaux:
I hate to say, given that I was the one who pointed in this direction, but that's quite horrible. It's going to break the notebooks. I was hoping that we could use the if name == main on only a small part of the script.

I thought that joblib isolated from this problem. But it's maybe created by the "resource_used" call.

If you remove the resource_used, does the problem still happen under windows? I'm asking this question not as a suggestion to remove the resource_used permanently, but rather as a debugging investigation.

GaelVaroquaux:
The rendered notebook can be seen here
https://670-124964482-gh.circle-artifacts.com/0/doc/auto_examples/04_dimension_reduction_and_performance.html
(thanks for the artifact redirector, that was just fixed!!)

It is indeed quite broken

Phaide:
> I hate to say, given that I was the one who pointed in this direction, but that's quite horrible. It's going to break the notebooks. I was hoping that we could use the if name == main on only a small part of the script.
> 
> I thought that joblib isolated from this problem. But it's maybe created by the "resource_used" call.
> 
> If you remove the resource_used, does the problem still happen under windows? I'm asking this question not as a suggestion to remove the resource_used permanently, but rather as a debugging investigation.

I tried my chance with the simple solution, obviously that was *too* easy !

I removed the ``resource_used`` decorator, and unsurprisingly it works without.

I'll try to have the faulty blocks only in ``if __name__ ...`` blocks (I did earlier and it gave some questionable results).

Phaide:
Just tested having all the ``resource_used`` calls in ``if __name__ == "__main__":`` blocks ; it is still causing some troubles: the code gets executed 4 times.
I also tried to add the block directly in the decorator, in which case it is executed twice before crashing :persevere:

GaelVaroquaux:
I have an idea: try implementing the decorator using the pattern explained here: http://gael-varoquaux.info/programming/decoration-in-python-done-right-decorating-and-pickling.html

Indeed, I am wondering whether the underlying problem is not one of pickling.

GaelVaroquaux:
Thinking about it, I wonder if the memory_profiler function that we are using works under windows. Can you check?

Another, possibly cleaner, option would be to avoid decoration, using another memory_profiler function and a context manager https://docs.python.org/3/library/contextlib.html

amy12xx:
It's a good question. I was looking for an alternate, and found the tracemalloc library which is part of the python library, and in particular https://docs.python.org/3/library/tracemalloc.html#tracemalloc.get_traced_memory ...but I found that the results are vastly different from what I get using memory_profiler, on windows. tracemalloc works on windows without enclosing in main().

Small example to see the difference between the two on Windows:
```
import numpy as np
import memory_profiler
import tracemalloc

def func():
    return np.ones((1000,1000))

memory_profiler.memory_usage(func, max_usage=True)
tracemalloc.start()
out = func()
first_size, first_peak = tracemalloc.get_traced_memory()
tracemalloc.stop()
first_peak = first_peak/(1024**2)
first_peak
```

This will print:
51.38671875 Mb(memory_profiler)
7.681624412 Mb(tracemalloc)

Phaide:
Hello

> I have an idea: try implementing the decorator using the pattern explained here: http://gael-varoquaux.info/programming/decoration-in-python-done-right-decorating-and-pickling.html
> 
> Indeed, I am wondering whether the underlying problem is not one of pickling.

Do you mean using ``functools.wraps()`` ?
Because it actually already is in use :)

> Thinking about it, I wonder if the memory_profiler function that we are using works under windows. Can you check?

According to its author, yes.
https://pypi.org/project/memory-profiler/
https://stackoverflow.com/a/10592072
I did some tests and it does work.

> It's a good question. I was looking for an alternate, and found the tracemalloc library which is part of the python library

Great finding !
I'm personally all for using the standard library as much as possible :)
I implemented ``tracemalloc`` in the notebook, following your example, and it works like a charm !
I want to make a few comparisons with other profiling tools and wait for @GaelVaroquaux 's opinion before submitting a PR.

I'll keep you updated

GaelVaroquaux:
> I want to make a few comparisons with other profiling tools and wait for @GaelVaroquaux 's opinion before submitting a PR.

Submit a PR, even if it's a draft. That way, we can have a look at it and
see if it's promising.

Phaide:
There you go !
Sorry for the horrible output in the visualization ; the only thing that changed is the decorator.

> I want to make a few comparisons with other profiling tools [...]

I did using [guppy3](https://pypi.org/project/guppy3/) and [memory_profiler](https://pypi.org/project/memory-profiler/), and while they all gave different values, tracemalloc's output was the closest from the system's (Windows').

GaelVaroquaux:
The circleCI runner didn't run :(.

GaelVaroquaux:
Also, I don't know what happened, but the diff is unmanageable: the whole file is displayed as changed, and not only the relevant lines.

Maybe a change in line endings? I'm not sure. Anyhow, we need to get at the root of this, and fix it, because having changes everywhere makes it hard to follow changes, will break history, and risks creating merge conflicts.

Phaide:
Hi,

I messed up my fork, so I made a new PR: #154 

See you there :)

PR #153, opened 2020-12-03
Adding Gamma poisson factorization

alexis-cvetkov:
Changes:

- Added `gamma_poisson_factorization.py` which implements online Gamma-Poisson factorization for encoding string variables.
- Added the corresponding tests in`test_gamma_poisson_factorization`.
- Modified examples 02 and 03 to include this method.
- Updated `CHANGES.rst` and `index.rst` to describe this new method.


GaelVaroquaux:
Wohoo!!

@Phaide : can you have a look when you have time?

Have a look at the code, the test (it's OK if you don't understand everything, this is a sophisticated mathematical algorithm. Check that you understand how to use this, that it feels natural, that the examples make sense, that it's well referenced in the docs (we should point from SimilarityEncoder to this for large datasets). Thanks!!!!

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,543 @@
...
+        Minimum number of iterations on the input data.
+
+    max_iter : int, default=5
+        Maximum number of iterations on the input data.
+
+    ngram_range : tuple, default=(2, 4)
+        The range of ngram length that will be used to build the
+        bag-of-n-grams representation of the input data.
+
+    analizer : str, default='char'.
```
Phaide, line 84: Wouldn't it be ``analyzer`` instead of ``analizer`` ?

alexis-cvetkov: You're right, I'll fix this !

alexis-cvetkov:
Following the error on travis, now I check for the sklearn version when importing the "k-means++" init method from `sklean.cluster.k_means_`, since its name has changed in the development version.

Phaide (commented):
Very good !
Thanks @alexis-cvetkov, good work !

The code is clean and fairly documented (though a few more inline comments to explain complex operations would be welcome!).
One last request: as mentioned by Gaël, could you please consider adding a little notice in the ``SimilarityEncoder`` docstring ?
Thanks again !

amy12xx:
Great to see this addition. I look forward to going through the code when I have a bit more time.

alexis-cvetkov:
Thank you for the review @Phaide, I have added more documentation in the code for gamma-poisson factorization, and a little docstring in `similarity_encoder.py`.
This PR looks ready for merge to me now !

Phaide (approved):
Looks good to me !
@amy12xx could you please have a quick look before merging ?
Thanks !

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
 Release 0.0.7
 =============
 
+* **OnlineGammaPoissonFactorization**: Added ``gamma_poisson_factorzation.py``
+  file that implement online Gamma-Poisson factorization through the
+  ``OnlineGammaPoissonFactorization`` class. This method allows for
+  intrerpretable encoding of string categorical variables, as it factorizes
```
amy12xx, line 7: Typo in intrerpretable

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+Online Gamma-Poisson factorization of string arrays.
+The principle is as follows:
+    1. Given an input string array X, we build its bag-of-n-grams
+       representation V (n_samples, vocab_size).
+    2. Instead of using the n-grams counts as encodings, we look for low-
+       dimensional representations by modeling n-grams counts as linear
+       combinations of topics V = HW, with W (n_topics, vocab_size) the topics
+       and H (n_samples, n_topics) the associated activations.
+    3. Assuming that n-grams counts follow a Poisson law, we fit H and W to
+       maximize the likelihood of the data, with a Gamma prior for the
```
amy12xx, line 11: Is it better to say " to maximize their likelihood on the data" instead?

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+       activations H to induce sparsity.
+    4. In practice, this is equivalent to a non-negative matrix factorization
+       with the Kullback-Leibler divergence as loss, and a Gamma prior on H.
+       We thus optimize H and W with the multiplicative update method.
+"""
+import numpy as np
+from distutils.version import LooseVersion
+from scipy import sparse
+import sklearn
```
amy12xx, line 20: It may be better to only include the modules we need instead of import sklearn

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+    gamma_shape_prior : float, default=1.1
+        Shape parameter for the Gamma prior distribution.
+
+    gamma_scale_prior : float, default=1.0
+        Scale parameter for the Gamma prior distribution.
+
+    rho : float, default=0.95
+        Weight parameter for the update of the W matrix.
+    
+    rescale_rho : boolean, default=False
```
amy12xx, line 64: If we are keeping to sklearn's naming conventions, then we can use bool instead of boolean.

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+        the corresponding bag-of-n-grams representation h.
+
+        Parameters
+        ----------
+        X : array-like, shape (n_samples, )
+            The string data to fit the model on.
+        H : array-like, shape (n_samples, n_vocab)
+            The corresponding bag-of-n-grams representations.
+        """
+        for x, h in zip(X, H):
```
amy12xx, line 186: Using a list comprehension will be faster.

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+                self.vocabulary = np.concatenate(
+                    (self.vocabulary, self.word_count.get_feature_names()))
+
+        _, self.n_vocab = unq_V.shape
+        # Init the topics W given the n-grams counts V
+        self.W_, self.A_, self.B_ = self._init_w(unq_V[lookup], X)
+        # Init the activations unq_H of each unique input string
+        unq_H = _rescale_h(unq_V, np.ones((len(unq_X), self.n_topics)))
+        # Map unique input strings to their activations
+        self.H_dict = dict()
```
amy12xx, line 213: Should we initialize in the init method instead?

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+        self._update_H_dict(unq_X, unq_H)
+        if self.rescale_rho:
+            # Make update rate per iteration independant of the batch_size
+            self.rho_ = self.rho ** (self.batch_size / len(X))
+        else:
```
amy12xx, line 218: I guess we can do away with the else block

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+        # Check if first item has str or np.str_ type
+        assert isinstance(X[0], str), "ERROR: Input data is not string."
+        # Make n-grams counts matrix unq_V
+        unq_X, unq_V, lookup = self._init_vars(X)
+        n_batch = (len(X) - 1) // self.batch_size + 1
+        del X
+        # Get activations unq_H
+        unq_H = self._get_H(unq_X)
+
+        for iter in range(self.max_iter):
```
amy12xx, line 318: We can use a different name since iter is a python keyword.

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+                epsilon=1e-3, max_iter=100,
+                rescale_W=self.rescale_W,
+                gamma_shape_prior=self.gamma_shape_prior,
+                gamma_scale_prior=self.gamma_scale_prior)
+        # Store and return the encoded vectors of X
+        self._update_H_dict(unq_X, unq_H)
+        return self._get_H(X)
+
+
+def _rescale_W(W, A, B):
```
amy12xx, line 504: Is there a reason to pass B? Right now it does not seem to be used.

amy12xx (commented):
This looks great @alexis-cvetkov . I left a few minor comments.

dirty_cat/test/test_gamma_poisson_factorization.py
```diff
@@ -0,0 +1,87 @@
...
+
+                    # Test same seed return the same output
+                    encoder = OnlineGammaPoissonFactorization(
+                        n_topics=n_topics, hashing=hashing, init=init,
+                        analyzer=analyzer, add_words=add_words,
+                        random_state=42)
+                    encoder.fit(X)
+                    y2 = encoder.transform(X)
+                    np.testing.assert_array_equal(y, y2)
+
```
GaelVaroquaux, line 40:
```suggestion


```
PEP8 asks for 2 empty lines between first-level functions or classes.

dirty_cat/test/test_gamma_poisson_factorization.py
```diff
@@ -0,0 +1,87 @@
...
+def test_input_type():
+    # Numpy array
+    X = np.array(['alice', 'bob'])
+    enc = OnlineGammaPoissonFactorization(n_topics=2)
+    enc.fit_transform(X)
+    # List
+    X = ['alice', 'bob']
+    enc = OnlineGammaPoissonFactorization(n_topics=2)
+    enc.fit_transform(X)
+
```
GaelVaroquaux, line 50:
```suggestion


```

PEP8 asks for 2 empty lines between first-level functions or classes.

doc/index.rst
```diff
@@ -16,6 +16,10 @@ The :class:`SimilarityEncoder` is a drop-in replacement for
 :class:`~sklearn.preprocessing.OneHotEncoder`.
 If speed and scalability are an issue, the :class:`MinHashEncoder` provides
 a fast encoding method.
```
GaelVaroquaux, line 18:
```suggestion
a fast encoding method.

```

An empty line is required to start a new paragraph.

dirty_cat/gamma_poisson_factorization.py
GaelVaroquaux: +1

GaelVaroquaux: I actually think that this line is not necessary.

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+else:
+    from sklearn.cluster._kmeans import kmeans_plusplus
+
+if LooseVersion(sklearn.__version__) < LooseVersion('0.22'):
+    from sklearn.decomposition.nmf import _beta_divergence
+else:
+    from sklearn.decomposition._nmf import _beta_divergence
+
+
+class OnlineGammaPoissonFactorization(BaseEstimator, TransformerMixin):
```
GaelVaroquaux, line 41: I would like to suggest to rename this class to a more marketting-friendly name: I fear that the current name will scare users into thinking that this is a highly-technical class, while actually this is really the class that they should be using.

I propose `GapEncoder`. `Gap` is a fairly classic abbreviation of Gamma Poisson.

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+        self.max_iter = max_iter
+        self.min_iter = min_iter
+        self.init = init
+        self.analyzer = analyzer
+        self.add_words = add_words
+        self.random_state = check_random_state(random_state)
+        self.rescale_W = rescale_W
+        self.max_iter_e_step = max_iter_e_step
+
+        # Init n-grams counts vectorizer
```
GaelVaroquaux, line 157: All the lines below should be done in the beginning of fit, and not the init. In scikit-learn style, inits have no other logic than storing parameters in attributes

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+        if self.hashing:
+            self.ngrams_count = HashingVectorizer(
```
GaelVaroquaux, line 159: In scikit-learn style, as this attibute is data-dependent, it must be named with a trailing underscore.
```suggestion
            self.ngrams_count_ = HashingVectorizer(
```

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+                 analyzer=self.analyzer, ngram_range=self.ngram_range,
+                 n_features=self.hashing_n_features,
+                 norm=None, alternate_sign=False)
+            if self.add_words: # Init a word counts vectorizer if needed
+                self.word_count = HashingVectorizer(
```
GaelVaroquaux, line 164: In scikit-learn style, as this attibute is data-dependent, it must be named with a trailing underscore.
```suggestion
                self.word_count_ = HashingVectorizer(
```

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+                     analyzer='word',
+                     n_features=self.hashing_n_features,
+                     norm=None, alternate_sign=False)
+        else:
+            self.ngrams_count = CountVectorizer(
```
GaelVaroquaux, line 169: In scikit-learn style, as this attibute is data-dependent, it must be named with a trailing underscore.
```suggestion
            self.ngrams_count_ = CountVectorizer(
```

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+                 analyzer=self.analyzer, ngram_range=self.ngram_range)
+            if self.add_words:
+                self.word_count = CountVectorizer()
```
GaelVaroquaux, line 172: In scikit-learn style, as this attibute is data-dependent, it must be named with a trailing underscore.
```suggestion
                self.word_count_ = CountVectorizer()
```

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+if LooseVersion(sklearn.__version__) < LooseVersion('0.22'):
+    from sklearn.decomposition.nmf import _beta_divergence
+else:
+    from sklearn.decomposition._nmf import _beta_divergence
+
+
+class OnlineGammaPoissonFactorization(BaseEstimator, TransformerMixin):
+    """
+    Online Gamma-Poisson Factorization by minimizing the
+    Kullback-Leibler divergence.
```
GaelVaroquaux, line 44: We need to give a bit of intuition here. The technical details can go in the "notes" section that goes at the end of the docstring.

For the top, I suggest something like:
```
This encoder can be understand as a continuous encoding on a set of latent categories estimated from the data.
The latent categories are built by capturing combinations of substrings that frequently co-occur.
```

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+
+    Parameters
+    ----------
+
+    n_topics : int, default=10
```
GaelVaroquaux, line 49: I know that everybody is going to hate me, but should we be using a different term than "topic". The term "topic" comes from the topic-modeling origin of the Gamma-Poisson Factorization, but here the user probably does not think in terms of topic modeling. Should we use `n_components`, as in the MinHashEncoder?

GaelVaroquaux (commented):
Here are a bunch of cosmetic comments.

I'll try to come back and check the algorithmics, but we had checked them back when we coded this originally.

Thanks for the hard work!

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: I agree, I just import the variable sklearn.__version__ now.

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: What do you have in mind ? I cannot use assignments like `self.H_dict[x] = h` in a list comprehension, so something like `[self.H_dict[x] = h for x, h in zip(X, H)]` doesn't work.

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: Yes, I moved it there!

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: Yes, I moved `self.rho_ = self.rho` in the init method.

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: You're right, removed!

dirty_cat/gamma_poisson_factorization.py
GaelVaroquaux: No, it should not be in the init, because if people change rho or batch_size before calling fit, the change won't be propagated. I guess what @amy12xx meant is that you can do:
```python
self.rho_ = rho
if self.rescale_rho:
       # Make update rate per iteration independant of the batch_size
       self.rho_ = self.rho ** (self.batch_size / len(X))
```

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: I agree! Changed into `GapEncoder`.

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: Yes, this will be consistent with the minhash encoder.

CHANGES.rst
```diff
@@ -1,6 +1,12 @@
...
+  character-level substring counts into linear combinations of latent topics.
```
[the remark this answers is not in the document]
alexis-cvetkov: Here is an attempt:
> Added online Gamma-Poisson factorization through the :class:\`GapEncoder\` class. This method discovers latent categories formed via combinations of substrings, and encodes string data as combinations of these categories. To be used if interpretability is important.

GaelVaroquaux:
> Here is an attempt:

>     Added online Gamma-Poisson factorization through the :class:`GapEncoder`
>     class. This method discovers latent categories formed via combinations of
>     substrings, and encodes string data as combinations of these categories. To
>     be used if interpretability is important.

This is great!!

dirty_cat/test/test_gamma_poisson_factorization.py
```diff
@@ -0,0 +1,87 @@
...
+from dirty_cat import OnlineGammaPoissonFactorization
+
+def test_OnlineGammaPoissonFactorization(n_samples=70):
+    X_txt = fetch_20newsgroups(subset='train')['data']
+    X = X_txt[:n_samples]
+    n_topics = 10
+    for hashing in [True, False]:
+        for init in ['k-means++', 'random', 'k-means']:
+            for analyzer in ['word', 'char', 'char_wb']:
+                for add_words in [True, False]:
```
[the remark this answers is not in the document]
alexis-cvetkov: This test takes about 40s to run, and less if I decrease the number of samples/components. Should it be done on only a fraction of possible combinations then ?

GaelVaroquaux:
> This test takes about 40s to run, and less if I decrease the number of samples/ components. Should it be done on only a fraction of possible combinations then ?

Yes please. I'm not sure that being exhaustive is crucial. I'd rather
work on a smaller number of well-chosen combinations. I would keep the
number of samples, so that the test is reliable.

dirty_cat/gamma_poisson_factorization.py
alexis-cvetkov: I'd say "likelihood of the data", because we tune H and W so that the observed data is the most plausible.

alexis-cvetkov:
CircleCI is not building anymore. I think this is because it now uses python 3.9 during the build, and some packages like scikit-learn do not support it yet.

Update: I pinned the version of python to 3.8 in `build_doc.sh`. I don't know if it's the right way to fix this but it works!

amy12xx (commented):
I had something like this in mind: self.H_dict = {x: h for x, h in zip(X, H)} but I think I assumed that H_dict would have the exact same elements as X. If this is not the case, you can ignore this comment.

dirty_cat/gamma_poisson_factorization.py
```diff
@@ -0,0 +1,600 @@
...
+    ----------
+
+    References
+    ----------
+    For a detailed description of the method, see
+    `Encoding high-cardinality string categorical variables
+    <https://hal.inria.fr/hal-02171256v4>`_ by Cerda, Varoquaux (2019).
+    """
+
+    def __init__(self, n_topics=10, batch_size=512, gamma_shape_prior=1.1,
```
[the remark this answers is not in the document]
alexis-cvetkov: Actually, I don't know. The code for the GapEncoder was initially implemented by Patricio Cerda and Gael, so I kept the default parameters. However when I fit this encoder to 100,000 input strings, the memory charge is about 500Mb/CPU thread, which is maybe a bit much. Decreasing the batch_size to 100 solves this, and I checked that it doesn't change the performance of GapEncoder on examples 02/03, so I propose to use that instead. Maybe @GaelVaroquaux has an opinion on that ?

GaelVaroquaux:
> Decreasing the batch_size to 100 solves this,

Good idea: try putting it to 128.

Thanks!

amy12xx:
Thanks @alexis-cvetkov . Looks good to me.

Phaide (approved):

dirty_cat/gap_encoder.py
```diff
@@ -0,0 +1,600 @@
...
+                rescale_W=self.rescale_W,
+                gamma_shape_prior=self.gamma_shape_prior,
+                gamma_scale_prior=self.gamma_scale_prior)
+        # Compute the KL divergence between V and HW
+        kl_divergence = _beta_divergence(
+            unq_V[lookup], unq_H[lookup], self.W_,
+            'kullback-leibler', square_root=False)
+        return kl_divergence
+
+    # def partial_fit(self, X, y=None):
```
GaelVaroquaux, line 403: Likewise: why is the partial_fit commented out? It's a very useful feature.

GaelVaroquaux (changes requested):
Thanks a lot for the PR. I finally had time to review it in depth.

I made a bunch of inline comments. On top of these, I have a few bigger picture comments:

- We need to add a comment at the top of the "Scalability considerations" example saying that the GapEncoder is naturally scalable and should be preferred for large-scale data.
- We should restore the get_feature_names method and demo it in one of the examples
- We should restore the partial_fit method and demo it in a new example on scalability

Thanks!!!

dirty_cat/gap_encoder.py
```diff
@@ -0,0 +1,600 @@
...
+        self.init = init
+        self.analyzer = analyzer
+        self.add_words = add_words
+        self.random_state = check_random_state(random_state)
+        self.rescale_W = rescale_W
+        self.max_iter_e_step = max_iter_e_step
+        self.H_dict = dict()
+        self.rho_ = self.rho
+
+        # Init n-grams counts vectorizer
```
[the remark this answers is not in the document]
alexis-cvetkov: Is there a specific reason to initialize `H_dict` in `fit` rather than in `__init__` ? Also, it means that multiple calls of `fit` will update the same `H_dict` and not retrain the model from scratch. Is that what we want ?

GaelVaroquaux:
> Is there a specific reason to initialize H_dict in fit rather than in __init__ ?

It implies that every call to fit has a clean start, independent of the past.

> Also, it means that multiple calls of fit will update the same H_dict and not retrain the model from scratch. Is that what we want ?

I'm confused: isn't it the other way around: initialize in fit means that
multiple calls of fit will not update the same H_dict (partial_fit should
behave differently), no?

alexis-cvetkov:
> It implies that every call to fit has a clean start, independent of the past.

Okay, I was confused by the test you proposed:
```
if not hasattr(self, "H_dict"):
    ...
```
In this case I don't need to test if `H_dict` exist but I can simply re-initialize it everytime `fit` is called

GaelVaroquaux:
> Okay, I was confused by the test you proposed:

OK, this was an error on my side. Thanks for noting it.

GaelVaroquaux:
@alexis-cvetkov : careful, I am pushing to your branch. If you work on this, pull first.

GaelVaroquaux:
I believe that this is now good to go. I will merge when the tests have run. Thanks!!!

alexis-cvetkov:
Great ! Thank you !

PR #155, opened 2021-02-04
Refactor of the fetching system

LilianBoulard:
Follows #147 

<hr />

This is a refactor of the dataset fetching system.
Historically, datasets were fetched from various websites.

Our aim with this update is to only use OpenML.org's API, through Scikit-learn's [`fetch_openml()`](https://scikit-learn.org/stable/modules/generated/sklearn.datasets.fetch_openml.html) function.
This allows us to have a much more reliable and unified interface, and avoids losing access to datasets due to deletion, renaming, etc.

For instance, with the current system, 4 on the 7 datasets are unavailable (403 Access Denied, website down, etc.).

The user-interface stays the same with the functions `fetch_*()` (e.g `fetch_open_payments()`), still returning a similar dictionary.
The major difference is that this dictionary returns, among other information, a `path`, where a CSV file is located, and must be loaded (using for instance pandas' `read_csv()` function).

<hr />

TL;DR of the previous thread:
the way `fetch_openml()` is used here makes pandas a requirement.


LilianBoulard:
So, little info on the (I hope last) error we got here:
The test gets stopped on a fetch, which raises this error:
```
raise JSONDecodeError("Expecting value", s, err.value) from None
```
Now printing the values, we get `s=This is not JSON-encoded data!`.
This is the expected behavior: as it doesn't receive JSON-encoded data, it raises an error.
Now the thing that I don't quite understand is how does this string ends up here...

It comes from a mock at [this line](https://github.com/dirty-cat/dirty_cat/pull/155/files#diff-8f0f3ae7f57ff22fa657e96be3abe55cce37d98be950884e0d981173752436f3R167) (which is contained using a `with` statement), but the test raises at [this line](https://github.com/dirty-cat/dirty_cat/pull/155/files#diff-8f0f3ae7f57ff22fa657e96be3abe55cce37d98be950884e0d981173752436f3R59) (another function entirely), so I'm assuming this is a mock leak.

I'm investigating the issue :)

LilianBoulard:
So, for this (I hope last) error, we got an issue with the JSON decoding. 
It gets raised inside `sklearn.datasets.fetch_openml`, and unfortunately, this is linked to the version, as the tests pass with `scikit-learn>=0.21.0`.

Unless told to, I won't try to dig inside sklearn's code to figure out at which step it goes wrong, and perhaps monkeypatch a solution.
So meanwhile, we have two options: 
- bump sklearn's minimal version for dirty_cat to 0.21.0
- raise a clean error inviting the user to upgrade his version if he tries to fetch a dataset

What do you suggest ?

GaelVaroquaux:
Is this error happening for all the datasets, or only some?

GaelVaroquaux:
By the way, I checked, sklearn 0.21.0 is a couple years old. It's fine to bump the requirement.

LilianBoulard:
> Is this error happening for all the datasets, or only some?

All of them, as the error is linked to `fetch_openml` and they all use it. See this screenshot:
image

LilianBoulard:
> By the way, I checked, sklearn 0.21.0 is a couple years old. It's fine to bump the requirement.

Great ! I'll make a separate PR

LilianBoulard:
Finally the tests pass !

The current error on CircleCI seems to be caused by insufficient memory : https://support.circleci.com/hc/en-us/articles/115014359648-Exit-code-137-Out-of-memory

For the record, I ran benchmarks on the notebooks and here are the results :

Notebook 1: 82.694251MB
Notebook 2: 149.632461MB
Notebook 3: 174.760288MB
Notebook 4: 88.800206MB
Notebook 5: 820.054484MB
Notebook 6: 1331.590108MB
Total: ~2.643 GB

As we can see, notebooks 5 and 6 stand out in terms of memory usage.
I'm not sure whether CicleCI flushes memory between each notebook.

I don't know if that's possible, but we should add these stats to the footer on the HTML render (along with `Total running time of the script: (...)`). Note: I used `tracemalloc`.

I removed the duplicate 3rd notebook (see #179), and fixed new notebook 3, the one with the SuperVectorizer, we'll see how it goes.

GaelVaroquaux:
Tell me when this is ready for review

LilianBoulard:
So the memory issue we had with the CI comes from the download process, specifically with the dataset `traffic_violation`, which is massive (see https://www.openml.org/d/42132).
I adapted the 6th notebook to use open_payments, which is way lighter.
The only downside I see is that we lost a section on the benefit of using `float32` instead of `float64`, as there are no numeric variables in the dataset. 

For the record, it was removed in 17163b7.

dirty_cat/datasets/fetching.py
```diff
@@ -1,493 +1,380 @@
...
-        if provided, verifies the integrity of the file using a hash
-    uncompress: bool
-        whether to uncompress the content of the url
-
-    show_progress:
-        if ``True``, displays a progressbar during the downloading of the
-        dataset. Warning: ``clint`` needs to be implemented and is not in the
-        requirements for now
+# Future notes:
+# - Watch out for ``fetch_openml()`` API modifications: as of january 2021, the function is marked as experimental.
```
GaelVaroquaux, line 12: PEP8 coding standards tell us that lines should be no more than 79 char long

dirty_cat/datasets/fetching.py
```diff
@@ -1,493 +1,380 @@
...
+MIDWEST_SURVEY_ID = 42805
+MEDICAL_CHARGE_ID = 42720
+EMPLOYEE_SALARIES_ID = 42125
+TRAFFIC_VIOLATIONS_ID = 42132
+DRUG_DIRECTORY_ID = 43044
 
-    # link dead.
-    # return fetch_dataset(EMPLOYEE_SALARIES_CONFIG, show_progress=False)
 
+def fetch_openml_dataset(dataset_id: int, data_directory: Path = get_data_dir()) -> dict:
```
GaelVaroquaux, line 57: I think that this line has a couple of PEP8 violations:
- too long
- spaces in front and after the equal sign

You should consider using a pep8 violation highlighter such as flake8

dirty_cat/datasets/fetching.py
```diff
@@ -1,493 +1,396 @@
...
+    features_gz_path = data_directory / FEATURES_DIRECTORY / f'{dataset_id}.gz'
+
+    if not details_gz_path.is_file() or not features_gz_path.is_file():
+        # If the details file or the features file don't exist,
+        # download the dataset.
+        warnings.warn(
+            f"Could not find the dataset {dataset_id} locally. "
+            "Downloading it from OpenML; this might take a while... "
+            "If the process is interrupted, files will be invalid/incomplete. "
+            "To fix this problem, delete the CSV file if it exists. "
```
GaelVaroquaux, line 110: Maybe we should print also the path of the CSV file to help the user

LilianBoulard: You're right ! I don't like this message altogether, it's too vague and inaccurate, I'll change it.

dirty_cat/datasets/fetching.py
```diff
@@ -1,493 +1,396 @@
...
+    if not data_gz_path.is_file():
+        # This is a double-check.
+        # If the data file does not exist, download the dataset.
+        _download_and_write_openml_dataset(dataset_id=dataset_id,
+                                           data_directory=data_directory)
+
+    if not csv_path.is_file():
+        # If the CSV file does not exist, use the dataset
+        # downloaded by ``fetch_openml()`` to construct it.
+        features = _get_features(features_gz_path)
```
GaelVaroquaux, line 133: This seems slightly suboptimal: we are calling _get_details which reads the json from the compressed file, and latter calling _get_features, which itself also reads the compressed file.

LilianBoulard: Yes, because these are two separate files :)

GaelVaroquaux:
> Yes, because these are two separate files :)

Right... (me go and hide)

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -1,238 +1,352 @@
...
+            returned_info = fetch_openml_dataset(dataset_id=test_dataset["id"],
+                                                 data_directory=test_data_dir)
+
+        except URLError:
+            # No internet connection, or the website is down.
+            # We abort the test.
+            #
+            # One could try to manually recreate the tree structure
+            # created by ``fetch_openml()``,  and the
+            # ``.gz`` files within in order to finish the test.
```
GaelVaroquaux, line 70: I'm not sure that we should do this totally silently.

I think that we should:
1. Pop a warning (or something similar)
2. Skip using some of pytests' functionality for this (for instance calling pytest.skip)

GaelVaroquaux: The goal is to have some track that the test is being skipped, and not silently loose track of it

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -1,238 +1,352 @@
...
+                                               "owner", "recipient"])
+
+    test_data_dir = get_test_data_dir()
+
+    # We test the function to see if it behaves correctly when files exists
+    mock_pathlib_path_isfile.return_value = True
+
+    fetch_openml_dataset(50, test_data_dir)
+
+    mock_download       .assert_not_called()
```
GaelVaroquaux, line 116: Strange formatting. I don't think that this is PEP8 (but correct me if I'm wrong)

examples/02_fit_predict_plot_employee_salaries.py
```diff
@@ -16,23 +16,24 @@
...
-employee_salaries = fetch_employee_salaries()
-print(employee_salaries['DESCR'])
+
+info = fetch_employee_salaries()
+print(info['description'])
 
 
 ################################################################################
 # Then we load it:
 import pandas as pd
```
GaelVaroquaux, line 26: I know that it's ugly, but I would rather that you:
1. print the info['read_csv_kwargs']
2. pass them explicitly to read_csv
3. add a comment saying that such operation is feasible with the syntax ...**info['read_csv_args']

The reason is that I am trying to reach out to people who do not understand this syntax

examples/04_feature_interpretation_gap_encoder.py
```diff
@@ -16,14 +16,14 @@
...
-print(employee_salaries['DESCR'])
+info = fetch_employee_salaries()
+print(info['description'])
 
 
 ################################################################################
 # Then we load it:
 import pandas as pd
-df = employee_salaries['data']
+df = pd.read_csv(info['path'], **info['read_csv_kwargs'])
```
GaelVaroquaux, line 26: Same comment on the ** syntax

examples/05_scaling_non_linear_models.py
```diff
@@ -435,7 +435,7 @@ def encode(X, y_int, one_hot_encoder, column_transformer, rbf_sampler):
 
 iter_csv = pd.read_csv(
     info['path'], nrows=online_train_set_size, chunksize=batchsize,
-    skiprows=1, names=columns_names, sep='\t')
+    skiprows=1, names=columns_names, **info['read_csv_kwargs'])
```
GaelVaroquaux, line 438: Same comment :)

GaelVaroquaux (commented):
I finished reviewing this and made all my comments. After these small things, this will be ready for merge.

LilianBoulard:
Good to go !

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -65,12 +66,14 @@ def test_fetch_openml_dataset():
...
-            # One could try to manually recreate the tree structure
-            # created by ``fetch_openml()``,  and the
-            # ``.gz`` files within in order to finish the test.
+            warnings.warn(
+                "No internet connection or the website is down, test aborted."
+            )
+            pytest.skip(
+                "One could try to manually recreate the tree structure "
+                "created by ``fetch_openml()``,  and the "
+                "``.gz`` files within in order to finish the test."
```
GaelVaroquaux, line 75: I think that the content of this should go in a comment, and the message in the "skip" call should be more understandable to someone who is looking only at the logs. Something like "URLError: skipping this test as the failure might be due to an internet connection problem".

examples/01_investigating_dirty_categories.py
```diff
@@ -16,7 +16,14 @@
 
 info = datasets.fetch_employee_salaries()
 print(info['description'])
-data = pd.read_csv(info['path'], **info['read_csv_kwargs'])
+data = pd.read_csv(
+    info['path'],
+    quotechar=info['read_csv_kwargs']['quotechar'],
+    escapechar=info['read_csv_kwargs']['escapechar'],
+    na_values=info['read_csv_kwargs']['na_values'],
```
GaelVaroquaux, line 23: I would litterally hard code the values: "na_values='na'"

dirty_cat/datasets/tests/test_fetching.py
LilianBoulard: Ah, I see ! I'll change that

examples/01_investigating_dirty_categories.py
LilianBoulard: Sure, why not, it should make the code easier to read

LilianBoulard:
These most recent changes improve the user interface.
Now, instead of reading the CSV from the disk, it is directly loaded in memory if `load_dataframe=True`.
It also returns a namedtuple instead of a dictionary, which has an object-oriented interface.

alexis-cvetkov:
This PR looks good to me, I think it can be merged.
Thanks Lilian !

GaelVaroquaux:
I was about to merge, but don't we need a an entry summarizing things in the changelog?

alexis-cvetkov:
Yes you're right, I always forget that.

GaelVaroquaux:
Yey! Merged

PR #156, opened 2021-02-06
Fix CircleCI checks [doc build]

alexis-cvetkov:
CircleCI checks didn't pass after the last PR. When building example 05 in the documentation, an import error is raised when executing `from sklearn.linear_model.stochastic_gradient import SGDClassifier`. Since sklearn 0.22, it should be imported with `from sklearn.linear_model import SGDClassifier`instead, so I changed that line of code.

However I wonder why this error wasn't raised before. Maybe when merging PR to master, CircleCI doesn't build examples that weren't modified.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,9 @@ jobs:
       - run: ./build_tools/circle/checkout_merge_commit.sh
       - restore_cache:
           key: v1-datasets-{{ .Branch }}
-      - run: ./build_tools/circle/build_doc.sh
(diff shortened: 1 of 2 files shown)


alexis-cvetkov:
I have also changed `.circleci/config.yml` to increase the `no_output_time` to 30 min. By default, jobs with no output for more than 10 min fail. In our case example 05 took 13 min to run and thus raised the error.

GaelVaroquaux:
Good job, thanks!!

PR #162, opened 2021-03-31
Adding pretrained fastText models

alexis-cvetkov:
Introduces the new class `PretrainedFastText` that allows encoding of categorical data with pretrained fastText models.


dirty_cat/pretrained_fastText.py
```diff
@@ -0,0 +1,203 @@
...
+        self.language = language
+        self.bin_dir = bin_dir
+        # Load the model from binary file
+        if file_name == None:
+            file_name = f"cc.{language}.{n_components}.bin"
+        self.file_path = os.path.join(bin_dir, file_name)
+        self.load_model()
+        return
+    
+    def download_model(self):
```
GaelVaroquaux, line 73: There should not be an empty line here

setup.py
```diff
@@ -39,5 +39,5 @@
           packages=find_packages(),
           package_data={'dirty_cat': ['VERSION.txt', 'data/midwest_survey/*.csv']},
           install_requires=['scikit-learn>=0.20', 'numpy', 'scipy', 'requests',
-                            'joblib'],
+                            'joblib', 'fasttext>=0.9.2'],
```
GaelVaroquaux, line 42: Can we make it so that fasttext is an optional dependency: it is loaded only inside the methods of the class that need it.

alexis-cvetkov: I can move `import fasttext` inside the methods that use it. Regarding `setup.py`, should I declare optional dependencies (see below) or just remove fasttext from `install_requires` ?
```
extras_require={
              "fasttext": ["fasttext>=0.9.2"]  
            }
```

If we declare optional dependencies with `extras_require`, then should we add `python-Levensthein` too ? It is presented as an optional dependency in `README.rst` but isn't mentioned in `setup.py`.

alexis-cvetkov:
The coverage of the code is smaller than usual because some methods cannot be tested (i.e. `download_model` which would require to download few Gb of data). Codecov is complaining but I don't think we can do better.

GaelVaroquaux:
> should I declare optional dependencies (see below)

Yes!!

> If we declare optional dependencies with extras_require, then should we add python-Levensthein too ?

Indeed!

alexis-cvetkov:
I have merged the latest changes in dirty_cat. Everything works well, except for the test coverage of the new `PretrainedFastText` encoder. As I said before, this is because some of its methods cannot be tested (e.g. `download_model` that requires downloading few Gb of data). Does it look good to you @GaelVaroquaux

GaelVaroquaux:
I've started having a look again at this PR. I worry that we have added a 1.6Mb file to the package (the dbpedia file), which makes it much heavier than it is currently, while the file is useful only for a single functionality.

The good solution would be to add a downloader for this file.

alexis-cvetkov:
That's a good point! I have added a downloader for the fastText test model.

GaelVaroquaux:
We should not add an optional dependency on fasttext (optional dependencies make packages hard to grasp and use: you never know what you have).

I'm not sure how to move this forward without the optional dependency

Issue #54, opened 2018-10-29
Automatic column vectorizer depending on the column structure


pcerda:
Automatically select the type of encoder for a given column depending on its structure:

- If the column type is 'float' or 'int', pass through.

- If the column is type is 'categorical', select the encoder depending on the cardinality 'k' of the variable. For example: If k<=20, select one-hot encoding; if not, select similarity encoding.

GaelVaroquaux:
As a side note, we cannot name this object "AutoEncoder", as an autoencoder is already a well-identified concept in machine learning that differs quite strongly from what we do here.

Maybe, as a proposed name: DataFrameEncoder

amy12xx:
Adding some more to the mix: AutoSelectEncoder, BestEncoder and MetaEncoder.

GaelVaroquaux:
This has been implemented.

PR #167, opened 2021-06-22
Add SuperVectorizer

LilianBoulard:
Hello

This PR aims at implementing an "automatic column vectorizer" as suggested by #54 (solves #54) : the SuperVectorizer.

It can be used to automatically apply encoders to columns, depending on implicit characteristic (such as the column `dtype`).
Under the hood, it is an interface for sklearn's `ColumnTransformer`.

Currently, it selects columns based on their `dtype`, and divides them into several groups:
- Numerical variables
- Datetime variables
- Low cardinality categorical variables
- High cardinality categorical variables
- Low cardinality string variables
- High cardinality string variables

As you noticed, we differentiate categorical and string.  
The threshold used to categorize low or high cardinality can be specified by the user.

An experimental feature currently implemented is the `auto_cast`, which will try to cast each column to the best possible data type.

Please submit feedback and suggestions, I'd be happy to discuss it !


dirty_cat/super_vectorizer.py
```diff
@@ -0,0 +1,341 @@
...
+            array[col].replace(to_replace='?', value='', inplace=True)
+            if dtype_name.startswith('int') or dtype_name.startswith('float'):
+                array[col].fillna(0, inplace=True)
+            else:
+                array[col].fillna('', inplace=True)
+        array = array.reset_index(drop=True)
+        return array
+    elif isinstance(array, np.ndarray):
+        # Replace missing values for numpy
+        array[np.where(np.isnan(array))] = 0
```
GaelVaroquaux, line 69: Here you are replace the values inplace. This is in general dangerous (though certainly not frowned upon) if the programmer is not aware.

The challenge of this function is that it does a bit of both: it modifies the input data, but what it returns can be different. It's a bit bad style, but I have no proposal that is as memory-friendly :).

This should at least be mentioned clearly in a very short docstring, so that later contributors don't make mistakes.

dirty_cat/super_vectorizer.py
```diff
@@ -0,0 +1,341 @@
...
+
+    def _transform(self, X) -> pd.DataFrame:
+        # Convert to pandas DataFrame if not already.
+        if not isinstance(X, pd.DataFrame):
+            X = pd.DataFrame(X)
+
+        # Detect if the array contains missing values.
+        if _has_missing_values(X):
+            if self.handle_missing == '':
+                X = _replace_missing(X)
```
GaelVaroquaux, line 282: Actually, ideally I think that we would not handle missing values in the supervectorizer, as often downstream (transformers or even the final learner) can do this better.

dirty_cat/super_vectorizer.py
```diff
@@ -0,0 +1,341 @@
...
+        return ColumnTransformer.transform(self, X)
+
+    def fit_transform(self, X, y=None):
+        X = self._transform(X)
+
+        # Select columns by dtype
+        numeric_columns = X.select_dtypes(include=['int', 'float']).columns.to_list()
+        string_columns = X.select_dtypes(include=['string', 'object']).columns.to_list()
+        categorical_columns = X.select_dtypes(include='category').columns.to_list()
+        datetime_columns = X.select_dtypes(include='datetime').columns.to_list()
```
GaelVaroquaux, line 307: Shouldn't we check that we have selected all the columns with the above? And issue a warning or an error if not.

dirty_cat/super_vectorizer.py
```diff
@@ -0,0 +1,341 @@
...
+
+        # Divide string and categorical columns by cardinality
+        low_card_str_columns = [col for col in string_columns if X[col].nunique() < self.cardinality_threshold]
+        high_card_str_columns = [col for col in string_columns if X[col].nunique() >= self.cardinality_threshold]
+        low_card_cat_columns = [col for col in categorical_columns if X[col].nunique() < self.cardinality_threshold]
+        high_card_cat_columns = [col for col in categorical_columns if X[col].nunique() >= self.cardinality_threshold]
+
+        # Construct the transformers
```
GaelVaroquaux, line 315: I don't really understand what the code below is doing (forgive my tired brain).

These lines need at least a high-level comment before them describing them.

Also, I am surprised, it seems that they are using a dataframe to build a list. I wonder whether it would not be more readable using a combination of for loops and the append or extend method of lists.

dirty_cat/super_vectorizer.py
```diff
@@ -0,0 +1,341 @@
...
+        # True: non-empty columns field
+        valid_columns_mask = transformers_df['columns'].apply(lambda x: True if x != [] else False)
+
+        mask = valid_encoders_mask & valid_columns_mask
+        self.transformers = transformers_df[mask].values.tolist()
+
+        if not self.transformers:
+            raise RuntimeError('No transformers could be generated !')
+
+        if self.verbose is not False:
```
GaelVaroquaux, line 337:
```suggestion
        if self.verbose:
```
"is" in Python tests the exact identity of objects. You seldom want to use it. "verbose=0" is a valid way of specifying no verbosity, and it would not pass the "is" test.

dirty_cat/super_vectorizer.py
LilianBoulard: Great question !  
Technically yes, but we currently don't include some data types (i.e.  `timedelta`, and perhaps others I forgot).

dirty_cat/super_vectorizer.py
LilianBoulard: Sure, I'll try to improve the doc  
I used this process because it was the lightest code-wise. I'll implement the same thing in vanilla Python and compare the two.

dirty_cat/super_vectorizer.py
LilianBoulard: Oh, you're right ! I didn't think about the edge case `verbose=0`

GaelVaroquaux:
:heart:

You have a failing CI. It will need attention at some point.

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,126 @@
+"""
+Automatic pre-processing with the SuperVectorizer
+=================================================
+
+In this notebook, we will illustrate the use of the `SuperVectorizer`, which
+can automatically classify columns of a dataset based on their type, and apply
+transformers based on this classification.
+To demonstrate that, we will use the `employee salaries` dataset.
```
GaelVaroquaux, line 8:
```suggestion
In this notebook, we illustrate the `SuperVectorizer`, which automatically
applies transformers to the different columns of a dataset.
We demonstrate it onthe `employee salaries` dataset.
```

Less words are better (and future tense is not needed)

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,126 @@
...
+        'full_name',  # Not relevant to the analysis
+        '2016_gross_pay_received',  # Too linked with target
+        '2016_overtime_pay',  # Too linked with target
+        'date_first_hired'  # Redundant with "year_first_hired"
+    ],
+    axis=1,
+    inplace=True
+)
+
+X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.15, random_state=42)
```
GaelVaroquaux, line 38:
```suggestion
X_train, X_test, y_train, y_test = train_test_split(X, y, test_size=0.15,
                                                                            random_state=42)
```
I think that this line is a bit long, no?

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,126 @@
...
+
+###############################################################################
+# Under the hood
+# --------------
+# Let's now break down what the SuperVectorizer did.
+#
+# Once it has been trained on data,
+# we can print the assignation it did:
+
+print(sup_vec.transformers_)
```
GaelVaroquaux, line 66: Break the cell here, so that we can see the output

GaelVaroquaux, line 66: Break the cell here, so that we can see the output

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,126 @@
...
+# Fit the SuperVectorizer
+X_train_enc = sup_vec.fit_transform(X_train, y_train)
+X_test_enc = sup_vec.transform(X_test)
+# And the regressor
+regressor.fit(X_train_enc, y_train)
+
+
+###############################################################################
+# Under the hood
+# --------------
```
GaelVaroquaux, line 60:
```suggestion
# Inspecting the features created by the SuperVectorizer
# -------------------------------------------------------------------------
```

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,126 @@
...
+# Let's now break down what the SuperVectorizer did.
+#
+# Once it has been trained on data,
+# we can print the assignation it did:
```
GaelVaroquaux, line 64:
```suggestion
# we can print the transformers and the columns assignment it create:
```

LilianBoulard:
> You have a failing CI. It will need attention at some point.

Yes, it's a work in progress !

LilianBoulard:
Okay, I found the source of the error.
As shown by the CI, it comes from code
```
', '.join(map(str, sorted(set(dtypes_orig))))
```
`dtypes_orig` is a list of types, e.g. `[Int64Dtype, np.float64]`

I first though it was a type problem, but actually, it comes from the `sorted`: the expression isn't properly ordered.
This way it works: (and does the same thing)
```
', '.join(sorted(map(str, set(dtypes_orig))))
```

To fix this issue, I just replaced the StandardScaler from the workflow in the test by `"passthrough'`.

So, note that using a StandardScaler, and perhaps other transformers on older versions of sklearn, might raise this error.

GaelVaroquaux:
> So, note that using a StandardScaler, and perhaps other transformers on older versions of sklearn, might raise this error.

Then we need to fix it

LilianBoulard:
> Then we need to fix it

Addressed in bd90180 :)

CHANGES.rst
```diff
@@ -11,6 +11,8 @@ Major changes
   - SciPy (>= 1.2)
   - scikit-learn (>= 0.20.0)
 
+* Add SuperVectorizer v1
```
GaelVaroquaux, line 14: Desribe here what the SuperVectorizer does

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,178 @@
...
+plt.yticks(range(n), labels, size=15)
+plt.tight_layout(pad=1)
+plt.show()
+
+###############################################################################
+# We can deduce a few things from this data:
+# the three factors that define the most the salary are: being a manager,
+# being hired for a long time, and have a permanent, full-time job.
+#
+# Finally, let's make a pipeline, perform a cross-validation
```
GaelVaroquaux, line 153: Move this above as a section of itself, in the beginning of the notebook.

The section could be "A machine-learning pipeline with the SuperVectorizer", and then there would be a second section called "Understanding the features created by the SuperVectorizer" with 2 subsections.

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,178 @@
...
+        tree.feature_importances_
+        for tree in regressor.estimators_
+    ],
+    axis=0
+)
+indices = np.argsort(importances)[::-1]
+
+# Plotting the results:
+
+plt.figure(figsize=(18, 9))
```
GaelVaroquaux, line 138:
```suggestion
plt.figure(figsize=(12, 9))
```
Make the figure more narrow, so that the text are more visible

CHANGES.rst
LilianBoulard: Addressed in 6aa83cf

examples/07_automatic_preprocessing_with_the_supervectorizer.py
LilianBoulard: Addressed in acff116

doc/index.rst
```diff
@@ -43,7 +47,7 @@ ______
 API documentation
 =================
 
-Encoders
+Encoders / Vectorizers
 ------------
```
GaelVaroquaux, line 51: I think that you need to expend the title line below

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,193 @@
...
+from sklearn.ensemble import RandomForestRegressor
+
+
+sup_vec = SuperVectorizer(auto_cast=True)
+regressor = RandomForestRegressor(n_estimators=25, random_state=42)
+# Fit the SuperVectorizer
+X_train_enc = sup_vec.fit_transform(X_train, y_train)
+X_test_enc = sup_vec.transform(X_test)
+# And the regressor
+regressor.fit(X_train_enc, y_train)
```
GaelVaroquaux, line 117: Maybe we should move this part after we have printed the feature names

examples/07_automatic_preprocessing_with_the_supervectorizer.py
```diff
@@ -0,0 +1,193 @@
...
+print(feature_names[:8])
+print(len(feature_names))
+
+###############################################################################
+# As we can see, it created a new column for each unique value.
+# This is because we used |SimilarityEncoder| on the column "division",
+# which was classified as a high cardinality string variable.
+# (default values, see `SuperVectorizer`'s docstring).
+# In total, we have 1212 encoded columns.
+#
```
GaelVaroquaux, line 153: Here we could move the import and fit of the random forest, in a section call "Feature importance in the statistical model"

examples/07_automatic_preprocessing_with_the_supervectorizer.py
LilianBoulard: You mean for the RFR ?
To get the feature names, the SV must be fitted.

examples/07_automatic_preprocessing_with_the_supervectorizer.py
LilianBoulard: Addressed in 4f7585e

doc/index.rst
LilianBoulard: Addressed in 01cf92c

GaelVaroquaux:
CI ran, I'm merging!

GaelVaroquaux:
This is exciting!!

Issue #168, opened 2021-06-22
AttributeError: 'SimilarityEncoder' object has no attribute 'drop_idx_'


LilianBoulard:
Hello

While working today with the SimilarityEncoder, I found this error :
```
/usr/bin/python3.8 /home/lboulard/.config/JetBrains/PyCharmCE2021.1/scratches/Issue_with_similarity_encoder.py
Traceback (most recent call last):
  File "Issue_with_similarity_encoder.py", line 11, in <module>
    feature_names = enc.get_feature_names()
  File "~/.local/lib/python3.8/site-packages/sklearn/preprocessing/_encoders.py", line 590, in get_feature_names
    if self.drop_idx_ is not None and self.drop_idx_[i] is not None:
AttributeError: 'SimilarityEncoder' object has no attribute 'drop_idx_'
```

Code to reproduce : 
```
from dirty_cat import SimilarityEncoder
from dirty_cat.datasets import fetch_employee_salaries

employee_salaries = fetch_employee_salaries()

# Get only one column
X = employee_salaries['data']['employee_position_title'].to_numpy().reshape(-1, 1)

enc = SimilarityEncoder()
enc.fit(X)
feature_names = enc.get_feature_names()
print(feature_names)
```

While looking at the code, I found that, in `sklearn.OneHotEncoder` - that SimilarityEncoder inherits - there is this line in the `fit` method :

[`self.drop_idx_ = self._compute_drop_idx()`](https://github.com/scikit-learn/scikit-learn/blob/2beed55847ee70d363bdbfe14ee4401438fba057/sklearn/preprocessing/_encoders.py#L425)

Which is missing from `SimilarityEncoder.fit()`.

**Easy fix: add the exact same line in the method, which fixes the issue.**

Disclaimer: I haven't looked very much in depth into what `OneHotEncoder._compute_drop_idx()` does. 
I'll do when I have a little more time :)  
So... might not be the best fix (tough in my case it works).

Please let me know if this looks good to you, I have a PR ready.

GaelVaroquaux:
Seems like a good fix.

Please send a PR with the fix and a test. Thanks a lot!!

GaelVaroquaux:
Fixed by #169

PR #169, opened 2021-06-22
Fix AttributeError drop_idx_ SimilarityEncoder

LilianBoulard:
Hello

Solves #168

### DIFF
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -402,6 +402,8 @@ class SimilarityEncoder(OneHotEncoder):
                 self.vocabulary_count_matrices_.append(vocabulary_count_matrix)
                 self.vocabulary_ngram_counts_.append(vocabulary_ngram_count)
 
+        self.drop_idx_ = self._compute_drop_idx()
+
         return self
 
     def transform(self, X, fast=True):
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
Darn, we have an error in the build which looks unrelated.

I believe that it is a problem with a change in the acceptable arguments for sphinx-gallery configuration

GaelVaroquaux:
I tried to fix the problem in #170

GaelVaroquaux:
I merged a fix. You can rebase on master to fix the problem.

However, I am worried that we no longer have a CI running on the tests ( https://github.com/dirty-cat/dirty_cat/issues/171 ). I wonder if I would not want that problem addressed first.

dirty_cat/similarity_encoder.py
```diff
@@ -402,6 +402,8 @@ def fit(self, X, y=None):
                 self.vocabulary_count_matrices_.append(vocabulary_count_matrix)
                 self.vocabulary_ngram_counts_.append(vocabulary_ngram_count)
 
+        self.drop_idx_ = self._compute_drop_idx()
```
alexis-cvetkov, line 405: It seems that `drop_idx` and `_compute_drop_idx` were added in sklearn 0.21, so that's why for skleanr 0.20 the tests failed. You can use the code below to check that sklearn version is at least 0.21 and in that case call the function `compute_drop_idx`.

```
if LooseVersion(sklearn_version) > LooseVersion('0.20'):
    self.drop_idx_ = self._compute_drop_idx()
```

GaelVaroquaux (approved):
LGTM. Thanks a lot!!

Issue #171, opened 2021-06-22
Move our tests from travis to azure pipelines


GaelVaroquaux:
Travis is not longer free for open source projects. Hence our tests no longer run.

We need to switch to azure pipelines.

Nilearn and scikit-learn already did this. We can borrow from them.

GaelVaroquaux:
Fixed by #172

PR #172, opened 2021-06-23
Enable testing on GitHub actions

NicolasGensollen:
See #171 

This PR proposes to run tests on Github actions.
Note: I've used the structure we use for Nilearn but I can adapt if needed.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -1,0 +1,88 @@
+name: 'build'
+
+on:
+  push:
+    branches:
+      - "master"
+  pull_request:
+    branches:
+      - '*'
+
+jobs:
+  check_skip:
+    runs-on: ubuntu-latest
+    outputs:
+      skip: ${{ steps.result_step.outputs.ci-skip }}
+    steps:
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - id: result_step
+        uses: mstachniuk/ci-skip@master
+        with:
+          commit-filter: '[skip ci];[ci skip];[skip github]'
+          commit-filter-separator: ';'
+
+  min_requirements:
+    needs: check_skip
+    if: ${{ needs.check_skip.outputs.skip == 'false' }}
+    runs-on: "ubuntu-latest"
+    name: Python 3.6 with minimal requirements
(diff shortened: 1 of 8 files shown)


NicolasGensollen:
@LilianBoulard it looks like the tests don't run with minimum dependencies (I used the ones from the README). 
Do you know if these were tested before? 
Note: I tried bumping to numpy==1.16, scipy==1.2, and sklearn==0.20 and it runs better

LilianBoulard:
Hi @NicolasGensollen  

Just checked on a prior PR that used Travis, and while there is a worker that is supposed to launch the tests with the minimal dependencies, there was a misconfiguration with it, and he did not 😅.
See https://travis-ci.org/github/dirty-cat/dirty_cat/jobs/760701080 and https://travis-ci.org/github/dirty-cat/dirty_cat/jobs/760701079

@GaelVaroquaux what version do you reckon we should bump the minimal dependencies to ?
Perhaps `numpy==1.16`, `scipy==1.2`, and `sklearn==0.20` as suggested by Nicolas ?

LilianBoulard:
Thanks for launching the tests again, looks like it's working this time, though there is a small error with the `GapEncoder`.  
@alexis-cvetkov, could you please have a look ?

alexis-cvetkov:
I was able to reproduce the error on a local environment. I am comparing two outputs that should be the same (or at least almost equal) but for some reason on the "minimal requirements" environment they are not exactly identical but very close (the max absolute / relative error is 1e-12 / 1e-16).
For such cases we can use the function `np.testing.assert_almost_equal` that check if two inputs are equal up to a desired precision (by default 1e-7). We use it in other tests and I think we can simply use that instead. I am pushing this.

GaelVaroquaux:
> For such cases we can use the function np.testing.assert_almost_equal that check if two inputs are equal up to a desired precision (by default 1e-7). We use it in other tests and I think we can simply use that instead.

+1

GaelVaroquaux (changes requested):
Looks great!

One remaining thing: to update the version numbers in the readme

GaelVaroquaux (approved):
:heart:

Merging. Thanks!!

PR #181, opened 2021-07-16
Fix type hinting in generated doc

### LINKED ISSUE #177: Function signatures hard to read due to type hinting
On current builds of the documentation, the type-hinted signatures are very hard to read.
See https://github.com/sphinx-doc/sphinx/issues/5868 for a similar issue.

Resources for future fix : 
https://pypi.org/project/sphinx-autodoc-typehints/

LilianBoulard:
Fixes #177

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -260,7 +260,7 @@ ogp_use_first_image = True
 ogp_site_name = "Dirty cat"
 
 # -- sphinxext.napoleon configuration --------------------------------------
-napoleon_use_param = True
+napoleon_use_param = False
 
 # -- sphinxext.typehints configuration -------------------------------------
 autodoc_typehints = "description"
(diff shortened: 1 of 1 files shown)


LilianBoulard:
https://881-124964482-gh.circle-artifacts.com/0/doc/generated/dirty_cat.SuperVectorizer.html#dirty_cat.SuperVectorizer

Works for the uppermost paragraph, but we have "Parameters" twice. Kinda weird, I'll look into it.

GaelVaroquaux:
I think that the "parameters" twice is because the '__init__' is documented also.

GaelVaroquaux:
We also still have type hints poping up for the object attributes ("transformers_", "columns_"). That's hard to read.

Finally (but that's a problem that was already there). The attributes and methods are not labelled as such. Add a separation or a section title describing what they are would be great. Maybe this can be fixed using css?

GaelVaroquaux:
I'm merging this because it is a net improvement. We can always iterate more later.

PR #187, opened 2021-07-19
DOC: Add binder

### LINKED ISSUE #186: Configure sphinx-gallery to enable binder usage
nilearn has an example of how to do this.

GaelVaroquaux:
Closes #186

Unfortunately, I believe that this needs to be merged in master to be tested.


GaelVaroquaux:
I'll merge this to see if it works in master. I don't think that I can debug further.

PR #189, opened 2021-07-19
Remove unused file

### LINKED ISSUE #149: Codecov for ``count_3_grams.py`` at 0%
Hi,

On the [codecov report](https://codecov.io/gh/dirty-cat/dirty_cat/tree/master/dirty_cat), I can see that it considers the file ``count_3_grams.py`` as not tested (coverage at 0%).

However, when we take a peak into the file, we can see there are actually test units inside.
\
It is also said in the docstring:
> Run unit tests with "pytest count_3_grams.py"

In order for the code coverage to improve, I would advise splitting the file in two so that the test units move into ``dirty_cat/tests/``.

Please tell me if it's fine to do that, and I'll submit a PR.

Thanks
Lilian

LilianBoulard:
Removes the file `count_3_ngrams.py` as it is a benchmark and is not used anywhere.
For the record, only the docstring seems interesting :

```
Quick benchmarking seems to show that pure Python code is faster when
for strings less that 1000 characters, and numpy versions is faster for
longer strings.
Very long strings would benefit from probabilistic counting (bloom
filter, count min sketch) as implemented eg in the "bounter" module.
```

Fixes #149


GaelVaroquaux:
Great! Merging!

Issue #164, opened 2021-04-02
Run scikit-learn's common tests on dirty_cat's estimators


LilianBoulard:
As suggested by Gaël, we should try and run scikit-learn's common tests to be sure dirty_cat's estimators follow sklearn's conventions.  
Resources: 
https://scikit-learn.org/stable/developers/develop.html#rolling-your-own-estimator

LilianBoulard:
Solved by #274

PR #197, opened 2021-07-21
Fix SV not raising error when not fitted

### LINKED ISSUE #193: Better error message when SuperVectorizer's transform method is called before fit
Scikit-learn has a mechanism to check that an estimator is fitted, and if not raise a good error message. We should use it.

LilianBoulard:
Fixes #193

### DIFF
--- a/dirty_cat/super_vectorizer.py
+++ b/dirty_cat/super_vectorizer.py
@@ -171,8 +171,6 @@ class SuperVectorizer(ColumnTransformer):
         self.transformer_weights = transformer_weights
         self.verbose = verbose
 
-        self.columns_ = []
-
     def _auto_cast_array(self, X):
         """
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
Can we add a test that fails without this PR, so that we don't fall back to the same problem after a refactoring.

LilianBoulard:
Done, although I think it should get addressed by #164

GaelVaroquaux:
We have a failing test.

Maybe we can skip for old versions of scikit-learn (if this is the source of the problem).

LilianBoulard:
Found a workaround for older versions by passing all the attributes of the object.
Not sure it's the way the function is supposed to be used, the docstring doesn't say much more, but the test should be passing :)

GaelVaroquaux:
LGTM. Merging. Thanks

PR #198, opened 2021-07-21
Set auto_cast to True by default

### LINKED ISSUE #192: Consider switching auto_cast to True in the SuperVectorizer
Every single one of our examples uses auto_cast=True. That tells us that it should probably be the default.

LilianBoulard:
Fixes #192


LilianBoulard:
I believe it's already done, otherwise can you please specify which examples ?
Thanks

GaelVaroquaux:
My mistake. Merging. Thank you

Issue #196, opened 2021-07-21
Using labels for issues


LilianBoulard:
I noticed that labels are unavailable to "default" users when they open new issues on the GitHub repo (they cannot assign labels to issues they open). It's probably a setting of the repo. Is it intended ?
I also noticed we generally don't use them very much :)
We should ! That's part of the default GH toolbox, and might help in categorizing issues.

What do you think ?

LilianBoulard:
As pointed out by @NicolasGensollen above, random users cannot assign labels by default, and now that I think about it, I believe it should stay that way. My bad :)

PR #199, opened 2021-07-22
Add issue templates

NicolasGensollen:
Related to #196 

@LilianBoulard AFAIK, random users cannot put labels on issues (I believe only contributors can...). I think the reason behind this is that users are not supposed to know your labeling conventions, and therefore they are not the best persons to decide which labels should be used.
However, most users opening an issue know whether it is to report a bug, ask for a new feature, say something about the documentation, and so on...
This PR proposes to add a few issue templates with predefined text and labels in order to ease this process. Namely:

- a bug report template
- a documentation template
- a feature request template

Note that you can easily add/remove them (maybe you'll also need a usage question template and so on...).

Feel free to close this if you don't think it's relevant :wink:

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -1,0 +1,25 @@
+---
+name: Bug report
+about: Something not working as described? This is the place.
+title: ''
+labels: 'bug'
+assignees: ''
+
+---
+<!--Provide a brief description of the bug.-->
+
+
+<!--Please fill in the following information, to the best of your ability.-->
+dirty_cat version:
+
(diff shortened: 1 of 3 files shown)


NicolasGensollen:
Should look like that:

templates

LilianBoulard:
Oh... now that I think about it, you're right !
https://docs.github.com/en/issues/using-labels-and-milestones-to-track-work/managing-labels
Thanks for the clarification

LilianBoulard (approved):

GaelVaroquaux:
> @LilianBoulard approved this pull request.

@LilianBoulard: feel free to merge if you think that this is right.

Issue #180, opened 2021-07-07
Drop 'Current Annual Salary' in employee_salaries loader


GaelVaroquaux:
Our examples all drop the ''Current Annual Salary' when loading the employee_salaries dataset (because it is too linked to the target). To simplify the example, we should consider adding an option, on by default, to drop it in the loader.

In the same logic, we should probably also drop '2016_gross_pay_received',  '2016_overtime_pay'

And we should also drop the 'full name' column.

PR #200, opened 2021-07-23
Add parameter for dropping pointless columns

LilianBoulard:
Quickfix for #180 

Should also be addressed properly in #155

### DIFF
--- a/dirty_cat/datasets/fetching.py
+++ b/dirty_cat/datasets/fetching.py
@@ -7,12 +7,16 @@ meaning are directly copied. The rest is annoted.
 """
 # -*- coding: utf-8 -*-
 import os
+
 import requests
 import shutil
 import urllib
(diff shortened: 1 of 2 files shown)


LilianBoulard:
Ouch... We're having the exact same issue as in #155...
A few solutions are proposed in this PR, could you please have a look @GaelVaroquaux ?

Meanwhile, I can remove the test, which only penalizes codecov, as the feature runs fine otherwise.

GaelVaroquaux:
Hum, it seems that we are still having a failure. :$

LilianBoulard:
I've spent some time this morning investigating the issue, and it seems to be raised for two datasets, `road_safety` and `open_payments`.
The error is fixed by #155, so if you don't mind, I'll try to work on this PR instead.

PR #201, opened 2021-07-23
Don't handle missing values in the SuperVectorizer

### LINKED ISSUE #178: Missing values replaced by the SuperVectorizer
Currently, the ``SuperVectorizer`` replaces missing values, however, as pointed out by @GaelVaroquaux in the original PR, the transformers / estimators behind often handle them by default.

Therefore, the ``SuperVectorizer`` should not replace these values.

LilianBoulard:
Fixes #178


dirty_cat/super_vectorizer.py
```diff
@@ -24,29 +24,26 @@
...
+
 def _has_missing_values(array: pd.DataFrame) -> bool:
     """
     Returns True if `array` contains missing values, False otherwise.
     """
     return any(array.isnull())
 
 
-def _replace_missing(array: pd.DataFrame) -> pd.DataFrame:
+def _replace_missing(array: pd.DataFrame, value: str = "missing") -> pd.DataFrame:
```
GaelVaroquaux, line 37: I think that pep8 wants
```suggestion
def _replace_missing(array: pd.DataFrame, value: str="missing") -> pd.DataFrame:
```

GaelVaroquaux: Also, to me it is surprising that a variable of type DataFrame is named "array". In this function, I would name it "df", as this is a typical convention for dataframes.

dirty_cat/super_vectorizer.py
```diff
@@ -197,20 +198,30 @@ def _transform(self, X) -> pd.DataFrame:
...
+                        self.high_card_str_transformer,
+                        self.low_card_cat_transformer,
+                        self.high_card_cat_transformer,
+                    ]
+                )
+                if using_ohe and _sklearn_loose_version < LooseVersion('0.24'):
+                    # If we use a OneHotEncoder and we're using a version
+                    # of sklearn prior to 0.24, replace missing values,
+                    # otherwise, leave them in.
+                    X = _replace_missing(X, self.handle_missing)
```
GaelVaroquaux, line 218: Shouldn't we do this only on the columns where we are using the OneHotEncoder?

dirty_cat/super_vectorizer.py
LilianBoulard: When using this format `def func(argument: type=default_value)`, my IDE (PyCharm) raises warnings E251/E252.
See https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes

I found no written rule about that, nonetheless I believe the consensus is 
```
Don't use spaces around the '=' sign when used to indicate a keyword argument 
or a default parameter value, except when using a type annotation.
```

For the argument named `array`, you're right, I'll rename it `df`. It stayed that way because historically we also managed numpy arrays.

LilianBoulard:
Updated original comment to describe the changes of d5f6193
I still have a bit of work to do, especially improving the tests and documenting the new properties.

dirty_cat/super_vectorizer.py
```diff
@@ -130,27 +125,35 @@ class SuperVectorizer(ColumnTransformer):
...
                  auto_cast: bool = True,
+                 impute_missing: str = 'auto',
                  # Following parameters are inherited from ColumnTransformer
-                 handle_missing: str = '',
-                 remainder='passthrough',
-                 sparse_threshold=0.3,
-                 n_jobs=None,
+                 remainder: str = 'passthrough',
+                 sparse_threshold: float = 0.3,
+                 n_jobs: int = None,
```
GaelVaroquaux, line 154: AFAIC, coding standards (PEP8) tell us that in keyword arguments, there should be no space before and after the equal sign:
```suggestion
                 n_jobs: int=None,
```
I am making this comment here, but it apply to many other places.

dirty_cat/super_vectorizer.py
```diff
@@ -162,70 +165,86 @@ def __init__(self, *,
...
-                # for earlier versions of sklearn
-                for column in X:
-                    dtype = X[column].dtype
-                    if issubclass(dtype.__class__, ExtensionDtype):
-                        try:
-                            X[column] = X[column].astype(dtype.type)
-                        except TypeError:
-                            pass
+            X.columns = self.columns_
+            X = X.astype(self.types_)
```
GaelVaroquaux, line 243: Maybe if you move this line above, you won't be modifying the input dataframe, as this line assigns a new object to the dataframe. Hence, you will not need to copy the dataframe to address my comment above.

dirty_cat/super_vectorizer.py
```diff
@@ -257,8 +275,20 @@ def fit_transform(self, X, y=None):
             usually because transformers passed do not match any column.
             To fix the issue, try passing the least amount of None as encoders.
         """
-        X = self._transform(X)
+        # Create a copy to avoid altering the original data.
+        X = X.copy()
```
GaelVaroquaux, line 279: Copies are expensive, we try to do as little as possible (but still, copy when necessary). Maybe it would be possible to move this copy in an "else" block, right below, as I below that the "if" statement is inducing a copy.

dirty_cat/super_vectorizer.py
```diff
@@ -257,8 +275,20 @@ def fit_transform(self, X, y=None):
...
+        # Convert to pandas DataFrame if not already.
+        if not isinstance(X, pd.DataFrame):
+            X = pd.DataFrame(X)
+
+        # X = self._transform(X)
```
GaelVaroquaux, line 284: Do not leave left-over commented code :)

dirty_cat/super_vectorizer.py
```diff
@@ -162,70 +165,86 @@ def __init__(self, *,
...
-                                 "Expected any of {'', 'error'}, "
-                                 f"got {self.handle_missing}")
+            # Check the number of columns matches the fitted array's.
+            if X.shape[1] != len(self.columns_):
+                raise ValueError("Passed array does not match column count of "
+                                 f"fitter array. Got X.shape={X.shape}")
+            # If the DataFrame does not have named columns already,
+            # apply the learnt columns
+            if isinstance(X.columns, pd.RangeIndex):
+                X.columns = self.columns_
```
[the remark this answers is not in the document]
LilianBoulard: I don't think we're altering the data as we make a copy at the beginning of the method.

dirty_cat/super_vectorizer.py
LilianBoulard: Unfortunately it won't work in this context because we only use `X.astype` when `auto_cast` is `True`, but right after, we replace missing values regardless.   
We already perform a copy at the beginning as pointed out in a previous comment, so I'll leave it that way for now.

dirty_cat/super_vectorizer.py
```diff
@@ -24,30 +23,25 @@
...
-        if dtype_name.startswith('int') or dtype_name.startswith('float'):
-            array[col] = array[col].fillna(0)
-        else:
-            array[col] = array[col].fillna('')
-    array = array.reset_index(drop=True)
-    return array
+    dtype_name = df.dtype.name
+    if dtype_name == 'category' and (value not in df.cat.categories):
+        df = df.cat.add_categories(value)
+    df = df.fillna(value=value)
```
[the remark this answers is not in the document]
LilianBoulard: Indeed. This is an issue, I'm working on it !

dirty_cat/super_vectorizer.py
```diff
@@ -110,11 +104,12 @@ class SuperVectorizer(ColumnTransformer):
...
-    handle_missing: str, default=''
-        One of the following values: 'error' or '' (empty).
-        Defines how the encoder will handle missing values.
-        If set to 'error', will raise ValueError.
-        If set to '', will impute the missing values (pd.NA) with blank strings.
+    impute_missing: str, default='auto'
+        By-column missing values imputation parameter.
+        'auto' will impute missing values if it's considered appropriate.
+        'force' will impute all missing values.
+        'skip' will not impute at all.
```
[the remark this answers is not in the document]
LilianBoulard: The render isn't so bad, but not too great either.

image
https://1040-124964482-gh.circle-artifacts.com/0/doc/generated/dirty_cat.SuperVectorizer.html#dirty_cat.SuperVectorizer

It's readable but I agree it could use some better style :)

dirty_cat/test/test_super_vectorizer.py
```diff
@@ -18,90 +19,171 @@ def check_same_transformers(expected_transformers: dict, actual_transformers: li
...
+    """
+    Creates a simple DataFrame with missing values.
+    """
+    return pd.DataFrame({
+        'int': pd.Series([15, 56, 63, 12, 44], dtype='int'),
+        'float': pd.Series([5.2, 2.4, 6.2, 10.45, 9.], dtype='float'),
+        'str1': pd.Series(['public', np.nan, 'private', 'private', 'public'], dtype='object'),
+        'str2': pd.Series(['officer', 'manager', np.nan, 'chef', 'teacher'], dtype='object'),
+        'cat1': pd.Series([np.nan, 'yes', 'no', 'yes', 'no'], dtype='object'),
+        'cat2': pd.Series(['20K+', '40K+', '60K+', '30K+', np.nan], dtype='object'),
```
GaelVaroquaux, line 46: For recent pandas version, we should also test with Na on ints
 https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html#integer-dtypes-and-missing-data

dirty_cat/test/test_super_vectorizer.py
```diff
@@ -18,90 +19,171 @@ def check_same_transformers(expected_transformers: dict, actual_transformers: li
...
     check_same_transformers(expected_transformers_plain, vectorizer_cast.transformers)
     # With numpy
     vectorizer_cast.fit_transform(X_str.to_numpy())
-    check_same_transformers(expected_transformers_np, vectorizer_cast.transformers)
+    check_same_transformers(expected_transformers_np_cast, vectorizer_cast.transformers)
 
 
+def test_with_clean_data():
+    X = _get_clean_dataframe()
+    # Define the transformers we'll use throughout the test.
```
GaelVaroquaux, line 105: Tell us a bit more in plain English in the comments how you built these list, so that someone coming here in a few years understands the logic of the test :)

dirty_cat/test/test_super_vectorizer.py
```diff
@@ -18,90 +19,171 @@ def check_same_transformers(expected_transformers: dict, actual_transformers: li
...
+        expected_transformers_np_no_cast,
+        expected_transformers_series,
+        expected_transformers_plain,
+        expected_transformers_np_cast,
+    )
+
+
+def test_with_dirty_data():
+    X = _get_dirty_dataframe()
+    # Define the transformers we'll use throughout the test.
```
GaelVaroquaux, line 148: Same comment: explain the logic of these lists

GaelVaroquaux (commented):
OK, I finished going through all the code, and I put in all my comments. This PR will be good to go.

dirty_cat/super_vectorizer.py
LilianBoulard: For standard keywords, yes, but not for type hinted ones.
See PEP484, section "Support for singleton types in union", second block for an example:
https://www.python.org/dev/peps/pep-0484/#support-for-singleton-types-in-unions

GaelVaroquaux:
> For standard keywords, yes, but not for type hinted ones.

Thank you for correcting me. My knowledge is getting too old :)

dirty_cat/super_vectorizer.py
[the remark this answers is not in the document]
LilianBoulard: It was fixed in the fit_transform and not directly in this function for technical reasons :)

alexis-cvetkov:
I finished reviewing the PR:
- we need to update `CHANGES.rst`
- maybe add a test to ensure that `fit_transform(X)` is equivalent to `fit(X).transform(X)`
- is there a reason to define a separate encoder for string column and categorical columns in the parameters of the SuperVectorizer ?

GaelVaroquaux:
The questions of Alexis are very relevant. Once they are addressed, I am +1 for merge.

PR #206, opened 2021-07-27
Fix memory leak in notebook 6

LilianBoulard:
As pointed out in #155, there is a memory leak responsible for the CI errors.

This commit should lower the used memory of notebook 6 from ~1.3 GB to ~200 MB (measured with `tracemalloc` and `memory_profiler`), as well as making the build a little faster.


GaelVaroquaux:
Nice!! Thanks. Great solution. Will merge when tests are done.

GaelVaroquaux:
Merged!! Good job!

PR #207, opened 2021-07-28
Handle missing in GapEncoder

LilianBoulard:
Similar to #128

Add missing values handling in the GapEncoder.
Required for #178

### DIFF
--- a/dirty_cat/gap_encoder.py
+++ b/dirty_cat/gap_encoder.py
@@ -24,6 +24,7 @@ from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.feature_extraction.text import CountVectorizer, HashingVectorizer
 from sklearn.cluster import KMeans
 from sklearn.neighbors import NearestNeighbors
+from sklearn.utils.fixes import _object_dtype_isnan
 import pandas as pd
 from .utils import check_input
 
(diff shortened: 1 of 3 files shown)


dirty_cat/gap_encoder.py
```diff
@@ -514,6 +515,14 @@ class GapEncoder(BaseEstimator, TransformerMixin):
     max_iter_e_step : int, default=20
         Maximum number of iterations to adjust the activations h at each step.
 
+    handle_missing : 'error' or '' (default)
+        Whether to raise an error or impute with blank string '' if missing
```
GaelVaroquaux, line 519: Maybe "passthrough" instead of ''

GaelVaroquaux:
I made a little comment, but aside from this commit, this PR is good for me. Thanks!!

alexis-cvetkov (approved):

dirty_cat/gap_encoder.py
LilianBoulard: Well I guess we're not passing through, as we're imputing the missing data. Am I missing something ?
Thanks :)

GaelVaroquaux:
> Well I guess we're not passing through, as we're imputing the missing data. Am
> I missing something ?

Point taken, I'll remove this.

dirty_cat/gap_encoder.py
GaelVaroquaux: Maybe "zeroout"?

I just don't love the empty string. It doesn't really tell what it does to a casual reader.

alexis-cvetkov: I agree the empty string is not clear, maybe we can simply set `handle_missing = "impute"` ?

GaelVaroquaux: OK, then "zero_impute". I don't like "impute", because people typically expect conditional imputations.

LilianBoulard: I changed the default parameter to `zero_impute`, while still imputing with empty strings

LilianBoulard:
Merging !

GaelVaroquaux:
Thanks!! Sorry, I'm teaching all day. Little feedback today

PR #210, opened 2021-09-28
[MinHashEncoder] Fix handle_missing parameter and standardize input data shape

alexis-cvetkov:
I have made small modifications to the MinHashEncoder:

- The documentation for the `handle_missing` parameter was not accurate since we actually do not impute missing values with an empty string. Instead we assign them a encoding vector filled with zeros.
- Since other encoders take as input data with shape `(N_samples, 1)` (to be consistent with scikit-learn), I updated the MinHashEncoder to behave in the same way.

I also fixed the doc for the `handle_missing` parameter in GapEncoder.


alexis-cvetkov:
Merging !

PR #211, opened 2021-09-28
Tune SuperVectorizer

LilianBoulard:
Following some benchmarks we made, I've tuned the default parameters of the SuperVectorizer to reflect these results.
Finer tuning is possible, but it already took almost two weeks to get these results, so for now it'll do.


GaelVaroquaux:
Thank you. Merging!

Issue #214, opened 2021-10-08
Page contents in sidebar


GaelVaroquaux:
With the new examples, we are getting documentation pages that are longer (and this is good).

We should add a TOC (table of content) in the sidebar, as on http://dirtydata.science/python/gen_notes/02_dirty_categories.html#sphx-glr-gen-notes-02-dirty-categories-py

The page above is actually generate from a setup very similar to the doc of dirtycat. To port these changes to dirty cat, we need to:
- Add the localtoc template back to the alabster theme https://github.com/dirty-data-science/python/blob/main/conf.py#L99 We want to add this localtoc at the end of the list
- Add the scrolltoc.js to the javascript https://github.com/dirty-data-science/python/tree/main/_static and in https://github.com/dirty-data-science/python/blob/main/conf.py
- Merge the CSS in dirty-cat's CSS: https://github.com/dirty-data-science/python/blob/main/_static/custom.css

PR #217, opened 2021-10-11
[doc-build] Add table of content in notebooks

LilianBoulard:
Solves #214


GaelVaroquaux:
That problem could be due to CSS. You can try by brutally replacing dirty-cat's CSS by the other project's, to see if it fixes the pb

GaelVaroquaux:
I think that that "Table of Content" should be replaced by "This page". You can do this by finding the right "template" of the Alabaster theme, modifying it, and adding it to the "_templates" directory.

Issue #230, opened 2021-12-08
Add an handle_unknown parameter for SuperVectorizer.


LeoGrin:
- I think there should be three options {"error", "ignore", "error_strict"}. If "error" or "ignore", this option would be passed to the low_card_cat_transformer. If "error_strict", this would pass the "error" option to both low_card_cat_transformer and high_card_cat_transformer.
- I would use "error" as the default value.

GaelVaroquaux:
Sounds good!

LeoGrin:
Made a PR #235

LeoGrin:
Closing this issue after a discussion in #235 (it's seems simpler to simply set this parameter in the low_card_transformer argument of the SuperVectorizer)

PR #235, opened 2021-12-15
Add a handle_unknown parameter to the SuperVectorizer

LeoGrin:
#230 

I only use this parameter for low_card_cat_encoder, because I think that raising error for unknown categories in the high cardinality setting isn't reasonable.

If the user specifies an encoder with handle_unknown != error for low_card_cat_encoder, we don't overwrite his choice.
This idea is that this means that the user wanted to change default parameters.

I made the assumptions that for low_card_cat_encoder, we only needed to support:
- OneHotEncoder
- TargetEncoder
- OrdinalEncoder (in this case, if handle_unknown = "ignore"), the missing category is replaced by np.nan

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,9 @@ Major changes
 * Add a method "get_feature_names_out" for the **GapEncoder** and the **SuperVectorizer**,
   since "get_feature_names" will be depreciated in scikit-learn 1.2 (#216).
 
+* Add a `handle_unknown` parameter to the `SuperVectorizer` to deal with categories
+  that are not in the training set.
+
 Notes
 -----
 
--- a/dirty_cat/super_vectorizer.py
+++ b/dirty_cat/super_vectorizer.py
@@ -50,18 +50,22 @@ def _set_handle_unknown(transformer: BaseEstimator, handle_unknown: str) -> Base
     Sets the `handle_unknown` parameter of a transformer. Input is in {"error", "ignore"},
     converts it if for OrdinalEncoder.
     """
-    if isinstance(transformer, OneHotEncoder) or isinstance(transformer, TargetEncoder):
-        return transformer.set_params(handle_unknown=handle_unknown)
-    elif isinstance(transformer, OrdinalEncoder):
-        if handle_unknown == "error":
-            return transformer.set_params(handle_unknown="error")
-        elif handle_unknown == "ignore":
-            return transformer.set_params(handle_unknown="use_encoded_value", unknown_value=np.nan)
-
-    else:
-        warn("handle_unknown parameter cannot be set with the transformer you chose for low_card_cat_transformer."
-             "Using default.")
+    # If low_card_cat_transformer is a string, it's drop or passthrough
+    if isinstance(transformer, str):
         return transformer
+    if handle_unknown == "error":
+        # Either the transformer has default value and we don't need to change it
+        # Or the user has changed the default value and we don't want to overwrite it
+        return transformer
+    elif handle_unknown == "ignore":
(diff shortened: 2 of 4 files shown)


CHANGES.rst
```diff
@@ -32,6 +32,9 @@ Major changes
 * Add a method "get_feature_names_out" for the **GapEncoder** and the **SuperVectorizer**,
   since "get_feature_names" will be depreciated in scikit-learn 1.2 (#216).
 
+* Add a `handle_unknown` parameter to the `SuperVectorizer` to deal with categories
+  that are not in the training set.
+
```
GaelVaroquaux, line 37: Release 0.2.0 is already out.

You should create a new section "Release 0.2.1" and add this there.

dirty_cat/super_vectorizer.py
```diff
@@ -46,6 +45,30 @@ def _replace_missing_in_col(df: pd.Series, value: str = "missing") -> pd.Series:
...
+        return transformer
+    elif handle_unknown == "ignore":
+        if isinstance(transformer, OneHotEncoder) or isinstance(transformer, TargetEncoder):
+            return transformer.set_params(handle_unknown="ignore")
+        elif isinstance(transformer, OrdinalEncoder):
+            return transformer.set_params(handle_unknown="use_encoded_value", unknown_value=np.nan)
+        else:
+            warn("handle_unknown parameter cannot be changed with the "
+                 "transformer you chose for low_card_cat_transformer.")
+            return transformer
```
GaelVaroquaux, line 68: I fear that a typo in "handle_unknown" will lead to a silent or incomprehensible error.

Here we need an "else" clause which explicitly raises a error (I would use a "ValueError"). To improve the user experience, the message of the error should indicate the value that was given, and the acceptable values for instance f'{repr(handle_unknown)} was given, while the valid values are "error" or "ignore"')

dirty_cat/super_vectorizer.py
```diff
@@ -145,12 +181,13 @@ def __init__(self, *,
...
                  auto_cast: bool = True,
                  impute_missing: str = 'auto',
+                 handle_unknown: str = 'error',
                  # Following parameters are inherited from ColumnTransformer
                  remainder: str = 'passthrough',
                  sparse_threshold: float = 0.3,
                  n_jobs: int = None,
                  transformer_weights=None,
-                 verbose: bool = False,
+                 verbose: bool = False
```
GaelVaroquaux, line 190: It's good practice to leave a comma at the end of an enumeration like this: it means that adding something at the end will require only modifying a single line, and thus decrease the chances of conflict.

dirty_cat/super_vectorizer.py
```diff
@@ -161,13 +198,16 @@ def __init__(self, *,
...
         self.impute_missing = impute_missing
+        self.handle_unknown = handle_unknown
 
         self.remainder = remainder
         self.sparse_threshold = sparse_threshold
         self.n_jobs = n_jobs
         self.transformer_weights = transformer_weights
         self.verbose = verbose
 
+        self.low_card_cat_transformer = _set_handle_unknown(self.low_card_cat_transformer, handle_unknown)
```
GaelVaroquaux, line 209: By scikit-learn conventions, no input values should be modified during "__init__". This line should be moved to the beginning of fit (that way, if "handle_unknown" is modified after the init, it gets propagated).

GaelVaroquaux (changes requested):
I've added a few comments.

I do have a big-picture comment (sorry, it did not come to my mind when we had discussed this issue originally): why do it this way, rather than pass the right arguments to the transformers given as the "low_card_cat_transformer" to the SuperVectorizer? The proposed solution seems more ad-hoc.

Issue #232, opened 2021-12-08
Replace columns indexes by columns names in SuperVectorizer.transformers_


LeoGrin:
Right now, the transformers_ attribute of SuperVectorizer contains either the columns names (for low_card_cat_transformer and high_card_cat_transformer) or the columns indexes (for remainder for instance). I think we could use columns names for every transformer., as it would be more readable for users.

GaelVaroquaux:
To make sure that I understand well, can you add an example of the
current sub-optimal behavior?

Thanks!

LeoGrin:
For instance in the [SuperVectorizer example](https://dirty-cat.github.io/stable/auto_examples/01_dirty_categories.html#example-super-vectorizer):

`sup_vec.transformers_`

`[('low_card_cat', OneHotEncoder(), ['gender', 'department', 'department_name', 'assignment_category']), ('high_card_cat', GapEncoder(n_components=30), ['division', 'employee_position_title', 'underfilled_job_title']), ('remainder', 'passthrough', [7])]`

To understand that that column 7 is year_first_hired, you need to go back to your dataframe.

GaelVaroquaux:
OK, +1 for suggested fix.

Thanks!!!

LeoGrin:
Fixed

PR #236, opened 2021-12-15
add a get_transformers() method to the SuperVectorizer()

LeoGrin:
Solving #232 , i.e the columns associated with the "remainder" transformer should be displayed by their names, not their index.

I tried to change the output of `sup_vec.transformers_` but it seems to be hard without changing the code of the ColumnTransformer, or doing weird hacks.

Instead, I created a new method `get_transformers()` which outputs the right result, but I'm not sure if creating a new method is a good idea.

### DIFF
--- a/dirty_cat/super_vectorizer.py
+++ b/dirty_cat/super_vectorizer.py
@@ -439,3 +439,14 @@ class SuperVectorizer(ColumnTransformer):
         get_feature_names.
         """
         return self.get_feature_names()
+
+    def get_transformers(self):
+        res  = []
+        for name, trans, cols in self.transformers_:
+            if name == "remainder":
+                res.append((name, trans,
+                            list(map(lambda i:self.columns_[i], cols))))
+            else:
+                res.append((name, trans, cols))
+        return res
+
--- a/dirty_cat/test/test_super_vectorizer.py
+++ b/dirty_cat/test/test_super_vectorizer.py
@@ -249,6 +249,13 @@ def test_transform():
     x_trans = sup_vec.transform(x)
     assert (x_trans == [[1, 0, 0, 0, 1, 0, 0, 0, 1, 0, 0, 0, 0, 1, 34, 5.5]]).all()
 
+def test_get_transformers():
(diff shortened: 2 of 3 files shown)


dirty_cat/super_vectorizer.py
```diff
@@ -439,3 +439,14 @@ def get_feature_names_out(self, input_features=None) -> List[str]:
         get_feature_names.
         """
         return self.get_feature_names()
+
+    def get_transformers(self):
+        res  = []
+        for name, trans, cols in self.transformers_:
+            if name == "remainder":
+                res.append((name, trans,
+                            list(map(lambda i:self.columns_[i], cols))))
```
GaelVaroquaux, line 448: List comprehensions are easier to read than pure functional programming for most people
```suggestion
                            [self.columns_[i] for i in cols])
```

CHANGES.rst
```diff
@@ -32,6 +32,9 @@ Major changes
 * Add a method "get_feature_names_out" for the **GapEncoder** and the **SuperVectorizer**,
   since "get_feature_names" will be depreciated in scikit-learn 1.2 (#216).
 
+* Add a method "get_transformers" to the **SuperVectorizer**, to display the
+  transformers used in the SuperVectorizer, with the associated column names.
```
GaelVaroquaux, line 36: Same comment as in the other PR: this should be added in a new section corresponding to release 0.2.1, as version 0.2.0 has already been released.

GaelVaroquaux (changes requested):
I've done a review and made a few comments to improve the coding. However, this is not the right fix: we don't want to be adding a new method that does not exist elsewhere. In general, good object-oriented style is to stick to as little public methods as possible, and to methods that are shared across different classes (one of the reason, but not the only one, is that the users won't keep in mind too many methods).

In general, the right solution would be to do the expansion in the fit.

There are two ways forward:
1) convincing upstream (scikit-learn) that the "remainder" should be expanded in the fit, and proposing a PR for this.
2) doing this in the fit of the SuperVectorizer

Issue #234, opened 2021-12-09
Handle string in auto_cast (SuperVectorizer)


LeoGrin:
# Suboptimal behavior

Right now the `auto_cast` method from the SuperVectorizer uses the pandas function `convert_dtypes`.

But this function doesn't  try to infer input type from string columns. For instance:

```
import pandas as pd

df = pd.DataFrame({"float_column":["1.3", "4.12", "7.12", "9.12", "1128.12323"],
                   "int_column":["10", "400", "1265", "12", "753"],
                   "bool_column":["true", "false", "true", "false", "false"]})

df = df.convert_dtypes()

for col in df.columns:
    print(df[col].convert_dtypes().dtypes)

```

outputs

```
string
string
string
```

# Use cases

I encountered this problem when dealing with survey data, where people had to enter a number, but some people just entered an empty string. I'm not sure how useful it is in general for the use cases we have in mind.

# Possible solution

We could do something akin to what I suggested in  #233 for dates. For instance, to infer int or float columns, we could use `pd.to_numeric` and check if it worked on enough rows.

GaelVaroquaux:
I think that pandas must have code to do the conversion that you are
looking for, because it does this when loading csvs. Inspecting the code
source of pandas (starting from the read_csv function), might give you a
hint of how to call code from pandas to improve the behavior.

Issue #233, opened 2021-12-08
Handling date columns in SuperVectorizer


LeoGrin:
# Inferring that a column is a date
Pandas `convert_dtypes` function that we use to infer column type doesn't work for inferring date type.
Plan to infer if a column contains a date or a datetime:
- First convert columns with `convert_dtypes` if auto_cast==True (as we do now)
- For every string column, use `pd.to_datetime(errors=coerce)`, and check how many rows have been successfully converted to date (i.e check how many new missing values we have created).
- If enough rows have been converted (> 90% ?), use the converted column, otherwise keep it as string.

We should check if the runtime is acceptable when the user has a lot of string columns. If it's too high, we can apply this on a random subset of rows.

# Using a date column
- If the number of unique date values is low (< 15?), consider it as a categorical variable, otherwise, convert it to a time since epoch.
- Add an option `enrich_from_date` which control if we should create new features from the date columns. Features we could create from date.
    - Year
    - Month
    - Day of the month
    - Day of the week
    - Indicator of holyday? (maybe to country dependant)
    
 - Possible options design for  `enrich_from_date`:
   -  {"all", "none", List, Dic}: date columns for which we want to create new features. If List, list of columns. If Dic, this kind: {"col_2":["year", "month", "weekday"], "col_5": ["weekday"]}. Maybe the Dic option is too complicated.

LeoGrin:
Solved by #239

Issue #231, opened 2021-12-08
Replace strings which indicates missing values by NaNs  (for SuperVectorizer)


LeoGrin:
I think it would be nice for the SuperVectorizer to convert strings which indicates missing values by NaNs before handling missing values.

Strings I have in mind:

- whitespaces (\s)
- "?"
- "missing"

GaelVaroquaux:
I'm not sure how useful that would be: for categorical data, treating
missing values as a special category is a very good option; and I believe
that the current state of the code will lead to that.

LeoGrin:
I agree for categorical data. I actually encountered this issue when dealing with a column which contained numerical data, entered by users in a survey: the values in the column were strings containing either an integer, or a space. This was breaking Pandas type detection.

GaelVaroquaux:
OK, so it's a type detection problem.

Then I'm fine with the solution, but it should be explained and
documented as relevant for type detection (not in these terms, as they
are not meaningful for most users).

Thanks!!

LeoGrin:
Actually I checked and the type detection problem is more general: pandas function convert_dtypes doesn't infer type from string columns. I've made a new issue #234  and I'm closing this one (as solving #234 should solve this issue at the same time).

PR #238, opened 2022-02-14
Change auto_cast in SuperVectorizer

LeoGrin:
Uses pandas [`TextParser`](https://github.com/pandas-dev/pandas/blob/06d230151e6f18fdb8139d09abf539867a8cd481/pandas/io/parsers/readers.py#L1289) (used in `read_csv`, as per @GaelVaroquaux suggestion in #234) instead of `convert_dtypes` in the SuperVectorizer `_auto_cast` method. This has several advantages:

- Type detection works on string columns, e.g `["3", "2", "1"]` would get detected as integer. Solves #234 
-  Missing values encoded as strings get treated as missing values, and don't prevent type detection. For instance,` [1, 2, "", 4]` is detected as integers. (solves  #231). Which strings get detected as Nans is easy to change (see below).
- Datetime columns are converted to datetimes, and the format is inferred by Pandas. Solve the first part of #233.

Some comments:
- TODO: I'd like to enable the user to specify a specific date format if he wants to.
- Strings detected as NaNs are the default [`STR_NA_VALUES`](https://github.com/pandas-dev/pandas/blob/main/pandas/_libs/parsers.pyx#L1350), and I've added `[None, " ", "?", "..."]`. I think we could find a more principled way to choose which strings to include.
- I haven't properly measured the speed of the conversion.
- To use TextParser, I need to convert the input data into a list of list. I don't think this is a problem, but tell me if you think otherwise.
- All variables which are not numerics or datetimes are converted to object (and thus handled as categorical in the SuperVectorizer).
- Right now, if a datetime column contains mixed timezones, it is encoded as an object.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -1,3 +1,17 @@
+Release 0.2.1
+=============
+
+Major changes
+-------------
+
+Notes
+-----
+
+* Improvements to the SuperVectorizer
+
+    - Type detection works better: handles dates, numerics columns encoded as strings, or numeric columns containing strings for missing values
+
+
 Release 0.2.0
 =============
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,7 +7,7 @@ Major changes
 Notes
 -----
 
-* Improvements to the SuperVectorizer
+* Improvements to the :class:`SuperVectorizer`
 
     - Type detection works better: handles dates, numerics columns encoded as strings, or numeric columns containing strings for missing values
 
--- a/dirty_cat/super_vectorizer.py
+++ b/dirty_cat/super_vectorizer.py
@@ -23,7 +23,6 @@ from sklearn.preprocessing import OneHotEncoder
 
 from dirty_cat import GapEncoder
 
-
 _sklearn_loose_version = LooseVersion(sklearn.__version__)
 
 
@@ -169,29 +168,43 @@ class SuperVectorizer(ColumnTransformer):
         self.verbose = verbose
 
     @staticmethod
-    def _auto_cast(X: pd.DataFrame) -> pd.DataFrame:
+    def _auto_cast(X) -> pd.DataFrame:
         """
-        Takes a pandas DataFrame and tries to convert its columns to the best
+        Takes an array-like or a dataframe and tries to convert its columns to the best
         possible data type.
 
         Parameters
         ----------
-        X: pd.DataFrame
-            Input data as a pandas DataFrame.
+        X : {array-like, dataframe} of shape (n_samples, n_features)
+            The data to be transformed.
 
         Returns
         -------
         array
             The same array, with its columns casted to the best possible
             data type.
-            Columns with missing values won't be modified.
         """
         from pandas.core.dtypes.base import ExtensionDtype
+        # To use TextParser, we need to have a list of rows
+        colnames = None  # save the column names which we lose during the conversion
+        if isinstance(X, pd.DataFrame):
+            colnames = X.columns
+            X = X.to_numpy().tolist()
+        elif isinstance(X, np.ndarray):
+            X = X.tolist()
+
+        parser = pd.io.parsers.TextParser(X,
+                                          header=None,
+                                          na_values=[None, " ", "?", "..."],  # additional values to be considered as NA
+                                          infer_datetime_format=True,  # speed-up the date conversion
+                                          parse_dates=list(range(len(X[0]))))  # parse for all columns
+        X = parser.read()
+
+        if colnames is not None:
+            X.columns = colnames  # restore the column names
+
         for col in X.columns:
             dtype = X[col].dtype
-            contains_missing: bool = _has_missing_values(X[col])
-            if not contains_missing:
-                X[col] = X[col].convert_dtypes()
             # Cast pandas dtypes to numpy dtypes
             # for earlier versions of sklearn
             if issubclass(dtype.__class__, ExtensionDtype):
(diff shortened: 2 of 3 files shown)


dirty_cat/super_vectorizer.py
```diff
@@ -169,45 +168,61 @@ def __init__(self, *,
...
         Parameters
         ----------
-        X: pd.DataFrame
-            Input data as a pandas DataFrame.
+        X : {array-like, dataframe} of shape (n_samples, n_features)
+            The data to be transformed.
 
         Returns
         -------
         array
```
LilianBoulard, line 183:
```suggestion
        pd.DataFrame
```

dirty_cat/super_vectorizer.py
```diff
@@ -169,45 +168,61 @@ def __init__(self, *,
...
             The same array, with its columns casted to the best possible
```
LilianBoulard, line 184:
```suggestion
            The same array as a pandas DataFrame, with its columns casted to the best possible
```

dirty_cat/super_vectorizer.py
```diff
@@ -169,45 +168,61 @@ def __init__(self, *,
...
+            X = X.tolist()
+
+        parser = pd.io.parsers.TextParser(X,
+                                          header=None,
+                                          na_values=[None, " ", "?", "..."])  # additional values to be considered as NA
+        X = parser.read()
+
+        for col in X.columns:
+            # only convert string, object or categorical columns to datetimes
+            if X[col].dtype != "int64" and X[col].dtype != "float64":
```
LilianBoulard, line 203: Should it be useful to account for columns with type `int32` or similar ?
Somewhere in the code, I've used [is_numeric_dtype](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.api.types.is_numeric_dtype.html), perhaps it could also make the code easier to read ?

dirty_cat/super_vectorizer.py
```diff
@@ -229,30 +244,23 @@ def transform(self, X) -> np.ndarray:
...
+        # Check the number of columns matches the fitted array's.
+        if X.shape[1] != len(self.columns_):
+            raise ValueError("Passed array does not match column count of "
+                             f"array seen at fit time. Got {X.shape[0]} "
+                             f"columns, expected {len(self.columns_)}")
+
         # If the DataFrame does not have named columns already,
         # apply the learnt columns
-        if isinstance(X.columns, pd.RangeIndex):
+        if X.columns.dtype == "int64":
```
LilianBoulard, line 261: Same comment as above

LilianBoulard (commented):
Thanks for the PR, it looks nice !

LilianBoulard (approved):
LGTM !

dirty_cat/super_vectorizer.py
```diff
@@ -169,45 +168,62 @@ def __init__(self, *,
...
+            The same array as a pandas DataFrame, with its columns casted to the best possible
             data type.
-            Columns with missing values won't be modified.
         """
         from pandas.core.dtypes.base import ExtensionDtype
+        # To use TextParser, we need to have a list of rows
+        colnames = None  # save the column names which we lose during the conversion
+        if isinstance(X, pd.DataFrame):
+            colnames = X.columns
+            X = X.to_numpy().tolist()
```
GaelVaroquaux, line 192: I fear that on large arrays, this will blow the memory. Can we do type detection only on the first few thousand lines?

LeoGrin:
In trying to take into account @GaelVaroquaux's [comment](https://github.com/dirty-cat/dirty_cat/pull/238#discussion_r810191767), I've actually realized that instead of using TextParser, it would be simpler to go back to @LilianBoulard's version of autocast, and:
- replace convert_dtypes by pd.to_numeric and pd.to_datetime
- Replace every missing value with np.nan *before* type conversion

GaelVaroquaux:
Very interesting!! You know these questions better than us, now, so your input is super useful. Tell us when you want a second review.

GaelVaroquaux:
Darn, the tests errors are showing us that the strategy here is probably quite sensitive to the version of pandas.

dirty_cat/test/test_super_vectorizer.py
```diff
@@ -18,6 +18,17 @@ def check_same_transformers(expected_transformers: dict, actual_transformers: li
     actual_transformers_dict = dict([(name, cols) for name, trans, cols in actual_transformers])
     assert actual_transformers_dict == expected_transformers
 
+def type_equality(expected_type, actual_type):
+    """
+    Checks that the expected type is equal to the actual type, assuming object and str types
+    are equivalent (considered as categorical by the supervectorizer).
+    """
+    if (isinstance(expected_type , object) or  isinstance(expected_type, str))\
```
GaelVaroquaux, line 26: I don't need to protect a line return if it is in the middle of enclosing parenthesis.

LilianBoulard (commented):
Looks good !

CHANGES.rst
```diff
@@ -1,3 +1,17 @@
+Release 0.2.1
+=============
+
+Major changes
+-------------
+
+Notes
+-----
+
+* Improvements to the SuperVectorizer
```
GaelVaroquaux, line 10:
```suggestion
* Improvements to the :class:`SuperVectorizer`
```

dirty_cat/super_vectorizer.py
```diff
@@ -208,7 +198,30 @@ def _auto_cast(X: pd.DataFrame) -> pd.DataFrame:
                 if pd.api.types.is_numeric_dtype(X[col]):
                     X[col] = X[col].astype(np.float64)
                 X[col].fillna(value=np.nan, inplace=True)
+        STR_NA_VALUES = ['null', '', '1.#QNAN', '#NA', 'nan', '#N/A N/A', '-1.#QNAN', '<NA>', '-1.#IND', '-nan', 'n/a',
+                         '-NaN', '1.#IND', 'NULL', 'NA', 'N/A', '#N/A', 'NaN']  # taken from pandas.io.parsers (version 1.1.4)
+        X = X.replace(STR_NA_VALUES + [None, "?", "..."],
```
GaelVaroquaux, line 203: I'm surprised that this works if X is a numpy array. I actually think that it doesn't work.

It seems that we are indeed testing properly the SuperVectorizer with numpy arrays. I'm a bit confused :)

GaelVaroquaux (commented):
A few thoughts (I'm too tired to finish tonight).

GaelVaroquaux (approved):
LGTM. Merging. Thanks a lot!!

PR #239, opened 2022-02-19
Datetime encoder

LeoGrin:
Creates a new encoder which transform datetime columns into several numerical features (year, month, day...). Solves the second part of #233.

### DIFF
--- a/dirty_cat/datetime_encoder.py
+++ b/dirty_cat/datetime_encoder.py
@@ -1,0 +1,97 @@
+from sklearn.base import BaseEstimator, TransformerMixin
+import numpy as np
+import pandas as pd
+
+
+class DatetimeEncoder(TransformerMixin, BaseEstimator):
+    def __init__(self,
+                 stop_extraction="minute",
+                 add_day_of_the_week=True,
+                 add_holidays=False):
+        # TODO validate_keywords
+        # TODO doc
+        """
+        Parameters
+        ----------
+        stop_extraction : {"year", "month", "day", "hour", "minute", "second",
+        "millisecond", "microsecond", "nanosecond"}, default="minutes"
+            Extract up to this granularity. For instance, if you specify "days", only "year", "month", and "day"
+            features will be created.
+        add_day_of_the_week: bool, default=True
+            Add day of the week feature (if day is extracted).
+        add_holidays : bool, default=False
+            Whether to add a categorical variable encoding if the day of the date is a holiday. For now uses pandas
+            calendar, which only support US holidays.
+        """
+        to_extract = ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]
+        self.to_extract_full = to_extract[:to_extract.index(stop_extraction) + 1]
+        self.add_day_of_the_week = add_day_of_the_week
+        self.add_holidays = add_holidays
+        # number of new columns created per feature, before removing constant columns
+        self.n_features_per_col_full = len(self.to_extract_full) + self.add_day_of_the_week + self.add_holidays
+
+    def _extract_from_date(self, date_series, feature):
+        if feature == "year":
+            return pd.DatetimeIndex(date_series).year.to_numpy()
+        elif feature == "month":
+            return pd.DatetimeIndex(date_series).month.to_numpy()
+        elif feature == "day":
+            return pd.DatetimeIndex(date_series).day.to_numpy()
+        elif feature == "hour":
+            return pd.DatetimeIndex(date_series).hour.to_numpy()
+        elif feature == "minute":
+            return pd.DatetimeIndex(date_series).minute.to_numpy()
+        elif feature == "second":
+            return pd.DatetimeIndex(date_series).second.to_numpy()
+        elif feature == "millisecond":
+            return pd.DatetimeIndex(date_series).microsecond.to_numpy()
+        elif feature == "microsecond":
+            return pd.DatetimeIndex(date_series).microsecond.to_numpy()
+        elif feature == "nanosecond":
+            return pd.DatetimeIndex(date_series).nanosecond.to_numpy()
+        elif feature == "dayofweek":
+            return pd.DatetimeIndex(date_series).dayofweek.to_numpy()
+        elif feature == "holiday":
+            return None
+
+    def _create_datetime_features(self, X):
+        """Create datetime features
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, 1)
+            The input samples, containing 1 datetime column.
+        Returns
+        -------
+        X_new : array-like of shape (n_samples, n_features_out)
+            The transformed dataset.
+        """
+        X_ = np.empty((X.shape[0], self.n_features_per_col_full), dtype=np.int64)
+        for i, feature in enumerate(self.to_extract):
+            X_[:, i] = self._extract_from_date(X, feature)
+        return X_
+
+    def fit(self, X, y=None):
+        self.to_extract = {}  # Features to extract for each column, after removing constant features
+        for i in range(X.shape[1]):
+            self.to_extract[i] = []
+        # Check which columns are constant
+        for i in range(X.shape[1]):
+            for feature in self.to_extract_full:
+                if self._extract_from_date(X[:, i], feature).std() > 0:
+                    self.to_extract[i].append(feature)
+            if "day" in self.to_extract[i]:
+                if self.add_day_of_the_week:
+                    self.to_extract[i].append("dayofweek")
+                if self.add_holidays:
+                    self.to_extract[i].append("holiday")
+
+        self.n_features_out = len(np.concatenate(list(self.to_extract.values())))
+
+    def transform(self, X, y=None):
+        X_ = np.empty((X.shape[0], self.n_features_out), dtype=np.int64)
+        idx = 0
+        for i in range(X.shape[1]):
+            for j, feature in enumerate(self.to_extract[i]):
+                X_[:, idx + j] = self._extract_from_date(X[:, i], feature)
+            idx += len(self.to_extract[i])
+        return X_
--- a/dirty_cat/datetime_encoder.py
+++ b/dirty_cat/datetime_encoder.py
@@ -1,6 +1,7 @@
 from sklearn.base import BaseEstimator, TransformerMixin
 import numpy as np
 import pandas as pd
+from dirty_cat.utils import check_input
 from pandas.tseries.holiday import USFederalHolidayCalendar as calendar
 
 
@@ -9,6 +10,7 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
     """
     This encoder transforms each datetime column into several numeric columns corresponding to temporal features,
     e.g year, month, day...
+    If the dates are timezone aware, all the features extracted will correspond to the provided timezone.
     Parameters
     ----------
     extract_until : {"year", "month", "day", "hour", "minute", "second",
@@ -76,19 +78,22 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
             return np.isin(date_series, holidays).astype(int)
         elif feature == "other":
             # Gather all the variables below the extract_until into one numerical variable
-            res = (date_series - pd.to_datetime(pd.DatetimeIndex(date_series).floor(
+            res = (pd.to_datetime(date_series) - pd.to_datetime(pd.DatetimeIndex(date_series).floor(
                 self.word_to_alias[self.extract_until]))).to_numpy()
             # Convert to the extract_until unit (e.g if I extract until "minute", then convert to minutes)
             return res / pd.to_timedelta(1, self.word_to_alias[self.extract_until])
 
     def fit(self, X, y=None):
+        X = check_input(X)
         self.to_extract = {}  # Features to extract for each column, after removing constant features
         for i in range(X.shape[1]):
             self.to_extract[i] = []
         # Check which columns are constant
         for i in range(X.shape[1]):
-            print(i)
             for feature in self.to_extract_full:
+                print(feature)
+                print(self._extract_from_date(X[:, i], feature))
+                print(self._extract_from_date(X[:, i], feature).dtype)
                 if np.nanstd(self._extract_from_date(X[:, i], feature)) > 0:
                     self.to_extract[i].append(feature)
             if "day" in self.to_extract[i]:
@@ -102,6 +107,7 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
         return self
 
     def transform(self, X, y=None):
+        X = check_input(X)
         # Create a new dataframe with the extracted features, choosing only features that weren't constant during fit
         X_ = np.empty((X.shape[0], self.n_features_out), dtype=np.float64)
         idx = 0
--- a/dirty_cat/datetime_encoder.py
+++ b/dirty_cat/datetime_encoder.py
@@ -11,6 +11,7 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
     """
     This encoder transforms each datetime column into several numeric columns corresponding to temporal features,
     e.g year, month, day...
+    If one of the feature is constant, this feature will be dropped.
     If the dates are timezone aware, all the features extracted will correspond to the provided timezone.
 
     Parameters
@@ -20,15 +21,15 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
         Extract up to this granularity, and gather the rest into the "other" feature.
         For instance, if you specify "day", only "year", "month", "day" and "other" features will be created.
         The "other" feature will be a numerical value expressed in the "extract_until" unit.
-    add_day_of_the_week: bool, default=True
+    add_day_of_the_week: bool, default=False
         Add day of the week feature (if day is extracted). This is a numerical feature from 0 to 6.
     add_holidays : bool, default=False
         Whether to add a numerical variable encoding if the day of the date is a holiday.
-        Uses pandas calendar, which for now only supports US holidays.
+        Uses pandas calendar, which for now only contains US holidays.
     """
     def __init__(self,
                  extract_until="hour",
-                 add_day_of_the_week=True,
+                 add_day_of_the_week=False,
                  add_holidays=False):
         to_extract = ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]
         self.extract_until = extract_until
@@ -100,11 +101,10 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
             for feature in self.to_extract_full:
                 if np.nanstd(self._extract_from_date(X[:, i], feature)) > 0:
                     self.to_extract[i].append(feature)
-            if "day" in self.to_extract[i]:
-                if self.add_day_of_the_week:
-                    self.to_extract[i].append("dayofweek")
-                if self.add_holidays:
-                    self.to_extract[i].append("holiday")
+            if self.add_day_of_the_week:
+                self.to_extract[i].append("dayofweek")
+            if self.add_holidays:
+                self.to_extract[i].append("holiday")
 
         self.n_features_out = len(np.concatenate(list(self.to_extract.values())))
 
--- a/dirty_cat/datetime_encoder.py
+++ b/dirty_cat/datetime_encoder.py
@@ -7,6 +7,8 @@ from pandas.tseries.holiday import USFederalHolidayCalendar as calendar
 
 class DatetimeEncoder(TransformerMixin, BaseEstimator):
     """
+    This encoder transforms each datetime column into several numeric columns corresponding to temporal features,
+    e.g year, month, day...
     Parameters
     ----------
     extract_until : {"year", "month", "day", "hour", "minute", "second",
@@ -23,8 +25,6 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
                  extract_until="hour",
                  add_day_of_the_week=True,
                  add_holidays=False):
-        # TODO validate_keywords
-        # TODO doc
         to_extract = ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]
         self.extract_until = extract_until
         self._validate_keywords()
@@ -34,7 +34,7 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
         self.add_holidays = add_holidays
         # number of new columns created per feature, before removing constant columns
         self.n_features_per_col_full = len(self.to_extract_full) + self.add_day_of_the_week + self.add_holidays
-
+        # Some functions need aliases
         self.word_to_alias = {"year": "Y", "month": "M", "day": "D", "hour": "H", "minute": "min", "second": "S",
                               "millisecond": "ms", "microsecond": "us", "nanosecond": "N"}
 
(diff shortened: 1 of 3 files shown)


dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,173 @@
...
+from pandas.tseries.holiday import USFederalHolidayCalendar as calendar
+from typing import Union, Optional, List
+
+
+
+class DatetimeEncoder(TransformerMixin, BaseEstimator):
+    """
+    This encoder transforms each datetime column into several numeric columns corresponding to temporal features,
+    e.g year, month, day...
+    If one of the feature is constant, this feature is dropped.
```
LilianBoulard, line 14:
```suggestion
    Constant extracted features are dropped ; for instance, if the year is always the same in a feature, the extracted "year" column won't be added.
```

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,173 @@
...
+            for j, feature in enumerate(self.features_per_column[i]):
+                X_[:, idx + j] = self._extract_from_date(X[:, i], feature)
+            idx += len(self.features_per_column[i])
+        return X_
+
+    def get_feature_names(self) -> List[str]:
+        """
+        Returns clean feature names with format "<column_name>_<new_feature>"
+        if the original data has column names, otherwise with format
+        "<column_indice>_<new_feature>". new_feature is one of ["year", "month",
```
LilianBoulard, line 157:
```suggestion
        "<column_index>_<new_feature>". new_feature is one of ["year", "month",
```

dirty_cat/test/test_datetime_encoder.py
```diff
@@ -0,0 +1,183 @@
...
+    expected_result = np.array([[2020, 1, 1, 10, 12 / 60 + 1 / 3600, 2],
+                                [np.nan, np.nan, np.nan, np.nan, np.nan, np.nan],
+                                [2022, 1, 1, 23, 23 / 60 + 43 / 3600, 5],
+                                [2023, 2, 3, 11, 12 / 60 + 12 / 3600, 4]])
+    enc.fit(X)
+    X_trans = enc.transform(X)
+    assert np.allclose(X_trans, expected_result, equal_nan=True)
+
+    # Datetimes with TZ
+    # If the date are timezone-aware, all the feature extraction should be done
```
LilianBoulard, line 171:
```suggestion
    # If the dates are timezone-aware, all the feature extractions should be done
```

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,173 @@
...
+        Fit the DatetimeEncoder to X. In practice, just stores which extracted features
+        are not constant.
+
+        Parameters
+        ----------
+        X : array-like, shape [n_samples, n_features]
+            Data where each column is a datetime feature.
+        Returns
+        -------
+        self
```
LilianBoulard, line 108:
```suggestion
        self
            Fitted DatetimeEncoder instance.
```

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,173 @@
...
+        elif feature == "hour":
+            return pd.DatetimeIndex(date_series).hour.to_numpy()
+        elif feature == "minute":
+            return pd.DatetimeIndex(date_series).minute.to_numpy()
+        elif feature == "second":
+            return pd.DatetimeIndex(date_series).second.to_numpy()
+        elif feature == "millisecond":
+            return pd.DatetimeIndex(date_series).microsecond.to_numpy()
+        elif feature == "microsecond":
+            return pd.DatetimeIndex(date_series).microsecond.to_numpy()
```
LilianBoulard, line 80: I don't get why it returns the same exact same thing when specifying "millisecond" and "microsecond", is there a reason ?

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,173 @@
...
+        if the original data has column names, otherwise with format
+        "<column_indice>_<new_feature>". new_feature is one of ["year", "month",
+        "day", "hour", "minute", "second", "millisecond", "microsecond",
+        "nanosecond", "dayofweek", "holiday"]
+        """
+        feature_names = []
+        for i in self.features_per_column.keys():
+            prefix = str(i) if self.colnames is None else self.colnames[i]
+            for feature in self.features_per_column[i]:
+                feature_names.append(prefix + "_" + feature)
```
LilianBoulard, line 165:
```suggestion
                feature_names.append(f"{prefix}_{feature}")
```

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,173 @@
...
+
+    def _validate_keywords(self):
+        if self.extract_until not in ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]:
+            msg = (
+                """extract_until should be one of ["year", "month", "day", "hour", "minute", "second", "millisecond", 
+                "microsecond", "nanosecond"], got {0}.""".format(
+                    self.extract_until
+                )
+            )
+            raise ValueError(msg)
```
LilianBoulard, line 62: I'd advise using modern python f-strings, which are usually nicer to read :)

```suggestion
        if self.extract_until not in ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]:
            raise ValueError(
                f'"extract_until" should be one of ["year", "month", "day", "hour", "minute", '
                f'"second", "millisecond", "microsecond", "nanosecond"], got {self.extract_until}.'
            )
```

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,173 @@
...
+    """
+    def __init__(self,
+                 extract_until="hour",
+                 add_day_of_the_week=False,
+                 add_holidays=False):
+        to_extract = ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]
+        self.extract_until = extract_until
+        self._validate_keywords()
+        self.to_extract_full = to_extract[:to_extract.index(extract_until) + 1]
+        self.to_extract_full.append("other")
```
LilianBoulard, line 47: Following the previous comment, you should probably move these lines somewhere in your methods, to avoid raising an IndexError.

LilianBoulard (changes requested):
Here's bunch of small adjustments, but overall this is very good ! Thank you :)

dirty_cat/datetime_encoder.py
LeoGrin: Oooops

LeoGrin:
Thank you for the review @LilianBoulard !

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+                 extract_until="hour",
+                 add_day_of_the_week=False,
+                 add_holidays=False):
+        self.to_extract = ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]
+        self.extract_until = extract_until
+        self.add_day_of_the_week = add_day_of_the_week
+        self.add_holidays = add_holidays
+        # Some functions need aliases
+        self.word_to_alias = {"year": "Y", "month": "M", "day": "D", "hour": "H", "minute": "min", "second": "S",
+                              "millisecond": "ms", "microsecond": "us", "nanosecond": "N"}
```
GaelVaroquaux, line 49: The convention in scikit-learn is that the init should do nothing other than set parameters handed in as inputs. One reason is to have a very well understood statefulness of the objects.

These can be set at the beginning of fit.

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+
+    def _validate_keywords(self):
+        if self.extract_until not in ["year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"]:
```
GaelVaroquaux, line 52: You could probably define a module-level constant (constants are in ALLCAPS) and reuse it in the file, as is seems that you are using this set quite often.

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+        ----------
+        X : array-like, shape [n_samples, n_features]
+            Data where each column is a datetime feature.
+        Returns
+        -------
+        self
+            Fitted DatetimeEncoder instance.
+        """
+        self._validate_keywords()
+        self.to_extract_full = self.to_extract[:self.to_extract.index(self.extract_until) + 1]
```
GaelVaroquaux, line 106: If you are never modifying self.to_extract, you can put it as a class variable rather than set it in the init.

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+                self.features_per_column[i].append("dayofweek")
+            if self.add_holidays:
+                self.features_per_column[i].append("holiday")
+
+        self.n_features_out_ = len(np.concatenate(list(self.features_per_column.values())))
+
+        return self
+
+    def transform(self, X, y=None):
+        """ Transform X by replacing each column with corresponding numerical features.
```
GaelVaroquaux, line 131: I think that you are missing an empty line here for this to be valid docstring formatting.

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+        Parameters
+        ----------
+        X : array-like, shape (n_samples, ) or (n_samples, 1)
+            The data to transform, where each column is a datetime feature.
```
GaelVaroquaux, line 135: Also missing an empty line here

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+
+        self.n_features_out_ = len(np.concatenate(list(self.features_per_column.values())))
+
+        return self
+
+    def transform(self, X, y=None):
+        """ Transform X by replacing each column with corresponding numerical features.
+        Parameters
+        ----------
+        X : array-like, shape (n_samples, ) or (n_samples, 1)
```
GaelVaroquaux, line 134: If the line below is correct, than the line above is incorrect: X can actually be of shape (n_samples, n_features)

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+        self
+            Fitted DatetimeEncoder instance.
+        """
+        self._validate_keywords()
+        self.to_extract_full = self.to_extract[:self.to_extract.index(self.extract_until) + 1]
+        self.to_extract_full.append("other")
+        if isinstance(X, pd.DataFrame):
+            self.colnames = X.columns
+        else:
+            self.colnames = None
```
GaelVaroquaux, line 111: Given that `self.colnames` is determined from the data, it should end with an underscore: `self.colnames_`.

In addition, we prefer to separate words with underscores (to make them easier to read): `self.col_names_`

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+        X = check_input(X)
+        self.features_per_column = {}  # Features to extract for each column, after removing constant features
```
GaelVaroquaux, line 113: This should also end with an underscore

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+        ----------
+        X : array-like, shape (n_samples, ) or (n_samples, 1)
+            The data to transform, where each column is a datetime feature.
+        Returns
+        -------
+        array, shape (n_samples, n_features_out_)
+            Transformed input.
+        """
+        X = check_input(X)
+        # Create a new dataframe with the extracted features, choosing only features that weren't constant during fit
```
GaelVaroquaux, line 142:
```suggestion
        # Create a new array with the extracted features, choosing only features that weren't constant during fit
```

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+        "nanosecond", "dayofweek", "holiday"]
+        """
+        feature_names = []
+        for i in self.features_per_column.keys():
+            prefix = str(i) if self.colnames is None else self.colnames[i]
+            for feature in self.features_per_column[i]:
+                feature_names.append(f"{prefix}_{feature}")
+        return feature_names
+
+    def get_feature_names_out(self, input_features=None) -> List[str]:
```
GaelVaroquaux, line 166: Can you invert get_feature_names and get_feature_names_out: get_feature_names will be removed in a little while, when we no longer depend on old versions of scikit-learn.

dirty_cat/test/test_datetime_encoder.py
```diff
@@ -0,0 +1,183 @@
+import pandas as pd
+from dirty_cat.datetime_encoder import DatetimeEncoder
+import numpy as np
+
+
+def get_date_array():
+    return np.array([pd.to_datetime(["2020-01-01", "2020-01-02", "2020-01-03"]),
+                  pd.to_datetime(["2021-02-03", "2020-02-04", "2021-02-05"]),
+                  pd.to_datetime(["2022-01-01", "2020-12-25", "2022-01-03"]),
+                  pd.to_datetime(["2023-02-03", "2020-02-04", "2023-02-05"])])
```
GaelVaroquaux, line 10: In terms of indentation, it would be great if you indented a bit more this list, to align the "pd" to the above "pd" on the first line (comment also applies below in a few places)

dirty_cat/test/test_datetime_encoder.py
```diff
@@ -0,0 +1,183 @@
...
+
+def get_datetime_array():
+    return np.array([pd.to_datetime(["2020-01-01 10:12:01", "2020-01-02 10:23:00", "2020-01-03 10:00:00"]),
+                  pd.to_datetime(["2021-02-03 12:45:23", "2020-02-04 22:12:00", "2021-02-05 12:00:00"]),
+                  pd.to_datetime(["2022-01-01 23:23:43", "2020-12-25 11:12:00", "2022-01-03 11:00:00"]),
+                  pd.to_datetime(["2023-02-03 11:12:12", "2020-02-04 08:32:00", "2023-02-05 23:00:00"])])
```
GaelVaroquaux, line 16: PEP8 style guide says that there should be 2 empty lines between first-level functions (comment applies in quite a few places)

GaelVaroquaux (changes requested):
Nice!

I left a few comments, but none of them are major.

The new object should be added to the list of encoders on the first page (doc/index.rst).

We also need an example demonstrating the new object.

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
+from sklearn.base import BaseEstimator, TransformerMixin
+import numpy as np
+import pandas as pd
+from dirty_cat.utils import check_input
+from pandas.tseries.holiday import USFederalHolidayCalendar as calendar
```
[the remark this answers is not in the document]
LeoGrin: Yes I’m not sure. Maybe it’s okay as it is deactivated by default, and the documentation states that it only supports US holidays for now? Otherwise, we can maybe use https://github.com/dr-prodigy/python-holidays. Do you think a new dependency is worth it?

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+    Constant extracted features are dropped ; for instance, if the year is always the same in a feature, the extracted "year" column won't be added.
+    If the dates are timezone aware, all the features extracted will correspond to the provided timezone.
+
+    Parameters
+    ----------
+    extract_until : {"year", "month", "day", "hour", "minute", "second",
+    "millisecond", "microsecond", "nanosecond"}, default="hour"
+        Extract up to this granularity, and gather the rest into the "other" feature.
+        For instance, if you specify "day", only "year", "month", "day" and "other" features will be created.
+        The "other" feature will be a numerical value expressed in the "extract_until" unit.
```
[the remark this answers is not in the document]
LeoGrin: Hmm interesting. I get you point, but here's why I chose to do it this way:
- I fear that full_time_to_epoch would be very collinear with the highest-level feature, hurting interpretability. For a worst case scenario, we can imagine a dataset of different hours during the same day for different years.
- It’s true that with my choice, the learner has to learn the weird algebra. But with your proposition, the learner would still need to learn this weird algebra if it wants to use my `other` feature (which may be more important for prediction since it’s information which isn’t contained in the other variables), because `other` = full_time_to_epoch - (year - 1970) * 365 * 24 * 3600 + …
Maybe using both could indeed make the learning easier, but wouldn't you worry about collinearity?
All in all I'm not sure.

GaelVaroquaux:
> Do you think a new dependency is worth it?

No, I don't think that a new dependency is worth it.

I'm wondering if we should simply drop this feature (less features = less code = less problems :D )

dirty_cat/datetime_encoder.py
```diff
@@ -0,0 +1,171 @@
...
+    """
+    This encoder transforms each datetime column into several numeric columns corresponding to temporal features,
+    e.g year, month, day...
+    Constant extracted features are dropped ; for instance, if the year is always the same in a feature, the extracted "year" column won't be added.
+    If the dates are timezone aware, all the features extracted will correspond to the provided timezone.
+
+    Parameters
+    ----------
+    extract_until : {"year", "month", "day", "hour", "minute", "second",
+    "millisecond", "microsecond", "nanosecond"}, default="hour"
```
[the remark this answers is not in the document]
LeoGrin: It was a mess indeed :)

dirty_cat/datetime_encoder.py
LeoGrin: I've made it a module-level constant

LeoGrin:
> I'm wondering if we should simply drop this feature
I've removed it.

LilianBoulard (approved):

dirty_cat/datetime_encoder.py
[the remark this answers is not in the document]
GaelVaroquaux: I do machine learning. I don't worry about collinearity :).

It does not hurt for prediction. I can hurt for interpretability.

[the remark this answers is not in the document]
GaelVaroquaux: My reason for favoring the full_time_to_epoch is that I think that, in general, it is more likely to be useful than the other end, and we are more likely to have to reconstruct it.

GaelVaroquaux (changes requested):
I think that:
- I still prefer to have the full time to epoch
- We are lacking an example (examples are crucial). We probably need to add a new example about this specific topic.

LeoGrin:
Additions:
- Replaced the "other" feature by the "full" features, which contains the full time to epoch (see above discussion with @GaelVaroquaux)
- Made the DatetimeEncoder the default for datetime columns in the SuperVectorizer
- Added a simple example
- Small fixes

LeoGrin:
Apparently using fetch_traffic_violations() in the example takes too much memory for circleCI

LeoGrin:
I've changed the example to use another dataset, I think the PR is ready for review @LilianBoulard @GaelVaroquaux @jovan-stojanovic

examples/06_datetime_encoder.py
```diff
@@ -0,0 +1,103 @@
...
+###############################################################################
+# One-liner with the SuperVectorizer
+# ----------------------------
+# The DatetimeEncoder is used by default in the SuperVectorizer, which automatically detects datetime features.
+from dirty_cat import SuperVectorizer
+
+sup_vec = SuperVectorizer()
+X_ = sup_vec.fit_transform(X)
+feature_names = sup_vec.get_feature_names_out()
+print(feature_names)
```
GaelVaroquaux, line 92:
```suggestion
sup_vec.get_feature_names_out()
```
I think that the code will be easier to read and the formatting of the example will be nicer like this.

examples/06_datetime_encoder.py
```diff
@@ -0,0 +1,103 @@
...
+###############################################################################
+# We can see that the SuperVectorizer is indeed using a DatetimeEncoder for the datetime features.
+sup_vec.transformers_
+
+###############################################################################
+# If we want the day of the week, we can just replace SuperVectorizer's default
+sup_vec = SuperVectorizer(datetime_transformer=DatetimeEncoder(add_day_of_the_week=True))
+X_ = sup_vec.fit_transform(X)
+feature_names = sup_vec.get_feature_names_out()
+print(feature_names)
```
GaelVaroquaux, line 103:
```suggestion
sup_vec.get_feature_names_out()
```

examples/06_datetime_encoder.py
```diff
@@ -0,0 +1,103 @@
+"""
+Handling datetime features with the DatetimeEncoder
+==========================================
+
+We illustrate here how to handle datetime features with the DatetimeEncoder. The DatetimeEncoder
+breaks down each datetime features into several numerical features, by extracting relevant information from the
+datetime features, such as the month, the day of the week, the hour of the day, etc. Used in
+the SuperVectorizer, which automatically detects the datetime features, the DatetimeEncoder allows
+to handle datetime features easily.
```
GaelVaroquaux, line 9: Would it be possible to do a small prediction model at the end, add a section, to showcase this in a prediction pipeline?

For didactic purposes it would be important to have not only the cross-validation (using https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.TimeSeriesSplit.html ) but also the plot of the extrapolated time series (compared to the real one)

dirty_cat/datetime_encoder.py
```diff
@@ -20,9 +20,9 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
     Parameters
     ----------
     extract_until : {"year", "month", "day", "hour", "minute", "second", "millisecond", "microsecond", "nanosecond"}, default="hour"
-        Extract up to this granularity, and gather the rest into the "other" feature.
-        For instance, if you specify "day", only "year", "month", "day" and "other" features will be created.
-        The "other" feature will be a numerical value expressed in the "extract_until" unit.
+        Extract up to this granularity. If all features have not been extracted, add the "full" feature, which contains
```
GaelVaroquaux, line 23: Maybe call it "total_time" rather than "full". I had not understood what "full" meant.

GaelVaroquaux (changes requested):
A few changes requested. Most are minor. I hope that the prediction example won't ask for too much work.

LeoGrin:
Thank you for the comments! I've added a new section to the example

GaelVaroquaux:
Oups, I just saw that there is a "examples/.DS_Store" that needs to be removed. You can add it in the gitignore if it's helpful.

examples/06_datetime_encoder.py
```diff
@@ -0,0 +1,159 @@
...
+categorical_columns = ["city"]
+
+encoder = make_column_transformer((cat_encoder, categorical_columns),
+                                  (datetime_encoder, datetime_columns),
+                                  remainder="drop")
+
+
+###############################################################################
+# Transforming the input data
+# ----------------------------
```
GaelVaroquaux, line 49:
```suggestion
# ..........................
```
Decreasing the number of top-level sections (to make the table of content in the side bar more readable)

examples/06_datetime_encoder.py
```diff
@@ -0,0 +1,159 @@
...
+# We can see that the encoder is working as expected: the date feature has
+# been replaced by features for the month, day, hour, and day of the week.
+# Note that the year and minute features have been removed by the encoder because they are constant.
+X_ = encoder.fit_transform(X)
+feature_names = encoder.get_feature_names_out()
+feature_names
+
+###############################################################################
+# Features importance
```
GaelVaroquaux, line 58: I think that I would move the feature importance section below the prediction example. It feels more natural to understand feature importances in such contexte

GaelVaroquaux (commented):
Still cosmetic comments on the example :)

GaelVaroquaux:
Hum, I merged this, but I hadn't seen that the lines are too long (I'll fix it, I wanted to do cosmetics on the example anyhow).

It does have a practical consequence: it forces horizontal scrolling in the example, which is bad for readability.

GaelVaroquaux:
> This actually made me wonder: in which case should I use a class attribute vs a module-level constant?

Use a class attribute if you might want to override it in a subclass.

PR #241, opened 2022-03-18
Follow scikit-learn API change

GaelVaroquaux:
get_feature_names -> get_feature_names_out

### DIFF
--- a/dirty_cat/gap_encoder.py
+++ b/dirty_cat/gap_encoder.py
@@ -14,6 +14,7 @@ The principle is as follows:
        with the Kullback-Leibler divergence as loss, and a Gamma prior on H.
        We thus optimize H and W with the multiplicative update method.
 """
+import warnings
 import numpy as np
 from distutils.version import LooseVersion
 from scipy import sparse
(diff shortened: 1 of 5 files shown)


LilianBoulard:
I think we should check whether the installed version of sklearn is > 1.0, and warn only if it is, as somebody using sklearn < 1.0  would be using the previously standard method (`get_feature_names`), and get these warnings.
What do you think ?

GaelVaroquaux:
Thanks for the review!

Actually, I think that I want to push people to change their code ASAP, because it is dangerous to have code using APIs that are deprecated. For dirty-cat, if they see this warning, they can do the change.

LilianBoulard (approved):

lmeyerov:
@GaelVaroquaux This seems to have missed https://github.com/dirty-cat/dirty_cat/blob/fb0a1c4216533034e7516efc0698c7e4477b0243/dirty_cat/super_vectorizer.py#L471


```
pygraphistry/lib/python3.8/site-packages/dirty_cat/super_vectorizer.py:471: in get_feature_names_out
    trans_feature_names = trans.get_feature_names(cols)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
args = (OneHotEncoder(), ['user', 'type']), kwargs = {}
    @functools.wraps(fun)
    def wrapped(*args, **kwargs):
>       warnings.warn(msg, category=FutureWarning)
E       FutureWarning: Function get_feature_names is deprecated; get_feature_names is deprecated in 1.0 and will be removed in 1.2. Please use get_feature_names_out instead.
pygraphistry/lib/python3.8/site-packages/sklearn/utils/deprecation.py:87: FutureWarning
```

PR #261, opened 2022-06-07
Remove warnings created by version checks

### LINKED ISSUE #260: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
Since december 2021, `distutils.version.Version` (and thus `LooseVersion`) [is deprecated](https://github.com/pypa/setuptools/commit/1701579e0827317d8888c2254a17b5786b6b5246).
We use it in [several components](https://github.com/dirty-cat/dirty_cat/search?q=LooseVersion), and therefore we have a number of warnings in the CI logs ([for example](https://github.com/dirty-cat/dirty_cat/runs/6667648188?check_suite_focus=true), under "Run tests")

Sample:
```
dirty_cat/test/test_similarity_encoder.py: 118 warnings
  /Users/runner/work/dirty_cat/dirty_cat/dirty_cat/similarity_encoder.py:328: DeprecationWarning: distutils Version classes are deprecated. Use packaging.version instead.
    if LooseVersion(sklearn.__version__) > LooseVersion('0.21'):
```

as suggested by the warning, we could use `packaging.version.Version` instead.

LilianBoulard:
Fixes #260 by replacing `distutils.version.LooseVersion` by `packaging.version.Version`.  
Therefore, it also adds `packaging` to the requirements.


LilianBoulard:
Following the previous point, I've added a little replacement class that works pretty much like the two other packages.
I believe it should be enough for our needs !

GaelVaroquaux:
I like the idea, but we have a lot of failing tests right now :)

PR #263, opened 2022-06-09
Add introductory video on the website

LilianBoulard:
For a congress I recently participated in, I realized a video presenting some encoders implemented in dirty_cat.
The video [is available on YouTube](https://youtu.be/_GNaaeEI2tg) and I though we could add it to the website if you think this is relevant :)

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -49,6 +49,10 @@ pipeline you can tweak.
     </div>
     </div>
 
+.. raw:: html
+
+    <iframe src="https://youtu.be/_GNaaeEI2tg" frameborder="0" allowfullscreen></iframe>
+
 For a detailed description of the problem of encoding dirty categorical data,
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
I think that it's a great idea to add it to the website.

I however think that it is not positioned in the right place: it breaks the flow of information.

Maybe we could do a two-column layout where it is currently positioned (using a proper responsive design that falls back to a different layout on narrow screens).

LilianBoulard:
Thanks for the feedback, I adjusted the style :)
What do you think [about this](https://output.circle-artifacts.com/output/job/c367a704-6e96-4169-bbd4-99a406963c54/artifacts/0/doc/index.html) ?
Would be better if the columns where switched ?

GaelVaroquaux:
I think that this is good. Thanks! Merging!

PR #266, opened 2022-06-29
Display column names instead of indices for remainder columns in the transformers_ attribute of the SuperVectorizer

LeoGrin:
Solves #232 
Following discussion with @GaelVaroquaux in previous PR (#236), I've made the modification in the fit method of the SuperVectorizer.

### DIFF
--- a/dirty_cat/super_vectorizer.py
+++ b/dirty_cat/super_vectorizer.py
@@ -430,10 +430,11 @@ class SuperVectorizer(ColumnTransformer):
         res = super().fit_transform(X, y)
 
         # for the "remainder" columns, the ColumnTransformer transformers_ attribute
-        # contains the index instead of the column name so we convert it to the column name.
+        # contains the index instead of the column name, so we convert it to the column name
+        # if there is less than 20 columns in the remainder.
         for i, tup in enumerate(self.transformers_):
             name, enc, cols = tup  # Unpack
-            if name == "remainder":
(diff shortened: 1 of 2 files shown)


dirty_cat/super_vectorizer.py
```diff
@@ -427,7 +427,16 @@ def fit_transform(self, X, y=None):
         if self.verbose:
             print(f'[SuperVectorizer] Assigned transformers: {self.transformers}')
 
-        return super().fit_transform(X, y)
+        res = super().fit_transform(X, y)
+
+        # for the "remainder" columns, the ColumnTransformer transformers_ attribute
+        # contains the index instead of the column name so we convert it to the column name.
+        for i, tup in enumerate(self.transformers_):
+            name, enc, cols = tup  # Unpack
```
GaelVaroquaux, line 435: Maybe let's do this only if len(cols) < 20, if you agree?

GaelVaroquaux (changes requested):
Thank you. That's a great feature. I made a minor suggestion.

dirty_cat/super_vectorizer.py
LeoGrin: Oh yes, agreed!

GaelVaroquaux (approved):

GaelVaroquaux:
Merged! Thanks

PR #267, opened 2022-06-29
ENH MinHash parallel

LeoGrin:
Compute the min hash transform method in parallel, as suggested by @alexis-cvetkov.

We no longer use the self.hash_dict attribute, so the fit method does nothing now.


LeoGrin:
I don't understand why the coverage has changed. It seems that the function called by joblib.Parallel (`compute_hash`) is not counted in the coverage, but I may be reading codecov wrong.

GaelVaroquaux:
I worry that the self.hash_dict was really useful to speed things up by avoiding recomputation of repeated entries

LeoGrin:
We use np.unique, so for one transform we don't recompute repeated entries.
Using self.hash_dict would indeed speed things up if we transform several
inputs with common entries, using the same encoder. Do you think that's
likely to happen / worth the additional memory usage ?

GaelVaroquaux:
> Do you think that's likely to happen / worth the additional memory usage ?

Yes: it's very frequent. Typically, the entries are repeated many times.

LeoGrin:
@GaelVaroquaux just want to make sure I understood what you were saying before putting the hash_dict back in the code.

LeoGrin:
Following discussion with @GaelVaroquaux : using the same encoder to transform several Xs may happen in online learning settings, for instance with a big X.
I've put the hash_dict back in

jovan-stojanovic (changes requested):
To improve coverage, maybe you can add a test for the min_hash asserting that the parameters need to have the right value.
If not, the appropriate error is raised (for instance when the `hashing` parameter is neither `fast` nor `murmur`).

Otherwise looks good!

Thanks!

dirty_cat/minhash_encoder.py
```diff
@@ -172,56 +177,47 @@ def transform(self, X):
...
-                    elif x not in self.hash_dict:
-                        X_out[i, k * self.n_components:counter] = self.hash_dict[x] = self.minhash(
+        if not (X == X).all():  # contains at least one missing value
+            if self.handle_missing == 'error':
+                raise ValueError("Found missing values in input data; set "
+                                 "handle_missing='zero_impute' to encode with missing values")
+            elif self.handle_missing == 'zero_impute':
+                X[~(X == X)] = "NAN"  # this will be replaced by a vector of zeroes by compute_hash
+
+        def compute_hash(x):
```
GaelVaroquaux, line 197: Please put this function outside as a first level citizen and not nested inside. Making a function easy to important facilitates tests, reuse, and persistence.

dirty_cat/minhash_encoder.py
```diff
@@ -172,56 +177,47 @@ def transform(self, X):
...
-            raise ValueError(msg)
-        return X_out
+                        raise ValueError("hashing function must be 'fast' or"
+                                         "'murmur', got '{}'"
+                                         "".format(self.hashing))
+            return self.hash_dict[x]
+
+        # Compute the hashes for unique values
+        unique_x, indices_x = np.unique(X, return_inverse=True)
+        unique_x_trans = Parallel(n_jobs=self.n_jobs)(delayed(compute_hash)(x) for x in unique_x)
```
GaelVaroquaux, line 219: I think that we should apply things on batches rather than on single rows. I fear that single rows will lead to overhead.

Maybe we should to to use 
https://scikit-learn.org/stable/modules/generated/sklearn.utils.gen_even_slices.html
as in
https://github.com/scikit-learn/scikit-learn/blob/703bee65e2122ad273cf0b42460c5c28ca638af8/sklearn/decomposition/_lda.py#L460

We should benchmark this to see what the fastest / best option. Talk to @jjerphan, he has good experience in benchmarking parallel computing.

dirty_cat/minhash_encoder.py
```diff
@@ -47,6 +47,9 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
     handle_missing : 'error' or 'zero_impute' (default)
         Whether to raise an error or encode missing values (NaN) with
         vectors filled with zeros.
+    n_jobs : int, default=-1
```
jovan-stojanovic, line 50: The default n_jobs should be `None` as inspired from [here (Jo Blib)](https://joblib.readthedocs.io/en/latest/generated/joblib.Parallel.html)

dirty_cat/minhash_encoder.py
LeoGrin: Benchmarking really quickly on my Mac M1 with 8 cores, the rows version is about twice faster than the batched version, but we should do a more serious benchmark.

GaelVaroquaux:
> Benchmarking really quickly on my Mac M1 with 8 cores, the rows version is about twice faster than the batched version, but we should do a more serious benchmark.

On a dataset of which size (number of rows)?

dirty_cat/_minhash_encoder.py
```diff
@@ -155,6 +162,41 @@ def get_fast_hash(self, string: str) -> np.array:
                 ]
             )
 
+    def compute_hash(self, x) -> np.array:
```
LilianBoulard, line 165: Type hint the variable (should also be renamed, `x` is not descriptive), fix return type (`np.array` is a function, `np.ndarray` is the type it returns), and make the method private.
```suggestion
    def _compute_hash(self, x: str) -> np.ndarray:
```

dirty_cat/_minhash_encoder.py
```diff
@@ -155,6 +162,41 @@ def get_fast_hash(self, string: str) -> np.array:
...
+                elif self.hashing == "murmur":
+                    self.hash_dict_[x] = self.minhash(
+                        x,
+                        n_components=self.n_components,
+                        ngram_range=self.ngram_range
+                    )
+                else:
+                    raise ValueError("hashing function should be either 'fast' or"
+                                     "'murmur', got '{}'"
+                                     "".format(self.hashing))
```
LilianBoulard, line 197: f-strings for the win :)
```suggestion
                    raise ValueError("hashing function should be either 'fast' or"
                                     f"'murmur', got '{self.hashing!r}'")
```

dirty_cat/_minhash_encoder.py
```diff
@@ -192,70 +234,33 @@ def transform(self, X) -> np.array:
...
-                self.n_components % 2 == 0
-            ), "n_components should be even when minmax_hash=True. "
-        if self.hashing == "murmur":
-            assert (
-                not self.minmax_hash
-            ), 'minmax_hash is not implemented with hashing="murmur". '
+            if self.n_components % 2 != 0:
+                raise ValueError("n_components should be even when using"
+                                 "minmax_hash encoding, got {}"
+                                 "".format(self.n_components))
```
LilianBoulard, line 240:
```suggestion
                raise ValueError("n_components should be even when using"
                                 f"minmax_hash encoding, got {self.n_components!r}")
```

dirty_cat/_minhash_encoder.py
```diff
@@ -192,70 +234,33 @@ def transform(self, X) -> np.array:
...
+        if self.hashing == 'murmur':
+            if self.minmax_hash:
+                raise ValueError("minmax_hash encoding is not supported"
+                                 "with murmur hashing function")
+        if self.handle_missing not in ['error', 'zero_impute']:
+            template = ("handle_missing should be either 'error' or "
+                        "'zero_impute', got %s")
+            raise ValueError(template % self.handle_missing)
```
LilianBoulard, line 248:
```suggestion
            raise ValueError("handle_missing should be either 'error' or "
                             f"'zero_impute', got {self.handle_missing!r}")
```

dirty_cat/tests/test_minhash_encoder.py
```diff
@@ -170,3 +171,100 @@ def test_min_hash_encoder_hashing_fast(idx):
...
+
+DEFAULT_JOBLIB_BACKEND = joblib.parallel.get_active_backend()[0].__class__
+class MyBackend(DEFAULT_JOBLIB_BACKEND):  # type: ignore
+    def __init__(self, *args, **kwargs):
+        self.count = 0
+        super().__init__(*args, **kwargs)
+
+    def start_call(self):
+        self.count += 1
+        return super().start_call()
```
LilianBoulard, line 201: Could use another name, and also a short docstring describing why we need this, and what it does.

dirty_cat/tests/test_minhash_encoder.py
```diff
@@ -170,3 +171,100 @@ def test_min_hash_encoder_hashing_fast(idx):
...
+    print("{:.4} seconds".format(
+        profile_encoder(MinHashEncoder, hashing='murmur')))
+    print("Parallelism test")
+    test_parallelism()
+    print('joblib backend test')
+    test_backend_respected()
+    print('correct arguments test')
+    test_correct_arguments()
+
+    print('Done')
```
LilianBoulard, line 270: We ditched this section in other tests, so we should too in this one :)

CHANGES.rst
```diff
@@ -60,6 +60,9 @@ Notes
 * The ``transformers_`` attribute of the SuperVectorizer now contains column
   names instead of column indices for the "remainder" columns.
 
+* The MinHashEncoder now supports a `n_jobs` parameter to parallelize
+the hashes computation.
+
```
LilianBoulard, line 65: I think this could be added to the "Major changes" of version 0.4

LilianBoulard (changes requested):
A few remarks, but otherwise that's a good PR!
Also, please pull the main branch, there are some conflicts.
I'll be working on some benchmarks; you're welcome to take part in that :)

jjerphan:
As discussed with @jovan-stojanovic privately regarding understanding the current implementation performance: one can profile it using profilers like [`py-spy`](https://github.com/benfred/py-spy) and visually search for bottlenecks using [Speed-scope](https://github.com/jlfwong/speedscope)

In that case, one can get a `py-spy` profile for Speedcope using: 
```bash
py-spy record --native -o py-spy-profile.svg -f speedscope -- python ./script.py
```

where `script.py` is the script using `MinHash` on a suitable dataset for `dirty_cat`.

If `MinHash` relies on `joblib` for parallelization, it might be better to use a sole thread (by setting `n_jobs=1`) so as to better understand the execution when inspecting the profile.

After having uploaded the profile on [Speedscope's UI](https://speedscope.app/), you will be able to have access to various representations:
 - `"Left Heavy"` to get a flame graph recapitulating the total runtime of each functions, we the heavier first
 - `"Time Order"` to understand the order of function call in their context.
 - `"Sandwich"` for fine-grained information for each interfaces

jovan-stojanovic:
The changelog test apparently works fine :) You should add the PR number to CHANGES.rst

LilianBoulard (approved):

LeoGrin:
- I've run the benchmark (thanks Lilian) on Margaret, and it confirms that the batched version seems to always be better. I've also added to the benchmark a `batch_per_job` parameter, but it had no influence on the results.

- I've put the minhashencoder code with all options in the bench_minhash function.

- In the original minhashencoder file, I've only kept the batched version, and removed the parameter to choose.

Ready for review!

LilianBoulard (approved):
Much better!

LilianBoulard:
Just waiting for Jovan's review, and we'll be ready to merge :)

benchmarks/bench_minhash.py
```diff
@@ -0,0 +1,460 @@
+#################
+# The MinHashEncoder version used for the benchmark
+# On the main branch, we only kept the best version of the MinHashEncoder
+# which is the batched version
+# with batch_per_job=1
+# (the batch_per_job parameter has no effect on the results)
```
jovan-stojanovic, line 6: From what I see, batches are useful in the end.
The detailed benchmarking results also depend on the number of batches per job, is it because the gain is not that big that we don't leave it as a choice?

jovan-stojanovic (commented):
Thanks! Just two minor comments/questions.

benchmarks/bench_minhash.py
LeoGrin: Yes the impact of `batch_per_job` is very small, and the impact of `batched` is large but seems quite constant. In both cases, I don't think it is worth it to have additional parameters, both for ease of use and ease of maintenance.

jovan-stojanovic:
Merging! Thanks @LeoGrin

PR #274, opened 2022-07-04
Pass scikit learn estimator tests

LeoGrin:
(Partially) solving #164 
Various changes to pass scikit learn `check_estimator` tests, including:

- No more transformers as default parameters in SuperVectorizer (they should be mutable)
- No more variables other than parameters set in _init_ in the dirty cat transformers
- No more variable create in fit which are not private (\_variable) or public (variable\_).
- More robustness to various inputs
- Better input checks

Currently the tests are not run on the dirty cat transformers, because of the tag X_types: ["categorical"]. Indeed, scikit-learn for now only supports running the tests with X_types: [2darray].
The tests could run on the SuperVectorizer, which doesn't have this tag, but I've commented this test for now: it seems that some parts of the scikit learn test do not work when converting the input to a pandas dataframe.


dirty_cat/super_vectorizer.py
```diff
@@ -346,7 +374,9 @@ def fit_transform(self, X, y=None):
             self.types_ = {c: t for c, t in zip(X.columns, X.dtypes)}
 
         # Select columns by dtype
-        numeric_columns = X.select_dtypes(include=['int', 'float']).columns.to_list()
+        numeric_columns = X.select_dtypes(include=['int', 'float', np.float64, np.float32, np.float16,
+                                                   np.int64, np.int32, np.int16, np.uint64, np.uint32,
+                                                   np.uint16]).columns.to_list()
```
LilianBoulard, line 379:
```suggestion
        numeric_columns = X.select_dtypes(include=['int', 'float', 
                                                   np.float64, np.float32, np.float16,
                                                   np.int64, np.int32, np.int16, 
                                                   np.uint64, np.uint32, np.uint16]).columns.to_list()
```

dirty_cat/datetime_encoder.py
```diff
@@ -43,6 +46,12 @@ def __init__(self,
         self.extract_until = extract_until
         self.add_day_of_the_week = add_day_of_the_week
 
+    def _more_tags(self):
+        """
+        Used to ease the sklearn checks of the estimator.
```
LilianBoulard, line 51: Is it used internally by sklearn ?
If so, I think it would be a good thing to mention it, something like

```suggestion
        Used internally by sklearn to ease the estimator checks.
```

(also applicable to the 3 other instances of this method in this PR)

dirty_cat/datetime_encoder.py
```diff
@@ -136,7 +145,13 @@ def transform(self, X, y=None):
         array, shape (n_samples, n_features_out_)
             Transformed input.
         """
+        check_is_fitted(self, attributes=["n_features_in_", "n_features_out_", "features_per_column_"])
         X = check_input(X)
+        if X.shape[1] != self.n_features_in_:
+            raise ValueError(
+                f"Number of features in the input data ({X.shape[1]}) does not match the number of features "
+                f"seen during fit ({self.n_features_in_})."
```
LilianBoulard, line 153:
```suggestion
                f"The number of features in the input data ({X.shape[1]}) does not match the number of features "
                f"seen during fit ({self.n_features_in_})."
```

LilianBoulard (commented):
Thank you !

dirty_cat/super_vectorizer.py
```diff
@@ -202,6 +207,28 @@ def __init__(self, *,
...
+        """
+        Used to ease the sklearn checks of the estimator.
+        """
+        return {"allow_nan": [True]}
+
+    def _clone_transformers(self):
+        if self.low_card_cat_transformer is not None:
+            self.low_card_cat_transformer_ = clone(self.low_card_cat_transformer)
+        else:
+            self.low_card_cat_transformer_ = clone(OneHotEncoder())
```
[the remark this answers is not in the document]
LeoGrin: I'm not sure, I'm a bit confused too. I've taken this from https://github.com/scikit-learn/scikit-learn/blob/baf0ea25d6dd034403370fea552b21a6776bef18/sklearn/ensemble/_stacking.py#L76, called with a new instance here https://github.com/scikit-learn/scikit-learn/blob/baf0ea25d6dd034403370fea552b21a6776bef18/sklearn/ensemble/_stacking.py#L550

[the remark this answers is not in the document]
LeoGrin: I think you're right that in our case we don't need this, for the StackingRegressor they probably wanted _close_final_estimator to be more general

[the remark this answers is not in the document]
LilianBoulard: > they probably wanted _close_final_estimator to be more general
Indeed !

dirty_cat/super_vectorizer.py
```diff
@@ -65,17 +67,19 @@ class SuperVectorizer(ColumnTransformer):
         the parameters `low_card_cat_transformer` and
         `high_card_cat_transformer` respectively.
 
-    low_card_cat_transformer: Transformer or str or None, default=OneHotEncoder()
+    low_card_cat_transformer: Transformer or str or None, default=None
         Transformer used on categorical/string features with low cardinality
         (threshold is defined by `cardinality_threshold`).
+        `OneHotEncoder()` is used by default.
```
GaelVaroquaux, line 73: Maybe we need to be explicit that None is converted to a OneHotEncoder ?

dirty_cat/super_vectorizer.py
```diff
@@ -65,17 +67,19 @@ class SuperVectorizer(ColumnTransformer):
...
         Can either be a transformer object instance (e.g. `OneHotEncoder()`),
         a `Pipeline` containing the preprocessing steps,
         None to apply `remainder`, 'drop' for dropping the columns,
         or 'passthrough' to return the unencoded columns.
 
-    high_card_cat_transformer: Transformer or str or None, default=GapEncoder(n_components=30)
+    high_card_cat_transformer: Transformer or str or None, default=None
         Transformer used on categorical/string features with high cardinality
         (threshold is defined by `cardinality_threshold`).
+        `GapEncoder(n_components=30)` is used by default.
```
GaelVaroquaux, line 82: Same comment about being explicit that None is converted

dirty_cat/target_encoder.py
```diff
@@ -151,7 +161,7 @@ def fit(self, X, y):
                 le.classes_ = np.array(self.categories[j])
 
         self.categories_ = [le.classes_ for le in self._label_encoders_]
-        self.n = len(y)
+        self._n = len(y)
```
GaelVaroquaux, line 164: I think that typically we would use "n_", according to scikit-learn conventions.

GaelVaroquaux (commented):
Super useful!! I made a few comments

LilianBoulard (approved):
LGTM :)

GaelVaroquaux:
Thanks!!! Merging!

PR #277, opened 2022-07-05
Clean up warnings

LilianBoulard:
Cleans up a bunch of warnings we usually get in the CI.

### DIFF
--- a/dirty_cat/datasets/tests/test_fetching.py
+++ b/dirty_cat/datasets/tests/test_fetching.py
@@ -26,6 +26,7 @@ def get_test_data_dir() -> Path:
     return get_data_dir("tests")
 
 
+@pytest.mark.filterwarnings('ignore::UserWarning')
 def test_fetch_openml_dataset():
     """
     Tests the ``fetch_openml_dataset()`` function in a real environment.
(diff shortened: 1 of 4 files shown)


dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -26,6 +26,7 @@ def get_test_data_dir() -> Path:
     return get_data_dir("tests")
 
 
+@pytest.mark.filterwarnings('ignore::UserWarning')
```
GaelVaroquaux, line 29: I must admit that I don't really love silencing all the warnings on a test like this because it can silence legitimate warnings. I'd rather have at the right place (which can be in the test) a catch_warnings.

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -94,6 +95,7 @@ def test_fetch_openml_dataset():
         shutil.rmtree(path=str(test_data_dir), ignore_errors=True)
 
 
+@pytest.mark.filterwarnings('ignore::UserWarning')
```
GaelVaroquaux, line 98: Same comment here (and below)

GaelVaroquaux (changes requested):
I love this PR. I would like slight more focal warning captures in some places.

GaelVaroquaux (approved):
LGTM. Merging

PR #281, opened 2022-07-05
Normalize `get_feature_names` handling

### LINKED ISSUE #279: Inconsistencies regarding transition from `get_feature_names` to `get_feature_names_out`
As part of the work I'm putting in #277, I noticed the way we're handling the usage of `get_feature_names` since the transition to `get_feature_names_out` (sklearn 1.0) is inconsistent.

For example, in the `GapEncoder` and the `SuperVectorizer`, we raise a warning no matter the installed version of scikit-learn, and in the `DatetimeEncoder`, we don't warn at all.

In my opinion, we should be checking sklearn's version beforehand, and warn only if appropriate.

Here's a template for this behavior: 

```py
def get_feature_names(*args, **kwargs):
    if Version(sklearn.__version__) >= '1.0':
        warn(
            "Following the changes in scikit-learn 1.0, "
            "get_feature_names is deprecated. "
            "Use get_feature_names_out instead.",
            DeprecationWarning
        )
    return self.get_feature_names_out(*args, **kwargs)
```

LilianBoulard:
Fixes #279

### DIFF
--- a/dirty_cat/datetime_encoder.py
+++ b/dirty_cat/datetime_encoder.py
@@ -173,7 +173,8 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
                 "Following the changes in scikit-learn 1.0, "
                 "get_feature_names is deprecated. "
                 "Use get_feature_names_out instead.",
-                DeprecationWarning
+                DeprecationWarning,
+                stacklevel=2,
             )
         return self.get_feature_names_out()
 
--- a/dirty_cat/datetime_encoder.py
+++ b/dirty_cat/datetime_encoder.py
@@ -1,8 +1,10 @@
 import numpy as np
 import pandas as pd
+from sklearn import __version__ as sklearn_version
 from sklearn.base import BaseEstimator, TransformerMixin
+from warnings import warn
 from typing import List
-from dirty_cat.utils import check_input
+from dirty_cat.utils import check_input, Version
 
 # Some functions need aliases
 WORD_TO_ALIAS = {"year": "Y", "month": "M", "day": "D", "hour": "H", "minute": "min", "second": "S",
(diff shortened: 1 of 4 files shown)


dirty_cat/datetime_encoder.py
```diff
@@ -166,5 +168,12 @@ def get_feature_names(self, input_features=None) -> List[str]:
         Ensures compatibility with sklearn < 1.0, and returns the output of
         get_feature_names_out.
         """
+        if Version(sklearn_version) >= '1.0':
+            warn(
+                "Following the changes in scikit-learn 1.0, "
+                "get_feature_names is deprecated. "
+                "Use get_feature_names_out instead.",
+                DeprecationWarning
```
GaelVaroquaux, line 176: Use the "stacklevel" argument of "warnings.warn", when you call "warn", where the value has to be chosen for the warning to appear in user code rather than in library code. Typically 1.

LeoGrin:
Looks great to me!

dirty_cat/gap_encoder.py
```diff
@@ -762,12 +762,16 @@ def get_feature_names_out(self, col_names=None, n_labels=3):
...
-            DeprecationWarning,
+        Ensures compatibility with sklearn < 1.0.
+        Use `get_feature_names_out` instead.
+        """
+        if Version(sklearn_version) >= '1.0':
+            warnings.warn(
+                "Following the changes in scikit-learn 1.0, "
+                "get_feature_names is deprecated. "
+                "Use get_feature_names_out instead.",
+                DeprecationWarning
```
GaelVaroquaux, line 774: stacklevel also :)

dirty_cat/super_vectorizer.py
```diff
@@ -491,11 +490,15 @@ def get_feature_names_out(self, input_features=None) -> List[str]:
...
-            DeprecationWarning,
+        Ensures compatibility with sklearn < 1.0.
+        Use `get_feature_names_out` instead.
+        """
+        if Version(sklearn_version) >= '1.0':
+            warn(
+                "Following the changes in scikit-learn 1.0, "
+                "get_feature_names is deprecated. "
+                "Use get_feature_names_out instead.",
+                DeprecationWarning
```
GaelVaroquaux, line 502: stacklevel also (though actually the initial call did not have stacklevel, so I'm asking for an improvement.

GaelVaroquaux (changes requested):
A few comments on stacklevel (see the warnings.warn documentation), and after it's good to go! Thanks!

GaelVaroquaux (approved):

GaelVaroquaux:
Darn, conflicts to resolv

PR #282, opened 2022-07-06
Drop support for Jaro, Jaro-Winkler and Levenshtein distances

### LINKED ISSUE #226: Remove python-levenstein
Next major release, I think that we should remove python-levenstein and the corresponding codepaths: it is not central to anything that we do, and it will hold us back.

LilianBoulard:
Resolves #226.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,6 +8,9 @@ Major changes
     columns (year, month, day, hour, minute, second, ...). It is now the default transformer used
     in the SuperVectorizer for datetime columns.
 
+* Dropped support for Jaro, Jaro-Winkler and Levenshtein distances, in favor of ngram.
+    Therefore, the :class:`SimilarityEncoder` now exclusively takes `ngram` as the `similarity` argument.
+
 Notes
 -----
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -8,8 +8,9 @@ Major changes
     columns (year, month, day, hour, minute, second, ...). It is now the default transformer used
     in the SuperVectorizer for datetime columns.
 
-* Dropped support for Jaro, Jaro-Winkler and Levenshtein distances, in favor of ngram.
-    Therefore, the :class:`SimilarityEncoder` now exclusively takes `ngram` as the `similarity` argument.
+* Dropped support for Jaro, Jaro-Winkler and Levenshtein distances.
+    The :class:`SimilarityEncoder` now exclusively uses `ngram` for similarities,
+    and the `similarity` parameter is deprecated. It will be removed in 0.5.
 
 Notes
 -----
--- a/dirty_cat/similarity_encoder.py
+++ b/dirty_cat/similarity_encoder.py
@@ -26,7 +26,6 @@ from sklearn.utils import check_random_state
 from sklearn.utils.fixes import _object_dtype_isnan
 
 from dirty_cat.utils import Version
-from . import string_distances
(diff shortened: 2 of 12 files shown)


dirty_cat/similarity_encoder.py
```diff
@@ -455,21 +445,7 @@ def transform(self, X, fast=True):
...
-                unqX = np.unique(Xlist[j])
-                encoder_dict = {x: vect(x, cats.reshape(1, -1))
-                                for x in unqX}
-                encoder = [encoder_dict[x] for x in Xlist[j]]
-                encoder = np.vstack(encoder)
-                out.append(encoder)
-            return np.hstack(out)
-
-        elif self.similarity == 'ngram':
+        if self.similarity == 'ngram':
```
GaelVaroquaux, line 448: We need to deprecate the "similarity" argument: it's value by default should be "None", any other value should lead to raising a DeprecationWarning saying that it will be removed in release 0.5.

dirty_cat/similarity_encoder.py
```diff
@@ -173,20 +165,18 @@ class SimilarityEncoder(OneHotEncoder):
     The method is based on calculating the morphological similarities
     between the categories.
     The categories can be encoded using one of the implemented string
-    similarities: ``similarity='ngram'`` (default), 'levenshtein-ratio',
-    'jaro', or 'jaro-winkler'.
+    similarities. Currently, only ``similarity='ngram'`` is available.
```
GaelVaroquaux, line 168: Say that this is deprecated

dirty_cat/test/test_similarity_encoder.py
```diff
@@ -177,28 +177,12 @@ def test_similarity_encoder():
...
-                             string_distances.levenshtein_ratio,
-                             categories=category,
-                             n_prototypes=None)
-            _test_similarity('jaro-winkler', string_distances.jaro_winkler,
-                             categories=category, n_prototypes=None)
-            _test_similarity('jaro', string_distances.jaro,
-                             categories=category, n_prototypes=None)
             _test_similarity('ngram', string_distances.ngram_similarity,
                              categories=category, n_prototypes=None)
             _test_similarity('ngram', string_distances.ngram_similarity,
```
GaelVaroquaux, line 182: We need to stop passing the "similarity" argument, as we want to remove this argument.

CHANGES.rst
```diff
@@ -8,6 +8,9 @@ Major changes
     columns (year, month, day, hour, minute, second, ...). It is now the default transformer used
     in the SuperVectorizer for datetime columns.
 
+* Dropped support for Jaro, Jaro-Winkler and Levenshtein distances, in favor of ngram.
+    Therefore, the :class:`SimilarityEncoder` now exclusively takes `ngram` as the `similarity` argument.
```
GaelVaroquaux, line 12:
```suggestion
    Therefore, the `similarity` argument in the :class:`SimilarityEncoder` is depecrated and only the `ngram` is possible.
```

GaelVaroquaux (changes requested):
We need to deprecate the similarity argument

GaelVaroquaux:
There are still place where python-levenshtein is remaining: I found one in .travis.yml, where I believe that it should be removed.

Can you "git grep levenshtein", and check all the corresponding places that they do not need to be fixed.

GaelVaroquaux (approved):
LGTM. Thank you

Issue #288, opened 2022-07-08
Bump to Python 3.8


LilianBoulard:
This issue is linked with #287.

In this PR, I used the `Literal` type hint, which allows us to specify a set of possible values a parameter can take:

```py
from typing import Literal


def func(status: Literal['on', 'off']) -> bool:
    return {
        'on': True,
        'off': False,
    }[status]


func('on')
# True

func('off')
# False

func('down')
# ValueError, but if your IDE supports type hinting, it should have told you !
```

I realized this feature was introduced in `typing` (the builtin library for type hinting) in Python 3.8.

Our current minimal requirement for Python is version 3.6, which was released in [December 2016](https://www.python.org/downloads/release/python-360/), and [has stopped receiving updates in December 2021](https://pythonspeed.com/articles/stop-using-python-3.6/).

Considering this, and the fact that it was released over 5 years ago, I think we should update.
[Python 3.7](https://www.python.org/downloads/release/python-370/) (released June 2018) has a bunch of new features, notably dataclasses, which could be used for the dataset fetching code.
[Python 3.8](https://www.python.org/downloads/release/python-380/) (released October 2019) also has a bunch of new features, notably the walrus operator and the equal-ending f-strings.

And of course, the main reason why I'm arguing here, the `Literal` type !

As mentioned, the release is almost two and a half years old, [and so is our minimal requirement of numpy](https://github.com/scikit-learn/scikit-learn/releases/tag/0.22), so I think it should be safe to upgrade for dirty_cat 0.3.

What do you think ?

LilianBoulard:
PR is ready to be merged. Can you confirm this is fine @GaelVaroquaux ?

LilianBoulard:
Should have been closed by #289 (looks like "Solve" is not a closing keyword).

PR #289, opened 2022-07-12
Bump min Python version

LilianBoulard:
Solves #288


LilianBoulard:
Got confirmation, merging !

PR #294, opened 2022-08-02
Add regression/classification in datasets' descriptions

### LINKED ISSUE #293: Add the type of problem in the example datasets
Currently, on most (if not all) the example datasets we provide, we do not mention whether this is a classification or regression problem.
Reading the whole description is sometimes not enough to determine this information, and so it's annoying to look for it.

I think alike the `load_*` functions of `sklearn.datasets`, we should add in the first line what type of target we're dealing with.

LilianBoulard:
Fixes #293


LilianBoulard:
(I should've skipped the CI)
Looks good, merging !

PR #287, opened 2022-07-08
Maintenance

LilianBoulard:
This PR aims at improving the overall quality of the code and doc.

It has several purposes:
- Correct typos
- Reword unclear sentences
- Minor updates to the doc
- Some minor structural improvements, such as moving some functions to suiting modules
- Rename some variables for better readability
- Use modern language features for better readability and performance
- Simplify the code, while leaving functionalities intact (no bug-fixes)
- Make extensive use of type hinting, which serves two purposes:
  - Make the code easier to work with, especially when working with IDEs that support type hinting
  - Make the functions more efficient and less error-prone when using tools that enforce types, such as MyPy

In general, these are rather small modifications for which making unique PRs would be kind of overkill.

PR #295, opened 2022-08-03
Fix CI build

LilianBoulard:
On a few recent PRs, we are having the following error in CircleCI:
```
Extension error (sphinx_gallery.docs_resolv):
Handler <function embed_code_links at 0x7ff2f7af8af0> for event 'build-finished' threw an exception (exception: Expecting property name enclosed in double quotes: line 1 column 2 (char 1))
make: *** [Makefile:37: html-noplot] Error 2
```

As mentioned in #287, it probably comes from sphinx-gallery, but a similar bug that has been resolved upstream a couple of weeks ago (which lead to the release of sphinx-gallery 0.11).

I also tried to fix a few warnings.


LilianBoulard:
Hooray ! Merging :)

PR #297, opened 2022-08-05
Remove unnecessary `RuntimeError` in `SuperVectorizer`

### LINKED ISSUE #296: `RuntimeError` should be warning
When passing an already "clean" dataset to the `SuperVectorizer` (which is common in generic pipelines), it raises a `RuntimeError` as `No transformers could be generated!`.
The doc mentions
```
Raises
------
RuntimeError
    If no transformers could be constructed,
    usually because transformers passed do not match any column.
    To fix the issue, try passing the least amount of None as encoders.
```

First, I think the message could be reworded, as it's not very clean what "no transformers being generated" means.
Secondly, I think it should be a warning (at best), not an error.

LilianBoulard:
Fixes #296


jovan-stojanovic:
Great! Maybe we should add a small test to check that it all works fine without raising the error?

LilianBoulard:
So, the reason I commented out `check_input(X)` is because if the data is not numerical at the end of the `fit_transform`, `check_input` raises an exception.
However, we want the SV to return possibly raw (categorical) features, as the `ColumnTransformer` would.
I've also fixed the `clone_transformers` method, which crashed when the transformers variables were strings. I'll add a test to check that possible strings are correctly handled.

As for the current error on the CI, I have no idea why the diff says there's a difference, when there isn't any...
image

LilianBoulard:
I just realized, we lost the possibility to apply the `remainder` as a transformer, which was used when the transformer passed was `None`.
The only solution I see is adding another possible string: "remainder", which would use the value of the SV's `remainder`, but it sounds confusing.

dirty_cat/test/test_super_vectorizer.py
```diff
@@ -414,6 +414,30 @@ def test_fit_transform_equiv():
...
+        datetime_transformer='passthrough',
+        numerical_transformer='passthrough',
+        impute_missing='skip',
+        auto_cast=False,
+    )
+
+    X_enc_dirty = sv.fit_transform(X_dirty)
+    X_enc_clean = sv.fit_transform(X_clean)
+
+    assert X_dirty.to_numpy() == X_enc_dirty
```
jovan-stojanovic, line 437:
```suggestion
    assert X_dirty.to_numpy().ravel().tolist() == X_enc_dirty.ravel().tolist()
```

dirty_cat/test/test_super_vectorizer.py
```diff
@@ -414,6 +414,30 @@ def test_fit_transform_equiv():
...
+    assert X_clean.to_numpy() == X_enc_clean
```
jovan-stojanovic, line 438:
```suggestion
    assert (X_clean.to_numpy() == X_enc_clean).all()
```

jovan-stojanovic (commented):
I think the test failed here because what was compared were arrays of lists and not the elements. This worked:

jovan-stojanovic:
Oh so it seems that with minimal requirements, ``np.nan == np.nan`` is compared which is ``False`` [by default](https://stackoverflow.com/questions/27616824/python-equality-for-nan-in-a-list) and the test fails.
We should find a way to compare the ``X_enc_dirty`` and ``X_dirty`` without looking for equality of ``nan``'s...

GaelVaroquaux:
This PR needs to be finished, to catter for the failing test

dirty_cat/super_vectorizer.py
```diff
@@ -456,7 +453,7 @@ def fit_transform(self, X, y=None):
         if self.verbose:
             print(f'[SuperVectorizer] Assigned transformers: {self.transformers}')
 
-        check_input(X)
```
GaelVaroquaux, line 459: Hum, is this intended? If so, we should actually delete the line rather than comment it out.

If this is intended, we should understand what consequences it has to remove the line.

GaelVaroquaux (commented):
One comment, but aside from this, LGTM.

Thanks a lot!!

dirty_cat/super_vectorizer.py
LilianBoulard: Yes, it's something I've [explained earlier in this thread](https://github.com/dirty-cat/dirty_cat/pull/297#issuecomment-1217702118) :)

GaelVaroquaux: Ok, then just remove it completely

dirty_cat/test/test_super_vectorizer.py
```diff
@@ -434,6 +434,40 @@ def test_fit_transform_equiv() -> None:
     assert np.allclose(enc1_x2, enc2_x2, rtol=0, atol=0, equal_nan=True)
 
 
+def test_passthrough():
+    """
+    Tests that when passed no encoders, the SuperVectorizer
+    returns the dataset as-is.
+    """
+
+    def is_equal(elements: Tuple[Any, Any]) -> bool:
```
GaelVaroquaux, line 443: In general, I prefer avoid defining functions inside function (they cannot be reused, they cannot be tested, they cannot be pickled, hence they can lead to trouble)

GaelVaroquaux (changes requested):
One rather trivial comment and we're good to go

PR #300, opened 2022-08-08
Binary encoding for features with cardinality <= 2

### LINKED ISSUE #246: One hot encoding of Supervectorizer for columns with 2 values
Just something I observed using SuperVectorizer, on a dataset with columns having mostly 2 categorical values. As documented, it runs one hot encoding on them, and this blows up the feature space. This is probably an edge case, and the example I have here is maybe an extreme one. For just two values it seemed excessive to do one-hot encoding.

```
from dirty_cat import SuperVectorizer
from sklearn.datasets import fetch_openml
X, y = fetch_openml("kr-vs-kp", return_X_y=True)
X.shape
(3196, 36)
X.describe()
       bkblk bknwy bkon8 bkona bkspr bkxbq bkxcr bkxwp blxwp bxqsq cntxt dsopp dwipd  ... simpl skach skewr skrxp spcop stlmt thrsk wkcti wkna8 wknck wkovl wkpos wtoeg
count   3196  3196  3196  3196  3196  3196  3196  3196  3196  3196  3196  3196  3196  ...  3196  3196  3196  3196  3196  3196  3196  3196  3196  3196  3196  3196  3196
unique     2     2     2     2     2     2     2     2     2     2     2     2     2  ...     2     2     2     2     2     2     2     2     2     2     2     2     2
top        f     f     f     f     f     f     f     f     f     f     f     f     l  ...     f     f     t     f     f     f     f     f     f     f     t     t     n
freq    2839  2971  3076  2874  2129  1722  2026  2500  1980  2225  1817  2860  2205  ...  1975  3185  2216  3021  3195  3149  3060  2631  3021  1984  2007  2345  2407

[4 rows x 36 columns]

sv = SuperVectorizer()
X_tx = sv.fit_transform(X)
X_tx.shape
(3196, 73)

```

LilianBoulard:
Fixes #246
Also refactored some tests for the `SuperVectorizer` to make the code cleaner.


LilianBoulard:
sklearn's `OrdinalEncoder` does not include a `get_feature_names` method, and only implemented `get_feature_names_out` in version 1.1.
How should we deal with this issue ?

One option would be to implement a `BinaryEncoder` that would inherit the `OrdinalEncoder`, and only implement what we need.

GaelVaroquaux:
I think that using the option 'drop="if_binary"' of the OneHotEncoder would be a cleaner way of doing things. I realize that it is available only since sklearn 0.23, but it makes code more future-proof.

dirty_cat/super_vectorizer.py
```diff
@@ -57,9 +57,11 @@ class SuperVectorizer(ColumnTransformer):
     ----------
 
     cardinality_threshold: int, default=40
-        Two lists of features will be created depending on this value: strictly
-        under this value, the low cardinality categorical values, and above or
-        equal, the high cardinality categorical values.
+        Two lists of features will be created depending on this value: between
+        3 (included) and this value (excluded) under this value,
```
jovan-stojanovic, line 61:
```suggestion
        3 (included) and this value (excluded),
```

dirty_cat/super_vectorizer.py
```diff
@@ -57,9 +57,11 @@ class SuperVectorizer(ColumnTransformer):
...
+        the low cardinality categorical values, and above or equal,
```
jovan-stojanovic, line 62:
```suggestion
        the low cardinality categorical features, and above or equal,
```

dirty_cat/super_vectorizer.py
```diff
@@ -57,9 +57,11 @@ class SuperVectorizer(ColumnTransformer):
...
+        the high cardinality categorical values.
```
jovan-stojanovic, line 63:
```suggestion
        the high cardinality categorical features.
```

dirty_cat/super_vectorizer.py
```diff
@@ -57,9 +57,11 @@ class SuperVectorizer(ColumnTransformer):
...
+        The features with a cardinality of <= 2 will get binary encoded.
```
jovan-stojanovic, line 64:
```suggestion
        The features with a cardinality of <= 2 will be binary encoded.
```

jovan-stojanovic (commented):
A few suggestions for the docs

GaelVaroquaux (changes requested):
Use the option 'drop="if_binary"' of the OneHotEncoder would be a cleaner way of doing things. 

It is available only since sklearn 0.23, but it makes code more future-proof.

LilianBoulard:
This feature requires a fix introduced in #303.

GaelVaroquaux:
> This feature requires a fix introduced in #303.

#303 is currently not mergeable because of failing tests and conflicts

dirty_cat/gap_encoder.py
```diff
@@ -35,21 +35,12 @@
 
 from .utils import check_input
 
-if Version(sklearn_version) == Version("0.22"):
-    with warnings.catch_warnings():
-        warnings.simplefilter("ignore")
-        from sklearn.cluster.k_means_ import _k_init
```
GaelVaroquaux, line 41: Good catch to clean this up!

dirty_cat/super_vectorizer.py
```diff
@@ -95,12 +95,26 @@ class SuperVectorizer(ColumnTransformer):
...
+        Two lists of features will be created depending on this value: between
+        3 (included) and this value (excluded), the low cardinality categorical
+        features, and above or equal, the high cardinality categorical features.
+        Different transformers will be applied to these two groups, defined by
+        the parameters `low_card_cat_transformer` and
         `high_card_cat_transformer` respectively.
+        The features with a cardinality of <= 2 will be binary encoded
+        (see `binary_cat_transformer`).
+
+    binary_cat_transformer : typing.Optional[typing.Union[sklearn.base.TransformerMixin, typing.Literal["drop", "remainder", "passthrough"]]], default=None  # noqa
```
GaelVaroquaux, line 108: Why do we need a special transformer for binary categories? Can we not simply add the option 'drop="if_binary"' to the transformer used in low_card_cat_transformer. It makes the interface and the code much simpler.

dirty_cat/super_vectorizer.py
```diff
@@ -287,6 +313,15 @@ def _clone_transformers(self):
...
         """
+        if isinstance(self.binary_cat_transformer, sklearn.base.TransformerMixin):
+            self.binary_cat_transformer_ = clone(self.binary_cat_transformer)
+        elif self.binary_cat_transformer is None:
+            self.binary_cat_transformer_ = OneHotEncoder(drop="if_binary")
+        elif self.binary_cat_transformer == "remainder":
+            self.binary_cat_transformer_ = self.remainder
+        else:
+            self.binary_cat_transformer_ = self.binary_cat_transformer
+
```
GaelVaroquaux, line 324: I don't understand why we don't just add "drop='if_binary'" to the OneHotEncoder a couple lines below. It would lead to much simpler code.

GaelVaroquaux (changes requested):
I think that it would be much simpler not to introduce a new category of transformers but simply modify the default one for low_card_cat.

In addition, we need to document clearly this change as it is a backward incompatible change that might change predictive pipelines.

dirty_cat/super_vectorizer.py
```diff
@@ -470,22 +505,31 @@ def fit_transform(self, X, y=None):
         ).columns.to_list()
 
         # Classify categorical columns by cardinality
+        _nunique_values = {  # Cache results
+            col: X[col].nunique() for col in categorical_columns
+        }
+        binary_cat_columns = [
+            col for col in categorical_columns if _nunique_values[col] <= 2
```
jovan-stojanovic, line 512: Should this be only if there are binary categorization?
```suggestion
            col for col in categorical_columns if _nunique_values[col] == 2
```
I guess there is no point in encoding if `_nunique_values[col] <= 1`

LilianBoulard:
So the reason I did an independent "category" for the binary features is because while OneHot has a parameter to avoid this, other encoders might not.

dirty_cat/super_vectorizer.py
LilianBoulard: True, but what do you think we should do with the features with cardinality==1 ? I guess maybe drop them ?

jovan-stojanovic: Yes drop would be the best not to create confusion.

GaelVaroquaux:
Hum, maybe we should actually use drop="first", which we solve all problems in a single step.

@amy12xx, what do you think?

GaelVaroquaux:
Hum, maybe we should actually use drop="first", which we solve all problems in a single step.

GaelVaroquaux:
> So the reason I did an independent "category" for the binary features is because while OneHot has a parameter to avoid this, other encoders might not.

But the original issue only applies to the OneHotEncoder

GaelVaroquaux:
> Hum, maybe we should actually use drop="first", which we solve all problems in a single step.

After many discussions, including with the scikit-learn team, the consensus is that "drop='first'" raises real interpretability issues  and that we should rather go for if_binary

CHANGES.rst
```diff
@@ -7,7 +7,14 @@ Major changes
 * New encoder: :class:`DatetimeEncoder` can transform a datetime column into several numerical
     columns (year, month, day, hour, minute, second, ...). It is now the default transformer used
     in the :class:`SuperVectorizer` for datetime columns.
-* The :class:`SuperVectorizer` has seen some major improvements and bug fixes.
+* The :class:`SuperVectorizer` has seen some major improvements and bug fixes
+  - Fixes the automatic casting logic in ``transform``.
+  - To avoid dimensionality explosion when a feature has two unique values,
```
GaelVaroquaux, line 12:
```suggestion
  - **Behavior change** To avoid dimensionality explosion when a feature has two unique values,
```
We need to signal to users that this is an important change.

dirty_cat/super_vectorizer.py
LilianBoulard: So, for now, we decided with Gaël to not drop it for now (given we want to release very soon), but that's something we can discuss in the future.

GaelVaroquaux (commented):
Looks great.

One tiny additional suggestion and we're ready to merge

GaelVaroquaux:
I committed the suggestion myself to be able to merge ASAP.

Will merge once the tests pass

GaelVaroquaux (approved):

LilianBoulard (commented):
Okay, I had just one tiny change to the `fit_transform` docstring (I committed but hadn't finished writing it :p). LGTM too!

jovan-stojanovic (approved):

GaelVaroquaux:
Merged! Yey!

PR #301, opened 2022-08-08
Implement CodeQL

LilianBoulard:
Adds CodeQL, a tool for finding vulnerabilities and mistakes in the code.
It was recommended by GitHub, let's see if this can be useful for us.


LilianBoulard:
So, the output can be seen on the test, and the output has been exported to the `Security` tag of the repository.
No vulnerabilities nor mistakes have been found, which is good !
I don't think it is useful to run it for each PR, but I guess scanning the code before releases could be a good practice.
Let me know what you think !

GaelVaroquaux:
Looks good to me.

I'm merging

PR #302, opened 2022-08-17
Fetching: move `namedtuples` to `dataclasses`, and add target name to `DatasetInfoOnly`

LilianBoulard:
Trying to use the no-loading version of the fetchers, I realized we're not passing the target name, and it's annoying to hard-code in downstream code, so I've added it to the `DatasetInfoOnly` requirements.
Also, since we're now in Python 3.8+ territory, I've figured we should use dataclasses instead of namedtuples, as their syntax is a little nicer (though more verbose), but most importantly, they integrate type hinting !

### DIFF
--- a/dirty_cat/datasets/fetching.py
+++ b/dirty_cat/datasets/fetching.py
@@ -22,19 +22,11 @@ import warnings
 import pandas as pd
 
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


LilianBoulard:
As this is a simple bugfix, I'm merging !

PR #303, opened 2022-08-22
Rework SV logic

LilianBoulard:
I noticed some inconsistencies in the behavior of the SuperVectorizer, which this PRs aims to fix:
- The `_auto_cast` method was called on the data passed to `transform`, which meant that types of the `fit_transform` output data could be different from the types of the `transform` output data.
- The `types_` variable was overwritten during `transform`, which shouldn't happen as this is a mapping of the types learnt during fit.
- "False missing" values (for a lack of a better term) were only replaced in the `_auto_cast` function, which made sense as it was used in the `transform` as well. Now, it has been moved to its own function

### DIFF
--- a/dirty_cat/super_vectorizer.py
+++ b/dirty_cat/super_vectorizer.py
@@ -8,18 +8,16 @@ manually categorize them beforehand, or construct complex Pipelines.
 
 # Author: Lilian Boulard <lilian@boulard.fr> | https://github.com/LilianBoulard
 
-import sklearn
-
 import numpy as np
 import pandas as pd
 
 from warnings import warn
-from typing import Union, Optional, List
-
-from sklearn.base import BaseEstimator, clone
+from typing import Union, Optional, List, Tuple
+from sklearn.base import TransformerMixin, clone
 from sklearn.compose import ColumnTransformer
 from sklearn.preprocessing import OneHotEncoder
 from sklearn import __version__ as sklearn_version
+from pandas.core.dtypes.base import ExtensionDtype
 
 from dirty_cat import GapEncoder, DatetimeEncoder
 from dirty_cat.utils import Version, check_input
@@ -32,16 +30,29 @@ def _has_missing_values(df: Union[pd.DataFrame, pd.Series]) -> bool:
     return any(df.isnull())
 
 
-def _replace_missing_in_col(df: pd.Series, value: str = "missing") -> pd.Series:
+def _replace_false_missing(ser: pd.Series) -> pd.Series:
+    STR_NA_VALUES = [
+        'null', '', '1.#QNAN', '#NA', 'nan', '#N/A N/A',
+        '-1.#QNAN', '<NA>', '-1.#IND', '-nan', 'n/a', '-NaN',
+        '1.#IND', 'NULL', 'NA', 'N/A', '#N/A', 'NaN',
+    ]  # taken from pandas.io.parsers (version 1.1.4)
+    ser = ser.replace(STR_NA_VALUES + [None, "?", "..."], np.nan)
+    ser = ser.replace(r'^\s+$', np.nan, regex=True)  # Replace whitespaces
+    return ser
+
+
+def _replace_missing_in_cat_col(ser: pd.Series, value: str = "missing") -> pd.Series:
     """
     Takes a Series with string data, replaces the missing values, and returns it.
     """
-    dtype_name = df.dtype.name
+    ser = _replace_false_missing(ser)
+    if ser.dtype.name == 'category' and (value not in ser.cat.categories):
+        ser = ser.cat.add_categories([value])
+    ser = ser.fillna(value=value)
+    return ser
 
-    if dtype_name == 'category' and (value not in df.cat.categories):
-        df = df.cat.add_categories(value)
-    df = df.fillna(value=value)
-    return df
+
+OptionalTransformer = Optional[Union[TransformerMixin, str]]
 
 
 class SuperVectorizer(ColumnTransformer):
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
This PR has conflicts and failing tests

LilianBoulard:
So, the failing test was because of an issue that affects pandas version 1.1.5 (our min dependency until now), and possibly earlier versions (I haven't checked).
Pandas [1.2.0](https://github.com/pandas-dev/pandas/releases/tag/v1.2.0) was released on December 26th, 2020, only a few weeks after [1.1.5](https://github.com/pandas-dev/pandas/releases/tag/v1.1.5) (December 7th), and fixes this bug, so I guess bumping is not a big deal.

dirty_cat/super_vectorizer.py
```diff
@@ -95,11 +126,11 @@ class SuperVectorizer(ColumnTransformer):
         data type (dtype).
 
     impute_missing : str, default='auto'
-        When to impute missing values in string columns.
+        When to impute missing values in categorical columns.
```
GaelVaroquaux, line 129: Shouldn't it be "string" rather than categorical? This is applied to string columns, no? The notion of categorical is slightly different (there is for instance the notion of categorical dtype in pandas, that is not string)

GaelVaroquaux (commented):
This looks overall good, but we do have the problem of terminology of "cat" vs "str". It seems to me that it should be "str" quite often.

dirty_cat/super_vectorizer.py
LilianBoulard: Well, it's for both. Check the content of `_replace_missing_in_cat_col` to see why.
I renamed the method this way (it was `_replace_missing_in_col` before) to clarify that we're not imputing on numerical columns.

GaelVaroquaux: To my surprise the code works on categorical dtype (which tells me that my understanding of categorical dtype is limited).

Still, it seems important to me that the docstring says "string and categorical columns", because it works on strings too, and the "dirty categories" that dirty-cat focuses on are not considered categorical by everybody.

LilianBoulard:
So, the PR is overall stable, but there's a last thing I'm concerned about: the imputation strategy.

Currently, the logic is to impute missing values categorical / string with a custom value "missing", depending on the `impute_missing` parameter.

There is currently no imputation for missing values in datetime and numerical columns.
I'm kind of curious as to why this has not been a problem so far: do most models support missing values out of the box ?
What do you think @GaelVaroquaux ?

LilianBoulard:
Okay so these last few commits add some doc to clarify to the user what we impute and what we don't.
Other than that, the PR looks finished to me !

LilianBoulard (commented):
Everything looks good to me!
I'm waiting for your feedback, and I'll merge it right away :)

dirty_cat/super_vectorizer.py
```diff
@@ -250,20 +276,35 @@ def _clone_transformers(self):
...
         else:
             self.high_card_cat_transformer_ = self.high_card_cat_transformer
 
+        if isinstance(self.numerical_transformer, sklearn.base.TransformerMixin):
+            self.numerical_transformer_ = clone(self.numerical_transformer)
+        elif self.numerical_transformer is None:
+            self.numerical_transformer_ = "passthrough"
+        elif self.numerical_transformer == "remainder":
+            self.numerical_transformer_ = self.remainder
+        else:
```
GaelVaroquaux, line 299: I'm not sure that I am following the logic fully. If I understand things right, if I put numerical_transformer="pasthrough" (a typo on "passthrough", the numerical_transformer ends up being the high_card_cat_transformer. This sounds like a dangerous behavior, as it will lead to people having bugs in their pipelines.

LilianBoulard: Ah, no, sorry that was en error, it should be `self.numerical_transformer_ = self.numerical_transformer`

dirty_cat/super_vectorizer.py
```diff
@@ -250,20 +276,35 @@ def _clone_transformers(self):
...
+            self.numerical_transformer_ = self.high_card_cat_transformer
```
GaelVaroquaux, line 300: I'm confused here. If I read this correctly, we are applying a transformer designed for categorical data to numerical features? This seems incorrect.

dirty_cat/super_vectorizer.py
LilianBoulard: And so if there was a typo, it will crash down the line when the ColumnTransformer's `fit_transform` will call `_validate_transformers`.

dirty_cat/super_vectorizer.py
LilianBoulard: It's an error indeed, see message above :)

dirty_cat/super_vectorizer.py
GaelVaroquaux: OK, thanks. Please add a comment that says that.

dirty_cat/super_vectorizer.py
```diff
@@ -26,7 +28,11 @@ def _has_missing_values(df: Union[pd.DataFrame, pd.Series]) -> bool:
     return any(df.isnull())
 
 
-def _replace_false_missing(ser: pd.Series) -> pd.Series:
```
GaelVaroquaux: Please add a small docstring to describe to the developer what this function does.
It does not need to be a properly formatted docstring describing all parameters, as this is an internal function, we just need to describe why this function

GaelVaroquaux (changes requested):
Two things to modify:
- The docstring to add
- The error on the numerical vs categorical transformer

After this, we merge. Don't do more changes, or I'll find other comments :)

PR #308, opened 2022-09-02
DEV Add a simple `pre-commit` setup

jjerphan:
Fixes https://github.com/dirty-cat/dirty_cat/issues/304.

This adds a simple pre-commit setup with `black`, `flake8`, `isort` and `mypy`.

Note that the code-base needs to be formatted using this setup.
IMO, this might better be done in another PR with a single commit to better ignore formatting changes (see [this blog](https://michaelheap.com/git-ignore-rev/) post for introduction).

Also, a GitHub action workflow can then be introduced to check that changes made in Pull Requests comply with this setup.

PR #314, opened 2022-09-02
MAINT Format the code-base to be `pre-commit` compliant

jjerphan:
Follow-up of #308 and #317.

This applies the `pre-commit` formatting on the whole code-base

A few docstrings lines were exceeding the column limit but we're hardly breakable. I chose to put a `# noqa` directive so that black ignore them; but there might be a better solution (such as defining custom types and using them in the docstring if it makes sense).

A follow-up PR will add a `.git-blame-ignore-revs` file not to have the formatting disrupt `git blame` with some documentation to have it setup, see [scikit-learn's](https://github.com/scikit-learn/scikit-learn/blob/main/.git-blame-ignore-revs).

### DIFF
--- a/dirty_cat/__init__.py
+++ b/dirty_cat/__init__.py
@@ -3,17 +3,22 @@ dirty_cat: Learning on dirty categories.
 """
 import os
 
-version_file = os.path.join(os.path.dirname(__file__), 'VERSION.txt')
+from .datetime_encoder import DatetimeEncoder
+from .gap_encoder import GapEncoder
+from .minhash_encoder import MinHashEncoder
+from .similarity_encoder import SimilarityEncoder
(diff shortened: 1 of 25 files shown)


jjerphan:
I think the failure on CodeCov regarding the [Relative Coverage](https://docs.codecov.com/docs/codecov-delta#relative) (76.24% which does not meet the target of 85.33%) can be ignored due to the nature of this PR.

jjerphan:
@GaelVaroquaux and @LilianBoulard: this PR can be reviewed.

GaelVaroquaux (approved):
LGTM (fingers crossed that this doesn't create too much of a mess). Merging

jjerphan:
Thanks.

#325 has been opened, and once it's merged the entire setup is done.

PR #317, opened 2022-09-02
CI Add GitHub action to run pre-commit checks 

jjerphan:
Follow-up of #308.

This also remove mypy checks from pre-commit setup which creates confusion with version parsing via the private `Version` class.


GaelVaroquaux (approved):
LGTM. Merging. Thank you!!

PR #320, opened 2022-09-02
Remove buggy file

GaelVaroquaux:
The build is failing on master. Removing to clean stuff


GaelVaroquaux:
Merging. The bug was only triggered on master. Hopefully this will fix it.

PR #321, opened 2022-09-02
Add docstring tests to prepare meta issue

Vincent-Maladiere:
### Changes
Prepare the ground by adding a utils test file to:
1. Fetch methods of all estimators with `get_all_methods()`.
2. Run numpydoc test on a specific target or the entire project:

```shell
pytest dirty_cat/test/test_docstrings.py -k TargetEncoder-fit
```

or equivalently:

```shell
python dirty_cat/test/test_docstrings.py dirty_cat.target_encoder.TargetEncoder.fit
```

### DIFF
--- a/dirty_cat/test/test_docstrings.py
+++ b/dirty_cat/test/test_docstrings.py
@@ -1,0 +1,172 @@
+import re
+from inspect import signature
+import pytest
+import inspect
+from importlib import import_module
+from typing import Optional
+
+numpydoc_validation = pytest.importorskip("numpydoc.validate")
+
+
+def all_estimators():
+    module = import_module("dirty_cat")
+    classes = inspect.getmembers(module, inspect.isclass)
+    classes = [
+        (name, est_cls) for name, est_cls in classes if not name.startswith("_")
+    ]
+    return sorted(classes, key=lambda x: x[0])
+
+
+def get_all_methods():
+    estimators = all_estimators()
+    for name, Estimator in estimators:
+        if name.startswith("_"):
+            # skip private classes
+            continue
+        methods = []
+        for name in dir(Estimator):
+            if name.startswith("_"):
+                continue
+            method_obj = getattr(Estimator, name)
+            if hasattr(method_obj, "__call__") or isinstance(method_obj, property):
+                methods.append(name)
+        methods.append(None)
+
+        for method in sorted(methods, key=str):
+            yield Estimator, method
+
+
+def repr_errors(res, estimator=None, method: Optional[str] = None) -> str:
+    """Pretty print original docstring and the obtained errors
+
+    Parameters
+    ----------
+    res : dict
+        result of numpydoc.validate.validate
+    estimator : {estimator, None}
+        estimator object or None
+    method : str
+        if estimator is not None, either the method name or None.
+
+    Returns
+    -------
+    str
+       String representation of the error.
+    """
+    if method is None:
+        if hasattr(estimator, "__init__"):
+            method = "__init__"
+        elif estimator is None:
+            raise ValueError("At least one of estimator, method should be provided")
+        else:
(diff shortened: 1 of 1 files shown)


dirty_cat/test/test_docstrings.py
```diff
@@ -0,0 +1,172 @@
+import re
```
GaelVaroquaux, line 1: A think that a module-level docstring giving a short description of what is in this module would be useful (for the future developers discovering the codebase)

dirty_cat/test/test_docstrings.py
```diff
@@ -0,0 +1,172 @@
...
+from inspect import signature
+import pytest
+import inspect
```
GaelVaroquaux, line 4: It's a minor nitpick, but it's nice to order the imports in a way. Right now we have two imports of inspect, but not consecutive. The typical ordering that I know is: std lib, classic packages, less classic packages, local import. And inside a given category, alphabetic ordering.

GaelVaroquaux (commented):
Overall, I think that this looks great.

One thing that I do not fully understand is: is this already executed during the test? I'm surprised that we don't need a blacklist of things that fail.

dirty_cat/test/test_docstrings.py
jjerphan: FYI, `isort`, set in the `pre-commit` setup should sort the import canonically.

If it has not, you can run:

```bash
pre-commit run --from-ref=upstream/master --to-ref=prepare_docstring_meta_issue
```

dirty_cat/test/test_docstrings.py
Vincent-Maladiere: Okay, I added a small description, let me know what you think.

dirty_cat/test/test_docstrings.py
Vincent-Maladiere: Thanks for pointing this out and thanks to @jjerphan for the pre-commit tricks. It is now fixed.

Vincent-Maladiere:
Thanks for your feedback @GaelVaroquaux. The pytest main function was indeed missing, hence no docstring tests were triggered.

dirty_cat/test/test_docstrings.py
LilianBoulard: I think there was a misunderstanding here: from what I understand, your description explains why we use `re`, but I think what Gaël meant was to clarify what this file (`test_docstrings`) does.

I'm thinking of something along the lines of *"This test suite ensures the docstrings of functions and classes in dirty_cat are formatted according to numpydoc specifications"*.

dirty_cat/test/test_docstrings.py
LilianBoulard: I think there was a misunderstanding here: from what I understand, your description explains why we use `re`, but I think what Gaël meant was to clarify what this file (`test_docstrings`) does.

I'm thinking of something along the lines of *"This test suite ensures the docstrings of functions and classes in dirty_cat are formatted according to numpydoc specifications"*.

LilianBoulard (changes requested):
Thanks for the PR!
As mentioned in a conversation above, please fix the file docstring, and we should be good to merge :)

dirty_cat/test/test_docstrings.py
```diff
@@ -0,0 +1,245 @@
...
+to skip while running the validation tests, so that CI will
+not fail.
+
+Therefore, developpers having formatted methods to numpydoc
+should also remove their corresponding references from the list.
+"""
+import inspect
+
+# We use the regex module to detect camelCase characters,
+# corresponding to python class name.
```
LilianBoulard, line 15:
```suggestion
```

LilianBoulard (approved):

Issue #306, opened 2022-09-02
Deduplication


LilianBoulard:
As discussed with Gaël and Jovan, we could get some inspiration from the [dedupe library](https://github.com/dedupeio/dedupe) for the fuzzy join and deduplication features.

mjboos:
Summary of my current understanding of the different approaches one could take to this:

### Dedupe
Dedupe allows deduplication based on multiple fields and (simplifying a bit) computes a string similarity (levenshtein AFAICT) between fields for each pair of records. It then performs hierarchical clustering on the similarities between records to identify duplicates. There are three tricks/fancy things it uses on top of that:
- It scales this by pre-selecting blocks of data to compare with each other based on (more computationally efficient) "predicate functions", e.g. only do full string comparison of records if they share the first three digits (+ many more predicate functions and combinations thereof), see [here](https://docs.dedupe.io/en/latest/how-it-works/Making-smart-comparisons.html#).
- It computes a string similarity between two records for each field independently, i.e. 5 fields lead to 5 string similarities (between two records). Since not all fields are equally informative, it uses a logistic regression with similarity per field as features to predict whether something is a duplicate (and hence be able to down-weight certain fields via logistic reg. weights). It learns them via presenting the most uncertain examples to the user and collecting ground truth about whether they are duplicates from them.
- It selects the cut-off for hierarchical clustering by precision/recall based on the ground truth data collected from the user via the active inference above.

Disadvantage is clearly the complexity and maybe the reliance on user input, yet using logistic regression or similar for weighting different fields seems very useful.

### Simplest alternative using native dirty_cat and scipy

[This](https://mjboos.github.io/posts/2021/11/unsupervised-typo-finder/) example uses a similar principle, but much simplified using only dirty_cat and scipy clustering: For a single field compute similarities via `SimilarityEncoder`, then perform hierarchical clustering on this and let the user pick the cut-off. Basically the approach that dedupe takes using dirty_cat's string similarities without the fancy things above. It would struggle with scaling (always performs an all-to-all string similarity), doesn't have a principled way to do similarity across 2+ fields/columns jointly, and has no heuristic to pick the cut-off.

### Building on FuzzyJoin

It would be nice to build this on the fuzzyjoin PR, since this already handles matching on multiple columns - but we might want to have better string similarity than n-gram CountVectorizer and for clustering (and potentially weighting columns differently) we might need user input/ground truth labels.

### Open questions

IMO the biggest ones are:

[...]

PR #339, opened 2022-09-11
FEA Deduplication

mjboos:
This PR implements a basic `deduplicate` function (see discussion in #306 ).
It is based on computing n-gram tfidf distances between unique words in a column (same as for `fuzzy_join`), hierarchically clustering the distance matrix, then deduplicating by picking the most frequent exemplar in each cluster as the "correct" spelling.

Open design question: Should this be a function or a class? It clearly has a state (and could be useful when "trained" on a large set of dirty categories and then applied to new incoming data) - but saving a translation table in a class would be costly. An alternative would be to go from a clustered distance matrix to a K-means representation based on string distances (where each centroid is the most frequent cluster exemplar).

It's currently WIP because I still want to add an example/documentation.

FWIW this function was immediately useful in my work, but its usefulness depends on the structure of your data:
If there are few underlying categories, which might be corrupted by typos or alternative spellings, then it should work well out of the box.


jovan-stojanovic:
Hi @mjboos, thanks for opening this PR, exciting work ahead.
Just so you know, on Wednesday during the day we plan to have a sprint during which we will take a closer look at the problem.
We may come up with some comments and answers to your questions.

mjboos:
Thanks for the update @jovan-stojanovic !
I added a bare-bones example (examples/07_deduplication.py) to show a use case for this work, let me know if you have any questions/ideas for future directions.

mjboos:
@jovan-stojanovic @GaelVaroquaux Initially I wanted to extend the example, but think it's better to only do that after you gave some feedback on the PR/design questions I raised above.

Seems like the test is failing due to some missing approval/workflow configuration - let me know if it's something on my end.

GaelVaroquaux:
I approved running the tests.

GaelVaroquaux:
Trivial comment: the file should be named "_deduplicate.py", to signal that it is private and that people should not be importing from it.

dirty_cat/deduplicate.py
```diff
@@ -0,0 +1,177 @@
...
+        original_spelling.extend(sorted_spellings.tolist())
+        # assumes spelling that occurs most frequently in cluster is correct
+        corrected_spelling.extend(
+            np.repeat(sorted_spellings[0], sorted_spellings.shape[0])
+        )
+    pd_spell_correct = pd.Series(corrected_spelling, index=original_spelling)
+    return pd_spell_correct
+
+
+def compute_ngram_distance(
```
GaelVaroquaux, line 150: As this is used in the "deduplicate" function, convention goes that it should come before: define the functions that you need before using them. It used to be necessary for compilation, but it no longer is.

dirty_cat/test/test_deduplicate.py
```diff
@@ -0,0 +1,119 @@
...
+    Characters are misspelled with probability `prob_mistake_per_letter`"""
+
+    data = []
+    for example, n_ex in zip(examples, entries_per_example):
+        len_ex = len(example)
+        # generate a 2D array of chars of size (n_ex, len_ex)
+        str_as_list = np.array([list(example)] * n_ex)
+        # randomly choose which characters are misspelled
+        idxes = np.where(
+            np.random.random(len(example[0]) * n_ex) < prob_mistake_per_letter
```
GaelVaroquaux, line 28: You need to use a seeded local random number generator: at the beginning of the function, define
```python
rng = np.random.RandomState(seed)
```
and use it's methods

examples/08_deduplication.py
```diff
@@ -0,0 +1,163 @@
...
+_ = plt.ylabel("Counts")
+
+
+# We can now see clearly the structure of the data: The three original medications are the most common ones, however there are many spelling mistakes and hence many slight variations of the names of the original medications.
+#
+# The idea is to use the fact that the string-distance of each misspelled medication name will be closest to either the correctly or incorrectly spelled orginal medication name - and therefore form clusters.
+
+
+
+from dirty_cat.deduplicate import deduplicate, compute_ngram_distance
```
GaelVaroquaux, line 94: I would import the function deduplicate from the main `__init__` of dirty-cat. This is where we have been putting all our user-facing functions and objects so far.

examples/08_deduplication.py
```diff
@@ -0,0 +1,163 @@
...
+
+import seaborn as sns
+
+fig, axes = plt.subplots(1, 1, figsize=(12, 12))
+sns.heatmap(
+    square_distances, yticklabels=ex_series.index, xticklabels=ex_series.index, ax=axes
+)
+
+
+# ### Clustering to suggest corrections of misspelled names
```
GaelVaroquaux, line 115: Can you use sphinx-gallery markup here to do titles and line separation. A bit of doc here:
https://sphinx-gallery.github.io/stable/syntax.html#embedding-rst
(it's mostly using restructured text in the comments, with the proper separator at the top)

GaelVaroquaux:
I had a quick look, and overall things look good. I made a bunch of minor comments, but nothing on the overall design. Let me try to think about it more. Yet, so far, so good!

dirty_cat/deduplicate.py
```diff
@@ -0,0 +1,177 @@
...
+    for cluster in np.unique(clusters):
+        sorted_spellings = (
+            count_series.loc[clusters == cluster]
+            .sort_values(ascending=False)
+            .index.values
+        )
+        original_spelling.extend(sorted_spellings.tolist())
+        # assumes spelling that occurs most frequently in cluster is correct
+        corrected_spelling.extend(
+            np.repeat(sorted_spellings[0], sorted_spellings.shape[0])
```
GaelVaroquaux, line 144: Don't use np.repeat as you need a list. Use multiplication on a list:
n * [1, ], this is better because you need a list at the end of the day, rather than a numpy array

mjboos:
Thanks for the feedback @GaelVaroquaux I implemented the suggestions

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,175 @@
...
+    categories is significantly smaller than the number of observed spellings.
+
+    Parameters
+    ----------
+    data : Sequence[str]
+        The data to be deduplicated.
+    n_clusters : Optional[int], optional
+        number of clusters to use for hierarchical clustering, if None use the
+        number of clusters that lead to the lowest silhouette score,
+        by default None
```
jovan-stojanovic, line 69: On the docs, best is to see what was used before ([e.g. fuzzy_join](https://github.com/dirty-cat/dirty_cat/blob/master/dirty_cat/_fuzzy_join.py)) and try to implement the same. 
We try to stay consistent with work on previous features. For instance, this should be:
```suggestion
    n_clusters : typing.Optional[int], default=None
        Number of clusters to use for hierarchical clustering, if None use the
        number of clusters that lead to the lowest silhouette score.
```

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,175 @@
...
+        `CountVectorizer` analyzer for computing n-grams, by default "char_wb"
+    method : str, optional
+        linkage method to use for merging clusters, by default "average"
+    return_translation_table : bool, optional
+        Whether to return a translation table with original words as indices
+        and deduplicated words as values, by default False
+
+    Returns
+    -------
+    Union[List[str], Tuple[List[str], pd.Series]]
```
jovan-stojanovic, line 82: What (object name) is returned?

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,175 @@
...
+    if n_clusters is None:
+        n_clusters = guess_clusters(Z, distance_mat)
+    clusters = fcluster(Z, n_clusters, criterion="maxclust")
+
+    translation_table = create_spelling_correction(unique_words, counts, clusters)
+    unrolled_corrections = translation_table[data]
+    if return_translation_table:
+        return unrolled_corrections, translation_table
+    else:
+        return unrolled_corrections
```
jovan-stojanovic, line 102: I had the same idea for fuzzy_join, but some users pointed out that the simplest was to return one table, with an additional column containing "non-translated" values. It may be best to do here as well. From experience, the user may then more easily observe results and correct if needed.

jovan-stojanovic (changes requested):
Thanks for the great work! Some comments from my side as well.
On your open question, I believe this should be a function, so I agree with what you did.

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,175 @@
...
+
+import numpy as np
+import pandas as pd
+from scipy.cluster.hierarchy import fcluster, linkage
+from scipy.spatial.distance import pdist, squareform
+from sklearn.feature_extraction.text import CountVectorizer, TfidfTransformer
+from sklearn.metrics import silhouette_score
+
+
+def compute_ngram_distance(
```
jovan-stojanovic, line 16: I think the three support functions guess_clusters, compute_ngram_distance, create_spelling_correction should be made private (add a _ before the name)

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,175 @@
...
+    analyzer : str, optional
+        Analyzer to extract n-grams, by default "char_wb"
+
+    Returns
+    -------
+    np.ndarray
+        An n-times-(n-1)/2 matrix of n-gram TfIdf distances between `unique_words`.
+    """
+    enc = CountVectorizer(ngram_range=ngram_range, analyzer=analyzer)
+    encoded = TfidfTransformer().fit_transform(enc.fit_transform(unique_words))
```
jovan-stojanovic, line 40: This is equivalent to using scikit-learn's [TDIDF Vectorizer](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html).

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,175 @@
...
+    int
+        number of clusters that maximize the silhouette score.
+    """
+    max_clusters = distance_mat.shape[0]
+    n_clusters = np.arange(2, max_clusters)
+    # silhouette score needs a redundant distance matrix
+    redundant_dist = squareform(distance_mat)
+    silhouette_scores = []
+    for n_clust in n_clusters:
+        labels = fcluster(Z, n_clust, criterion="maxclust")
```
jovan-stojanovic, line 127: Why did you choose criterion `maxclust`?

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,175 @@
...
+    Parameters
+    ----------
+    unique_words : Union[Sequence[str], np.ndarray]
+        A sequence or array of unique words in the original data.
+    counts : Union[Sequence[int], np.ndarray]
+        A sequence or array of counts of how often each unique word appears in
+        the original data.
+    count_series : pd.Series
+        A series with unique words (in the original data) as indices and number
+        of occurrences of each word in the original data as values.
```
jovan-stojanovic, line 151: This is no longer a parameter, right?

jovan-stojanovic (commented):
Some more comments.This is going to be a great new feature, thanks.

dirty_cat/_deduplicate.py
mjboos: Not sure I understand, you mean returning translated data as a table with suggested translations (or non-translated values)?
Like `translation_table`, just for all data instead of the unique examples (`translation_table` contains unique strings/categories + suggested translations, or the original if no translation was suggested)?

dirty_cat/_deduplicate.py
mjboos: More out of ease-of-use than performance considerations.
I found it easier for me (or other users) to "guess" the number of unique categories that is present in the data, than have an intuition about minimum cluster distance etc.

mjboos:
Thanks for the comments @jovan-stojanovic !
I implemented your suggestions, but had some additional questions (took some time to get back to it, life happens).

When I used the function in the past month, I stumbled across another use case:
Often I notice dirty categorical data in a database/data warehouse, this data is usually too large to download and put into the function itself. So I already create a table of unique values and their counts and then use this as input (instead of the raw data).
To fix typos, I create the translation table and use it to join in the database.
I'm not sure if this is a workflow you want to support in dirty cat, but I could adapt the function to accept already unique strings + their counts (instead of the raw data/list of strings).
Trade off is that we overload the parameter (can be either unique values + counts or just list of potentially duplicate strings) and I'm not sure whether dirty cat is often used with data too large to fit into memory.

dirty_cat/_deduplicate.py
jovan-stojanovic: I actually meant returning a merged table `unrolled_corrections` + `translation_table` if `return_translation_table` is True.

jovan-stojanovic: Ok actually, I think the best is that this function only returns `unrolled_corrections table`. 
The main problem is that the output format changes depending on parameters value, which is usually not recommended.
Translation tables can be easily extracted then by dropping duplicates in the dirty data column..

jovan-stojanovic:
Thanks! Interesting use case but there is something I don't understand:
> So I already create a table of unique values and their counts and then use this as input (instead of the raw data). To fix typos, I create the translation table and use it to join in the database. 

Isn't this exactly what the `deduplicate` function does? So the only difference would be that you would manually provide `unique words` and `counts` instead of doing what is in the first line `unique_words, counts = np.unique(data, return_counts=True)`
But then again, you would need to do this to get the counts and uniques, only outside of the function?

dirty_cat/_deduplicate.py
mjboos: I agree, it should always return the same output type.
Okay, will return the table you described above with translated/corrected and uncorrected values (lmk if I misunderstood).

mjboos:
> Thanks! Interesting use case but there is something I don't understand:
> 
> > So I already create a table of unique values and their counts and then use this as input (instead of the raw data). To fix typos, I create the translation table and use it to join in the database.
> 
> Isn't this exactly what the `deduplicate` function does? So the only difference would be that you would manually provide `unique words` and `counts` instead of doing what is in the first line `unique_words, counts = np.unique(data, return_counts=True)` But then again, you would need to do this to get the counts and uniques, only outside of the function?

Exactly, (locally) I allow `counts` as optional argument, if it is present I assume `data` already refers to unique strings.
I get the counts & uniques via a cloud database (e.g., google cloud) and download only these. For me, this is helpful because I have very large datasets in the cloud and don't want to get the full, raw data - the correction then happens in the cloud too, by uploading the `translation_table` as a literal table and joining on it.
It's quite useful for me, but I'm not sure how many `dirty_cat` users will use a similar workflow (hence trade-off of overloading the parameters vs. supporting this workflow).

jovan-stojanovic:
I see, then best to do in my opinion is the following:
- finish this version of deduplication;
- add an issue mentioning this, to keep track of it and see if it affect others.

mjboos:
> I see, then best to do in my opinion is the following:
> 
> * finish this version of deduplication;
> * add an issue mentioning this, to keep track of it and see if it affect others.

Sounds good @jovan-stojanovic .
I changed the function to always return the deduplicated data and adapted the example to show how to re-create the translation table - that should make it easier to use.
Lmk if that's enough.

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,173 @@
...
+        Analyzer to extract n-grams
+
+    Returns
+    -------
+    np.ndarray
+        An n-times-(n-1)/2 matrix of n-gram TfIdf distances between `unique_words`.
+    """
+    encoded = TfidfVectorizer(ngram_range=ngram_range, analyzer=analyzer).fit_transform(
+        unique_words
+    )
```
LilianBoulard, line 41: Just a restyle
```suggestion
    encoded = (
        TfidfVectorizer(ngram_range=ngram_range, analyzer=analyzer)
        .fit_transform(unique_words)
    )
```

examples/06_deduplication.py
```diff
@@ -0,0 +1,186 @@
...
+            for i in np.random.choice(np.arange(26), len(idxes)).astype(int)
+        ]
+        # introduce spelling mistakes at right examples and right char locations per example
+        str_as_list[idxes // len_ex, idxes % len_ex] = replacements
+        # go back to 1d array of strings
+        data.append(np.ascontiguousarray(str_as_list).view(f"U{len_ex}").ravel())
+    return np.concatenate(data)
+
+
+np.random.seed(123)
```
LilianBoulard, line 67:
```suggestion
# set seed for reproducibility
np.random.seed(123)
```

examples/06_deduplication.py
```diff
@@ -0,0 +1,186 @@
...
+# We can visualize the distribution of categories in the deduplicated data:
+
+deduplicated_unique_examples, deduplicated_counts = np.unique(
+    deduplicated_data, return_counts=True
+)
+deduplicated_series = pd.Series(deduplicated_counts, index=deduplicated_unique_examples)
+
+deduplicated_series.plot.barh(figsize=(10, 15))
+_ = plt.xlabel("Medication name")
+_ = plt.ylabel("Counts")
```
LilianBoulard, line 146: Is there a specific reason why we need to get/store the function output? Otherwise it's much cleaner to do
```suggestion
plt.xlabel("Medication name")
plt.ylabel("Counts")
```

examples/06_deduplication.py
```diff
@@ -0,0 +1,186 @@
...
+# and build a series out of them
+ex_series = pd.Series(counts, index=unique_examples)
+
+###############################################################################
+# Visualize the data
+# ------------------
+
+ex_series.plot.barh(figsize=(10, 15))
+_ = plt.xlabel("Medication name")
+_ = plt.ylabel("Counts")
```
LilianBoulard, line 90: Same thing here
```suggestion
plt.xlabel("Medication name")
plt.ylabel("Counts")
```

examples/06_deduplication.py
```diff
@@ -0,0 +1,186 @@
...
+# most common exemplars (the guessed correct spelling):
+
+
+def print_corrections(spell_correct):
+    correct = np.unique(spell_correct.values)
+    for c in correct:
+        print(
+            f"Guessed correct spelling: {c} for"
+            f" {spell_correct[spell_correct==c].index.values}"
+        )
```
LilianBoulard, line 176:
```suggestion
        print(
            f"Guessed correct spelling: {c!r} for"
            f" {spell_correct[spell_correct==c].index.values}"
        )
```

LilianBoulard (approved):
Excellent PR, thanks a lot!

examples/06_deduplication.py
mjboos: Just personal preference to not print the label to stdout (and have it show up in the cell).
But it's probably easier to get rid of this idiosyncrasy :)

examples/06_deduplication.py
mjboos: Not sure why `!r` is needed - `c` will be strings -> no `repr` call required?

mjboos:
Thanks @jovan-stojanovic @LilianBoulard
I accepted all suggestions apart from the one where I had an additional question :)

examples/06_deduplication.py
LilianBoulard: Good point, the repr call is actually a way to put the value between single quotes
```
>>> print(f"Got unexpected value: {s}")
Got unexpected value: value
>>> print(f"Got unexpected value: {s!r}")
Got unexpected value: 'value'
```

mjboos: I see, thanks for the explanation, makes sense

dirty_cat/_deduplicate.py
```diff
@@ -0,0 +1,174 @@
...
+
+import numpy as np
+import pandas as pd
+from scipy.cluster.hierarchy import fcluster, linkage
+from scipy.spatial.distance import pdist, squareform
+from sklearn.feature_extraction.text import TfidfVectorizer
+from sklearn.metrics import silhouette_score
+
+
+def _compute_ngram_distance(
```
jovan-stojanovic, line 16:
```suggestion
def compute_ngram_distance(
```
You need to make this function public (as it is used in the example).

examples/06_deduplication.py
```diff
@@ -0,0 +1,187 @@
...
+from scipy.spatial.distance import squareform
+
+ngram_distances = _compute_ngram_distance(unique_examples)
+square_distances = squareform(ngram_distances)
+
+###############################################################################
+# We can visualize the pair-wise distance between all medication names
+# --------------------------------------------------------------------
+#
+# Below I use a heatmap to visualize the pairwise-distance between medication names.
```
jovan-stojanovic, line 114:
```suggestion
# Below we use a heatmap to visualize the pairwise-distance between medication names.
```
'We' should used instead of 'I' in examples.

examples/06_deduplication.py
```diff
@@ -0,0 +1,187 @@
...
+# The idea is to use the fact that the string-distance of each misspelled medication
+# name will be closest to either the correctly or incorrectly spelled orginal
+# medication name - and therefore form clusters.
+
+
+from dirty_cat import deduplicate
+from dirty_cat._deduplicate import _compute_ngram_distance
+from scipy.spatial.distance import squareform
+
+ngram_distances = _compute_ngram_distance(unique_examples)
```
jovan-stojanovic, line 107: Function to be made public (see above)
```suggestion
from dirty_cat._deduplicate import compute_ngram_distance
from scipy.spatial.distance import squareform

ngram_distances = compute_ngram_distance(unique_examples)
```

examples/06_deduplication.py
```diff
@@ -0,0 +1,187 @@
...
+        # and randomly pick with which character to replace
+        replacements = [
+            string.ascii_lowercase[i]
+            for i in np.random.choice(np.arange(26), len(idxes)).astype(int)
+        ]
+        # introduce spelling mistakes at right examples and right char locations per example
+        str_as_list[idxes // len_ex, idxes % len_ex] = replacements
+        # go back to 1d array of strings
+        data.append(np.ascontiguousarray(str_as_list).view(f"U{len_ex}").ravel())
+    return np.concatenate(data)
```
jovan-stojanovic, line 64: Best is to move this function into [datasets._fetching](https://github.com/dirty-cat/dirty_cat/blob/main/dirty_cat/datasets/_fetching.py) and rename it to something like `make_deduplication` so that we use it whenever we want to generate an example for this.
And to have a one-line import rather than a function makes the important features of this example much more visible.

Note: I see you used the same function in the [tests](https://github.com/dirty-cat/dirty_cat/blob/09a88c4053a031965ec8f08c6331491cceec5a51/dirty_cat/tests/test_deduplicate.py#L18), so maybe import it there as well

dirty_cat/tests/test_deduplicate.py
```diff
@@ -0,0 +1,124 @@
...
+        # and randomly pick with which character to replace
+        replacements = [
+            string.ascii_lowercase[i]
+            for i in rng.choice(np.arange(26), len(idxes)).astype(int)
+        ]
+        # introduce spelling mistakes at right examples and char locations per example
+        str_as_list[idxes // len_ex, idxes % len_ex] = replacements
+        # go back to 1d array of strings
+        data.append(np.ascontiguousarray(str_as_list).view(f"U{len_ex}").ravel())
+    return np.concatenate(data)
```
jovan-stojanovic, line 40: Import this function from datasets, see comment in example section.

jovan-stojanovic (changes requested):
A few remarks came up to us just before closing this:

Hope this is it this time..

GaelVaroquaux:
> You need to make this function public (as it is used in the example).

It implies: changing the naming, importing it in the main namespace (init) of dirty_cat and documenting it.

> Best is to move this function into datasets._fetching and rename it to something like make_deduplication

I'd rather use as a name "make_deduplication_data"

LilianBoulard:
Could you also add a changelog major change entry for version 0.4.0? Thanks!

LilianBoulard:
Hey @mjboos
We are looking to release a new version of dirty_cat within the next two weeks, and we're excited to have the deduplication as part of 0.4! Thus we need to merge this PR :)
Do you need any help/guidance on how to operate the previously suggested changes?
Thanks!

mjboos:
Sure, I can prioritize it and get it done in the next few days :)

Am Mo., 9. Jan. 2023 um 14:50 Uhr schrieb Lilian ***@***.***>:

jovan-stojanovic:
Hey @mjboos, we are ready for a beta release, would be good to have this as well. Do you need assistance with the few changes left?

dirty_cat/tests/test_deduplicate.py
mjboos: I left the function, because it's slightly different - for ease of testing I pass the randomstate explicitly here, but for the user facing function in `datasets` this makes it unncecessarily complicated IMO.
If you prefer to make it DRY-er, I can add the randomstate argument to the function in `datasets` as well.

mjboos:
Sure, I accepted the suggestions and made the function public. I kept the different function for testing though @jovan-stojanovic and detailed why I did above (but can change it).
Also added a change log @LilianBoulard lmk if you want to have it another way or want to add attribution to suggesters of changes as well.

jovan-stojanovic (approved):
Thanks, will merge when the tests pass!

dirty_cat/tests/test_deduplicate.py
jovan-stojanovic: Fine for me. I think the most important was to make the example lighter. If this one is different you can keep it in the tests.

mjboos:
> The test_docstrings.py are failing (you should update the docstrings to meet the numpydoc standards).

Oops, fixed it.

jovan-stojanovic:
It's green, merging. Thanks!

PR #355, opened 2022-09-14
FIX use sklearn.utils.fixes.parse_version instead of private Version

glemaitre:
Solve the current issue.

I will check if the function is available for older scikit-learn


glemaitre:
`parse_version` is available in the oldest scikit-learn version supported.

LilianBoulard (approved):

Issue #330, opened 2022-09-07
Generated doc is broken


LilianBoulard:
Since we moved to numpydoc (#316), the documentation is kind of broken for the parameters types, [for example in the SuperVectorizer](https://dirty-cat.github.io/stable/generated/dirty_cat.SuperVectorizer.html#dirty_cat.SuperVectorizer)
![image](https://user-images.githubusercontent.com/33222654/188840814-c984030f-7989-49f8-a499-bcfeb9c26813.png)

Before that, the resolution was effective, so [we had that](https://output.circle-artifacts.com/output/job/d5c4f7e9-3cf8-4faf-b197-e769be49d585/artifacts/0/doc/generated/dirty_cat.SuperVectorizer.html#dirty_cat.SuperVectorizer)
![image](https://user-images.githubusercontent.com/33222654/188841117-2a2010a9-77bd-4fbc-8f27-221a42fe6fe4.png)

GaelVaroquaux:
Does this resolution work with numpydoc in another project?

LilianBoulard:
So, to answer the question, no I haven't!
Though I've noticed sometimes it works and sometimes it doesn't, depending on the section, and I don't necessarily understand why.
So in the end, by merely tweaking the docstring of the `SuperVectorizer` as an example and following carefully the [style guide](https://numpydoc.readthedocs.io/en/latest/format.html), I was able to get a good result, so it comes down to resolving #345.

PR #359, opened 2022-09-14
Fix fetching functions doc generation

### LINKED ISSUE #357: Fetching functions' style is not great
image

Two issue I notice:
- The title is overlapping with the logo (linked to #349)
- The quote doesn't work
- There is no link to `DatasetInfoOnly` and `DatasetAll` (linked to #330)

LilianBoulard:
Fixes #357

Also standardize `DatasetAll` and `DatasetInfoOnly` attributes.


GaelVaroquaux:
Can you resolve the conflicts please, so that CI runs and we can check the generated docs

GaelVaroquaux:
Darn, I made a comment that got lost (probably from my mobile phone).

So, one problem is that the links of DatasetAll and DatasetInfoOnly points nowhere. For them to point somewhere, there needs to be a page where these objects are documented in the docs (as the GapEnoder is documented on the front page). I would prefer not to add them on the front page, and to create a dedicated page for them. This page does not need to be linked anywhere and should focus on documenting those data structures with only a sentence in the beginning of the page.

dirty_cat/datasets/_fetching.py
```diff
@@ -61,16 +61,33 @@ class Features:
...
+
     name: str
     description: str
+    source: str
+    target: str
     X: pd.DataFrame
     y: pd.Series
-    source: str
     path: Path
+    read_csv_kwargs: Dict[str, Any]
```
jovan-stojanovic, line 77: You've explained really well below why `read_csv_kwargs` was added to `DatasetInfoOnly` but I don't see the utility for the user of adding it here. It multiplies information in this already complex class and is not used in this case.

jovan-stojanovic (changes requested):

dirty_cat/datasets/_fetching.py
LilianBoulard: I added a little clarification for why we pass this information, even though the dataset is loaded in memory.
Mainly, it's because we want to have this information in case the dataframe has to be read from the file again.

LilianBoulard:
So, regarding the missing link and generated doc of the `DatasetAll` and `DatasetInfoOnly`, adding `unlisted.rst` did generate the page for `DatasetAll` and `DatasetInfoOnly` ([eg](https://output.circle-artifacts.com/output/job/823926b3-546d-4b55-ab05-6f217fa3c19f/artifacts/0/doc/generated/dirty_cat.datasets.DatasetAll.html)). However, they are still not linked in the return types of the fetching functions.
This is now a problem linked to #330.

GaelVaroquaux:
OK, this is still an improvement. So I am merging it, even if some problems remain.

PR #331, opened 2022-09-07
Make unnecessary API private

LilianBoulard:
Fixes #315

PR #364, opened 2022-09-15
Restore test_docstrings.py

LilianBoulard:
Thanks to a comment on the numpydoc meta issue, I've noticed that the file `dirty_cat/tests/test_docstrings.py` was empty in master.
A bisect identified the #331 as the culprit, which is now obvious given the +/- diff.
I've reintroduced the file, which is not necessarily the best solution (a cherrypick might have been a better one, but I have next to no experience with it).


LilianBoulard:
I'm merging right away as to not block other contributors, we can fix git later!

Issue #312, opened 2022-09-02
Switch to `pyproject.toml`


LilianBoulard:
We should investigate switching to `pyproject.toml` for our packaging management.

LilianBoulard:
I also think we should move the dependency listing from `requirements.txt` and `requirements-min.txt` to somewhere else (maybe only the toml file, I'm not too familiar with it TBH).

Vincent-Maladiere:
Working on it!

LilianBoulard:
Solved by #365

PR #365, opened 2022-09-15
add deps version checks at runtime

Vincent-Maladiere:
### What does this implement?

Addresses #312 

1. Migrate metadata from `setup.py` to `setup.cfg` and prepare the ground for "everything on `pyproject.toml`".
2. During the setup in `setup.py`, we parse `setup.cfg` to extract the `required_dependencies`, and create a `dirty_cat/_config_requirements.py` file.
3.  Finally, when `dirty_cat` is imported, we check that the installed dependencies versions match the required versions from the `dirty_cat/_config_requirements.py` file and raise an `ImportError` otherwise.

- Also, some housekeeping by removing `requirements.txt` and `requirements-min.txt` that are no longer used thanks to the `setup.cfg` and updating `.github/workflow/testing.yml` accordingly.

### Comments

Prior to that PR, pip reduces the likelihood of making mistakes by raising an error when the user updates a package to a version that is out of the required version scope. However, after the error message, the updated version is installed anyway.

Screen Shot 2022-09-15 at 15 43 42

Therefore, this error is non-blocking and can easily be missed or skipped by the user, which will have weird bugs after importing the module.

This PR helps enforce the required versions by raising an error when importing `dirty_cat` with incorrect dependencies versions

```shell
$ python -c "import dirty_cat"
  File "<string>", line 1, in <module>
  File "/Users/vincentmaladiere/INRIA/dirty_cat/dirty_cat/__init__.py", line 42, in <module>
    _check_pack_version(dep, package, package_version, sign)
  File "/Users/vincentmaladiere/INRIA/dirty_cat/dirty_cat/__init__.py", line 15, in _check_pack_version
    raise ImportError(
ImportError: dirty_cat 0.3.0 requires scikit-learn>=0.23.0, but you have scikit-learn 0.22 which is incompatible.
```


Vincent-Maladiere:
@GaelVaroquaux, I think I have found a simple solution to our import issue with this PR, let me know what you think :)

Vincent-Maladiere:
@LilianBoulard this also fixes the issues you had with numpydoc today —however, codecov and CircleCI fail for some reasons that I believed are unrelated to this PR.

setup.py
```diff
@@ -1,49 +1,34 @@
...
+    setup_file = "setup.cfg"
+    config = ConfigParser()
+    config.read(setup_file)
+    deps = config["options"]["install_requires"].strip()
+    project_name = config["metadata"]["name"]
+    project_dir = _Path(__file__).parent / project_name
+    req_path = os.path.join(project_dir, "_config_requirements.py")
+    with open(req_path, "w+") as f:
+        body = "deps = " + str({"deps": deps})
+        f.write(body)
```
LilianBoulard, line 30: As we're using pathlib, I guess it could be
```suggestion
    req_path = project_dir / "_config_requirements.py"
    req_path.write_text(f"deps = { {"deps": deps} }")
```
(not 100% sure this will run as expected)
Also, is there a specific reason you used "w+"? As you're only writing to the file, I assumed you would only need "w".

pyproject.toml
```diff
@@ -1,3 +1,12 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = [
+    "setuptools<60.0",
```
LilianBoulard, line 4: Is there a specific reason for this?
And why not use `setuptools==60.0`?
Also, [more recent versions are available](https://pypi.org/project/setuptools/#history).

.github/workflows/testing.yml
```diff
@@ -23,35 +23,6 @@ jobs:
...
-      - shell: bash -l {0}
-        run: ./build_tools/github/install.sh
-        name: 'Install dirty-cat'
-      - shell: bash -l {0}
-        run: ./build_tools/github/test.sh
-        name: 'Run tests'
-      - uses: codecov/codecov-action@v1
-        if: success()
-        name: 'Upload coverage to CodeCov'
-
```
LilianBoulard, line 54: By removing this, I believe we don't test the package with the minimal requirements anymore (I don't see a corresponding action in the CI of this PR).

Vincent-Maladiere: Yes we do, the minimal requirements are now the default (see `setup.cfg`).

pyproject.toml
Vincent-Maladiere: I simply stuck to scikit-learn for this one, but we could definitely try to use it for more recent versions.

setup.py
Vincent-Maladiere: Yes, we should use `pathlib` here. And I think I put `"w+"` by pure reflex but yes we can limit it to `"w"`.

dirty_cat/_check_dependencies.py
```diff
@@ -0,0 +1,36 @@
...
+            f"dirty_cat {_version} requires {dep}, but you have "
+            f"{package_name} {installed_version} which is incompatible."
+        )
+
+
+parent_dir = _Path(__file__).parent
+with open(parent_dir / "VERSION.txt") as _fh:
+    _version = _fh.read().strip()
+
+deps = deps["deps"].split("\n")
```
Vincent-Maladiere, line 25: I will replace this part with this scikit-learn utils `from sklearn.externals._packaging.version import parse` that deals with version parsing way better than my simple implementation.

setup.py
```diff
@@ -1,49 +1,34 @@
...
+    required versions at run time, during the import of this package.
+    """
+    setup_file = "setup.cfg"
+    config = ConfigParser()
+    config.read(setup_file)
+    deps = config["options"]["install_requires"].strip()
+    project_name = config["metadata"]["name"]
+    project_dir = _Path(__file__).parent / project_name
+    req_path = project_dir / "_config_requirements.py"
+    with open(req_path, "w") as f:
```
Vincent-Maladiere, line 28: This part is quite hacky, but so is raising a version error during __init__. Let me know if you have suggestions of a more elegant solution.

dirty_cat/_check_dependencies.py
```diff
@@ -0,0 +1,15 @@
...
+    env = pkg_resources.Environment()
+    package = env[package_name][0]
+    dependencies = package.requires()
+    for dep in dependencies:
+        installed_dep = env[dep.name][0]
+        if installed_dep not in dep:
+            raise ImportError(
+                f"{package_name} {package.version} requires {dep} "
+                f"but you have {installed_dep}, which is not compatible."
+            )
```
fcharras, line 15: Few comments from our discussion earlier:

- replace `env[package_name][0]` with something like `next((installed_dep for installed_dep in env[dep.name] if installed_dep.project_name == dep.name))` else you might have issues if there are several packages with names that contain the string `dep.name`

- I think it's a good idea to keep the `setup.py` with just an empty `setup()` instruction, so you can keep using things like `python setup.py bdist_wheel`, I don't think there's a replacement for this syntax yet.

- Feature-wise, we check that the direct requirements of `dirty-cat` are met but we don't check the whole dependency tree, there still might be issues at lower level of the dependency tree. (I think it is the right thing to do but I wanted to mention it clearly somewhere)

And finally maybe ensure that it is safe to use `pkg_resources` (for me I can't think of environments where it wouldn't be available or wouldn't be working correctly)

jovan-stojanovic:
Do you think we need a .binder folder [just like in scikit-learn](https://github.com/scikit-learn/scikit-learn/tree/main/.binder), so that the notebooks in the example are well ? See #380

setup.cfg
```diff
@@ -1,3 +1,52 @@
...
+    pytest-cov
+    coverage
+    mypy
+    numpydoc
+    flake8
+doc =
+    sphinx-gallery
+    sphinxext-opengraph
+    matplotlib
+    seaborn
```
jovan-stojanovic, line 48: Please add `statsmodels` here as well (added as doc dependancy with the latest PR) :)

jovan-stojanovic (changes requested):

.github/workflows/testing.yml
GaelVaroquaux: OK, but do we test more recent versions of the dependencies?

Our CI scripts are a bit awkward and do not enable to easily test different combinaison of requirements (and this should chance, see #384 )

GaelVaroquaux: From what I can tell inspecting the tests of this PR, is seems that every test is now run with the min dependencies.

The tests should really sample a variety of packages versions. Granted, this is awkward to do with our current test system. But I really frown at the idea of merging something that decreases our already-thin test entropy.

The right thing to do would be to see how the version entropy is done in the tests for scikit-learn, and port it here.

pyproject.toml
```diff
@@ -1,3 +1,12 @@
+[build-system]
+# Minimum requirements for the build system to execute.
+requires = [
+    "setuptools",
+    "wheel",
+    "Cython>=0.28.5",
```
GaelVaroquaux, line 6: Do we need cython? I don't think that we use any cython, so I don't think that we need it here even for the build (unlike scikit-learn).

GaelVaroquaux (changes requested):
This PR looks really cool.

I did a minor comment (on Cython).

My big comment is with regards the variety of the versions installed in our test matrix. I realize that it is somewhat orthogonal to this PR (but not completely), but I really frown at merging a PR that decreases our test variety. Let's discuss what's the best way forward.

jovan-stojanovic (approved):

Vincent-Maladiere:
Our tests fail for Python 3.10 with minimal requirements. I must find the correct minimal version of numpy.

.github/workflows/testing.yml
```diff
@@ -61,7 +32,19 @@ jobs:
         matrix:
             os: ["ubuntu-latest", "macos-latest"]
             python-version: ["3.8", "3.9", "3.10"]
-    name: ${{ matrix.os }} with Python ${{ matrix.python-version }} and latest package versions
+            dependencies-version-type: ["min", "latest"]
+            include:
```
GaelVaroquaux, line 36: Add a comment to tell the reader where these things are defined.

setup.cfg
```diff
@@ -1,3 +1,69 @@
...
+    statsmodels
+# overwrite install requires for CI testing purposes
```
GaelVaroquaux, line 50: Here mention the file that points to this (the testing.yml)

GaelVaroquaux (commented):
Meta comment (I'm in a rush): this is a strong improvement. It just needs comments to help people discover how things are linked.

.github/workflows/testing.yml
```diff
@@ -61,7 +32,19 @@ jobs:
...
+              - dependencies-version: "dev, min-py38"
+                python-version: "3.8"
+                dependencies-version-type: "min"
+              - dependencies-version: "dev, min-py39"
+                python-version: "3.9"
+                dependencies-version-type: "min"
+              - dependencies-version: "dev, min-py310"
+                python-version: "3.10"
+                dependencies-version-type: "min"
+    name: ${{ matrix.os }} with Python ${{ matrix.python-version }} and $${{ matrix.dependencies-version-type }} package versions
```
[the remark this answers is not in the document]
Vincent-Maladiere: Thanks! I couldn't see it

setup.cfg
```diff
@@ -1,3 +1,69 @@
...
+min-py39 =
+    scikit-learn==0.23.0
+    numpy==1.17.3
+    scipy==1.6.0
+    pandas==1.2.0
+min-py310 =
+    scikit-learn==0.24.0
+    numpy==1.22.4
+    scipy==1.7.2
+    pandas==1.4.4
```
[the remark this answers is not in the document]
Vincent-Maladiere: Exactly, some package versions can't be installed with newer python versions.

Vincent-Maladiere:
We're almost good here, @LilianBoulard do you know why one test is failing for python 3.8? I can't reproduce it on my end. The error is about not being able to fetch a food dataset from openml but seems unrelated to the min dependencies. I will investigate this tomorrow.

LilianBoulard:
I'm not 100% sure, but I think the CircleCI error is linked to #389.
Try pushing renaming "master to "main" on your local env, and push it onto your fork.
Or maybe it has something to do with [`build.sh`](https://github.com/Vincent-Maladiere/dirty_cat/blob/migrate_to_pyproject/build_tools/circle/build_doc.sh)

LilianBoulard:
Hey, update: we fixed an issue with the CI on main, could you please pull the latest changes, see if that fixes CircleCI ?

Vincent-Maladiere:
@LilianBoulard the CI is still broken I'm afraid

LilianBoulard:
So strangely it seems to be an issue related to your user and not the project itself...
https://discuss.circleci.com/t/release-build-fails-because-deleted-the-circleci-oauth-app/45101
I'm eager to merge this so I guess we could try and play with fire by merging it anyway. I'm just waiting for https://github.com/Vincent-Maladiere/dirty_cat/pull/1 to be merged and we'll see how that goes.

LilianBoulard (approved):
Thanks a lot for the PR, merging! Let's cross fingers and hope CircleCI will work on main :crossed_fingers:

Issue #345, opened 2022-09-12
Ensure that docstrings pass numpydoc validation


LilianBoulard:
### Background / Objective

[Docstrings](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring) in Python are string literals that occur as the first statement in a module, function, class, or method definition.

These are some of the characteristics of a docstring:

- Triple quotes are used to encompass the docstring text.
- There is no blank line before or after the docstring.
- The docstring is a phrase ending in a period.
- [more details](https://www.python.org/dev/peps/pep-0257/#what-is-a-docstring)

[numpydoc](https://numpydoc.readthedocs.io/en/latest/format.html) is one set of criteria to check for consistent documentation structure.

### Validating docstrings in dirty_cat

To ensure consistent documentation structure in dirty_cat, we are using numpydoc validation. Currently, documentation tests are failing for various functions. As a temporary fix, we have suppressed error messages in [test_docstrings.py](https://github.com/dirty-cat/dirty_cat/blob/master/dirty_cat/tests/test_docstrings.py). Many of the functions in dirty_cat need to be updated to comply with numpy docstring validation. In the below issue, we provide step-by-step instructions on how contributors can test and update functions.

### Steps

1. Make sure you have the development dependencies and documentation dependencies installed (see [the contribution guidelines](https://github.com/dirty-cat/dirty_cat/blob/master/CONTRIBUTING.rst)).
2. Pick a function/class from the list below and leave a comment saying you are going to work on it. This way we can keep track of what everyone is working on.
2.1 Make sure you've created a separate branch from main before editing files for your new contribution. Refer to our [contributing guidelines](https://github.com/dirty-cat/dirty_cat/blob/master/CONTRIBUTING.rst) for more information.
3. Remove the function from the list `DOCSTRING_TEMP_IGNORE_SET` in [`test_docstring.py`](https://github.com/dirty-cat/dirty_cat/blob/master/dirty_cat/tests/test_docstrings.py#L22).
5. Let's say you picked `dirty_cat.gap_encoder.GapEncoder.fit`, run numpydoc validation as follows.
```commandline
pytest dirty_cat/tests/test_docstrings.py
```

6. If you see the test fails, please fix them by following the recommendation provided by the failing test.
7. If you see all the tests pass, you do not need to do any additional changes.
8. Commit your changes.
9. Open a Pull Request with an opening message Addresses #345. Note that each item should be submitted in a separate Pull Request.
[...]

PR #368, opened 2022-09-17
DOC Ensures that DatetimeEncoder.transform passes numpydoc validation

alexdt1982:
Addresses #345

### DIFF
--- a/dirty_cat/_datetime_encoder.py
+++ b/dirty_cat/_datetime_encoder.py
@@ -184,9 +184,7 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
         return self
 
     def transform(self, X, y=None) -> np.array:
-        """
-        Transform X by replacing each datetime column with
-        corresponding numerical features.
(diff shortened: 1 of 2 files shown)


jovan-stojanovic (approved):
Thanks for this PR @alexdt1982! Looks good to me (LGTM),

jovan-stojanovic:
However, just a small advice, as written https://github.com/dirty-cat/dirty_cat/issues/345#issue-1370087581 by @LilianBoulard, you should always create a separate branch for you PR's and not work from the master branch.

This may lead you into trouble when working in the future, for instance with multiple PR's, so it is always a good thing to leave the master branch clean and separate from the rest of the "experimental" work.

I admit that maybe the contributing guidelines were not clear enough on that, I will fix this.

jovan-stojanovic:
And if you notice something not clear/not mentionned in our [contributing guidelines](https://github.com/dirty-cat/dirty_cat/blob/master/CONTRIBUTING.rst), do not hesitate to mention it in an issue!

PR #378, opened 2022-10-06
[MRG] BUG: handle None in MinHashEncoder

### LINKED ISSUE #377: MinHashEncoder does not understand "None" as missing value
The following raises an error because the None is not captured as a NaN:
```python
import numpy as np

a = np.array([['a', 'b', None, 'c']], dtype=object).T

from dirty_cat import MinHashEncoder

enc = MinHashEncoder()
enc.fit_transform(a)
```

GaelVaroquaux:
Fixes #377

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -3,11 +3,20 @@
 Release 0.4.0
 =============
 
-Minor changes
+Major changes
 -------------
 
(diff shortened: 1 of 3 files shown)


jovan-stojanovic:
LGTM, merging!

PR #383, opened 2022-10-10
Better fuzzy join

GaelVaroquaux:
I wanted to improve the fuzzy_join example to first use fuzzy_join as it would be used by people using pandas outside of a predictive analysis, and then do the predictive analysis: hence separating X and y later.

I hit a problem that dtypes are not maintained by the merge. I created a simple failing test to illustrate the problem.


GaelVaroquaux:
The problem comes from the lines around https://github.com/dirty-cat/dirty_cat/blob/master/dirty_cat/_fuzzy_join.py#L235
We are using numpy to merge the tables. We should probably be using pd.merge rather, and try to piggy-back as much as possible on it.

GaelVaroquaux:
Hum, the tests passed. But they should have failed. I did something wrong

GaelVaroquaux:
Switching this from draft to final, because the tests are not running

jovan-stojanovic:
When I run them locally they pass as well. But I see the problem, will try to test them on some of my use cases.

GaelVaroquaux:
Oups, tests were not run (I think that @LilianBoulard renamed the test directory to "tests" a little while ago).

I just got them to run, but now...

jovan-stojanovic:
It seems it's not the problem, I see that also the test folder is named `tests` in [scikit-learn](https://github.com/scikit-learn/scikit-learn).

GaelVaroquaux:
Darn, you are right...

GaelVaroquaux:
I have pandas 1.3.5, our minimal requirement is 1.2. I don't understand.

Tests are absolutely failing for me.

GaelVaroquaux:
Due to the way our build matrix works, I cannot easily add the corresponding version to the CI to check that I can reproduce the problem on the CI (I added an issue: #384 )

jovan-stojanovic:
> I have pandas 1.3.5, our minimal requirement is 1.2. I don't understand.
> 
> Tests are absolutely failing for me.

That may be due to the fact that I added lines of code checking that the input columns contain strings and converting them to pandas categorical (after discussing their advantage with Julien): 
`
#### L203-209 Making sure that the column types are string and categorical:
    left_table_clean[left_col] = (
        left_table_clean[left_col].astype(str).astype("category")
    )
    right_table_clean[right_col] = (
        right_table_clean[right_col].astype(str).astype("category")
    )
`

GaelVaroquaux:
> That may be due to the fact that I added lines of code checking that the input columns contain strings and converting them to pandas categorical

That's great (though the output dtypes should match the input ones).

But still, I don't understand why tests are failing on my box but not on the CI

jovan-stojanovic:
Ok, I used `pd.merge` (which really simplified the code btw) so now it's easy to align most of the parameters with the ones from pandas.

I also tested locally with pandas 1.3.5. it seems that it's ok now, as I don't change the column types.

I am working on benchmarking on large datasets the effect of this change (as we saw that pandas may have problems with those).

jovan-stojanovic:
The fuzzy_join function with pd.merge takes less time to join tables. It's good!

dirty_cat/_fuzzy_join.py
```diff
@@ -230,49 +195,31 @@ def fuzzy_join(
...
-            ]
-        ),
-        axis=1,
-    )
+    left_table_clean["fj_idx"] = idx_closest
+    right_table_clean["fj_idx"] = right_table_clean.index
 
-    cols = list(left_table_clean.columns) + list(right_table_clean.columns)
-    df_joined = pd.DataFrame(joined, columns=cols).replace(r"^\s*$", np.nan, regex=True)
+    norm_distance = np.round(1 - (distance / 2), 6)
```
GaelVaroquaux, line 201: Why do we round?

dirty_cat/tests/test_fuzzy_join.py
```diff
@@ -130,41 +95,28 @@ def test_parameters_error(analyzer, how, suffixes, on):
...
+
+def test_missing_keys():
+    a = pd.DataFrame({"col1": ["aaa", "bbb"], "col2": [1, 2]})
+    b = pd.DataFrame({"col1": ["aaa_", "bbb_"], "col3": [1, 2]})
     with pytest.raises(
         ValueError,
-        match=(
-            "Parameter 'left_on', 'right_on' or 'on' has invalid type, expected string"
-        ),
+        match="Parameter 'left_on', 'right_on' or 'on' is missing",
```
GaelVaroquaux, line 108: This test will probably need to be updated from my suggestion.

GaelVaroquaux (commented):
A first pass on the code of @jovan-stojanovic 

Does the comment on the "unique" make sense?

dirty_cat/_fuzzy_join.py
jovan-stojanovic: You are right, not sure how this got in.

dirty_cat/_fuzzy_join.py
```diff
@@ -171,52 +158,30 @@ def fuzzy_join(
...
 
     # Make sure that the column types are string and categorical:
-    left_table_clean[left_col] = (
-        left_table_clean[left_col].astype(str).astype("category")
-    )
-    right_table_clean[right_col] = (
-        right_table_clean[right_col].astype(str).astype("category")
-    )
+    left_col_clean = left_table_clean[left_col].astype(str).astype("category")
+    right_col_clean = right_table_clean[right_col].astype(str).astype("category")
```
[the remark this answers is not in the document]
jovan-stojanovic: Just benchmarked this, you are right, unique is more efficient than categories, especially with duplicated entries! Thanks.

dirty_cat/_fuzzy_join.py
```diff
@@ -27,13 +27,12 @@ def fuzzy_join(
     left_on: Union[str, None] = None,
     right_on: Union[str, None] = None,
     on: Union[str, None] = None,
-    how: Literal["left", "right", "all"] = "all",
```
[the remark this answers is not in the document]
jovan-stojanovic: As I was aligning with pandas.merge, I found that [their `how` parameter](https://pandas.pydata.org/pandas-docs/stable/user_guide/merging.html#brief-primer-on-merge-methods-relational-algebra) has a different meaning (inner, outer, left or right join) so there can be confusion. 
Trying now to align with this, as I am not always sure what applies to the fuzzy matching compared to exact matching.

GaelVaroquaux:
> As I was aligning with pandas.merge, I found that their how parameter has a different meaning (inner, outer, left or right join) so there can be confusion. Trying now to align with this, as I am not always sure what applies to the fuzzy matching compared to exact matching.

I'm pretty sure that we can implement left and right, no? For the others maybe not.

Is my understanding wrong?

jovan-stojanovic:
Exactly, this is what I added to the code for now. For the others, the logic is a bit different so we might have to discuss it more.

GaelVaroquaux:
> For the others, the logic is a bit different so we might have to discuss it more.

Yey! That's great!

Is the PR ready for a new review?

jovan-stojanovic:
Yes, it is, thanks!

dirty_cat/_fuzzy_join.py
```diff
@@ -45,6 +46,8 @@ def fuzzy_join(
         A table to merge.
     right : pandas.DataFrame
         A table used to merge with.
+    how: typing.Literal["left", "right"], default=`left`
+        Type of merge to be performed.
```
GaelVaroquaux, line 50:
```suggestion
        Type of merge to be performed. Note that unlike pandas' merge, 
        only "left" and "right" are supported so far, as the fuzzy-join comes
        with its own mechanism to resolve lack of correspondence between
        left and right tables.
```

dirty_cat/_fuzzy_join.py
```diff
@@ -148,36 +151,53 @@ def fuzzy_join(
...
 
-    left_table_clean = left.reset_index(drop=True).copy()
-    right_table_clean = right.reset_index(drop=True).copy()
+    if how == "left":
+        left_table_clean = left.reset_index(drop=True).copy()
+        right_table_clean = right.reset_index(drop=True).copy()
+    else:
+        # We inverse the process so that the join is performed on the right column
+        left_table_clean = right.reset_index(drop=True).copy()
+        right_table_clean = left.reset_index(drop=True).copy()
```
GaelVaroquaux, line 172: Did we test that when we pass dataframes with an index, the function does approximately the same things as pd.merge?

GaelVaroquaux:
Let's do this in a separate PR.

GaelVaroquaux:
Merging!

PR #388, opened 2022-10-17
Add warning banner on development documentation

### LINKED ISSUE #361: Add a conditional CSS rule to display a warning on dev version of docs
When a user is accessing the dev version of the website, we should add a big read warning on the docs telling the user that she is accessing a development version and pointing to the stable docs. Scikit-learn has such a warning, see for instance the https://scikit-learn.org/dev/auto_examples/

The way to implement this is probably a combination of modifying our jinja templates (in doc/_templates, which are inherited from the alabster theme) and add a CSS conditional rule to display the warning only when the version is dev (https://www.w3.org/TR/css-conditional-3).

Some inspiration can be found in the way the version is injected in the sidebar (see https://github.com/dirty-cat/dirty_cat/blob/master/doc/_templates/about.html )

LilianBoulard:
Fixes #361

Adds a warning banner in the "about" section, which is the left panel on the website page.
It has the advantage of being visible wherever you are on the website, and as it's the only big red thing on the screen, it's pretty noticeable, so that should do the trick.
Adding it to page RST (for example in `class.rst`) was a hustle, so I ended up dropping this idea.

### DIFF
--- a/doc/_static/custom.css
+++ b/doc/_static/custom.css
@@ -48,6 +48,16 @@ div.body pre {
 
 }
 
+/* Banner for warning the user */
+.warn-banner {
+  background: pink;
+  border: 1px solid red;
+  padding-left: 0.75em;
+  padding-right: 0.75em;
+  text-align: center;
+  margin-bottom: 1em;
+}
+
 /* For title page */
 
 div.flex-content {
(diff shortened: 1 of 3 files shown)


jovan-stojanovic (approved):
Amazing! Looks great, thanks! Merging.

GaelVaroquaux:
Super  cool!! Thank you!!!

Gaël

On Oct 26, 2022, 11:20, at 11:20, Jovan Stojanovic ***@***.***> wrote:
***@***.*** approved this pull request.

PR #394, opened 2022-10-25
Fix generated doc push to website

### LINKED ISSUE #393: Website is not regenerating
It seems that we are no longer pushing the dev docs to the corresponding repo:
https://github.com/dirty-cat/dirty-cat.github.io
(last update 8 days ago)

Our build chain is somewhat broken

LilianBoulard:
Fixes #393.

As discussed over there, it seems some occurrences of `master` were not renamed and which causes the generated doc to not be pushed to the website repo.
I didn't because I wasn't sure what the environment variables were (e.g. `CIRCLE_BRANCH`). Should've checked :p
https://circleci.com/docs/built-in-environment-variables/

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,7 @@ jobs:
       - run: ls -ltrh doc/_build/html
       - deploy:
           command: |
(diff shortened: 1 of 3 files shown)


jovan-stojanovic:
Thanks for this. Ok, there is no error anymore at the beginning when building docs!

jovan-stojanovic (approved):

Issue #373, opened 2022-09-26
Add simple reproducible examples to features


jovan-stojanovic:
Our main encoders and other features are missing simple reproducible examples
that will appear on the website and give a first overview of what the feature can do.

### Examples to add

- [ ] `dirty_cat/_super_vectorizer.py`

- [ ] `dirty_cat._gap_encoder.GapEncoder`
- [ ] `dirty_cat/_minhash_encoder.py`
- [ ] `dirty_cat/_similarity_encoder.py`
- [ ] `dirty_cat/_target_encoder.py`

- [ ] `dirty_cat._datetime_encoder.DatetimeEncoder`

PR #399, opened 2022-10-27
DOC Ensures that `SuperVectorizer` passes numpydoc validation

LilianBoulard:
Part of #345.

### DIFF
--- a/dirty_cat/_super_vectorizer.py
+++ b/dirty_cat/_super_vectorizer.py
@@ -84,7 +84,8 @@ OptionalTransformer = Optional[
 
 
 class SuperVectorizer(ColumnTransformer):
-    """
+    """Easily transform a heterogeneous array to a numerical one.
+
     Easily transforms a heterogeneous data table
     (such as a :class:`pandas.DataFrame`) to a numerical array for machine
     learning. For this it transforms each column depending on its data type.
@@ -183,7 +184,7 @@ class SuperVectorizer(ColumnTransformer):
         Note that using this feature requires that the DataFrame columns
         input at :term:`fit` and :term:`transform` have identical order.
 
-    sparse_threshold: float, default=0.3
+    sparse_threshold : float, optional, default=0.3
         If the output of the different transformers contains sparse matrices,
         these will be stacked as a sparse matrix if the overall density is
         lower than this value. Use sparse_threshold=0 to always return dense.
@@ -202,7 +203,7 @@ class SuperVectorizer(ColumnTransformer):
 
     verbose : bool, default=False
         If True, the time elapsed while fitting each transformer will be
-        printed as it is completed
+        printed as it is completed.
 
     Attributes
     ----------
--- a/dirty_cat/_super_vectorizer.py
+++ b/dirty_cat/_super_vectorizer.py
@@ -4,7 +4,6 @@ transformers/encoders to different types of data, without the need to
 manually categorize them beforehand, or construct complex Pipelines.
 """
 
-
 from typing import Dict, List, Literal, Optional, Tuple, Union
 from warnings import warn
 
@@ -19,6 +18,9 @@ from sklearn.preprocessing import OneHotEncoder
 
 from dirty_cat import DatetimeEncoder, GapEncoder
 from dirty_cat._utils import Version
+
+# Required for ignoring lines too long in the docstrings
+# flake8: noqa: E501
 
 
 def _has_missing_values(df: Union[pd.DataFrame, pd.Series]) -> bool:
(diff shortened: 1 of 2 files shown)


LilianBoulard:
So my last commit actually removed the "See Also not found" from the checks, as it's not what I'd call required. Feel free to disagree :)
For "No example section found", I've added it to the ignore list for now, but as commented, the statement should be removed when #373 is resolved.

dirty_cat/_super_vectorizer.py
```diff
@@ -4,7 +4,6 @@
 manually categorize them beforehand, or construct complex Pipelines.
 """
 
-
 from typing import Dict, List, Literal, Optional, Tuple, Union
```
jovan-stojanovic, line 7: Ok, I see you removed here and everywhere else the `typing.Optional` (and maybe some other ones as well) and put directly the type name. If that's the case you should also remove the imports associated. To be consistent, if this is the way to go, we should remove `typing.Optional` in other indicators as well?
```suggestion
from typing import Dict, List, Literal, Tuple, Union
```

jovan-stojanovic (commented):
We should decide what to do with the typing, as it is kept in some cases (`get_feature_names_out`) and removed in other ones.

dirty_cat/_super_vectorizer.py
LilianBoulard: Yes I removed these occurrences in the docstring, but not in the code.
For example `typing.Optional` is still used [here](https://github.com/dirty-cat/dirty_cat/blob/d634f0e3845e31414e25260d764a67d456d2f231/dirty_cat/_super_vectorizer.py#L81). Also, if we had unused imports, the `Run pre-commit checks` test would fail.
The point of this PR is to make the [generated doc](https://output.circle-artifacts.com/output/job/7d81266c-bf80-49e6-90fc-68f6855743d1/artifacts/0/doc/generated/dirty_cat.SuperVectorizer.html#dirty_cat.SuperVectorizer) cleaner by formatting the docstring to the numpydoc guidelines.

dirty_cat/_super_vectorizer.py
```diff
@@ -180,31 +184,31 @@ class SuperVectorizer(ColumnTransformer):
...
         ``None`` means 1 unless in a :obj:`joblib.parallel_backend` context.
         ``-1`` means using all processors.
 
-    transformer_weights : dict, default=None
+    transformer_weights : dict, optional
         Multiplicative weights for features per transformer. The output of the
         transformer is multiplied by these weights. Keys are transformer names,
         values the weights.
 
     verbose : bool, default=False
```
jovan-stojanovic, line 204: Should you add an `optional` here as well?
```suggestion
    verbose : bool, optional, default=False
```

dirty_cat/_super_vectorizer.py
```diff
@@ -180,31 +184,31 @@ class SuperVectorizer(ColumnTransformer):
...
-    sparse_threshold: float, default=0.3
+    sparse_threshold : float, optional, default=0.3
         If the output of the different transformers contains sparse matrices,
         these will be stacked as a sparse matrix if the overall density is
         lower than this value. Use sparse_threshold=0 to always return dense.
         When the transformed output consists of all dense data, the stacked
         result will be dense, and this keyword will be ignored.
 
-    n_jobs : int, default=None
+    n_jobs : int, optional, default=None
```
jovan-stojanovic, line 194: Why did you leave here `default=None` here and removed it everywhere else?

dirty_cat/_super_vectorizer.py
```diff
@@ -104,19 +108,19 @@ class SuperVectorizer(ColumnTransformer):
         Note: currently, missing values are counted as a single unique value
```
jovan-stojanovic, line 108: Should you add an `optional` for the `cardinality_threshold` as well?

jovan-stojanovic (commented):
Ok, got it. Here are my comments then:

dirty_cat/_super_vectorizer.py
LilianBoulard: Good point! There's actually a specific reason I left it, and it's because it's explicitly mentioned in the `n_jobs` doc (just below) that `None means 1 unless in a joblib.parallel_backend context.`, so essentially it's a way to indicate to the user "by default, it's 1".

dirty_cat/_super_vectorizer.py
LilianBoulard: Indeed!

dirty_cat/_super_vectorizer.py
LilianBoulard: Looking back at the numpy style guide, I noticed this is not the expected way of doing it, so I modified it :)

dirty_cat/_super_vectorizer.py
LilianBoulard: Yes, done!

dirty_cat/_super_vectorizer.py
LilianBoulard: Looking back at the numpy style guide, I noticed this is not the expected way of doing it, so I modified it :)

LilianBoulard:
Thanks for the review, merging!

Issue #326, opened 2022-09-05
Add test for `CHANGES.rst`


LilianBoulard:
I notice that on most PRs (including mine), we often forget to modify `CHANGES.rst`, however this is important.
I think we should add a small test to check that the file has been modified in the PR (compared to master).

GaelVaroquaux:
Scikit-learn has some logic for that that we should import

LilianBoulard:
Might also be interesting to implement some of [what has been done in GAMA](https://github.com/openml-labs/gama/blob/master/.github/workflows/changelog.yaml)

LeoGrin:
I think this can be closed.

PR #405, opened 2022-11-06
add a github workflow to check if CHANGES.rst has been changed

LeoGrin:
Solves #326


LeoGrin:
I think there is a permission issue which prevent the github workflow from commenting on a PR. Could someone with access to the settings try to create a Personal Access Token and see if it solves the issue? (https://stackoverflow.com/questions/70435286/resource-not-accessible-by-integration-on-github-post-repos-owner-repo-ac)

LilianBoulard:
Hi, as discussed, using a message might be a bit overwhelming, but using comprehensive check description might be the way to go :)
There's probably some inspiration to take from sklearn: https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/check-changelog.yml

LeoGrin:
It seems to be working! I kept scikit-learn specificities like:
- Only requires changelog to be updated if the PR modifies a test file. This makes it simpler to update docs without being bothered by this check.
- This also checks that the pr is referenced in the changelog, which we doesn't do right now. I think that's good but tell me if you think it's too much.

.github/workflows/changelog.yml
```diff
@@ -1,31 +1,68 @@
...
+            echo "A Changelog entry is missing."
+            echo ""
+            echo "Please add an entry to the changelog at 'CHANGES.rst'"
+            echo "to document your change assuming that the PR will be merged"
+            echo "in time for the next release of dirty-cat."
+            echo ""
+            echo "Look at other entries in that file for inspiration and please"
+            echo "reference this pull request using the ':pr:' directive and"
+            echo "credit yourself (and other contributors if applicable) with"
+            echo "the ':user:' directive."
```
LilianBoulard: Part of this comment is not applicable to the way we format our changelog
```suggestion
            echo "Look at other entries in that file for inspiration."
```

.github/workflows/changelog.yml
```diff
@@ -1,31 +1,68 @@
...
+            echo ""
+            echo "If you see this error and there is already a changelog entry,"
+            echo "check that the PR number is correct."
```
LilianBoulard, line 62: So, that's interesting, it seems there is a check that the :pr: directive mentioned above is in the changelog.
It seems to be on line 32
```bash
if [[ "$all_changelogs" =~ :pr:\`$PR_NUMBER\` ]]
```
but I'm not sure we're willing to keep that functionality.

.github/workflows/changelog.yml
```diff
@@ -1,31 +1,68 @@
...
+            echo ""
+            echo "If you believe that this PR does not warrant a changelog"
+            echo "entry, say so in a comment so that a maintainer will label"
+            echo "the PR with 'No Changelog Needed' to bypass this check."
```
LilianBoulard, line 66: And this one seems to indicate that there's a github label dedicated! I guess we can create it, we just have to check if there's logic linked to it.

Edit: logic there is! Line 13
```bash
if: ${{ contains(github.event.pull_request.labels.*.name, 'No Changelog Needed') == 0 }}
```

LilianBoulard (commented):
Some comments and logic inherited from sklearn, up to debate on whether we should keep it :)

.github/workflows/changelog.yml
LeoGrin: Yes the current PR suppose we start adding pr numbers to the changelog

.github/workflows/changelog.yml
LeoGrin: I think that's good but it may be a bit much at this stage.

.github/workflows/changelog.yml
LilianBoulard: Yes, my bad, I didn't see your previous message at the time of reviewing :)

LilianBoulard:
So to follow-up on the previous messages, I think the label logic is cool and fits our workflow, but I have quite mixed feelings about adding the PRs references to the current changelog. 
I don't know, it's up for debate, but if we decide to use the references, I think we should modify the `CHANGES.rst` retroactively.

LeoGrin:
> So to follow-up on the previous messages, I think the label logic is cool and fits our workflow, but I have quite mixed feelings about adding the PRs references to the current changelog. I don't know, it's up for debate, but if we decide to use the references, I think we should modify the `CHANGES.rst` retroactively.

@GaelVaroquaux do you have an opinion on this?

GaelVaroquaux:
Long term, light PR references are a good thing (this is how scikit-learn does things). It helps power users track the changes, which is useful to make power users happy and hopefully help convert them to contributors.

GaelVaroquaux:
By the way, I think that the check on whether tests are modified is interesting. It's may be considered a bit too restrictive (one could maybe check for modifications in the source code but not the docs, though it's debatable).

I think that something like this could be adapted / ported to scikit-learn. Pinging some scikit-learn folks who might be interested: @jjerphan @thomasjpfan @cmarmo

LeoGrin:
> I think that something like this could be adapted / ported to scikit-learn. Pinging some scikit-learn folks who might be interested: @jjerphan @thomasjpfan @cmarmo
I actually stole this from scikit-learn workflows https://github.com/scikit-learn/scikit-learn/blob/main/.github/workflows/check-changelog.yml

LeoGrin:
@LilianBoulard I've added retroactively the pr numbers and users to the Changelog up to version 0.1.0. I think the test errors are unrelated to the PR, so for me it is ready to merge!

LilianBoulard:
Thanks for adding that in!
Sorry to bother again, but there's still a bit of work on the render: [PRs and users are not resolved to GH](https://output.circle-artifacts.com/output/job/0e71a46b-a724-4a71-95de-fe1e379d79f7/artifacts/0/doc/CHANGES.html). 
[It works fine for sklearn](https://scikit-learn.org/stable/whats_new/v1.2.html#version-1-2-0).

LeoGrin:
Fixed 🥳 https://output.circle-artifacts.com/output/job/c06e6faa-6e02-4db9-8a0f-a24fea3dcf11/artifacts/0/doc/CHANGES.html

LilianBoulard (approved):

PR #410, opened 2022-11-14
Use `sklearn.utils.parse_version` instead of `Version`

### LINKED ISSUE #275: `Version` might not work with special versions
I worked earlier this week on version comparison, and the `Version` class was implemented in the utils. See #261.

Today I realized that the way it is currently implemented, it will probably raise errors when checking for special versions, such as pre-releases, release candidates, dev versions, and others.
This is because it separates the version based on the dot character, and uses the two first (or only the first) values as the major and minor versions, which it converts to integers.

I tested a few scikit-learn versions (taken from [PyPI](https://pypi.org/project/scikit-learn/#history)), and it seems to work :

To reproduce:
```commandline
$ pip install --upgrade scikit-learn==0.24.dev0
$ python -c "import sklearn; print(sklearn.__version__);"
0.24.dev0

$ pip install --upgrade scikit-learn==1.0rc2
$ python -c "import sklearn; print(sklearn.__version__);"
1.0.rc2

$ pip install --upgrade scikit-learn==1.1.0rc1
$ python -c "import sklearn; print(sklearn.__version__);"
1.1.0rc1
```

If the version returned were to be e.g. `1.0rc1`, the method would raise a `ValueError` because it would try to convert `0rc1` (the minor) to an integer.

Currently, we only use this class to compare versions of scikit-learn, so it shouldn't be a problem for now.
This is however something to look out for in the future !

LilianBoulard:
Replaces #261.
Fixes #275.

### DIFF
--- a/dirty_cat/_datetime_encoder.py
+++ b/dirty_cat/_datetime_encoder.py
@@ -5,10 +5,9 @@ import numpy as np
 import pandas as pd
 from sklearn import __version__ as sklearn_version
 from sklearn.base import BaseEstimator, TransformerMixin
-from sklearn.utils import parse_version
 from sklearn.utils.validation import check_is_fitted
 
-from dirty_cat._utils import check_input
+from dirty_cat._utils import check_input, parse_version
 
 # Required for ignoring lines too long in the docstrings
 # flake8: noqa: E501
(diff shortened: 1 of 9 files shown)


LilianBoulard:
Unlike what has been mentioned in #355, it seems to me that `parse_version` was introduced as part of sklearn 1.0 (PR https://github.com/scikit-learn/scikit-learn/pull/19826), and therefore isn't accessible, which is why the minimal requirements (with sklearn 0.23) tests fail.
Looking at the file in version 0.23, it seems we could try to use `_parse_version`, even if it seems way less robust than the current version (though it has pretty much the same functionality as the `Version` class I previously implemented).
What do you think @glemaitre?

jovan-stojanovic (approved):

PR #413, opened 2022-11-15
Fix random fetching error

### LINKED ISSUE #406: Fetching tests raise errors at random
As can be seen in some CI runs recently ([such as in](https://github.com/dirty-cat/dirty_cat/actions/runs/3387729832/jobs/5628850397) #365, [or on main](https://github.com/dirty-cat/dirty_cat/actions/runs/3386153511/jobs/5625193761)), the test `test_fetch_world_bank_indicator` randomly raises errors, invalidating otherwise correct code.
We should investigate what's going on.

LilianBoulard:
Attempts to fix #406.

### DIFF
--- a/dirty_cat/datasets/_fetching.py
+++ b/dirty_cat/datasets/_fetching.py
@@ -7,6 +7,7 @@ Scikit-Learn's ``fetch_openml()`` function.
 # - Watch out for ``fetch_openml()`` API modifications:
 # as of january 2021, the function is marked as experimental.
 
+from __future__ import annotations
 
 import gzip
 import json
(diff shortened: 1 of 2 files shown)


LilianBoulard:
Currently, this PR adds better error messages, and some minor code improvements.
I haven't been able to reproduce the random bug, despite running the CI 3 times (so 60 tests passed successfully).
I guess we could merge as it is, so that it can help us better understand the error if it happens in the future.

dirty_cat/datasets/_fetching.py
```diff
@@ -217,17 +219,16 @@ def _fetch_world_bank_data(
               saved as a CSV file.
 
     """
-    # See if file available locally :
     path = f"{data_directory}/{indicator_id}.csv"
-    csv_path = Path(path)
+    csv_path = Path(path).resolve()
```
jovan-stojanovic, line 223: Here we should identify if the file is present locally.

jovan-stojanovic (commented):
Ok, I'm fine with this. I still think that the problem is not the fetching function in itself but errors that are provoked by the worldbank data server.

This will happen very rarely as we read the csv file if it saved locally (which should be after the first time a user fetches the data). Can you add a test that verifies that this is done correctly?

dirty_cat/datasets/_fetching.py
LilianBoulard: `resolve` only converts relative paths to absolute paths, without the file necessarily existing.
We check if the file actually exists line 225.

jovan-stojanovic: Sorry, the question got lost in the comment below.

jovan-stojanovic (approved):
LGTM Merging, thanks, this is really good!

PR #416, opened 2022-11-16
Fix codecov

LilianBoulard:
Fixes codecov not ignoring some things incorrectly.
Followup to #362.


LilianBoulard:
It looks like it hasn't changed anything... I guess codecov already ignores tests in their report. Anyway since it's not changing much (and the information is correct), I'll be merging right away.

PR #420, opened 2022-12-07
Added HashingVectorizer to fuzzy_join.

### LINKED ISSUE #418: fuzzy_join is slow with big number of categories
When I tried to join my table with the [embeddings table](https://soda-inria.github.io/ken_embeddings/) that contain 2.5 million entities, I realised that the [fuzzy_join](https://dirty-cat.github.io/dev/generated/dirty_cat.fuzzy_join.html#dirty_cat.fuzzy_join) is getting increasingly slow with a big number of categories to join.

When I inspected more closely, the most expensive operation is encoding the categories with the [Count Vectorizer](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html). To be more precise, it's the [_count_vocab](https://github.com/scikit-learn/scikit-learn/blob/f3f51f9b6/sklearn/feature_extraction/text.py#L1192) that takes too long.

An alternative may be to use the [HashingVectorizer](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.HashingVectorizer.html) when the number of categories is bigger than a certain threshold. Similar to what is done in the [SuperVectorizer](https://dirty-cat.github.io/stable/generated/dirty_cat.SuperVectorizer.html#dirty_cat.SuperVectorizer) with the encoders.

rcap107:
Fix #418


jovan-stojanovic:
Thanks! Congratulations on your first open source project pull request :)

The threshold for when the HashingVectorizer is currently 1_000_000, which seems reasonable from experience.

Please add a test for the case when number of categories > 1 000 000 so we see that everything works fine in this case.

LeoGrin:
I think it would be worthwile to do a small benchmark to check that we're not impairing accuracy, and that 1e6 is the right cut-off. I can help if needed.

rcap107:
> Thanks! Congratulations on your first open source project pull request :)
> 
> The threshold for when the HashingVectorizer is currently 1_000_000, which seems reasonable from experience.
> 
> Please add a test for the case when number of categories > 1 000 000 so we see that everything works fine in this case.

I am working on a test for it, however I am not sure how to check that the correct encoder has been chosen, without exposing any variable other than the joined dataframe.

jovan-stojanovic:
Please do not forget also to add a small description of what was changed to [CHANGES.rst](https://github.com/dirty-cat/dirty_cat/blob/main/CHANGES.rst).

LilianBoulard:
To follow up on how to test that the HashingVectorizer was used instead of the CountVectorizer:
```py
def test_using_optimizations():
    cv_path = "sklearn.feature_extraction.text.CountVectorizer.__init__"
    hv_path = "sklearn.feature_extraction.text.HashingVectorizer.__init__"
    with (mock.patch(cv_path), mock.patch(hv_path)) as (cv, hv):
        # For a low number of categories,
        # we're expecting to be using a CountVectorizer
        fuzzy_join(...)
        cv.assert_not_called()
        hv.assert_called_once()
        
        cv.reset_mock()
        hv.reset_mock()
        
        # For a high number of categories, we should be expecting that the
        # fuzzy join uses a HashingVectorizer instead of the CountVectorizer
        fuzzy_join(...)
        cv.assert_called_once()
        hv.assert_not_called()
```

rcap107:
Hi,

I'm working on the count/hashing vectorizer part we were discussing yesterday, and I am not sure how to check for the correctness of the type of a generic vectorizer, since the `_VectorizerMixin` class is supposed to be private... I have the same problem with typechecking in the method signature and the docstring. 

How should I test the feature? We were talking about sending a custom `vectorizer` object that would raise a known exception if called, but I am not sure how to go about implementing it.

dirty_cat/_fuzzy_join.py
```diff
@@ -31,6 +36,7 @@ def fuzzy_join(
     analyzer: Literal["word", "char", "char_wb"] = "char_wb",
     ngram_range: Tuple[int, int] = (2, 4),
     return_score: bool = False,
+    encoder: Literal["hashing", "count"] = "hashing",
```
LilianBoulard, line 39: The type hint would probably be 
```suggestion
    encoder: Union[Literal["hashing", "count"], _VectorizerMixin] = "hashing",
```
That's true that it forces us to import from the private sklearn API. However, it's not a problem in this case: the public API of a library is *what can be imported by the user in a production environment*. The public API ensures the changes to be documented and to some extent, backward-compatible. Everything private is internal API to the lib, thus is not assured to be backward-compatible, nor useful for the standard user. We are developing a library, so it's a different use-case of sklearn.
I hope my explaination makes sense :sweat_smile:

dirty_cat/_fuzzy_join.py
```diff
@@ -61,8 +67,8 @@ def fuzzy_join(
         Must be found in both DataFrames. Use only if `left_on`
         and `right_on` parameters are not specified.
     analyzer : typing.Literal["word", "char", "char_wb"], default=`char_wb`
-        Analyzer parameter for the CountVectorizer used for the string
-        similarities.
+        Analyzer parameter for the CountVectorizer/HashingVectorizer used for
+        the string similarities.
```
LilianBoulard, line 71:
```suggestion
        Analyzer parameter for the CountVectorizer/HashingVectorizer
        (passed by `encoder`) used for the string similarities.
```

dirty_cat/_fuzzy_join.py
```diff
@@ -74,6 +80,7 @@ def fuzzy_join(
     return_score : boolean, default=True
         Whether to return matching score based on the distance between
         nearest matched categories.
+    encoder :
```
LilianBoulard, line 83: Missing a doc?

dirty_cat/_fuzzy_join.py
```diff
@@ -164,6 +171,13 @@ def fuzzy_join(
             f"analyzer should be either 'char', 'word' or 'char_wb', got {analyzer!r}",
         )
 
+    if encoder not in ["hashing", "count"]:
+        # TODO: implement check for generic vectorizer
+        raise ValueError(
+            "encoder should be either 'hashing', 'count', or a vectorizer object, got"
+            f" {encoder!r}"
+        )
```
LilianBoulard, line 179: Should add an if statement with something like `if issubclass(encoder.__class__, _VectorizerMixin)`

LilianBoulard (commented):
Here's a little review in regards to the previous comment.

To implement the custom vectorizer, you will want to implement your own by subclassing either the count of hash vectorizer, and overwrite for example the `fit` method:

```py

class TestVectorizer(HashingVectorizer):
    """
    Implements a custom vectorizer to check if the `encoder`
    parameter uses the passed vectorizer as expected.
    Should raise an assertion error when `fit` is called.
    """

    def fit(self, X):
        raise AssertionError("Custom vectorizer used as expected!")
```
Hope that helps you out!

dirty_cat/_fuzzy_join.py
rcap107: Shouldn't it be "passed *to* encoder" instead?

dirty_cat/_fuzzy_join.py
rcap107: Added

dirty_cat/_fuzzy_join.py
```diff
@@ -1,8 +1,9 @@
 """
 Fuzzy joining tables using string columns.
 The principle is as follows:
-  1. We embed and transform the key string columns using CountVectorizer
-  and TfifdTransformer.
+  1. We embed and transform the key string columns using CountVectorizer/
+  HashingVectorizer and TfifdTransformer. HashingVectorizer is used if more than
+  1_000_000 categories are found.
```
LilianBoulard, line 6: I think this is outdated now?

dirty_cat/tests/test_fuzzy_join.py
```diff
@@ -201,3 +202,52 @@ def test_fuzzy_join_pandas_comparison():
...
+        """Implements a custom vectorizer to check if the `encoder`
+        parameter uses the passed instance as expected.
+        Raises an error when `fit` is called.
+
+        Args:
+            HashingVectorizer (_type_): _description_
+        """
+
+        def __init__(self):
+            HashingVectorizer.__init__(self)
```
LilianBoulard, line 220: Since we don't do anything in this `__init__`, we can leave the implementation to the parent class (so remove the method).
```suggestion
```
Also, by convention we prefer
`super().__init__(*args, **kwargs)` to `<ClassName>.__init__(self, *args, **kwargs)` (notice we also remove the `self`) :)

dirty_cat/tests/test_fuzzy_join.py
```diff
@@ -201,3 +202,52 @@ def test_fuzzy_join_pandas_comparison():
...
+def test_correct_encoder():
+    """Test that the encoder error checking is working as intended."""
+
+    class TestVectorizer(HashingVectorizer):
+        """Implements a custom vectorizer to check if the `encoder`
+        parameter uses the passed instance as expected.
+        Raises an error when `fit` is called.
+
+        Args:
+            HashingVectorizer (_type_): _description_
```
LilianBoulard, line 216: Is there a reason for adding this?

LilianBoulard (commented):
Thanks, that's better!
I added a few comments.
Also, while we test that passing a custom vectorizer works, we don't test that passing "count" or "hashing" does anything. 
The best way of doing that would be to expect an significant difference, but I admit I'm not sure how to best test that.
The suboptimal way would be to monkey patch the encoder as discussed previously (like, if "count" is passed, we are expecting the `CountVectorizer` to be called, and not the `HashingVectorizer`, and vice-versa).

jovan-stojanovic (approved):
Thanks! Congrats on your first contribution :)

Issue #422, opened 2022-12-12
get_feature_names_out as an array make test fail


jovan-stojanovic:
It seems that one of the [test of the SuperVectorizer](https://github.com/dirty-cat/dirty_cat/actions/runs/3669076049/jobs/6204016053) fails after the new release of scikit-learn.

This is because the [get_feature_names_out](https://scikit-learn.org/stable/glossary.html#term-get_feature_names_out) is an array of strings and this seems not to be taken into account by the test.

LeoGrin:
(The fact that an array was returned instead of list was because an error condition was triggered and the supervectorizer was returning some default column names)

PR #423, opened 2022-12-12
fix bug on get_features_out from SuperVectorizer with new sklearn versions

LeoGrin:
The bug appeared because on newer versions of sklearn, ColumnTransformer has no attribute get_features_names. With the existing code, this meant that we used the non-transformed columns names. I've changed the logic to just distinguish if we call get_features_names or get_feature_name_out based on sklearn version.


LeoGrin:
Fix #422

jovan-stojanovic (approved):
Thanks!

LilianBoulard (approved):

PR #425, opened 2022-12-13
Implement benchmark mini-framework

### LINKED ISSUE #404: Better benchmarks
I've been thinking that we we should clean up our benchmark folder.

It might be interesting to create a mini-framework to normalize inputs and/or outputs of the benchmarks, as well as make the results public by pushing them in a `results/` subfolder.

LilianBoulard:
Fixes #404.
Work done as part of #267, moved here for better traceability.


LeoGrin:
Super cool thanks! I think it would be worth it to include a README on how to launch benchmarks

LeoGrin:
It would also be cool to have requirements for benchmarks. For instance you need a recent enough version of tracemalloc.

benchmarks/README.md
```diff
@@ -0,0 +1,59 @@
...
+If you're trying to benchmark an experimental feature,
+and as benchmark files are "stand-alone" (they use `import dirty_cat`),
+you'll have to make your copy of dirty_cat available to your Python executable.
+
+To do that, the recommended way is to create a development environment as
+explained in the project's contributing guidelines, and, if you're on linux,
+add the path to your dirty_cat local repository in the Python path with the command
+
+```bash
+export PYTHONPATH="$PYTHONPATH:</path/to/repository/>"
```
LeoGrin, line 27: Do you need to do this if you're doing `pip install -e .` ?

setup.cfg
```diff
@@ -47,6 +47,12 @@ doc =
     matplotlib
     seaborn
     statsmodels
+benchmarks =
```
LeoGrin, line 50: Right now you need python>=3.9 to run the benchmark (otherwise I get an error with tracemalloc.reset_...)

benchmarks/README.md
```diff
@@ -0,0 +1,59 @@
...
+```bash
+pip install -e .[benchmarks]
+```
+
+### Analyzing results
+
+The results of the benchmarks ran by maintainers are pushed in the `results/` folder.
+
+As mentioned earlier, benchmarks implement a ``--plot`` parameter used
+to display the results visually.
```
LeoGrin, line 59:
```suggestion
to display the results visually. Using ``--plot`` without ``--run`` allows you to plot the results without re-running the benchmark.
```

benchmarks/README.md
LilianBoulard: No, you're right!

LeoGrin (approved):

setup.cfg
LilianBoulard: I've added a note on the readme, but I don't know if it's possible to specific as a config that Python 3.9+ is required.

PR #424, opened 2022-12-13
Link source in doc to Github

LeoGrin:

PR #426, opened 2022-12-13
Fix sphinx issues

LeoGrin:
Fixed most of Sphinx errors/warnings


LeoGrin:
To be merged after #405 and #424

jovan-stojanovic (commented):
Thanks, this is great. I am only not sure if we should delete the README of the examples.. What was the error?

LeoGrin:
I was getting a `CRITICAL: Unexpected section title.` refering to the content of the README in `auto-examples/index.rst`. I'm not exactly sure of what was happening, but I don't think what was in the README was useful, right?

GaelVaroquaux:
> Thanks, this is great. I am only not sure if we should delete the README of the examples..

That's probably a bad idea.

LeoGrin:
The thing is that anything we put in this README is included in the *Usage* section on the main page (or Sphinx tries to include it and returns an error, which was what was happening). That's why I made the README empty (it's not deleted). There may be a better solution, but I'm not sure why making the README empty is problem.

GaelVaroquaux:
> There may be a better solution, but I'm not sure why making the README empty is a problem.

Empty readme is not a problem. No readme will probably break the build.

Sorry, catching up randomly with a heap of messages...

dirty_cat/_datetime_encoder.py
```diff
@@ -198,7 +198,7 @@ def transform(self, X, y=None) -> np.array:
 
         Returns
         -------
-        np.array, shape (n_samples, n_features_out_)
+        np.array, shape (n_samples, n_features_out\_)
```
LilianBoulard, line 201:
```suggestion
        np.array, shape (n_samples, n_features_out_)
```
Might be a mistake, this causes a warning in the current build:
```
 dirty_cat/_datetime_encoder.py:190
  /home/runner/work/dirty_cat/dirty_cat/dirty_cat/_datetime_encoder.py:190: DeprecationWarning: invalid escape sequence \_
```

examples/README.txt
```diff
@@ -1,3 +0,0 @@
-
-Using dirty_cat
-===============
```
LilianBoulard, line 3: This is used as the title for the example page:
https://output.circle-artifacts.com/output/job/8b13cf5a-1866-482f-b0d6-728a852d5100/artifacts/0/doc/auto_examples/index.html

LeoGrin: You're right, I've put it back and found a way for the README not to appear on the main page (changing the `start_line` in ìnclude`)

dirty_cat/_datetime_encoder.py
```diff
@@ -218,7 +218,7 @@ def transform(self, X, y=None) -> np.ndarray:
 
         Returns
         -------
-        np.ndarray, shape (n_samples, n_features_out_)
+        np.ndarray, shape (n_samples, n_features_out\\_)
```
GaelVaroquaux, line 221: Oh no, that's horrible. I fixes the warning but makes the docstring unreadable in plain text.

Can we try something like `n_features_out_` rather? (I don't know if it fixes the warning)

GaelVaroquaux (commented):
That's useful, thank you.

I do have a comment on one aspect.

dirty_cat/_datetime_encoder.py
GaelVaroquaux, line 221:
```suggestion
        np.ndarray, shape (n_samples, `n_features_out_`)
```
Like this?

LeoGrin:
No Sphinx warning 🥳

A few questions: 
- Do we want to delete `development.rst`? I don't think it's used anywhere. Right now I've just added an orphan tag.
- The `Usage` button was broken, I fixed it by linking to the Usage section on the front page. Tell me if you think it is better to link to the [example page](https://output.circle-artifacts.com/output/job/2493eca4-f58c-41ac-ba7a-5e4050650270/artifacts/0/doc/auto_examples/index.html).

LilianBoulard:
Great!

Last few things before merging:
- I looks like `development.rst` is empty and not referenced anywhere, so I guess it could be removed yeah
- There's one warning left:
  - `UserWarning: Usage of dash-separated 'python-requires' will not be supported in future versions. Please use the underscore name 'python_requires' instead`, which can be fixed in `setup.cfg`
- There's some conflicts with main

Thanks!

LilianBoulard (approved):

jovan-stojanovic (approved):

Issue #371, opened 2022-09-22
Improve coverage of main features


jovan-stojanovic:
Here are some of the code coverage results:
``` 
------------------------------------------------------------------------
dirty_cat/_datetime_encoder.py             88%
dirty_cat/_gap_encoder.py                  88%
dirty_cat/_minhash_encoder.py              94%
dirty_cat/_similarity_encoder.py           90%
dirty_cat/_string_distances.py             91%
dirty_cat/_super_vectorizer.py             85%
dirty_cat/_target_encoder.py               78%
dirty_cat/_utils.py                        93%
dirty_cat/datasets/_fetching.py            97%
...
------------------------------------------------------------------------
TOTAL                                     89%
```
It may be good to work on improving those so that we have clear standards on what coverage is needed for new features.

### Tests to Update

- [ ] `dirty_cat._datetime_encoder.DatetimeEncoder`
- [ ] `dirty_cat._gap_encoder.GapEncoder`
- [ ] `dirty_cat/_minhash_encoder.py` #379 
- [ ] `dirty_cat/_similarity_encoder.py`
- [ ] `dirty_cat/_string_distances.py`
- [ ] `dirty_cat/_super_vectorizer.py`
- [ ] `dirty_cat/_target_encoder.py` #375 
- [ ] `dirty_cat/_utils.py`
- [ ] `dirty_cat/datasets/_fetching.py`

jeromedockes:
tablevectorizer & datetimeencoder are at 100% after #902 so we checked those in the list. @TheooJ will open a PR to remove the unused function in _utils.py

PR #429, opened 2022-12-15
 Improve fetching test

LilianBoulard:
Improves the fetching test a little bit by using temporary directories.


jovan-stojanovic:
Related to #371.

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -32,6 +32,23 @@
 from dirty_cat.datasets._utils import get_data_dir as _get_data_dir
 
 
+class TemporaryDirectory(_TemporaryDirectory):
```
GaelVaroquaux, line 35: Do we really need this? Can we not add the "Path" later in the enclose code? Or use the required functionality outside of Path?

GaelVaroquaux (commented):
I think that I would prefer a solution that does not subclass a standard library object, if it is not too hard to do.

dirty_cat/datasets/tests/test_fetching.py
LilianBoulard: It looks to me that if we don't want to subclass like it's currently done, we either have to add support for `str` paths in the fetching code, or add boilerplate where it's called in the tests like that:

From

```py
with TemporaryDirectory() as temp_dir:
    ...
```

to

```py
with TemporaryDirectory() as temp_dir:
    temp_dir = Path(temp_dir).resolve()
    ...
```

which IMO is cumbersome and a bit error-prone.

Note: I though of using a function such as
```py
def get_temp_dir() -> Path:
    with TemporaryDirectory() as temp_dir:
        return Path(temp_dir).resolve()
```

but first, we lose the `with` ability with the function (can't do `with get_temp_dir() as temp_dir`), and most importantly, as soon as the path is returned, the `with` in the func exists, thus the directory is removed and can't be used afterwards.

LilianBoulard: Maybe the best solution would be to refactor the fetching code, as a parallel issue solved by #432 would allow users to pass a directory at fetching time, and while using pathlib is good practice, I don't think it's common practice for our userbase. Most people probably work with `os.path` and thus use strings.

GaelVaroquaux:
> add boilerplate where it's called in the tests like that:

>     temp_dir = Path(temp_dir).resolve()

I'd rather have this boilerplate code as it is more explicit

LilianBoulard:
Superseded by #453.

PR #433, opened 2022-12-15
Replace np.array with np.ndarray in type hints

LilianBoulard:
Because ``np.array`` is a function, and ``np.ndarray`` is the class.


LeoGrin (approved):

jovan-stojanovic (approved):
Is this finished @LilianBoulard ?

LilianBoulard:
> Is this finished @LilianBoulard ?

Yes!

PR #435, opened 2022-12-16
Fix em dash in website titles

LilianBoulard:
Currently, page titles are rendered with the code ``&mdash`` in the title, which is very ugly.
The dash is already rendered by sphinx or whatnot, so we should remove it.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -20,6 +20,7 @@ import os
 import shutil
 from datetime import datetime
 import sys
+
 # If extensions (or modules to document with autodoc) are in another
 # directory, add these directories to sys.path here. If the directory
 # is relative to the documentation root, use os.path.abspath to make it
@@ -174,7 +175,7 @@ html_static_path = ["_static"]
 
 
 # Modify the title to get good social-media links
-html_title = "&mdash; Dirty cat"
+html_title = "— Dirty cat"
 
 # -- Options for HTMLHelp output ------------------------------------------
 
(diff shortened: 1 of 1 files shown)


LeoGrin (approved):
Indeed the current google result preview is a bit ugly because of this. Thanks !

doc/conf.py
```diff
@@ -174,7 +175,7 @@
 
 
 # Modify the title to get good social-media links
-html_title = "&mdash; Dirty cat"
+html_title = "dirty_cat"
```
GaelVaroquaux, line 178:
```suggestion
html_title = "— dirty_cat"
```

GaelVaroquaux (approved):
LGTM (but I am adding back another a dash to mark the separation)

LilianBoulard:
Thanks for the reviews. Regarding Gaël's addition, would should make sure it renders correctly before merging, but right now the worker failed.
I feel like merging main should solve the issue, but I don't have my dev env right now (I'm on vacation).

doc/conf.py
```diff
@@ -174,7 +175,7 @@
 
 
 # Modify the title to get good social-media links
-html_title = "&mdash; Dirty cat"
+html_title = "— dirty_cat"
```
GaelVaroquaux, line 178:
```suggestion
html_title = "— dirty_cat"

```
Making a trivial change to trigger a merge + CI

GaelVaroquaux:
It seems that the tests are now running, thanks to @LeoGrin 's fixes on main.

Merging, thank you!

LilianBoulard:
I'd like to reopen this as it looks like the title now has a double long dash
image
I'll see that on monday

LeoGrin:
https://socialsharepreview.com/?url=https://dirty-cat.github.io/stable/

PR #446, opened 2022-12-23
Benchmark fuzzy join hash vectorizer

LeoGrin:
• Add the possibility for the benchmark function to return a dictionnary, which is added to the results by the `monitor` decorator
• Benchmark different encoders for fuzzy_join (issue #418, related to #420) . It seems that using the `HashingVectorizer` instead of `CountVectorizer` is always better (no cost for f1 score, and almost always faster, see plot). If the user want to tradeoff performance for speed, it seems better to play with the `ngram_range` than changing the encoder. Therefore I recommend to just use the `HashingVectorizer` for `fuzzy_join`, instead of using it only for big datasets like in #420.

Looking forward to hearing what other people think!
image


benchmarks/utils/monitor.py
```diff
@@ -34,13 +34,23 @@ def parse_func_repr(representation: str) -> Tuple[str, tuple, dict]:
...
-        args.append(arg)
+    arg_list = args_repr.split(", ")
+    # Extract the keyword arguments from the positional arguments
+    for index in range(len(arg_list)):
+        # Check that the "," is not inside a parenthesis
+        # (this would mean that it is part of a string)
+        # If it is, then we merge the current string with the next one
+        # Until we find a string that is not inside a parenthesis
+        if arg_list[index].count("(") != arg_list[index].count(")"):
+            arg_list[index + 1] = arg_list[index] + ", " + arg_list[index + 1]
```
LilianBoulard, line 45: I'm having a hard time understanding the logic of this section, could you please provide an example that didn't work with the old implementation and works with this one?

benchmarks/utils/monitor.py
```diff
@@ -34,13 +34,23 @@ def parse_func_repr(representation: str) -> Tuple[str, tuple, dict]:
...
-    for arg in args_repr.split(", "):
-        if "=" in arg:
-            keyword, argument = arg.split("=", 1)
-            if keyword.isidentifier():
-                kwargs.update({keyword: argument})
-                continue
-        args.append(arg)
+    arg_list = args_repr.split(", ")
+    # Extract the keyword arguments from the positional arguments
+    for index in range(len(arg_list)):
```
LilianBoulard, line 39: Using `enumerate` would be a little cleaner
```py
    for index, value in enumerate(arg_list):
```

benchmarks/utils/monitor.py
```diff
@@ -165,7 +175,11 @@ def exec_func(*args, **kwargs) -> Dict[str, List[float]]:
                     if time:
                         t0 = perf_counter()
 
-                    func(*args, **kwargs)
+                    res_dic = func(*args, **kwargs)
+
+                    if not res_dic is None:
```
LilianBoulard, line 180: Python style recommends
```suggestion
                    if res_dic is not None:
```

benchmarks/benchmark_fuzzy_join_minhash.py
```diff
@@ -0,0 +1,433 @@
...
+            "ngram_range",
+            "dataset_name",
+            "time",
+            "memory",
+            "precision",
+            "recall",
+            "f1",
+            "time_fj",
+        ],
+    )
```
LilianBoulard, line 396: Now this is some annoying code! :joy:
Having to write that looks to me like `monitor` needs to be improved.

jovan-stojanovic:
Maybe also change the name of your benchmark file to benchmark_fuzzy_join_hash.py, I got confused when I saw the `minhash`, I thought you used the [MinHash encoder](https://github.com/dirty-cat/dirty_cat/blob/main/dirty_cat/_minhash_encoder.py) :smiley:

benchmarks/utils/monitor.py
LeoGrin: It was to handle parameters with tuple parameters (like ngram_range), but it's not useful with the new changes.

jovan-stojanovic (approved):
Thanks for this, I think this is ready now?

LilianBoulard (approved):
Thanks for the additions! I've also improved a few things :)
Looks good to merge.

PR #449, opened 2023-01-02
Hashing vectorizer in fuzzy join

LeoGrin:
Following #446 (which seems to show that using `HashingVectorizer` is almost always faster than using `CountVectorizer`, without any apparent accuracy tradeoff) and discussion, this PR adds a `vectorizer` parameter to the `fuzzy_join` function, which defaults to `hashing`, i.e using `HashingVectorizer`.

Replaces #420.

I think someone should check my benchmark in #446 before we consider merging this PR.


LeoGrin:
Could someone add a "No changelog needed" tag? As the fuzzy_join feature is not out yet, I don't think changing the changelog is needed.

LeoGrin:
Closing as it is now redundant with #420

PR #453, opened 2023-01-06
Support paths as strings in public fetching API

LilianBoulard:
- Adds support for paths as strings in the public fetching API
- Uses cleaner temporary directories in the fetching test


LilianBoulard:
Changelog entry taken care of by #432

jovan-stojanovic (commented):
It seems there is a problem with the fetch_openml function and the temporary directory where you store downloaded files..

GaelVaroquaux:
Darn, there are test failures, and they seem related to this PR

LilianBoulard:
Yes, and it's system-specific, which is very annoying to debug... I'll try and tackle that today

Issue #462, opened 2023-01-16
Uninformative error when calling SuperVectorizer's transform before fit


LeoGrin:
Calling transform before fit with `SuperVectorizer` gives an uninformative error.

dirty_cat version:
'0.4.dev0'

Expected behavior:
Error explaining that the vectorizer is not fitted.

Observed behavior:
Error: `AttributeError: 'SuperVectorizer' object has no attribute 'columns_'

Steps and code to reproduce the bug:

```python
import dirty_cat
import numpy as np
X = np.array([["a", "b", "c"], ["a", "c", "b"]])
sv = dirty_cat.SuperVectorizer()
sv.transform(X)
```
gives the error `AttributeError: 'SuperVectorizer' object has no attribute 'columns_'` instead of explaining that the vectorizer is not fitted.

PR #463, opened 2023-01-16
Use `check_is_fitted` for all dirty_cat transformers.

LeoGrin:
This gives an intelligible error if someone calls `transform`. Solves #462


jovan-stojanovic (approved):
Great, thanks a lot!

jovan-stojanovic:
As having good, informative errors is very important, I would include this in the next release.

PR #465, opened 2023-01-17
ENH: more didactic deduplicate example

GaelVaroquaux:
- Add the functions to the generated docs
- Incorporate the random number generation seed accordingly to scikit-learn's standard
- A bit of phrasing


examples/06_deduplication.py
```diff
@@ -90,6 +85,12 @@
 # a lighter color means a larger distance. We can see that we are dealing with three
 # clusters - the original medication names and their misspellings that cluster around them.
 
+from dirty_cat._deduplicate import compute_ngram_distance
```
jovan-stojanovic, line 88:
```suggestion
from dirty_cat import compute_ngram_distance
```
This function was actually made public and added to the init, so there is no need to import it from a private submodule #466

jovan-stojanovic (changes requested):
Simple solution for the first point in #466

GaelVaroquaux:
I applied the suggested change and resolved conflicts. Thanks!

jovan-stojanovic:
Merged, thanks!

Issue #455, opened 2023-01-09
`SuperVectorizer`'s default `low_card_cat_transformer` raises exceptions on unseen categories at test time


lsorber:
The default `SuperVectorizer` will raise an exception on unseen categories at test time. For the default behaviour, I believe it would be more friendly to try to handle those unseen categories as best as possible (e.g. encoding them as an infrequent or missing category) instead of throwing the issue back to the user.

Workaround:
```python
ohe = OneHotEncoder(drop="if_binary", handle_unknown="infrequent_if_exist")
supervec = dirty_cat.SuperVectorizer(low_card_cat_transformer=ohe)
```

dirty_cat version: 0.3.0

Expected behavior: No exception raised for unseen categories

Observed behavior: Exception raised for unseen categories

LeoGrin:
Thanks for the suggestion! I agree that it would be a better default. The only issue is that the `infrequent_if_exist` option requires sklearn >= 1.1, whereas we support sklearn from 0.23. One solution would be to use this default if sklearn version is greater than 1.1, but it may be confusing for users. Curious what other people think.

LeoGrin:
I think this should be closed

PR #473, opened 2023-01-23
Use handle_unknown=ignore in SuperVectorizer

LeoGrin:
Change default `low_card_cat_transformer` in SuperVectorizer to use handle_unknown="ignore".

Solves #455

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,7 +40,7 @@ Minor changes
   which can be used to specify where to save and load from datasets.
   :pr:`432` by :user:`Lilian Boulard <LilianBoulard>`
 
-* The :class:`SuperVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults
+* The :class:`TableVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults
   to `handle_unknown="ignore"` instead of `handle_unknown="error"` (for sklearn >= 0.24.2).
   This means that categories seen only at test time will be encoded by a vector of zeroes instead of raising an error. :pr:`473` by :user:`Leo Grinsztajn <LeoGrin>`
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,8 +40,9 @@ Minor changes
   which can be used to specify where to save and load from datasets.
   :pr:`432` by :user:`Lilian Boulard <LilianBoulard>`
 
-* The :class:`SuperVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults to `handle_unknown="ignore"` and `drop=None` instead of `handle_unknown="error"` and `drop=if_binary`. This means
-  that categories seen only at test time will be encoded by a vector of zeroes instead of raising an error, and that no category will be dropped. :pr:`473` by :user:`Leo Grinsztajn <LeoGrin>`
+* The :class:`SuperVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults
+  to `handle_unknown="ignore"` instead of `handle_unknown="error"` (for sklearn >= 0.24.2).
+  This means that categories seen only at test time will be encoded by a vector of zeroes instead of raising an error. :pr:`473` by :user:`Leo Grinsztajn <LeoGrin>`
 
 Bug fixes
 ---------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,8 +40,8 @@ Minor changes
   which can be used to specify where to save and load from datasets.
   :pr:`432` by :user:`Lilian Boulard <LilianBoulard>`
 
-* The :class:`SuperVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults to `handle_unknown="ignore"` instead of `handle_unknown="error"`, which means
-  than categories seen only at test time will be encoded by a vector of zeroes instead of raising an error. :pr:`473` by :user:`Leo Grinsztajn <LeoGrin>`
+* The :class:`SuperVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults to `handle_unknown="ignore"` and `drop=None` instead of `handle_unknown="error"` and `drop=if_binary`. This means
+  that categories seen only at test time will be encoded by a vector of zeroes instead of raising an error, and that no category will be dropped. :pr:`473` by :user:`Leo Grinsztajn <LeoGrin>`
 
 Bug fixes
 ---------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,7 +41,7 @@ Minor changes
   :pr:`432` by :user:`Lilian Boulard <LilianBoulard>`
 
 * The :class:`TableVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults
-  to `handle_unknown="ignore"` instead of `handle_unknown="error"` (for sklearn >= 0.24.2).
+  to `handle_unknown="ignore"` instead of `handle_unknown="error"` (for sklearn >= 1.0.0).
   This means that categories seen only at test time will be encoded by a vector of zeroes instead of raising an error. :pr:`473` by :user:`Leo Grinsztajn <LeoGrin>`
 
 Bug fixes
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,9 @@ Minor changes
   which can be used to specify where to save and load from datasets.
   :pr:`432` by :user:`Lilian Boulard <LilianBoulard>`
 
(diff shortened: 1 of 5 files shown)


LeoGrin:
Hmm it seems that we can't both specify `drop='if_binary`, and `handle_unknown="ignore"`, because it would create some ambiguity on what a zero vector means. I would say it's better to keep what we have now, that is `drop='if_binary` and `handle_unknown="error"`, but I'd like to hear other opinions.

GaelVaroquaux:
> Hmm it seems that we can't both specify drop='if_binary, and handle_unknown= "ignore", because it would create some ambiguity on what a zero vector means. I would say it's better to keep what we have now, that is drop='if_binary and handle_unknown="error", but I'd like to hear other opinions.

I would rather have a pipeline that runs and does not raise error, from a practical point of view.

dirty_cat/_super_vectorizer.py
```diff
@@ -427,6 +429,15 @@ def _apply_cast(self, X: pd.DataFrame) -> pd.DataFrame:
         for col in self.imputed_columns_:
             X.loc[:, col] = _replace_missing_in_cat_col(X[col])
         for col, dtype in self.types_.items():
+            # if categorical, add the new categories to prevent
+            # them to be encoded as nan
+            if pd.api.types.is_categorical_dtype(dtype):
+                knwon_categories = dtype.categories
+                new_categories = pd.unique(X[col])
+                dtype = pd.CategoricalDtype(
+                    categories=knwon_categories.union(new_categories)
```
jovan-stojanovic, line 438:
```suggestion
                known_categories = dtype.categories
                new_categories = pd.unique(X[col])
                dtype = pd.CategoricalDtype(
                    categories=known_categories.union(new_categories)
```
I guess is what you meant

jovan-stojanovic (commented):
Thanks, just a typo to correct I think.

jovan-stojanovic (approved):

CHANGES.rst
```diff
@@ -38,6 +38,9 @@ Minor changes
   which can be used to specify where to save and load from datasets.
   :pr:`432` by :user:`Lilian Boulard <LilianBoulard>`
 
+* The :class:`SuperVectorizer`'s default `OneHotEncoder` for low cardinality categorical variables now defaults to `handle_unknown="ignore"` and `drop=None` instead of `handle_unknown="error"` and `drop=if_binary`. This means
+  that categories seen only at test time will be encoded by a vector of zeroes instead of raising an error, and that no category will be dropped. :pr:`473` by :user:`Leo Grinsztajn <LeoGrin>`
```
GaelVaroquaux, line 42: Why did we get rid of "drop=if_binary"? It had been added as a specific request to avoid having too big encoded matrices.

dirty_cat/_super_vectorizer.py
```diff
@@ -427,6 +429,15 @@ def _apply_cast(self, X: pd.DataFrame) -> pd.DataFrame:
...
         for col, dtype in self.types_.items():
+            # if categorical, add the new categories to prevent
+            # them to be encoded as nan
+            if pd.api.types.is_categorical_dtype(dtype):
+                known_categories = dtype.categories
+                new_categories = pd.unique(X[col])
+                dtype = pd.CategoricalDtype(
+                    categories=known_categories.union(new_categories)
+                )
+                self.types_[col] = dtype
```
GaelVaroquaux, line 440: That part is really subtle (I would have gotten it wrong). Do we have tests that fail if it is not coded right?

CHANGES.rst
GaelVaroquaux: I understand that the semantics of the two options somewhat clash (maybe something to discuss with the scikit-learn team) and bad things will happen if we have binary and unknown in the same column, but it is quite unlikely.

LeoGrin: Potential solutions to keep using drop="binary":
- keep handle_unknown="ignore" for non-binary columns, but use handle_unknown="error" for binary columns
- create a subclass of OneHotEncoder which maps new columns encountered during transform to "infrequent" (quite similar to `infrequent_if_exists` but without needing sklearn > 1.1).

Tagging some scikit-learn people to have other opinions: @glemaitre @Vincent-Maladiere

glemaitre: I think this is related to: https://github.com/scikit-learn/scikit-learn/issues/18072 and resolved in https://github.com/scikit-learn/scikit-learn/pull/19041.

I think that the argument taken was the following: https://github.com/scikit-learn/scikit-learn/issues/18072#issuecomment-723934468

In the end, we are lenient and we are not capable of inverting properly the array.

GaelVaroquaux: I think that being lenient is fine for our purposes (the case is going to be very infrequent) and thus we should just turn on both options.

Thanks @glemaitre !

LeoGrin: Thanks @glemaitre ! I also think that that's the best solution. The only issue is that sklearn is lenient from 0.24.2, so we get an error for earlier version (from 0.23.0). I think we can just catch it.

dirty_cat/_super_vectorizer.py
LeoGrin: Good point. We have a test that checks that the output is right, but I'll add a test to check that the format is right too.

LeoGrin: Actually I've deleted this and converted every categorical column to `object` dtype, to avoid dealing with the categories attached to a dtype. I think it makes everything simpler, but I'm tagging @LilianBoulard to make sure this isn't wrong (I've deleted some of your code dealing with categories).

CHANGES.rst
LeoGrin: What I finally chose is to just turn on both options, but to use `handle_ignore="error"` if sklearn version is < 0.24.2, while issuing a warning. I know that we excluded using `infrequent_if_exists` because changing behaviour for different  sklearn version might be confusing for users, but I figured it would be okay in this case, as it only concerns sklearn versions between 0.23 and 0.24.2.

GaelVaroquaux:
> Actually I've deleted this and converted every categorical column to object dtype,

Often, in pandas things are faster (and consume less memory) with categorical dtype, rather than object dtype. It is in general a good idea to stick to categorical dtype.

GaelVaroquaux:
> What I finally chose is to just turn on both options, but to use handle_ignore= "error" if sklearn version is < 0.24.2, while issuing a warning.

Might lead to hacky code to issue the warning only for the right case.

LeoGrin:
> Often, in pandas things are faster (and consume less memory) with categorical dtype, rather than object dtype. It is in general a good idea to stick to categorical dtype.

Ah ok, I've reverted the change to keep categorical columns. If the speed gain is significant, would it make sense to convert columns with string or object dtypes to categorical? (maybe if the number of category is not too high)

LeoGrin:
> What I finally chose is to just turn on both options, but to use handle_ignore= "error" if sklearn version is < 0.24.2, while issuing a warning.
> Might lead to hacky code to issue the warning only for the right case.

Right now I issue a warning in `_clone_transformers` (called at the beginning of fit) if `low_card_cat_encoder` is None (default). So there may be a warning a bit more often what we could have (if we catched the `OneHotEncoder` errors for earlier sklearn versions), but on the other hand we avoid hacky code.

dirty_cat/tests/test_table_vectorizer.py
```diff
@@ -463,3 +464,60 @@ def test_check_name_change():
...
+
+
+def test_handle_unknown():
+    """
+    Test that new categories encountered in the test set
+    are handled correctly.
+    """
+    X = _get_clean_dataframe()
+    # Test with low cardinality and a StandardScaler for the numeric columns
+    sup_vec = SuperVectorizer(
```
GaelVaroquaux, line 476: You should be using TableVectorizer, and not SuperVectorizer which is no deprecated.

GaelVaroquaux (changes requested):
A couple of small changes are necessary, but the overall strategy is the right one. Thanks!!!

dirty_cat/tests/test_table_vectorizer.py
```diff
@@ -473,7 +473,7 @@ def test_handle_unknown():
     """
     X = _get_clean_dataframe()
     # Test with low cardinality and a StandardScaler for the numeric columns
-    sup_vec = SuperVectorizer(
```
GaelVaroquaux: I'm sorry, but it's kind of ugly to keep the name "sup_vec".

In a few years people will wonder what it is there for.

GaelVaroquaux (changes requested):
I'd like a bit more name change :(

dirty_cat/tests/test_table_vectorizer.py
LeoGrin: ooops

LeoGrin: I replaced it everywhere in dirty_cat (it also appeared in examples)

GaelVaroquaux (approved):
Excellent! Thank you!!

GaelVaroquaux:
Merging!

Issue #474, opened 2023-01-24
Uninformative error in GapEncoder when n_samples < n_components


LeoGrin:
When calling `GapEncoder` fit method with n_samples < n_components, we get the error `ValueError: n_samples=5 should be >= n_clusters=30.` raised by sklearn's kmean, which is hard to understand and fix.

dirty_cat version:
0.4.0b1

Expected behavior:
Informative error explaining what to change.

Observed behavior:
Uninformative error mentioning parameters not accessible to the user.

Steps and code to reproduce the bug:

```python
import numpy as np
from dirty_cat import SuperVectorizer
X = np.array([["officer", "manager", "lawyer", "chef", "teacher"]]).reshape(-1, 1)
sup_vec = SuperVectorizer(
        cardinality_threshold=4,
    )
sup_vec.fit_transform(X)
```

Sohaib90:
Can I look into this?

PR #475, opened 2023-01-24
More informative error when n_samples < n_components in GapEncoder

LeoGrin:
Solves #474

Previously, an error was raised from Kmeans, and talked about `n_clusters` instead of `n_components`.


LeoGrin:
I think this deserves a "No changelog needed", because it's a very small change.

jovan-stojanovic (approved):
Thanks!

PR #483, opened 2023-01-31
DOC: css for narrow screens

GaelVaroquaux:
Better layout on mobile phones

Should avoid the dangling list on the left, and the dangling "example" on the bottom right
image


GaelVaroquaux:
Hum, improvement, but imperfect, I need to work more on this.

GaelVaroquaux:
Ok, this should be ready to merge

jovan-stojanovic (approved):
Looks great, thanks!

PR #486, opened 2023-02-09
Add dynamic type checking

LilianBoulard:
Supersedes #443.

This implementation of a type checker is different than the one proposed in #443: instead of implementing an independent type system, it uses PEP 434 annotations to check for the types at execution time.

It still requires a little bit of work:
- [X] Fixed: Inspection is not possible with the main decorator (`validate_types`). It's temporarily addressed by the other decorator `validate_types_with_inspect`.
  - It in-turns causes issues with the class method `_get_param_names`, which uses inspect.
- [ ] The main decorator is currently almost 200 lines long, so dissecting it into smaller functions and/or into a more suitable data structure would be a good readability improvement.
- [ ] I still need to write a few tests.
- [ ] Add a constraint engine


LilianBoulard:
As discussed with the team, this feature is a little too constraining for users and complicated to extend and maintain (especially considering duck-typing), so we will stick to old-school checks.
Also, [beartype](https://github.com/beartype/beartype) does it better!

PR #489, opened 2023-02-13
Refactor fetching tests

LilianBoulard:
Following the CI errors of #453 and the conversation with the team, we agree that the tests I wrote about two years ago are too tightly coupled with the implementation.

This PR aims at refactoring these in order to have something more high-level, without caring too much of what's going on in the background.

### DIFF
--- a/dirty_cat/datasets/tests/test_fetching.py
+++ b/dirty_cat/datasets/tests/test_fetching.py
@@ -11,6 +11,17 @@ import pytest
 from dirty_cat.datasets import _fetching
 
 
+def fetch_employee_salaries(**kwargs):
+    """
+    Wrapper for `fetching.fetch_employee_salaries`, that sets the
+    optional parameters to False, so that the dataframe loaded in memory
+    and the dataframe on disk (which we will read) correspond.
+    """
+    return _fetching.fetch_employee_salaries(
+        drop_irrelevant=False, drop_linked=False, **kwargs
+    )
+
+
 @pytest.mark.parametrize(
     ("fetching_function",),
     [
@@ -20,7 +31,7 @@ from dirty_cat.datasets import _fetching
         (_fetching.fetch_open_payments,),
         (_fetching.fetch_traffic_violations,),
         (_fetching.fetch_drug_directory,),
-        (_fetching.fetch_employee_salaries,),
+        (fetch_employee_salaries,),
     ],
 )
 def test_openml_fetching(fetching_function: Callable):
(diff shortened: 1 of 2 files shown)


dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -1,400 +1,108 @@
...
+            assert dataset.y.shape > (5,)
+            assert dataset.X.shape[0] == dataset.y.shape[0]
+
+            # Less important checks, but might help finding errors
+            assert dataset.name == (
+                fetching_function.__name__[len("fetch_") :]
+                .replace("_", " ")
+                .capitalize()
+            )
+            assert "openml.org" in dataset.source
```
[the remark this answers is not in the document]
LilianBoulard: :eyes:

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -1,400 +1,120 @@
...
-
-        assert returned_info["path"].is_file()
-
-        dataset = pd.read_csv(
-            returned_info["path"], sep=",", quotechar="'", escapechar="\\"
+        else:
+            assert isinstance(dataset_wo_load, _fetching.DatasetInfoOnly)
+
+        # FIXME: An more elegant way of testing whether the dataset is loaded
+        #  in memory would be to monitor it and expect it to stay quite low.
```
GaelVaroquaux, line 58: Right, but that's hard to do reliably, so I wouldn't consider embarking on this too quickly

GaelVaroquaux (changes requested):
It seems to me that we have lost a lot of the testing logic. The tests are now much more lenient and don't really test the dowloaded data.

This seems to me like a regression.

To me, the goal of this refactor is to enable using path objects. I do not see why this goal implies giving up on testing what is actually returned.

LilianBoulard:
> It seems to me that we have lost a lot of the testing logic.

Yes, and it's on purpose, to de-couple the tests from the implementation - which is not important from the user's perspective: to me, we only care about the resulting data. It also makes the test much simpler to understand and maintain.

> The tests are now much more lenient and don't really test the dowloaded data.

You're right, and they should be improved in this regard before merging.
I'm looking for high-level properties which reflect that the data we got is correct. Checking things like shape and cardinality per column would be a good starting point IMO. It's hard-coded properties, but we don't expect these datasets to change anyway.

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -409,19 +129,26 @@ def test_fetch_world_bank_indicator():
...
+            returned_info = _fetching.fetch_world_bank_indicator(
+                indicator_id=test_dataset["id"],
+                directory=temp_dir,
+            )
 
-        except URLError:
+        except (ConnectionError, URLError):
             warnings.warn(
                 "No internet connection or the website is down, test aborted."
             )
```
jovan-stojanovic, line 154: We should maybe use pytest.skip as above?

jovan-stojanovic (commented):
Just a small remark, otherwise looks good!

dirty_cat/datasets/tests/test_fetching.py
LilianBoulard: I think it's already the case? I'll remove the warning though!

PR #492, opened 2023-02-15
Set default value to None for similarity in SimilarityEncoder

LilianBoulard:
When instantiating the `SimilarityEncoder`, it raises a warning by default, which is bad behavior.

```
>>> SimilarityEncoder()
UserWarning: The "similarity" argument is deprecated since dirty_cat 0.3, and will be removed in 0.5.The n-gram similarity is the only one currently supported.
```


GaelVaroquaux:
Good job!

PR #493, opened 2023-02-15
Parallelize parametrized tests

LilianBoulard:
Adds [`pytest-xdist`](https://github.com/pytest-dev/pytest-xdist) to the development dependencies to parallelize the tests, thus speeding them up.


LilianBoulard:
With with `-n auto`, we observe an average reduction of 12.25 seconds per test between main and this PR. That amounts to 2m27s saved overall (14m47s -> 12m20s).

LilianBoulard:
Using logical cores took 13m14s, which is slightly worse than auto (on a sample size of 1). Let's stick with the latter for now.

PR #494, opened 2023-02-15
Fix long CI install

LilianBoulard:
On some versions specified in the CI matrix, the installation of dirty_cat takes more than 10 minutes, which is very annoying and consumes resources pointlessly.
image
This PR attempts to fix that.

### DIFF
--- a/build_tools/github/install.sh
+++ b/build_tools/github/install.sh
@@ -1,3 +1,3 @@
 #!/bin/bash -e
 
-pip install --progress-bar off --upgrade ".[$DEPS_VERSION]" --use-pep517
+pip install --progress-bar off --only-binary :all: --upgrade ".[$DEPS_VERSION]"
--- a/build_tools/github/install.sh
+++ b/build_tools/github/install.sh
@@ -1,3 +1,3 @@
 #!/bin/bash -e
 
-pip install --progress-bar off --upgrade ".[$DEPS_VERSION]"
+pip install --progress-bar off --upgrade ".[$DEPS_VERSION]" --use-pep517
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
Did this PR improve things?

LilianBoulard:
No, and it's even worse by 5 minutes (49m28s -> 55m21s)...

LilianBoulard:
As suggested by Loïc, using `--no-binary` forces using wheels, but then the tests fail if the package is not available.
Looks to me like the better option is to keep `--no-binary` and bump the requirements.

GaelVaroquaux:
The failure is:
```
ERROR: Cannot install dirty-cat[dev,min-py310]==0.4.0b2 because these package versions have conflicting dependencies.

The conflict is caused by:
    dirty-cat[dev,min-py310] 0.4.0b2 depends on numpy>=1.17.3
    dirty-cat[dev,min-py310] 0.4.0b2 depends on numpy==1.21.2; extra == "min-py310"
```
I really don't understand it

GaelVaroquaux:
I think that the problem is because there is no python 3.10 binary for numpy 1.21.2: https://pypi.org/project/numpy/1.21.2/#files

GaelVaroquaux:
If I read things correctly, there is one for numpy 1.12.3 https://pypi.org/project/numpy/1.21.3/#files

LilianBoulard:
This PR is ready, it fixes all issues with the missing wheels.

jovan-stojanovic (approved):

GaelVaroquaux:
Definitely an improvement. I'm merging!!

We will also need to profile our tests to see where the time is spent. They are 20mn-long. It seems a bit long compared to the amount of code that we have.

PR #518, opened 2023-03-02
Fix `fetch_openml` parser warning

### LINKED ISSUE #517: `sklearn.datasets.fetch_openml` parser warning
Seen on [example 1](https://dirty-cat.github.io/stable/auto_examples/01_dirty_categories.html#sphx-glr-auto-examples-01-dirty-categories-py)
```
/home/circleci/project/dirty_cat/datasets/_fetching.py:586: UserWarning: Could not find the dataset 42125 locally. Downloading it from OpenML; this might take a while... If it is interrupted, some files might be invalid/incomplete: if on the following run, the fetching raises errors, you can try fixing this issue by deleting the directory /home/circleci/project/dirty_cat/datasets/data.
  info = _fetch_openml_dataset(dataset_id, data_directory)
/home/circleci/project/miniconda/envs/testenv/lib/python3.8/site-packages/sklearn/datasets/_openml.py:932: FutureWarning: The default value of `parser` will change from `'liac-arff'` to `'auto'` in 1.4. You can set `parser='auto'` to silence this warning. Therefore, an `ImportError` will be raised from 1.4 if the dataset is dense and pandas is not installed. Note that the pandas parser may return different data types. See the Notes Section in fetch_openml's API doc for details.
  warn(
```

LilianBoulard:
Fixes #517


GaelVaroquaux:
The proposed fix cannot work on older versions of scikit-learn (and the test are failing).

The correct fix is to test for the version of scikit-learn and add the argument only if the version of scikit-learn is recent enough.

LilianBoulard:
I know but even with the most recent version, the tests fail, and I'm trying to understand why.
I agree the fix I submitted was lazy, I'll look into the error :)

GaelVaroquaux:
> I agree the fix I submitted was lazy, I'll look into the error :)

Lazy is good when it works :)

LilianBoulard:
So the issue we have is related to scikit-learn/scikit-learn#25478.
The PR scikit-learn/scikit-learn#25511 has been submitted and should be ready for sklearn 1.2.2. I can preemptively submit a fix, but we should leave this PR open until it's released.

GaelVaroquaux:
Tests are passing. Is this mergeable? I noted that the PR is still in draft mode.

dirty_cat/datasets/_fetching.py
```diff
@@ -426,11 +428,28 @@ def _download_and_write_openml_dataset(dataset_id: int, data_directory: Path) ->
     # which behaves just like a ``namedtuple``.
     # However, we do not want to save this data into memory:
     # we will read it from the disk later.
-    fetch_openml(
-        data_id=dataset_id,
-        data_home=str(data_directory),
-        as_frame=True,
-    )
```
GaelVaroquaux, line 433: Cosmetic: to minimize code duplicate, I would rather use a kwarg dictionary: initialize it empty, populate it with the right key, value pairs in "if/then" clause, and call "fetch_openml" expending this dict with "**kwargs".

GaelVaroquaux (changes requested):
One small suggestion

GaelVaroquaux (approved):
LGTM, thanks a lot!

GaelVaroquaux:
The code looks just right!

We have test failing on the main branch (not sure how we got there, we screwed up at some point). Out of good practice, I'll wait for these to be fixed, and rebase, so as to merge with passing tests.

jovan-stojanovic (approved):
Merging now the tests have passed.

PR #524, opened 2023-03-03
Shorten fetching tests

### LINKED ISSUE #523: Fetching tests are halting CI
Since #489, the fetching tests take about 20 minutes to run, and the MacOS workers are getting stuck, which slows development times.
This is due to the fact that we download all datasets kind of sequentially (it's technically process-parallelized since #493, but we would need concurrent-parallelization in this case, and it wouldn't even fix the issue entirely).

image

Instead of downloading all datasets, a more efficient heuristic would be to download one small dataset, test the assumptions we have on it, and just check a few internals with the other datasets, without downloading them.

LilianBoulard:
Fixes #523.

Implements the heuristic discussed over there.


jovan-stojanovic:
Some tests are failing, I think because of the added `openml` dependency. Is it faster if we use `DatasetInfoOnly` or this also downloads all datasets?

LilianBoulard:
The reason `openml` is now a requirement is because we use `openml.datasets.check_datasets_active` to ensure the dataset exist on the platform. The error is quite weird, I don't understand why it works on my machine (ubuntu) but not on the CI...

LilianBoulard:
By looking at the [CI of OpenML](https://github.com/openml/openml-python/actions/runs/4330220401/jobs/7561309943), I was able to figure out that we needed to allow building liac-arff with `--no-binary liac-arff`. It works fine now!

GaelVaroquaux:
The problem with this approach, if I understand it correctly, is that if a dataset become unavailable upstream (these things happen) we don't see it.

I would rather have this approach + the old one, but running only on one of the workers (for instance via an environment variable, or another clean way of skipping tests)

LilianBoulard:
> If a dataset become unavailable upstream (these things happen) we don't see it.

There is actually a dedicated test (`test_openml_datasets_exist`) that checks the datasets are still available upstream.

dirty_cat/datasets/tests/test_fetching.py
```diff
@@ -1,111 +1,106 @@
 from tempfile import TemporaryDirectory
-from typing import Callable, Tuple
 from unittest import mock
 from urllib.error import URLError
 
 import pandas as pd
 import pytest
+from openml.datasets import check_datasets_active
```
GaelVaroquaux, line 7: Should we do an optional import and optional test so that the lone developer can run tests on her laptop without having this packaged installed.

GaelVaroquaux:
> > If a dataset become unavailable upstream (these things happen) we don't see it.
> 
> There is actually a dedicated test (`test_openml_datasets_exist`) that checks the datasets are still available upstream.

Sorry, I had missed it indeed. That's great! Congratulations!

GaelVaroquaux (commented):
One detail on making the dependency optional and it's good to go!

GaelVaroquaux (approved):
LGTM. Merging! Thanks!

PR #526, opened 2023-03-03
Fix duplicate CI workers

### LINKED ISSUE #525: Duplicated workers in CI
It seems that workers are duplicated on the CI, see for example [this run](https://github.com/dirty-cat/dirty_cat/actions/runs/4322419316/jobs/7544820670).

This seems to be the case since #502.

LilianBoulard:
Fixes #525.


LilianBoulard:
Works! Merging right away.

PR #535, opened 2023-03-29
DOC simplify `print_worst_matches` for `04_fuzzy_joining_and_FeatureAugmenter`

Vincent-Maladiere:
**What does this implement/fix?**

The `04_fuzzy_joining_and_FeatureAugmenter` example uses a `print_worst_matches` function to sort a DataFrame, while the same result could be obtained with a much more simple `df.sort_values`.


LilianBoulard:
LGTM! The errors are in main and should be fixed by #543. Merging!

PR #538, opened 2023-03-31
DOC Configure filename_pattern to run all examples during doc build

### LINKED ISSUE #527: Examples not rendering on the dev website version
See here: https://dirty-cat.github.io/dev/

It seems that this became a problem after merging https://github.com/dirty-cat/dirty_cat/pull/507

lesteve:
Fix #527

Since #507 `"filename_pattern": ""` was removed so this reverted to the default sphinx-gallery behaviour which is to run the examples starting with `plot_`. I used `".*"` to match any kind of files which is a bit more explicit than `""`, you could also use `\d+_.*\.py` or a variation of it if you prefer.

Locally this work, but I am going to push a commit with `[doc build]` to make sure that examples are indeed run.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -273,6 +273,7 @@ sphinx_gallery_conf = {
         # We don't specify the other modules as we use the intershpinx ext.
         # See https://sphinx-gallery.github.io/stable/configuration.html#link-to-documentation  # noqa
     },
+    "filename_pattern": ".*",
     "examples_dirs": "../examples",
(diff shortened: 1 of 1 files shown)


jovan-stojanovic (approved):
Thanks @lesteve! This seems to have resolved the issue.

GaelVaroquaux:
Thanks a lot @lesteve . I was indeed quite bothered by the fact that our dev website had a broken gallery.

PR #539, opened 2023-04-04
Add functions for exploring KEN embeddings

### LINKED ISSUE #495: Help users find relevant ken types
We should provide a helper function to help people tune the "type" argument, maybe a function called "get_ken_matching_types" that would return all the types that have the corresponding substring. This function should be exposed to users and mentioned in the relevant examples and related functions.

Complement to #487

LilianBoulard:
Fixes #495

### DIFF
--- a/dirty_cat/datasets/_ken_embeddings.py
+++ b/dirty_cat/datasets/_ken_embeddings.py
@@ -34,6 +34,10 @@ def get_ken_table_aliases() -> Set[str]:
         Helper function to search for entity types.
     :func:`get_ken_embeddings`
         Download Wikipedia embeddings by type.
+
+    Notes
+    -----
+    Requires and Internet connection to work.
     """
     correspondence = pd.read_csv(_correspondence_table_url)
     return set(["all_entities"] + list(correspondence["table"].values))
@@ -42,14 +46,19 @@ def get_ken_table_aliases() -> Set[str]:
 def get_ken_types(
     search: str = None,
     *,
+    exclude: Optional[str] = None,
     embedding_table_id: str = "all_entities",
 ):
     """Helper function to search for entity types.
+
+    The result can then be used with :func:`get_ken_embeddings`.
 
     Parameters
     ----------
     search : str, optional
         Substring pattern that filters the types of entities.
+    exclude : str, optional
+        Substring pattern to exclude from the search.
     embedding_table_id : str, optional, default='all_entities'
         Table of embedded entities from which to extract the embeddings.
         Get the supported tables with :func:`get_ken_table_aliases`.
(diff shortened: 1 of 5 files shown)


dirty_cat/datasets/_ken_embeddings.py
```diff
@@ -1,57 +1,139 @@
...
+    search: str = None,
+    *,
+    embedding_table_id: str = "all_entities",
 ):
+    """Helper function to search for entity types.
+
+    Parameters
+    ----------
+    search : str, optional
+        Substring pattern that filters the types of entities.
```
jovan-stojanovic, line 52: Maybe you can add a parameter exclude just like in the `get_ken_embeddings`. I found that this is very useful when you want to be a bit more precise: for instance search for "games" but exclude "card" games.

jovan-stojanovic: And to make it more compatible, have the same names for these parameters. Maybe replace `types` with `search` in the get_ken_embeddings or rename the two to `search_types`?

LilianBoulard:
> In a second iteration, we can add also simple reproducible examples to illustrate how people can easily use these functions.

Absolutely, and I think it could be included in the scope of #546 :)

dirty_cat/datasets/_ken_embeddings.py
```diff
@@ -1,57 +1,139 @@
...
     suffix : str, optional, default=""
         Suffix to add to the column names of the embeddings.
 
     Returns
     -------
-    embeddings: class:`~pandas.DataFrame`
+    :class:`~pandas.DataFrame`
         The embeddings of entities and the specified type from Wikipedia.
 
     See Also
```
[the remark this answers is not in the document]
LilianBoulard: Yep, thanks for pointing that out, I though I did!

dirty_cat/datasets/_ken_embeddings.py
LilianBoulard: I've ended up leaving as is because to me, in `get_ken_types` we *search* for available types, while in `get_ken_embeddings`, we get the embeddings matching some *types*.
I did add the `exclude` parameter though, it's indeed relevant :)

dirty_cat/datasets/_ken_embeddings.py
```diff
@@ -1,57 +1,157 @@
...
+    See Also
+    --------
+    :func:`get_ken_types`
+        Helper function to search for entity types.
+    :func:`get_ken_embeddings`
+        Download Wikipedia embeddings by type.
+
+    Notes
+    -----
+    Requires and Internet connection to work.
```
jovan-stojanovic, line 40:
```suggestion
    Requires an Internet connection to work.
```

jovan-stojanovic (approved):
One minor typo and LGTM!

PR #541, opened 2023-04-05
Fix _set_drop_idx bug for sklearn versions >=1.2.0 and < 1.2.2

LeoGrin:
In sklearn's `OneHotEncoder`, `_compute_drop_idx()` was replaced by `_set_drop_idx()` in V1.2.2 (see https://github.com/scikit-learn/scikit-learn/commit/1380e3c09b46568cd536ebc6f4720388f2f763d5), but we use it from V1.2.0, causing a bug on sklearn's versions 1.2.0 and 1.2.1. This PR fixes this.


LilianBoulard:
I can reproduce, LGTM!

PR #543, opened 2023-04-05
Improved datetime format inference

### LINKED ISSUE #540: TableVectorizer test failing with latest dependencies
We have an error when testing the TableVectorizer with the latest dependencies. See here:
https://github.com/dirty-cat/dirty_cat/actions/runs/4599113088/jobs/8124079078

Error:
```python
AssertionError: assert 'datetime64[ns]' == dtype('O')
```

LeoGrin:
Fix #540 
The bug in #540 was related to [pandas 2.0 update](https://pandas.pydata.org/pdeps/0004-consistent-to-datetime-parsing.html): as datetime format inference became stricter, a bug which was silent (inconsistent date format inferred for one column in the test) became loud.

This PR fixes this and improves date format detection (and thus datetime column detection in `TableVectorizer`). This replaces using `pd.datetime` directly for datetime format inference.

To do this, we try pandas's `guess_datetime_format` (on a subset) both with `dayfirst` being `True` and `False`, and see if one of these options finds a single format for all rows. If both work, we return the monthfirst format and raise a warning. If both return multiple formats, we give up.

This finds the right %d-%m-%Y  format for the failing test example, instead of failing or returning mixed format (see below).
```
  "11-12-2029",
  "02-12-2012",
  "11-09-2012",
  "13-02-2000",
  "10-11-2001"
```

### `pd.to_datetime` format inference
If I understand correctly, pandas datetime format inference (in `pd.to_datetime`) works like this:
#### Versions < 2.0
If not `infer_datetime_format`, the format of each row is inferred independently.
If `infer_datetime_format`, the format is inferred from the first non-missing example, and pandas *tries* to apply it to the other example (but can use another format otherwise).
This can easily create issues, for instance in our tests:
```
    "11-12-2029",
    "02-12-2012",
    "11-09-2012",
    "13-02-2000",
    "10-11-2001"
```
For all rows, the inferred format was %m-%d-%Y expect 13-02-2000, for which the format was %d-%m-%Y.
#### Version >= 2.0
The format is inferred from the first non-missing example, and pandas *tries* to apply it to the other example (and raise an error otherwise).


dirty_cat/_utils.py
```diff
@@ -66,3 +68,65 @@ def check_input(X) -> np.ndarray:
...
     return X_
+
+
+def _infer_date_format(date_column, n_trials=100) -> str:
+    """Infer the date format of a date column,
+    by finding a format which should work for all dates in the column.
+
+    Parameters
+    ----------
+    date_column : pandas.Series
```
LilianBoulard, line 79:
```suggestion
    date_column : :class:`~pandas.Series`
```

dirty_cat/_utils.py
```diff
@@ -66,3 +68,65 @@ def check_input(X) -> np.ndarray:
             )
 
     return X_
+
+
+def _infer_date_format(date_column, n_trials=100) -> str:
```
LilianBoulard, line 73:
```suggestion
def _infer_date_format(date_column: pd.Series, n_trials: int = 100) -> Optional[str]:
```

dirty_cat/_utils.py
```diff
@@ -66,3 +68,65 @@ def check_input(X) -> np.ndarray:
...
+    Parameters
+    ----------
+    date_column : pandas.Series
+        A column of dates, as strings.
+    n_trials : int, default=100
+        Number of rows to use to infer the date format.
+
+    Returns
+    -------
+    date_format : str
```
LilianBoulard, line 86:
```suggestion
    Optional[str]
```

dirty_cat/_utils.py
```diff
@@ -66,3 +68,65 @@ def check_input(X) -> np.ndarray:
...
+        The date format inferred from the column.
+        If no format could be inferred, returns None.
+    """
+    if len(date_column) == 0:
+        return None
```
LilianBoulard, line 91:
```suggestion
        return
```

dirty_cat/_utils.py
```diff
@@ -1,7 +1,9 @@
 import collections
+import warnings
 from typing import Any, Hashable
```
LilianBoulard, line 3:
```suggestion
from typing import Any, Hashable, Optional
```

dirty_cat/_utils.py
```diff
@@ -66,3 +68,65 @@ def check_input(X) -> np.ndarray:
...
+        warnings.simplefilter("ignore")
+        date_format_monthfirst = date_column_sample.apply(
+            lambda x: guess_datetime_format(x)
+        )
+        date_format_dayfirst = date_column_sample.apply(
+            lambda x: guess_datetime_format(x, dayfirst=True),
+        )
+    # if one row could not be parsed, return None
+    if date_format_monthfirst.isnull().any() or date_format_dayfirst.isnull().any():
+        return None
```
LilianBoulard, line 108:
```suggestion
        return
```

dirty_cat/_utils.py
```diff
@@ -66,3 +68,65 @@ def check_input(X) -> np.ndarray:
...
+                stacklevel=2,
+            )
+        return date_format_monthfirst[0]
+    elif date_format_dayfirst.nunique() == 1:
+        # only this format works for all the rows
+        return date_format_dayfirst[0]
+    else:
+        # more than two different formats were found
+        # TODO: maybe we could deal with this case
+        return None
```
LilianBoulard, line 132:
```suggestion
        return
```

LilianBoulard (commented):
Thanks for the PR, a few tweaks to merge and it's good for me!

dirty_cat/_utils.py
LilianBoulard, line 1: Maybe a remark on the `_infer_date_format` function: since it's only used in the `TableVectorizer`, I think it should be moved over to its file.

dirty_cat/_utils.py
```diff
@@ -1,7 +1,9 @@
 import collections
-from typing import Any, Hashable
+import warnings
+from typing import Any, Hashable, Optional
 
 import numpy as np
+from pandas._libs.tslibs.parsing import guess_datetime_format
```
LilianBoulard, line 6:
```suggestion
import numpy as np
import pandas as pd
from pandas._libs.tslibs.parsing import guess_datetime_format
```

LilianBoulard (approved):
LGTM, thanks again!

PR #546, opened 2023-04-07
Examples overhaul

### LINKED ISSUE #139: Comparison to factorization methods (e.g. SVD)
I used the notebook 02_fit_predict_plot_employee_salaries.html to compare the performance using SVD (with same number of components as MinHashEncoder). This could be added to the current or different notebook, if you agree. 

one-hot encoding
r2 score:  mean: 0.856; std: 0.034

target encoding
r2 score:  mean: 0.774; std: 0.033

similarity encoding
r2 score:  mean: 0.915; std: 0.012

minhash encoding
r2 score:  mean: 0.753; std: 0.025

svd encoding
r2 score:  mean: 0.856; std: 0.014

### LINKED ISSUE #341: Example 1: use DatetimeEncoder instead of isolating year
In [example 1](https://dirty-cat.github.io/stable/auto_examples/01_dirty_categories.html#sphx-glr-auto-examples-01-dirty-categories-py), we use the `year first hired` feature to predict wages. 

In the second part, when we use dirty_cat to encode variables, we may use the `DatetimeEncoder` instead of just passing through this variable. It is also possible not to drop the `date_first_hired`  variable and encode it instead.

To see what results this would give.

Also, would the SuperVectorizer (used in the second part) identify the `date_first_hired` column to be automatically encoded by the `DatetimeEncoder`? 
It does not at the moment for years.

### LINKED ISSUE #500: Show the results of deduplicate quicker in the example
In the deduplicate example, we should print the output of the deduplicate function right after calling it and discuss it (explain what it means)
https://dirty-cat.github.io/stable/auto_examples/06_deduplication.html#deduplication-suggest-corrections-of-misspelled-names

LilianBoulard:
Significantly changes how examples are organized, and their contents.

Here's the logic:
- I think example 1 should be an example that provides an overview of the problem of dirty data, and the tools the library provides to solve it. It should be a document that answers the question "what is dirty_cat?".
  - It was already its purpose, I just streamlined it, delegating the detailed explanations to the other notebooks
  - For now, I removed the feature importance section, which IMO didn't add much value to this new notebook. It's interesting in the context of the learning task, but the point of the example has deviated a bit.
- Example 2 should be about having a deeper understanding of the encoders the library currently features
  - It was already about that, I just improved the flow between the sections, removed the bloat and added a section for the `MinHashEncoder` (which I think is fun to get an intuition of)
- For the rest of the examples, for now I just improved the style and condensed some parts.

Fixes #139
Fixes #341
Fixes #500


examples/01_dirty_categories.py
```diff
@@ -1,274 +1,289 @@
 """
-Dirty categories: machine learning with non normalized strings
-==============================================================
+dirty_cat: Tackling Non-Normalized Categories In Machine Learning Settings
+==========================================================================
```
GaelVaroquaux, line 3:
```suggestion
Representing non-normalized categories for analysis
==============================================
```

examples/01_dirty_categories.py
```diff
@@ -1,274 +1,289 @@
...
-learning workflow, predicting the 'Current Annual Salary' with gradient
-boosted trees. First we manually assemble a complex encoder for the full
-dataframe, after which we show a much simpler way, albeit with less fine
-control.
+Typically, we encounter several types of morphological variants in categories:
+- **Typos**: "France" vs "Frqnce"
+- **Abbreviations**: "USA" vs "United States of America"
+- **Duplicated values**: "France/FR"
+- **Alternate values**: "Bordeaux, FR" & "Bordeaux"
+- **Variations**: "avocat" & "avocate" (masculine & feminine of lawyer in French)
```
GaelVaroquaux, line 20: How about we use "waiter" and "waitress" to stay in English (we need to keep this package international.

examples/01_dirty_categories.py
```diff
@@ -1,274 +1,289 @@
...
+This notebook offers an overview of *dirty_cat*, a library designed to handle
+dirty categorical data in machine learning settings.
+We will explore the following topics:
 
- .. |TV| replace::
-     :class:`~dirty_cat.TableVectorizer`
+1. Comparing the performance of a traditional encoder, the |OneHotEncoder|,
+   with the encoders provided by the *dirty_cat* library
+2. Introducing a simpler approach to assembling a machine-learning pipeline
+3. Demonstrating how to join and augment dirty tables effectively
```
GaelVaroquaux, line 38: I worry that the scope of this example is too big. Maybe I have the wrong feeling, but I'd like to be to say: one page describes one idea.

I see different ideas here, that I believe should probably be in different pages:
- Encoders: representing data, including in messy data
- Representing strings even when non-normalized
- Augmenting data (that definitely another track).

GaelVaroquaux (commented):
I really think that an overhaul is needed, but we should discuss the big picture together.

I started a quick look, but I ran out of time.

One comment: I wouldn't see example 1 as a "what is dirty-cat?" example, because we are never going to be able to keep up with the broadening of the scope of dirty-cat. I would rather focus the message on one aspect of dirty-cat, here "representing non-normalizing categories"

examples/01_dirty_categories.py
LilianBoulard: I like this example, thanks!

examples/01_dirty_categories.py
```diff
@@ -1,274 +1,289 @@
...
 
-Here we look at a dataset on wages [#]_ where the column 'Employee
-Position Title' contains dirty categories. On such a column, standard
-categorical encodings leads to very high dimensions and can lose
-information on which categories are similar.
+..note:
+    A variable is considered categorical when it can take on one of a
+    limited, and usually fixed (discrete) number of possible values.
+    It is opposed to continuous variables and free text,
+    which can take an infinite number of unique values.
```
jovan-stojanovic, line 13:
```suggestion
..note:
    A variable is categorical when it takes a
    fixed and limited number of possible values, 
    as opposed to continuous variables and free text,
    which can take large or infinite number of unique values.
```
Trying to make the note shorter (as this is just on the side)

examples/01_dirty_categories.py
jovan-stojanovic: I agree that this example should be focusing on point 1 and 2 noted here, which are the features of the "early" versions of dirty-cat.

It is not that easy to link all the features of the package in one example, as they have also different logic (e.g. encoding vs deduplicating).

We have other examples for 3 (fuzzy joining) and 4 (deduplicating)

jovan-stojanovic (commented):
Trying to sum up the logic behind the examples:

encoding:
- 1: dirty-cat encoders for dirty categories
- 2: understanding these encoders
- 3: special case: datetime encoding

augmenting:
- 4: fuzzy joining for dirty tables
- 5: joining with ken embeddings example (I would move it in this order)

cleaning:
- 6: deduplicating dirty categories

So I would consider just inverting examples 5 and 6 so that it makes more sense.

LilianBoulard:
Hi, thanks for the feedback!

Since the migration, the structure has changed a bit and so did the examples.
The original logic was centered around the fact that, to me, when stumbling upon the website, it was not clear what dirty_cat did, and so having a complete example that links everything together was a no-brainer.

But now, thanks to the [user guide](http://skrub-data.org/stable/documentation.html), we can have exactly that in a more structure manner :)

Here's the updated structure:
- Examples are divided in three categories (the ones we advertise on the website):
  - Assembling
  - Encoding
  - Cleaning
- The example are more atomic - they each cover a specific aspect/feature of the library.
- For now, I left the main example that was supposed to unify everything (it's named `introducing_skrub`) because I spent a good amount of time trying to make the sections flow as fluently as possible, and I think the user guide could get inspired by some parts (cc @jovan-stojanovic, following #576).

There is still a fair bit of work to be done on the examples - I'll tackle that tomorrow. 
Meanwhile, I'm happy to receive feedback and suggestions, I think this is gonna be great :smile:

examples/assembling/creating_a_ken_embedding_subset.py
```diff
@@ -0,0 +1,20 @@
+"""
+.. _example_create_ken_embedding_subset:
+
+===============================
+Creating a KEN embedding subset
+===============================
+
+In this example, we will see how to use the full KEN embeddings to create
```
jovan-stojanovic, line 8: Hey, thanks for this. In the discussion #578, I was more thinking of adding this to the get_ken_embeddings function description http://skrub-data.org/stable/generated/skrub.datasets.get_ken_embeddings.html, and refer to this in the KEN example.

LilianBoulard: Thanks for the comment, I've answered over there :)

LilianBoulard:
Sorry for the mess with the diff, I'll try and fix that

LilianBoulard:
This PR has grown too big for its own good, and with all the changes to the examples that are taking place in the background, it's hard to maintain up to date.
Thus, I've broken it down into atomic PRs.

GaelVaroquaux:
> Thus, I've broken it down into atomic PRs.

I think that this was a good move.

PR #548, opened 2023-04-18
Don't continue fitting in `GapEncoder.transform`

### LINKED ISSUE #188: Topic labels that changes over multiple calls of get_feature_names in GapEncoder
This is not a problem in itself, but could be an unexpected behaviour. Calling `GapEncoder.get_feature_names()` several times in a row can in rares cases return different labels between calls. This is because `get_feature_names` calls `transform` to encode potential labels and decide which ones are the most relevant for each topic. However `transform` returns slightly different activations H over multiple calls on the same input X, because it keeps the fitted activations H from the previous call to continue the optimization, instead of starting from scratch every time.
These slightly different encodings H can lead to small changes in topic labels.

LilianBoulard:
Fixes #188

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -7,10 +7,14 @@ Release 0.4.1
 
 Major changes
 -------------
+* :func:`GapEncoder.transform` will not continue fitting of the instance anymore.
+  It makes functions that depend on it (:func:`~GapEncoder.get_feature_names_out`,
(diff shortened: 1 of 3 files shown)


jovan-stojanovic (approved):
This is as we discussed, seems fine to me for the minor release.
I think though we should double check the implications of fixing the activations H on the results later.

PR #553, opened 2023-05-04
Fix CircleCI artifacts not rendering

LilianBoulard:
I've noticed that since #545 was merged, the artifacts (doc) do not render.

This is necessary for reviewing #551.

I've noticed updates on the [github action repository](https://github.com/larsoner/circleci-artifacts-redirector-action), which might be the reason why it's not working anymore.


jovan-stojanovic:
This might work only after merging it, WDYT? @LilianBoulard

LilianBoulard:
I don't think so, but we might as well try! (can't be worse)

PR #563, opened 2023-05-16
new website

GaelVaroquaux:
A work in progress to design the new website


doc/conf.py
```diff
@@ -58,6 +58,7 @@
     "sphinx.ext.linkcode",
     "sphinx.ext.autodoc.typehints",
     "sphinx_gallery.gen_gallery",
+    "sphinx_favicon",
```
GaelVaroquaux, line 61: Do we need this extension? A favicon can be easily added in the jinja theme. One less extension is one less problem.

jovan-stojanovic: True, we should remove it.

build_tools/circle/build_doc.sh
```diff
@@ -123,6 +123,7 @@ conda create -n $CONDA_ENV_NAME --yes --quiet python="${PYTHON_VERSION:-*}" \
 source activate testenv
 pip install sphinx-gallery
 pip install sphinxext-opengraph
```
GaelVaroquaux, line 125: Indeed, clearly the setup.cfg extra-requires doc section is not used.

Maybe we should modify the present script to use it?

GaelVaroquaux: Something like
```
pip install -e .[doc]
```
should work, and could allow replacing the list above

GaelVaroquaux:
Let's try to get this building ASAP, and merge it, so as to have something to build upon in an iterative way

GaelVaroquaux:
Merging as the website is rendering. We can improve from this

PR #568, opened 2023-05-30
Improve the layout of the home page

Vincent-Maladiere:
- Add correct padding to the feature blocks
- Improve centering by shrinking the edit and source buttons area on the right panel
- Fix the contributing link


LilianBoulard (approved):
LGTM!

PR #570, opened 2023-05-30
Use modern issue templates

LilianBoulard:
Uses `yml` templates instead of markdown. Renders are way cleaner!

Current:
image

Proposed:
image


jovan-stojanovic (approved):
I agree we should try this. Looks great, thanks!

PR #582, opened 2023-06-09
Parallelize the `GapEncoder`

LilianBoulard:
Implements parallelization of the `GapEncoder` column-wise, as it is currently a simple for loop.
This could speed up the process significantly for wide datasets (intuitively, to be benchmarked).

### DIFF
--- a/skrub/_gap_encoder.py
+++ b/skrub/_gap_encoder.py
@@ -648,8 +648,7 @@ class GapEncoder(BaseEstimator, TransformerMixin):
         be denoted as `None`.
     n_jobs : int, optional
         The number of jobs to run in parallel.
-        The Gamma-Poisson factorization cannot be parallelized as it is an
-        iterative process, instead, the process is parallelized column-wise,
+        The process is parallelized column-wise,
         meaning each column is fitted in parallel. Thus, having
         `n_jobs` > X.shape[1] will not speed up the computation.
     verbose : int, default=0
--- a/skrub/_gap_encoder.py
+++ b/skrub/_gap_encoder.py
@@ -8,6 +8,7 @@ from typing import Dict, Generator, List, Literal, Optional, Tuple, Union
 
 import numpy as np
 import pandas as pd
+from joblib import Parallel, delayed
 from numpy.random import RandomState
 from scipy import sparse
 from sklearn import __version__ as sklearn_version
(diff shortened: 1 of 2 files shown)


LeoGrin (approved):

Vincent-Maladiere (commented):
Hey @LilianBoulard! We might want to chose `n_jobs` based on the number of columns and the number of available core.

The `_partition_estimator` function from sklearn could be insightful :) 

https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/ensemble/_base.py#L216-L226

GaelVaroquaux:
> We might want to chose n_jobs based on the number of columns and the number of available core.

We never set parallelism on by default, because it can too easily lead to overcommit (eg with nested objects that have parallelism).

We have progressively evolved toward using n_jobs=None, which opens the door to joblib to make clever choices.
Our long-term view is that library users should rely on joblib.parallel_config to set parallel computing. But this is new and still to be experiment.

@TomMoral is the expert on these questions.

Vincent-Maladiere:
> We never set parallelism on by default, because it can too easily lead to overcommit (eg with nested objects that have parallelism).

Ok, so what about the predict methods of RandomForest in sklearn?

https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/ensemble/_forest.py#L866

In case the user *does* set `n_jobs`, wouldn't it be good to provide some heuristics?

GaelVaroquaux:
> In case the user does set n_jobs, wouldn't it be good to provide some heuristics?

That's a good point. The code in RandomForest is there to create bunches of estimators. This is useful when there are many more iterations than CPUs. I do not think that it will be useful here.

tomMoral:
Aha not sure if I am an expert but here is my 2cts:

`joblib` is auto batching so it should already  be not too bad. But if you find instances on which auto batching does not work well, please report them so we can try to improve the heuristic for every `joblib` users 😁

Also note that if possible, it is best to not change the `n_jobs` too much in different parts of the pipeline as this might cause respawning/resizing the executor, which is slow for fast running examples. If you really have an heuristic that works well, it is better to keep `n_jobs` the same but create smart batches (except if you intend to play with lower level parallelism, but in this case, tell me more 🙃)

skrub/_gap_encoder.py
```diff
@@ -645,6 +646,14 @@ class GapEncoder(BaseEstimator, TransformerMixin):
         (default is to impute).
         In :func:`~GapEncoder.inverse_transform`, the missing categories will
         be denoted as `None`.
+    n_jobs : int, optional
+        The number of jobs to run in parallel.
+        The Gamma-Poisson factorization cannot be parallelized as it is an
+        iterative process, instead, the process is parallelized column-wise,
+        meaning each column is fitted in parallel. Thus, having
+        `n_jobs` > X.shape[1] will not speed up the computation.
```
GaelVaroquaux, line 654:
```suggestion
        The process is parallelized column-wise,
        meaning each column is fitted in parallel. Thus, having
        `n_jobs` > X.shape[1] will not speed up the computation.
```

GaelVaroquaux (approved):
LGTM. I made a tiny suggestion, but I will just merge it, and than merge this PR

PR #584, opened 2023-06-09
Fix X shape in minhash_encoder docstrings

LeoGrin:
the `MinHashEncoder` works for data with multiple columns, but the documentation doesn't reflect this.


GaelVaroquaux (commented):
LGTM. Merging

PR #587, opened 2023-06-10
fall back to pandas if no datetime format is found

### LINKED ISSUE #555: Improving DateTime detection?
In the following dataset:
https://github.com/GaelVaroquaux/AIMS_sklearn/blob/main/challenge/train.csv

TableVectorizer fails to detect that HiredDT is a date. I wonder if it's possible to easily improve that.

LeoGrin:
Fix #555 

We now fall back on `pandas.to_datetime` format inference when we don't find a format with `_infer_datetime_format`. Indeed, this function relies on pandas' `guess_datetime_format`, which doesn't work for weirder date format, and `pandas.to_datetime` uses `dateutil` when `guess_datetime_format` doesn't work.

The code to catch warnings could maybe be simplified.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,9 +39,10 @@ Major changes
 Minor changes
 -------------
 * Improvement of date column detection and date format inference in :class:`TableVectorizer`. The
-  format inference now finds a format which works for all non-missing values of the column, instead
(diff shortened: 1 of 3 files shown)


GaelVaroquaux (approved):
LGTM. Merging

PR #589, opened 2023-06-11
DOC: landing page is jinja, not rst

GaelVaroquaux:
A variant of the doc build where the landing page is done via a jinja file (hence writing html), rather than a rst file.

This should make the life of a web-designed much easier: more flexibility.

By editing the jinja + CSS (we can insert landing-page specific css, eg to make the body wider)


GaelVaroquaux:
This is unfinished: there is clearly some html + css work to be done. I don't think that I'll be able to finish it, but I hope that it helps people

GaelVaroquaux:
I also added transparency to the gradients so that they give the right level of brightness both in white background and black background. I had to use inkscape to choose these colors as it has a more advanced colorpicker.

GaelVaroquaux:
I've fixed some of the css. The cards remain not quite right

GaelVaroquaux:
I improved the CSS a bit. Stopping here for lack of time. I think that this is quite OK and could be merged.

The lower part of the landing page still needs love.

Vincent-Maladiere (approved):

PR #592, opened 2023-06-12
Table vec parallelism

LeoGrin:


skrub/_table_vectorizer.py
```diff
@@ -495,7 +496,7 @@ def replace_passthrough(name, trans, columns):
                         # selection is done with one dimension
                         cols = cols[0]
 
-                yield (name, trans, cols, get_weight(name))
+                yield (name, copy.deepcopy(trans), cols, get_weight(name))
```
GaelVaroquaux, line 499: Should this be a copy or a clone?

GaelVaroquaux:
> It works on the latest sklearn version (1.2.2), [...]Do we still want to support only recent sklearn versions?

+1 for supporting only sklearn 1.2.2 and later (can we try 1.2.1 ?).

Can someone update our requirements and CI build matrix? @LilianBoulard ?

LeoGrin:
>+1 for supporting only sklearn 1.2.2 and later (can we try 1.2.1 ?).
Cool ! It also works for 1.2.1

LeoGrin:
Something I don't really like is that the `transformers_` attribute is a bit weird for high-cardinality columns:
```
[('datetime', DatetimeEncoder(), ['date_of_stop', 'time_of_stop']),
 ('low_card_cat',
  OneHotEncoder(drop='if_binary', handle_unknown='ignore'),
  ['agency',
   'subagency',
   'accident',
   'belts',
    ...
   'race',
   'gender',
   'driver_state',
   'dl_state',
   'arrest_type']),
 ('high_card_cat', GapEncoder(n_components=30), ['seqid']),
 ('high_card_cat', GapEncoder(n_components=30), ['description']),
 ('high_card_cat', GapEncoder(n_components=30), ['location']),
 ('high_card_cat', GapEncoder(n_components=30), ['search_reason_for_stop']),
 ('high_card_cat', GapEncoder(n_components=30), ['make']),
 ('high_card_cat', GapEncoder(n_components=30), ['model']),
 ('high_card_cat', GapEncoder(n_components=30), ['charge']),
 ('high_card_cat', GapEncoder(n_components=30), ['driver_city']),
 ('high_card_cat', GapEncoder(n_components=30), ['geolocation']),
 ('remainder',
  'passthrough',
  ['latitude', 'longitude', 'year', 'contributed_to_accident'])]
  ```
  but I'm not sure we can avoid it with this solution.

**Comment by LeoGrin** [comment]
@glemaitre do you have thoughts on this? Do you think it could be worth implementing the parallelisation directly in sklearn?

**Comment by glemaitre** [comment]
I am not convinced that the strategy is beneficial for all types of transformers.

To give a concrete example, I think it would be faster to run a `StandardScaler` on a matrix of several columns instead of paralleling `StandardScaler` for each column of the matrix.

So the proposed strategy would make sense if internally to the transformer, there is already an explicit `for` loop. But in this case, I would expect to get a single instance and pass a matrix instead of having multiple instances for each column. If the parallelization is beneficial for this transformer, I would then expect the transformer to internally parallelize.

**Comment by LeoGrin** [comment]
Thanks for your answer @glemaitre !

>I am not convinced that the strategy is beneficial for all types of transformers.
To give a concrete example, I think it would be faster to run a StandardScaler on a matrix of several columns instead of paralleling StandardScaler for each column of the matrix.

Agree! The idea would be to have a list of transformer classes for which it is useful.

>So the proposed strategy would make sense if internally to the transformer, there is already an explicit for loop. But in this case, I would expect to get a single instance and pass a matrix instead of having multiple instances for each column. If the parallelization is beneficial for this transformer, I would then expect the transformer to internally parallelize.
I might be missing some details to grasp the reason to dispatch a column per transformer here.

But right now the transformers used inside the `ColumnTransformer` are not internally parallelized, right? My first instinct was indeed to parallelize these transformers, but it would create nested parallelism. To avoid the nesting, we discussed favoring the inner loop and removing the `ColumnTransformer`'s own parallelism, but it was deemed too surprising for the user*. This is how we ended up dispatching a column per transformer (see discussion in #586), but there may be better solutiond.

*Now that I think of it, it is indeed suprising to set the `TableVectorize`'s `n_jobs` to 1 to remove the inherited `ColumnTransformer` parallelism, but it may work if implemented directly on sklearn? Maybe having a option to either parallelize all transformers internally and running them sequencially, or runing all transformers in parallel, but not parallelize them internally.

**Comment by glemaitre** [comment]
Yep, I see the issue regarding nested parallelism and oversubscription. But I don't know if we should special case or instead better handle the nested parallelism in `joblib` (which is a far harder task to do).

Having the nested parallelism should be somehow manageable but require the user to set properly the `n_jobs` of the different transformers and there is not a good default that just makes things work smoothly. But I don't know if this is worth more than having some internal hackish way to go around the problem.

I just thought now that UX-wise, having multiple instances of the same transformer for each column will not be user-friendly when it comes to inspection. We could always rebuild a single instance but it smells hackish from far away (and then there is the same issue regarding the parallelism at `transform` time).

**Comment by LeoGrin** [comment]
>I just thought now that UX-wise, having multiple instances of the same transformer for each column will not be user-friendly when it comes to inspection. We could always rebuild a single instance but it smells hackish from far away (and then there is the same issue regarding the parallelism at transform time).

Yes I agree

> Having the nested parallelism should be somehow manageable but require the user to set properly the n_jobs of the different transformers and there is not a good default that just makes things work smoothly. But I don't know if this is worth more than having some internal hackish way to go around the problem.

If the user provides a `n_jobs` for the `TableVectorizer` but don't touch the default internal transformers, do you think there a way to set the `n_jobs` of the different transformers to make it work nicely? Or it is what you mean by "there is not a good default"?

**Comment by LeoGrin** [comment]
Following a discussion with @glemaitre, @jovan-stojanovic and @Vincent-Maladiere, what we do now is to split the transformers (if instances of `GapEncoder` or `MinHashEncoder`) before `fit` or `transform` and merge them back afterward. It is a bit hackish, but it solves the UX issue.

File: skrub/_gap_encoder.py [hunk 2]
```diff
@@ -651,6 +651,36 @@ class GapEncoder(BaseEstimator, TransformerMixin):
     fitted_models_: list[GapEncoderColumn]
     column_names_: list[str]
 
+    @classmethod
+    def _merge(cls, transformers_list):
```
LilianBoulard, line 655: What's the type of `transformers_list`? Here's my guess, please correct it if necessary:
```suggestion
    def _merge(cls, transformers_list: list[GapEncoder]):
```

skrub/_minhash_encoder.py
```diff
@@ -120,6 +120,42 @@ class MinHashEncoder(BaseEstimator, TransformerMixin):
 
     _capacity: int = 2**10
 
+    @classmethod
+    def _merge(cls, transformers_list):
```
LilianBoulard, line 124: Same here
```suggestion
    def _merge(cls, transformers_list: list[MinHashEncoder]):
```

skrub/_table_vectorizer.py
```diff
@@ -453,6 +457,107 @@ def _clone_transformers(self):
 
         # TODO: check that the provided transformers are valid
 
+    def _split_univariate_transformers(self, split_fitted=False):
```
LilianBoulard, line 460:
```suggestion
    def _split_univariate_transformers(self, split_fitted: bool = False):
```

skrub/_table_vectorizer.py
```diff
@@ -453,6 +457,107 @@ def _clone_transformers(self):
...
+            self.transformers = self._transformers_original
+        # merge self.transformers_
+        new_transformers_ = []
+        new_transformer_to_input_indices = {}
+        # find all base names
+        base_names = []
+        for name, _, _ in self.transformers_:
+            base_name = name.split("_split_")[0]
+            if base_name not in base_names:
+                base_names.append(base_name)
```
LilianBoulard, line 531: Could be simplified using sets
```suggestion
        base_names = {
            name.split("_split_")[0]
            for name, _, _ in self.transformers_
        }
```

skrub/_table_vectorizer.py
```diff
@@ -453,6 +457,107 @@ def _clone_transformers(self):
...
+        # merge all transformers with the same base name
+        for base_name in base_names:
+            transformers = []
+            columns = []
+            names = []
+            for name, trans, cols in self.transformers_:
+                if name.startswith(base_name):
+                    columns.extend(cols)
+                    transformers.append(trans)
+                    names.append(name)
```
LilianBoulard, line 541: Could also be simplified:
```suggestion
            transformers = []
            columns = []
            names = []
            for name, trans, cols in self.transformers_:
                names, transformers, nested_columns = zip(*self.transformers_)
                # Flatten list[list[str]] to list[str]
                columns = [column for column in columns for columns in nested_columns]
```

skrub/_table_vectorizer.py
```diff
@@ -453,6 +457,107 @@ def _clone_transformers(self):
...
+                    (base_name, transformers[0].__class__._merge(transformers), columns)
+                )
+                new_transformer_to_input_indices[base_name] = list(
+                    np.concatenate(
+                        [self._transformer_to_input_indices[name] for name in names]
+                    )
+                )
+
+        self.transformers_ = new_transformers_
+        self._transformer_to_input_indices = new_transformer_to_input_indices
```
LilianBoulard, line 559: Its type should be added to the class attributes (see lines 356-359 in main).

skrub/_table_vectorizer.py
```diff
@@ -678,6 +783,11 @@ def fit_transform(self, X: ArrayLike, y: ArrayLike = None) -> ArrayLike:
         if self.verbose:
             print(f"[TableVectorizer] Assigned transformers: {self.transformers}")
 
+        # split the univariate transformers on each column
+        # to be able to parallelize the encoding
+        if self.n_jobs not in (None, 1):
```
LilianBoulard, line 788: Maybe this condition could be exported to a method? That would make this more readable and expandable in the future.
Something like that:
```py
    @property
    def is_parallelized(self) -> bool:
        return self.n_jobs not in (None, 1)
```
and you would do
```suggestion
        if self.is_parallelized:
```

skrub/_utils.py
```diff
@@ -37,6 +37,14 @@ def __contains__(self, key: Hashable):
         return key in self.cache
 
 
+def combine_LRUDicts(capacity: int, *LRUDicts: LRUDict) -> LRUDict:
+    combined_LRUDict = LRUDict(capacity)
+    for lru_dict in LRUDicts:
+        for key, value in lru_dict.cache.items():
+            combined_LRUDict[key] = value
+    return combined_LRUDict
```
LilianBoulard, line 45: Small improvements:
```suggestion
def combine_LRUDicts(capacity: int, *lru_dicts: LRUDict) -> LRUDict:
    combined_lru_dict = LRUDict(capacity)
    for lru_dict in lru_dicts:
        for key, value in lru_dict.cache.items():
            combined_lru_dict[key] = value
    return combined_lru_dict
```
Also the way the function is named does not follow the snake_case convention, it might be better named `combine_lru_dicts`, what do you think?

skrub/tests/test_gap_encoder.py
```diff
@@ -264,3 +267,107 @@ def test_transform_deterministic() -> None:
...
+        # check all attributes
+        attrs = ["rho_"]
+        for attr in attrs:
+            assert getattr(enc_list[i], attr) == getattr(enc, attr)
+        assert enc_list[i].column_names_ == [f"col{i}"]
+
+
+def test_split_and_merge_transformers() -> None:
+    # check that splitting the transformer after fitting
+    # and then merging the transformers doesn't the same result
```
LilianBoulard, line 347:
```suggestion
    # and then merging the transformers doesn't change the result
```

skrub/tests/test_table_vectorizer.py
```diff
@@ -645,3 +647,148 @@ def test_mixed_types():
...
+        lambda: GapEncoder(n_components=2, random_state=0),
+        OneHotEncoder,
+        lambda: MinHashEncoder(n_components=2),
+    ]:
+        table_vec_no_parallel = TableVectorizer(
+            high_card_cat_transformer=high_card_cat_transformer(),
+            cardinality_threshold=4,
+        )
+        X_trans = table_vec_no_parallel.fit_transform(X)
+        for joblib_backend in ["loky", "threading", "multiprocessing"]:
```
LilianBoulard, line 689: We probably don't care about running it with each backend, I'd stick with loky, since it's essentially the same thing as multiprocessing, and threading is limited by the GIL.

LilianBoulard (commented):
Thanks for the feature!
I'm still a bit divided on the specific of the implementation of the `_split` and `_merge` methods, a second review and opinion would be appreciated :)

skrub/_table_vectorizer.py
LeoGrin: I think it's better to keep the same order, which the set doesn't guarantee

LilianBoulard: Indeed, an equivalent option would be
```suggestion
    base_names = set()
    return [
        name for name, _, _ in self.transformers_
        if not (name in base_names)
        else seen.add(x)
    ]
```
From https://stackoverflow.com/a/480227

LilianBoulard:
Looks like there was a syntax error in my proposition! (probably wrong indentation)

skrub/_table_vectorizer.py
```diff
@@ -458,6 +471,105 @@ def _clone_transformers(self):
...
+            self.transformers = self._transformers_original
+        # merge self.transformers_
+        new_transformers_ = []
+        new_transformer_to_input_indices = {}
+        # find all base names
+        base_names = []
+        for name, _, _ in self.transformers_:
+            base_name = name.split("_split_")[0]
+            if base_name not in base_names:
+                base_names.append(base_name)
```
Vincent-Maladiere, line 545: Nitpick:
```suggestion
        base_names = pd.unique(
            [name.split("_split_")[0] for name, _, _ in self.transformers_]
        )
```
Note that `pd.unique()` respect the order

skrub/_table_vectorizer.py
```diff
@@ -458,6 +471,105 @@ def _clone_transformers(self):
...
+            self._transformers_fitted_original = self.transformers_
+            new_transformers_ = []
+            new_transformer_to_input_indices = {}
+            for name, trans, cols in self.transformers_:
+                if (
+                    trans.__class__.__name__ in UNIVARIATE_TRANSFORMERS
+                    and len(cols) > 1
+                ):
+                    splitted_transformers_ = trans._split()
+                    assert len(splitted_transformers_) == len(cols)
```
Vincent-Maladiere, line 498: Ditto for `assert`

skrub/_table_vectorizer.py
```diff
@@ -458,6 +471,105 @@ def _clone_transformers(self):
...
+                        new_transformers_.append(
+                            (
+                                name + "_split_" + str(i),
+                                splitted_transformers_[i],
+                                [col],
+                            )
+                        )
+                        new_transformer_to_input_indices[name + "_split_" + str(i)] = [
+                            self._transformer_to_input_indices[name][i]
+                        ]
```
Vincent-Maladiere, line 509: Nitpick (this also has the benefit of checking for sizes between the 3 lists)
```suggestion
                    for i, (col, trans, trans_to_mapping) in enumerate(zip(
                        cols,
                        splitted_transformers_,
                        self._transformer_to_input_indices[name]
                    ):
                        name_split = f"{name}_split_{i}"
                        new_transformers_.append(
                            (name_split, trans, [col])
                        )
                        new_transformer_to_input_indices[name_split] = [trans_to_mapping]
```

skrub/_table_vectorizer.py
```diff
@@ -761,6 +873,11 @@ def fit_transform(self, X: ArrayLike, y: ArrayLike = None) -> ArrayLike:
         if self.verbose:
             print(f"[TableVectorizer] Assigned transformers: {self.transformers}")
 
+        # split the univariate transformers on each column
+        # to be able to parallelize the encoding
+        if self.is_parallelized:
+            self._split_univariate_transformers(split_fitted=False)
```
Vincent-Maladiere, line 879: This is great! 👍

skrub/_gap_encoder.py
```diff
@@ -645,6 +646,36 @@ class GapEncoder(TransformerMixin, BaseEstimator):
...
 
+    @classmethod
+    def _merge(cls, transformers_list: list[GapEncoder]):
+        # merge GapEncoder fitted on different columns
+        # into a single GapEncoder
+        # useful for parallelization in the TableVectorizer
+        full_transformer = clone(transformers_list[0])
+        # assert rho_ is the same for all transformers
+        rho_ = transformers_list[0].rho_
+        assert all([transformers.rho_ == rho_ for transformers in transformers_list])
```
Vincent-Maladiere, line 657: We prefer writing a clear error message instead of using `assert`. In which scenarios could we have `transformers.rho_ != rho_`?

If such scenarios don't exist, you can safely remove this assertion.

skrub/_minhash_encoder.py
```diff
@@ -120,6 +121,42 @@ class MinHashEncoder(TransformerMixin, BaseEstimator):
...
     _capacity: int = 2**10
 
+    @classmethod
+    def _merge(cls, transformers_list: list[MinHashEncoder]):
+        # merge MinHashEncoder fitted on different columns
+        # into a single MinHashEncoder
+        # useful for parallelization in the TableVectorizer
+        full_transformer = clone(transformers_list[0])
+        capacity = transformers_list[0]._capacity
+        assert all(
```
Vincent-Maladiere, line 131: ditto on `assert`

Vincent-Maladiere (commented):
Hey @LeoGrin, thanks for this PR. This feature looks promising! Here is a first (long) review :)

skrub/_gap_encoder.py
```diff
@@ -1,6 +1,7 @@
 """
 Implements the GapEncoder: a probabilistic encoder for categorical variables.
 """
+from __future__ import annotations
```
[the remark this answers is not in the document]
LeoGrin: It's because `GapEncoder` has a class method (`_merge(cls, transformers_list: list[GapEncoder])`) whose type hint reference the class itself.

skrub/_gap_encoder.py
LeoGrin: Yes I don't have any scenario in mind, it was mostly to double check. I'm removing it.

skrub/_table_vectorizer.py
```diff
@@ -458,6 +471,105 @@ def _clone_transformers(self):
...
+                        ]
+                else:
+                    new_transformers_.append((name, trans, cols))
+                    new_transformer_to_input_indices[
+                        name
+                    ] = self._transformer_to_input_indices[name]
+            self.transformers_ = new_transformers_
+            self._transformer_to_input_indices = new_transformer_to_input_indices
+        else:
+            self._transformers_original = self.transformers
```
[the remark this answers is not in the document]
LeoGrin: That's a good point! Do you think we should worry about this situation: the user fits the `TableVectorizer` a first time (successfully), then changes e.g `tb.low_card_cat_transformer`, then call `fit_transform` a second time. In this case, `_transformer_original` would not be updated with the new transformer. I can changing an attribute directly is not good practice, but the error would be really confusing to debug.

skrub/tests/test_table_vectorizer.py
```diff
@@ -756,3 +758,148 @@ def test_changing_types_int_float():
...
+        table_vec_one_col.fit(X[[col]])
+        assert table_vec_one_col.get_feature_names_out() == [
+            feat
+            for feat in table_vec_all_cols.get_feature_names_out()
+            if feat.startswith(col)
+        ]
+
+
+@skip_if_no_parallel
+def test_parallelism() -> None:
```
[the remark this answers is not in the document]
LeoGrin: I'm not sure what you mean, it seems that the other tests also have "-> None" as a type hint (maybe it)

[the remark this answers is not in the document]
LilianBoulard: This is just for type hint, it's not very important for tests return types since it's always `None`.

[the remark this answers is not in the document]
LeoGrin: Oh actually most test in skrub have these type hint, but not all of them in `test_table_vectorizer`. For the sake of consistency, I'll put a type hint in all tests here.

skrub/tests/test_table_vectorizer.py
```diff
@@ -756,3 +758,148 @@ def test_changing_types_int_float():
...
+                        table_vec.get_feature_names_out(),
+                        table_vec_no_parallel.get_feature_names_out(),
+                    )
+                    # assert that get_params gives the same result expect for n_jobs
+                    # remove n_jobs from the dict
+                    params = table_vec.get_params()
+                    params.pop("n_jobs")
+                    params_no_parallel = table_vec_no_parallel.get_params()
+                    params_no_parallel.pop("n_jobs")
+                    assert str(params) == str(params_no_parallel)
```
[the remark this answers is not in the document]
LeoGrin: The problem is not the dictionnary, but the transformer values:
```
E                 Differing items:
E                 {'low_card_cat': OneHotEncoder(drop='if_binary', handle_unknown='infrequent_if_exist')} != {'low_card_cat': OneHotEncoder(drop='if_binary', handle_unknown='infrequent_if_exist')}
E                 {'high_card_cat': MinHashEncoder(n_components=2)} != {'high_card_cat': MinHashEncoder(n_components=2)}
```
Is there a cleaner whay do to this?

skrub/tests/test_table_vectorizer.py
```diff
@@ -756,3 +758,148 @@ def test_changing_types_int_float():
...
+                    assert table_vec.n_jobs == n_jobs
+                    # assert that all attributes are equal except for
+                    # the n_jobs attribute
+                    for attr in [
+                        "transformers_",
+                        "columns_",
+                        "types_",
+                        "imputed_columns_",
+                    ]:
+                        assert str(getattr(table_vec, attr)) == str(
```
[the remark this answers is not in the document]
LilianBoulard: Yep, unless repeating the code, which might actually look better.
```py
assert table_vec.transformers_ == table_vec_no_parallel.transformers_
...
```
which might also make the error code more descriptive in case the assertion returns False

skrub/tests/test_table_vectorizer.py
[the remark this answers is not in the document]
LilianBoulard: Mmm, and it might also make it more reliable, since dictionaries can have different orders.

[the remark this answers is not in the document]
Vincent-Maladiere: @LilianBoulard, according to the linked thread, Pytest knows how to handle dictionaries, regardless of the order.
@LeoGrin, I see. Like Lilian said in another comment, maybe the best solution is to test elements individually. Otherwise, there is no solution apart from defining `__eq__` methods.

skrub/_table_vectorizer.py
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, good catch! I think we should worry about it because, as Gaël said, this can happen. 

One way to deal with it could be to define a `__setattr__` method where in case of any of our transformer attributes (e.g. `"low_card_cat_transformer"`) being passed, we could do:
```python
if hasattr(self, "_transformers_original"):
    delattr(self, "_transformers_original")
```
WDYT? I don't like playing with `__setattr__` or `__getattr__` much, so there's probably a better solution

LeoGrin:
Ah that's a good idea, but I'm not super comfortable either playing with
__setattr__. I think I can actually get rid of `_transformers_original` and
reconstruct it from the splitted transformers, as I do for the
`transformers_` attribute.

Vincent-Maladiere:
Yes, that's probably the way to go!

skrub/tests/test_table_vectorizer.py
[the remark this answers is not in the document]
LeoGrin: @Vincent-Maladiere I think the simple `str` solution works well in this case, as sklearn puts the hyperparameters in the string representation. For the `transformers_` attribute, I've made a few functions to check equality, but they may be a bit overengineered. WDYT?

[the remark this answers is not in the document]
Vincent-Maladiere: Ok, let's keep it and move on! We can still correct it later.

skrub/tests/test_table_vectorizer.py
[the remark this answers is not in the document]
Vincent-Maladiere: This needs to be addressed :)

skrub/_table_vectorizer.py
```diff
@@ -546,9 +545,31 @@ def _split_univariate_transformers(self, during_fit: bool = False):
...
+                    base_name,
+                    new_trans,
+                    [
+                        col
+                        for name, _, cols in self.transformers
+                        for col in cols
+                        if name.startswith(base_name)
+                    ],
+                )
+            )
```
Vincent-Maladiere, line 571: mando

skrub/tests/test_table_vectorizer.py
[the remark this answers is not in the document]
LeoGrin: I think it is, right?

skrub/_table_vectorizer.py
```diff
@@ -474,6 +483,126 @@ def _clone_transformers(self):
...
+                        name
+                    ] = self._transformer_to_input_indices[name]
+            self.transformers_ = new_transformers_
+            self._transformer_to_input_indices = new_transformer_to_input_indices
+
+    def _merge_univariate_transformers(self):
+        # merge back self.transformers and self.transformers_
+        check_is_fitted(self, attributes=["transformers_"])
+        # merge self.transformers
+        new_transformers = []
```
Vincent-Maladiere, line 551: To improve readability, I suggest using two subfunctions outside of the scope of the class here. These subfunctions would respectively have `self.transformers` and `self.transformers_` for inputs. WDYT?

Vincent-Maladiere (commented):
Hey @LeoGrin, here is a couple of additional remarks :)

Vincent-Maladiere (commented):
Hey @LeoGrin, here is a couple of additional remarks :)

skrub/_table_vectorizer.py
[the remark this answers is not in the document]
LeoGrin: @Vincent-Maladiere I thought a bit more about this, and I think the original code should have worked: in `fit_transform`, we're always reconstructing `self.transformers` from the attributes `low_card_cat_transformer` etc., before splitting, which means that even if the `ColumnTransformer` fails before we can merge, the following `fit_transform` shouldn't be affected. Still, I like the solution we ended up implementing (without the additional `_transformers_original` attribute). In the meantime, I added tests to check for the two failure modes we were discussing, they may be useful if the implementation changes in the future

skrub/_table_vectorizer.py
```diff
@@ -146,6 +146,76 @@ def _replace_missing_in_cat_col(ser: pd.Series, value: str = "missing") -> pd.Se
...
+    Assert whether we want to parallelize the transformer over
+    the columns or not. We only want to parallelize if the transformer
+    is "univariate" (i.e. it can be duplicated for each column).
+    """
+    return (
+        (not isinstance(trans, str))
+        and trans._get_tags().get("univariate", False)
+        and len(cols) > 1
+    )
+
```
Vincent-Maladiere, line 160: It seems we can combine `_merge_fitted_transformers` and `_merge_unfitted_transformers` routines:

```python
from itertools import chain

def _merge_transformers(
    transformers_ list[tuple[str, str | TransformerMixin, list[str]]],
    is_fitted: bool,
    transformer_to_input_indices: dict[str, list[int]],
) -> tuple[list[TransformerMixin], dict[str, list[int]] | None]:
    """Merge a list of split transformers.
    """
    new_transformers = []
    new_transformer_to_input_indices = {}
    base_names = pd.unique([name.split("_split_")[0] for name, _, _ in transformers_])
    
    for base_name in base_names:
        # merge all transformers with the same base name
        transformers, columns, names = [], [], []
        for name, trans, cols in transformers:
            if name.startswith(base_name):
                columns.extend(cols)
                transformers.append(trans)
                names.append(name)

        new_trans = transformers[0]
	if is_fitted:
            if isinstance(new_trans, TransformerMixin):
                new_trans = clone(new_trans)
	else:
            if len(transformers) > 1:
                # merge transformers
                new_trans = new_trans.__class__._merge(transformers)
            new_transformer_to_input_indices[base_name] = list(
                chain.from_iterable([
                    _transformer_to_input_indices[name] for name in names
                ])
            )
        new_transformers.append(base_name, new_trans, columns)

    return new_transformers, new_transformer_to_input_indices
```

and then call it like:
```python
    def _merge_univariate_transformers(self):
        # merge back self.transformers and self.transformers_
        check_is_fitted(self, attributes=["transformers_"])
        self.transformers = _merge_transformers(self.transformers, is_fitted=False)
        (
            self.transformers,
            self._transformer_to_input_indices,
        ) = _merge_transformers(
            self.transformers_, is_fitted=True, self._transformer_to_input_indices
        )
```

WDYT?

skrub/_table_vectorizer.py
```diff
@@ -474,6 +553,78 @@ def _clone_transformers(self):
...
+        Parameters
+        ----------
+        during_fit : bool, default=False
+            Whether the method is called during `fit_transform` (True) or
+            during `transform` (False). This is used to determine whether
+            to split the self.transformers_ attribute (when False) or the
+            self.transformers attribute (when True).
+        """
+        if during_fit:
```
Vincent-Maladiere, line 570: Nitpick: For consistency, it may be better to put the logic in a subfunction outside the scope of the class, the way you did with `_merge_transformers`. WDYT?

Vincent-Maladiere (commented):
We're almost done, here are some final comments

skrub/_minhash_encoder.py
```diff
@@ -120,6 +121,39 @@ class MinHashEncoder(TransformerMixin, BaseEstimator):
 
     _capacity: int = 2**10
 
+    @classmethod
+    def _merge(cls, transformers_list: list[MinHashEncoder]):
```
Vincent-Maladiere, line 125: Could you add a docstring to _merge and _split? Ditto for GapEncoder

Vincent-Maladiere (approved):
Thanks, @LeoGrin, that's a great feature! Let's add some missing docstring, then LGTM :)

Vincent-Maladiere:
You have conflicts with the main branch

skrub/_table_vectorizer.py
```diff
@@ -147,6 +148,140 @@ def _replace_missing_in_cat_col(ser: pd.Series, value: str = "missing") -> pd.Se
     return ser
 
 
+def _parallel_on_columns(trans: TransformerMixin, cols: list[str]):
```
Vincent-Maladiere, line 151:
```suggestion
def _parallel_on_columns(trans: TransformerMixin, cols: list[str]) -> bool:
```

skrub/_table_vectorizer.py
```diff
@@ -147,6 +148,140 @@ def _replace_missing_in_cat_col(ser: pd.Series, value: str = "missing") -> pd.Se
...
+        and trans._get_tags().get("univariate", False)
+        and len(cols) > 1
+    )
+
+
+def _split_transformers(
+    transformers: list[tuple[str, str | TransformerMixin, list[str]]],
+    transformers_to_input_indices: dict[str, list[int]] | None = None,
+    during_fit: bool = False,
+):
```
Vincent-Maladiere, line 168: missing return type

Vincent-Maladiere (commented):
some last docstring comments

jovan-stojanovic (approved):
Thanks, LGTM!

PR #594, opened 2023-06-12
`GapEncoder` hyperparameter tuning

LeoGrin:


LeoGrin:
@LilianBoulard

LilianBoulard:
Thanks, this will be useful!
An approach I've taken with #593 is, instead of copying the whole code of the class we're trying to experiment with, inherit it, modify only what's needed, and specify the current commit on `main` in the file's header, which make it much simpler to analyze :smile:

benchmarks/bench_gap_encoder.py
```diff
@@ -0,0 +1,1332 @@
...
+
+    _args = ArgumentParser(
+        description="Benchmark for the batch feature of the MinHashEncoder.",
+        parents=[default_parser],
+    ).parse_args()
+
+    if _args.run:
+        print("Running benchmark")
+        ds = fetch_traffic_violations()
+        df = benchmark(ds)
```
LilianBoulard, line 1326: Passing both a `parametrize` in `monitor` and the dataset in the call is not supported (check `monitor`'s docstring).
Just load the dataset inside the benchmark function, it is loaded from disk after being downloaded during the first fetch call.

LeoGrin:
Closing as it's no redundant with #680

PR #598, opened 2023-06-13
Benchmark improvements

LeoGrin:
- Random search
- Simple tests for the monitor function
- joblib support (I think this won't work for memory monitoring, but should work for time monitoring. Would love feedback on this).

@LilianBoulard this might be useful for the gapencoder benchmark
@jovan-stojanovic

### DIFF
--- a/benchmarks/tests/test_monitor.py
+++ b/benchmarks/tests/test_monitor.py
@@ -1,5 +1,7 @@
 import pytest
 from benchmarks.utils import monitor, repr_func  # replace with the actual import path
+from sklearn.utils._testing import skip_if_no_parallel
+import time
 
 
 def test_repr_func():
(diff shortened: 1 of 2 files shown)


benchmarks/utils/monitor.py
```diff
@@ -184,13 +199,18 @@ def exec_func(*args, **kwargs) -> Dict[str, List[float]]:
...
-                parametrization = (parametrize, ())
+                parametrization = (
+                    parametrize
+                    if n_random_search is None
+                    else random.sample(parametrize, n_random_search),
+                    (),
+                )
             else:
                 parametrization = list(product(parametrize))
+                if not (n_random_search is None):
```
LilianBoulard, line 210: Can be simplified:
```suggestion
                if n_random_search is not None:
```

LilianBoulard (commented):
Thanks for the PR :)

On the parallelism of runs, I originally set them to run sequentially so the time and memory monitoring would be accurate.
If there is a reliable way to measure those while running the instances in parallel, I'm all for it! Though I don't think the current implementation is. I changed the inner structure quite a bit in #593, it might be easier to integrate parallelism with these.

I like the random search idea, though we should also have a way to set the seed, for reproducibility.

GaelVaroquaux:
Can you:
- merge with master
- add a changelog entry

benchmarks/tests/test_monitor.py
```diff
@@ -0,0 +1,95 @@
+import pytest
```
GaelVaroquaux, line 1: Please add a short module-level docstring describing what this does

benchmarks/tests/test_monitor.py
```diff
@@ -0,0 +1,95 @@
...
+from benchmarks.utils import monitor, repr_func  # replace with the actual import path
```
GaelVaroquaux, line 2: Out of curiosity: is this supposed to work without installation? I'm a bit surprised that it does.

benchmarks/utils/monitor.py
```diff
@@ -1,11 +1,13 @@
 import tracemalloc
+import random
```
GaelVaroquaux, line 2: Can we have a small module-level docstring that describe what this file does?

GaelVaroquaux (changes requested):
A few comments. Happy to discuss IRL if useful

benchmarks/tests/test_monitor.py
LilianBoulard: Yes, it does work, since there is an `__init__` file in `benchmarks/` :)

PR #514, opened 2023-03-01
Clean signatures for cleaner calls

LilianBoulard:
Fixes #510

PR #600, opened 2023-06-13
FEAT Add the AggJoiner and AggTarget transformers

Vincent-Maladiere:
**What does this PR introduce?**

This draft proposes a first POC for the Join Aggregator. It aims at aggregating auxiliary tables before merging them on the base table, in a 1:N fashion.

Its API follows a consistent logic to the one of Feature Augmenter:

```python
agg_joiner = AggJoiner(
    tables=[
        (aux_1_large, "Country Name", ["GDP per capita (current US$)"]),
        (aux_2_large, "Country Name", ["Life expectancy at birth, total (years)", "country"])
    ],
    main_key="Country",
    agg_ops=["mean", "min", "max", "mode", "value_counts", "hist(4)"],
)
agg_joiner.fit_transform(df)
```

It currently supports Pandas DataFrames, Polars DataFrames, and Polars LazyFrames. Therefore can be run lazily with Polars! My idea is to preserve the dataframe module of the input to avoid confusion, and refuse dataframes that mix backends. The Polars dependency is of course optional.

My next step will be to benchmark the RAM consumption and time to run for these 3 different dataframes on bigger datasets.

**Edit**: [Here is a demo](https://www.kaggle.com/code/vincentmaladiere/h-m-recsys-feature-engineering-with-skrub-polars?scriptVersionId=133614432) of the Join Aggregator applied to feature engineering for RecSys within a Kaggle Competition!

---

**How is it implemented?**

The API of Join Aggregator is straightforward and relies on specialized implementations of an abstract `AssemblingEngine` class to handle both Pandas and Polars dataframes, namely `PandasAssemblingEngine` and `PolarsAssemblingEngine`.

In the absence of fully fledged tests, here is a working example::

```python
import numpy as np
import pandas as pd
from skrub.datasets import fetch_world_bank_indicator
from skrub._join_aggregator import JoinAggregator


main = pd.read_csv(
    "https://raw.githubusercontent.com/dirty-cat/datasets/master/data/Happiness_report_2022.csv",
    thousands=",",
)
main = main[["Country", "Happiness score"]]

aux_1 = fetch_world_bank_indicator(indicator_id="NY.GDP.PCAP.CD").X
aux_2 = fetch_world_bank_indicator("SP.DYN.LE00.IN").X

# Duplicate rows to create 1:N conditions
def augmente(df, id_col, val_cols, n_repeat=5):
    dfs = []
    for val_col in val_cols:
        for id, el in df[[id_col, val_col]].values:
            repeated = np.random.normal(el, scale=el/100, size=n_repeat)
            df_ = pd.DataFrame({id_col: id, val_col: repeated})
            dfs.append(df_)
    return pd.concat(dfs)

aux_1_large = augmente(aux_1, "Country Name", ["GDP per capita (current US$)"])
aux_2_large = augmente(aux_2, "Country Name", ["Life expectancy at birth, total (years)"])

# Add a categorical column, arbitrarily
aux_2_large["country"] = aux_2_large["Country Name"]


# Pandas
join_agg = JoinAggregator(
    tables=[
        (aux_1_large, ["Country Name"], ["GDP per capita (current US$)"]),
        (aux_2_large, ["Country Name"], ["Life expectancy at birth, total (years)", "country"])
    ],
    main_key="Country",
    agg_ops=["mean", "min", "max", "mode"],
)
pandas_out = join_agg.fit_transform(main)

# Polars, eager
import polars as pl

join_agg = JoinAggregator(
    tables=[
        (pl.DataFrame(aux_1_large), ["Country Name"], ["GDP per capita (current US$)"]),
        (pl.DataFrame(aux_2_large), ["Country Name"], ["Life expectancy at birth, total (years)", "country"])
    ],
    main_key="Country",
    agg_ops=["mean", "min", "max", "mode"],
)
polars_eager_out = join_agg.fit_transform(pl.DataFrame(main))

# Polars, lazy
join_agg = JoinAggregator(
    tables=[
        (pl.DataFrame(aux_1_large).lazy(), ["Country Name"], ["GDP per capita (current US$)"]),
        (pl.DataFrame(aux_2_large).lazy(), ["Country Name"], ["Life expectancy at birth, total (years)", "country"])
    ],
    main_key="Country",
    agg_ops=["mean", "min", "max", "mode"],
)
polars_lazy_out = join_agg.fit_transform(pl.DataFrame(main).lazy())

```

WDYT?

cc @strayMat and his super helpful aggregation implementations [here](https://gitlab.com/strayMat/caumim/-/blob/main/caumim/variables/aggregation.py) and [here](https://gitlab.com/strayMat/event2vec/-/blob/main/event2vec/event_transformer.py#L430) :)


skrub/_join_aggregator.py
```diff
@@ -0,0 +1,354 @@
...
+
+    def agg(self, table, cols_to_join, cols_to_agg, agg_ops):
+        
+        num_cols, categ_cols = pandas_split_num_categ_cols(
+            table[cols_to_agg]
+        )
+        num_ops, categ_ops = split_num_categ_ops(agg_ops)
+
+        num_stats = pandas_get_agg_ops(num_cols, num_ops)
+        categ_stats = pandas_get_agg_ops(categ_cols, categ_ops)
```
GaelVaroquaux, line 132: I'm surprise that the above work, as I see the same code applied for categorical columns and numerical ones, but the dictionary above seemed to map only to categorical operations.

skrub/_join_aggregator.py
```diff
@@ -0,0 +1,354 @@
...
+
+def polars_get_agg_ops(cols, agg_ops):
+    stats, mode_cols = [], []
+    for col, op_name in product(cols, agg_ops):
+        op_dict = {
+            "mean": pl.col(col).mean().alias(f"{col}_{op_name}"),
+            "std": pl.col(col).std().alias(f"{col}_{op_name}"),
+            "min": pl.col(col).min().alias(f"{col}_{op_name}"),
+            "max": pl.col(col).max().alias(f"{col}_{op_name}"),
+            "mode": pl.col(col).mode().alias(f"{col}_{op_name}"),
```
GaelVaroquaux, line 167: Are these computed lazily? It seems that we are instantiating many objects that we might not be using. I hope that the instantiation is cheap.

skrub/_join_aggregator.py
```diff
@@ -0,0 +1,354 @@
...
+        Aggregation operations to perform on the auxilliary table.
+        Options: {'mean', 'std', 'min', 'max', 'mode'}. If set to None, 
+        ['mean', 'mode'] will be used.
+    """
+
+    def __init__(self, tables, main_key, agg_ops=None):
+        self.tables = tables
+        self.main_key = main_key
+        self.agg_ops = agg_ops
+        self.assembly_engine = AssemblingEngine.get_for(tables)
```
GaelVaroquaux, line 282: I would be prefer if this was done in the fit, as it would be legit to change the tables attribute.

In general the scikit-learn guidelines are to do as little as possible in the init, which only serves to store the configuration of the object, and move as much as possible to the fit.

GaelVaroquaux (commented):
Thanks for the PR. This is exciting!!

I must confess that I am having a hard time reviewing this PR, as the github diff is ridden with warnings from codecov complaining of missing test coverage.

The user-level API feels right to me, although without using it, I don't have a perfect feeling of the user experience.

I think that we should try to find a consistent naming across the FeatureAugmenter and the JoinAggregator. This may call for renaming the FeatureAugmenter

We need to write tests, as these will help us feel if we have the right API for the internal components: if things are easy to test, it's a good sign.

It would be nice to have an exemple where we don't duplicate row to create the 1 to many relation. Maybe @jovan-stojanovic can help here. Ideally, this example should also show that the aggregation is beneficial for prediction, and we should add it to the PR so that we can comment on it.

# Points for later:

Actual support of polars will require us to support it in every single function and class of skrub (else the user will be confused). This will be a bit of work, in particular it will require define impedence matching / adaptation to many functionality of pandas / polars.

I wonder if the pre-aggregation strategy is the good one. If my external table has many more entries on the common key than the main table, preaggregation will lead me to compute many aggregates that I don't need. We should note this and potentially adress it later.

skrub/_join_aggregator.py
Vincent-Maladiere: We're simply defining operators here, no computation happens, lazy or not. We give these to a table, lazy or not, and let Polars do the heavy lifting for us.

skrub/_join_aggregator.py
```diff
@@ -0,0 +1,354 @@
...
+        elif op_name in CATEG_OPS:
+            categ_ops.append(op_name)
+        else:
+            ValueError(
+                f"'ops' options are {ALL_OPS}, got: {op_name}."
+            )
+    return num_ops, categ_ops
+
+
+class AssemblingEngine:
```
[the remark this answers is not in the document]
Vincent-Maladiere: Hmm, what would this Dispatcher do if, according to your suggestion above, we replace `get_for` with a function?

skrub/_join_aggregator.py
Vincent-Maladiere: I had in mind [this design pattern from ArgKmin in scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/metrics/_pairwise_distances_reduction/_argkmin.pyx.tp#L343), but point taken let's change this

skrub/_join_aggregator.py
Vincent-Maladiere: Yes, the beauty of `pd.NamedAgg(col, op)` is being able to use `"mean"`, `np.mean` or even `pd.Series.mode` for `op` :)
Therefore, we can use most `agg_ops` as strings. `"mode"`, however, is a little bit trickier, hence the mapping.

skrub/_join_aggregator.py
```diff
@@ -0,0 +1,354 @@
...
+def pandas_split_num_categ_cols(table):
+    num_cols = table.select_dtypes("number").columns
+    categ_cols = table.select_dtypes(
+        ["object", "string", "category"]
+    ).columns
+    return num_cols, categ_cols
+    
+
+def pandas_get_agg_ops(cols, agg_ops):
+    pandas_ops_mapping = {
```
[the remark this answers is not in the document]
Vincent-Maladiere: To stay consistent, we'd like to also inline the `{pandas, polars}_get_agg_ops` and `{pandas, polars}_split_num_categ_cols` functions. 

However, some of these functions are long and might clutter the logic of the calling method. 

Maybe we should define these functions as `staticmethod`? So that things are close to each other.

WDYT?

GaelVaroquaux:
> Maybe we should define these functions as staticmethod? So that things are close to each other.

No, staticmethods are to be used only when they are important for inheritance reasons. Classes are not to be confused with modules. Classes serve to define inheritance. Objects serve to associate functions to data (attributes). Modules serve to group code (symbols) together.

Vincent-Maladiere:
Thanks for this detailed feedback!

> I must confess that I am having a hard time reviewing this PR, as the github diff is ridden with warnings from codecov complaining of missing test coverage.

Whoops sorry I didn't notice. Adding `skip-ci` in my commit messages should help here.

> I think that we should try to find a consistent naming across the FeatureAugmenter and the JoinAggregator. This may call for renaming the FeatureAugmenter

Yes, that sounds good. Something like "FuzzyJoiner"?

> We need to write tests, as these will help us feel if we have the right API for the internal components: if things are easy to test, it's a good sign.

Yes, that's next on my todo :)

> It would be nice to have an example where we don't duplicate rows to create the 1-to-many relation. 

I just made a demo on Kaggle, the base table is huge (30M rows) and needs Polars lazy mode :)

When using models like LambdaMart or BoostingTrees for Recommender Systems, you always end up aggregating the base table `| user_id | product_id | timestamp |` by `user_id` and also by `product_id` before joining these two aggregated tables back to the main one: this is an ideal and very useful application for the Join Aggregator!

Please have a look at it:

https://www.kaggle.com/code/vincentmaladiere/h-m-recsys-feature-engineering-with-skrub-polars

Of course, we also need a proper example in the documentation, as you mentioned. Why not use a lighter version of MovieLens and perform recommendations in our documentation? :)

### Points for later:

> Actual support of polars will require us to support it in every single function and class of skrub (else the user will be confused). This will be a bit of work, in particular it will require define impedence matching / adaptation to many functionality of pandas / polars.

Yes, how do you rank it as a priority? There is so much value in actually offering this, I'd be thrilled to start thinking about it very soon.

> I wonder if the pre-aggregation strategy is the good one. If my external table has many more entries on the common key than the main table, preaggregation will lead me to compute many aggregates that I don't need. We should note this and potentially address it later.

Very good point. We could filter our tables before aggregation on the right keys during `fit`!

Vincent-Maladiere:
> No, staticmethods are to be used only when they are important for inheritance reasons. Classes are not to be confused with modules. Classes serve to define inheritance. Objects serve to associate functions to data (attributes). Modules serve to group code (symbols) together.

What are you suggesting then?

Vincent-Maladiere:
I've just noticed that our TableVectorizer already handle polars input héhé (not lazily though)

```python
import polars as pl

main = pd.read_csv(
    "https://raw.githubusercontent.com/dirty-cat/datasets/master/data/Happiness_report_2022.csv",
    thousands=",",
)

tv = TableVectorizer()
tv.fit_transform(
    pl.DataFrame(main)
)
```

This is thanks to a fortunate cast to pandas during the `fit_transform`
```python
        # Convert to pandas DataFrame if not already.
        if not isinstance(X, pd.DataFrame):
            X = pd.DataFrame(X)
        else:
            # Create a copy to avoid altering the original data.
            X = X.copy()
```

A pleasant surprise 😄

GaelVaroquaux:
> Yes, that sounds good. Something like "FuzzyJoiner"?

Maybe just "Joiner"? I think that we will probably want to add fuzziness to almost everything that we do, so we shouldn't start putting it in every name.

> I just made a demo on Kaggle, the base table is huge (30M rows) and needs Polars lazy mode :)

> https://www.kaggle.com/code/vincentmaladiere/h-m-recsys-feature-engineering-with-skrub-polars

Nice! Great demonstration of the vision for skrub, this is material for an absolutely great example. Any chance that we can make this light and fast enough to be in our documentation? I would love to have it in our main docs, and to progressively build our features and API to keep making this example simpler and simpler.

> Yes, how do you rank it as a priority? There is so much value in actually
> offering this, I'd be thrilled to start thinking about it very soon.

Let's merge the present PR first :). After this, why not :).

GaelVaroquaux:
> What are you suggesting then?

Plain functions for now, and long term maybe two modules

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+col_trans = make_column_transformer(
+    ("passthrough", ["userId", "movieId", "rating"]),
+    (OneHotEncoder(sparse_output=False), ["day_name"]),
+    remainder="drop",
+    verbose_feature_names_out=False,
+)
+
+train = col_trans.fit_transform(train)
+train = pd.DataFrame(train, columns=col_trans.get_feature_names_out())
+train.head()
```
GaelVaroquaux, line 163: In recent scikit-learn, you can probably replace the 3 lines above by something like (not tested):
```suggestion
train = col_trans.set_output(transform="pandas").fit_transform(train)
train.head()
```

skrub/_join_aggregator.py
```diff
@@ -0,0 +1,405 @@
...
+        specifying the auxiliary dataframes and their columns for joining
+        and aggregation operations.
+
+        dataframe : :obj:`~pandas.DataFrame` or str
+            The auxiliary data to aggregate and join.
+            'X' can be provided as a string to perform self-aggregation on
+            the input data.
+
+        columns_to_join : str or array-like
+            Select the columns from the dataframe to use as keys during
```
GaelVaroquaux, line 227: Here you are talking about the column in the base table, no?

skrub/_join_aggregator.py
```diff
@@ -0,0 +1,405 @@
...
+            the join operation.
+
+        columns_to_agg : str or array-like
+            Select the columns from the dataframe to use as values during
```
GaelVaroquaux, line 231: While here this is the columns on the external table, no?

skrub/_join_aggregator.py
GaelVaroquaux: Ah, ok, I understand: this is the "one to many" key.

We are going to have to work on our API to make it readable / understandable as right now even someone who knows well both pandas and scikit-learn does not easily guess what is going on. I don't have a good suggestion right now, though. Will have to think about it.

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+# ***Join aggregator on `movieId` and `userId`***
+
+# In[11]:
+
+
+from skrub import JoinAggregator
+
+cols_day_name = list(set(train.columns) - set(["userId", "movieId"]))
+
+join_agg_rating = JoinAggregator(
```
GaelVaroquaux, line 175: Should we simplify the API by accepting only one external table?

(open question: I'm shuffling ideas around to make things more readable.

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+all_genres
+
+
+# In[14]:
+
+
+from sklearn.feature_extraction.text import CountVectorizer
+from skrub import MinHashEncoder
+
+vectorizer = CountVectorizer(
```
GaelVaroquaux, line 219: How does the GapEncoder work on the genres?

It might discover some of the genres by itself.

I'm trying to see everything that can make this example easier to read, because right now it is a bit challenging.

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+    vocabulary=all_genres,
+    ngram_range=(1, 1),
+    lowercase=False,
+)
+
+n_components = 10
+min_hash_encoder = MinHashEncoder(n_components=n_components)
+
+movie_transformer = make_column_transformer(
```
GaelVaroquaux, line 228: Once our TableVectorizer works well, we should be able to replace all this by the TableVectorizer (this should be a goal for skrub

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+# In[17]:
+
+
+def get_year_and_day_name(ratings):
+    timestamps = pd.to_datetime(ratings["timestamp"], unit="s")
+    day_names = timestamps.dt.day_name()
+    return pd.get_dummies(day_names)
+
+
+pipe_ohe_days = make_pipeline(
```
GaelVaroquaux, line 281: Could we get away with a DatetimeEncoder here? (if not, could we improve the DatetimeEncoder so that it works here?)

Trying to simplify this notebook as much as I can

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+from sklearn.preprocessing import OneHotEncoder
+
+
+def get_day_name(ratings):
+    timestamps = pd.to_datetime(ratings["timestamp"], unit="s")
+    day_names = timestamps.dt.day_name()
+    return day_names.to_frame()
+
+
+ft_get_day_name = FunctionTransformer(get_day_name)
```
GaelVaroquaux, line 298: Same question about the use of a DatetimeEncoder

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+
+
+# In[2]:
+
+
+import pandas as pd
+
+ratings["timestamp"] = pd.to_datetime(ratings["timestamp"], unit="s")
+ratings["year"] = ratings["timestamp"].dt.year
+ratings["day_name"] = ratings["timestamp"].dt.day_name()
```
jovan-stojanovic, line 29: I think the DatetimeEncoder does this once you give the `timestamp` column to the TableVectorizer

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+
+
+daily_volume = (
+    ratings["day_index"].value_counts().sort_index().to_frame().join(day_name_index)
+)
+
+make_barplot(
+    x=daily_volume["day_name"], y=daily_volume["count"], title="Daily volume of ratings"
+)
+
```
jovan-stojanovic, line 92: I think having the previous plot on the years is much more interesting and revealing time/n_ratings relationship than this one on individual days. For simplification and shortening purpose, I think removing this and going directly the next plot would not hurt.

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+    user_count_ratings["n_ratings"].max(),
+)
+
+ax = sns.histplot(user_count_ratings["n_ratings"])
+ax.set(
+    xlabel="Number of ratings given by users",
+    ylabel="Total number of users",
+    title="Number of ratings given by user",
+)
+
```
jovan-stojanovic, line 114: I think you need to add this so that it gets displayed in the example
```suggestion
plt.show()
```

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,438 @@
...
+        x=x,
+        y=y,
+        palette=cmap(norm(y)),
+    )
+    plt.xticks(rotation=30)
+    plt.xlabel(None)
+    plt.ylabel(None)
+    plt.tight_layout()
+    plt.title(title)
+
```
jovan-stojanovic, line 57: I think you need to add this so that the plot gets rendered in the example
```suggestion
    plt.show()
```

examples/07_join_aggregation.py
jovan-stojanovic: I find it hard to understand as well, I think it would be good to simplify the syntax if possible. Just like we did for the fuzzy_join, it would be interesting to ask around what users think is easier (maybe during the sprint?).

jovan-stojanovic: Also, for this first join-aggregation in the example, maybe it's best to do something simpler? For instance not to do both mean and sum or on multiple tables, just for better illustration purpose.

The dropping columns part below would also then disappear for now.

jovan-stojanovic (commented):
Thanks for this great new feature @Vincent-Maladiere! I have a few comments on the example.

examples/07_join_aggregation.py
jovan-stojanovic: And even before, you can test it directly using 

```python
TableVectorizer(high_card_cat_transformer= min_hash_encoder = MinHashEncoder(n_components=n_components))
```
From previous benchmarks, I don't think the `CountVectorizer` in itself does a good job here.

skrub/_join_aggregator.py
```diff
@@ -0,0 +1,405 @@
...
+        If set to None, we will use the table index, by looking at their
+        order in 'tables', e.g. for a duplicate columns: price, price_1, price_2.
+
+    See Also
+    --------
+    :class:`FeatureAugmenter` :
+        Augment a main table by automatically joining multiple
+        auxiliary tables on it.
+    """
+
```
jovan-stojanovic, line 257: It would be great to add here a small reproducible example with some generated data, so that people see the basic syntax to use when join-aggregating

rcap107:
Hello! This looks like a great addition! I was able to run it almost out of the box in my code, which is really convenient. 

Here are some of the potential features me and @Vincent-Maladiere discussed and that could/should be added to the JoinAggregator (some of them are inspired by the [METAM](https://arxiv.org/abs/2304.09068v1) paper):

- Rather than joining the entire table, it should be possible to join one column at a time, or only a subset of columns. 
- The selection of which columns should be used for joining should be done in a "screening" step, which measures various utility metrics between the left table and the candidate (right) table. 
  - Potential utility metrics include Mutual Information, semantic similarity, syntactic similarity, correlation
- It should be possible to decide (or let the user decide) what columns in the right table should be aggregated, and what to do with those that are not in that set (e.g., drop the extra columns, apply a simpler aggregation etc.) 
- The `screen` step should be executed in the `fit` and it should be possible to return the utility metrics for each join column as an additional return value (not sure how to implement this properly).
- If the aggregation is done in a train-test sequence of operations, the `screen` step should not be done when `JoinAggregator` is called from the "test" step because it might decide to drop columns that are used by the trained model. 
- Given the utility metrics in the `screen` function, there should be a threshold that is measured by combining the utility metrics measured within. This threshold should be used to drop columns whose "score" is lower than it, and ideally it should be possible to modify it with supervision. 

I'm aware that most of these points will be mid- to long-term, but I still think it's worth having them here so we can keep track. @GaelVaroquaux what do you think?

rcap107:
Some additional comments:

**On screening:**
- If we go with supervised metrics that rely on `y` (the target column in the left table), then we need to perform a join during the `fit`, since the screening that require those metrics is done at `fit` time. This join can be expensive, but since the screening is done one table at a time and we can drop the entire left table (with the exception of the left keys and `y`), it should still be manageable. 
- There should also be unsupervised screening metrics that drop columns based e.g., on the fraction of null values or the variance (`VarianceThreshold` from `sklearn.feature_selection`). Dropping "bad" columns can be done before executing the aggregation, which is another optimization. 

**On storing the info needed for the join**
- Some of the information is provided at init time (left table, left join key(s), right tables + keys, agg. ops, screening parameters), some is obtained during the operation (aggregated columns, screening results+metrics). 
- Keeping everything in memory can be expensive, especially for large joins. It will probably be necessary to store some of the data on disk and work out of core. 
- Lazy operations with polars will probably be an important feature.

Vincent-Maladiere:
This PR is now ready for review.

- It implements both `AggJoiner` and `AggTarget` with Pandas and partial support for Polars dataframes (aggregation operations limited to min, max, mode, mean, and sum for the latter).
- Reindexing operations like `histogram` and `value_counts` for Polars will be implemented in another PR.
- Lazyframes from Polars are currently not supported and will be the subject of another PR.
- The example has been reworked and lightened to showcase the feature engineering part. 
- Ideas for new aggregate functions can be found in this list: https://hackmd.io/2WFuejwNSj2FZ_pknkafWw.

Thanks to @GaelVaroquaux, @rcap107 and @jovan-stojanovic, that helped pushing this work forward :)

jovan-stojanovic (changes requested):
A few small comments for now :)

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,335 @@
...
+feature engineering in a simple way using |AggJoiner| and |AggTarget|.
+Note that our performance is higher than the baseline of using the mean
+rating per movies.
+
+
+.. |AggJoiner| replace::
+     :class:`~skrub.AggJoiner`
+
+.. |TargetJoiner| replace::
+     :class:`~skrub.TargetJoiner`
```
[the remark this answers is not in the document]
Vincent-Maladiere: Good eyes, thanks!

jovan-stojanovic (approved):

LilianBoulard:
Hi, thanks a lot for the impressive work, I'll try to review the PR this afternoon or tomorrow!

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+            Used to compute the correlation between the generated covariates
+            and the target for screening purposes.
+
+        Returns
+        -------
+        AggJoiner
+            Fitted :class:`AggJoiner` instance (self).
+        """
+        self.check_input(X)
+        agg_px, _ = get_namespace([table for (table, _, _) in self.tables_])
```
GaelVaroquaux, line 237: agg_px feels like a suboptimal name for this object: it implements much more than aggregation (for instance agg_px.aggregate, as below, seems strange)

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+        ----------
+        X : dataframe
+            The raw input to check.
+        """
+        if not hasattr(X, "__dataframe__"):
+            raise TypeError(f"X must be a dataframe, got {type(X)}.")
+
+        # Check main_keys
+        # XXX: Do we want to enable main_keys with multiple columns?
+        # e.g. [["user_id", "movie_id"], ["user_id"]]
```
GaelVaroquaux, line 311:
```suggestion
```
As we are not supporting this in this PR or in the immediate future, let's remove the XXX. We can always come back to it later.

skrub/_agg_pandas.py
```diff
@@ -0,0 +1,142 @@
...
+
+
+def aggregate(
+    table,
+    cols_to_join,
+    cols_to_agg,
+    num_ops,
+    categ_ops,
+    suffix,
+):
```
GaelVaroquaux, line 15: We need a fairly complete docstring on this: future developers will be trying to use it and need to understand all arguments.

Also: can any of these arguments take default values (become keyword arguments), to facilitate usage?

skrub/_agg_pandas.py
```diff
@@ -0,0 +1,142 @@
...
+    base_group.columns = [
+        f"{col}{suffix}" if col not in cols_to_join else col
+        for col in base_group.columns
+    ]
+    sorted_cols = sorted(base_group.columns)
+
+    return base_group[sorted_cols]
+
+
+def join(left, right, left_on, right_on):
```
GaelVaroquaux, line 61: Need docstring

skrub/_agg_pandas.py
```diff
@@ -0,0 +1,142 @@
...
+    return left.merge(
+        right,
+        how="left",
+        left_on=left_on,
+        right_on=right_on,
+    )
+
+
+def get_agg_ops(table, cols, agg_ops):
```
GaelVaroquaux, line 70: Need docstring

skrub/_agg_pandas.py
```diff
@@ -0,0 +1,142 @@
...
+
+def _parse_argument(op_name):
+    """Split a text input into a function name and its argument.
+
+    Examples
+    --------
+    >>> _parse_argument("hist(10)")
+    "hist", 10
+    >>> _parse_argument("hist([2, 3])")
+    "hist", [2, 3]
```
GaelVaroquaux, line 96: The fact that this is valid needs to be added to the documentation of which arguments are valid

GaelVaroquaux (commented):
Great job!

A first pass of comments.

We will also need to add polars to our build matrix.

I'll continue my review when I find some time, but this already gives some pointers.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+        self.main_key = main_key
+        self.agg_ops = agg_ops
+        self.suffixes = suffixes
+
+    def fit(self, X, y=None) -> "AggJoiner":
+        """Aggregate auxiliary tables based on the main keys.
+
+        Parameters
+        ----------
+        X : {pandas.Dataframe}
```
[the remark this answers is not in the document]
LilianBoulard: Also, check out the comments I made on the `DataFrameLike` API in my review, they're relevant :)

skrub/_agg_joiner.py
LilianBoulard: And what does `px` stand for? This needs to be more descriptive.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+        Parameters
+        ----------
+        X : dataframe
+            Must contains the columns names defined in ``main_key``.
+        y : dataframe or array-like
+            Must have the length of ``X``, with matching indices.
+            The target can be continuous or discrete, with multiple columns.
+
+            If the target is continuous, only numerical operations,
+            listed in NUM_OPS, are applied.
```
LilianBoulard, line 472:
```suggestion
            listed in NUM_OPS, can be applied.
```
Maybe? The way it's currently worded feels like all operations of `NUM_OPS` will be applied.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+        userId  movieId  rating   genre genre_mode_user  rating_mean_user  rating_mean_movie
+    0       1        1     4.0   drama           drama               4.0   3.5
+    1       1        3     4.0   drama           drama               4.0   3.0
+    2       1        6     4.0  comedy           drama               4.0   4.0
+    3       2        1     3.0      sf              sf               3.0   3.5
+    4       2        3     2.0  comedy              sf               3.0   3.0
+    5       2        6     4.0      sf              sf               3.0   4.0
+    """  # noqa: E501
+
+    def __init__(self, tables, main_key, agg_ops=None, suffixes=None):
```
LilianBoulard, line 213: Using keyword arguments could make readability way better:
```suggestion
    def __init__(self, *, tables, main_key, agg_ops=None, suffixes=None):
```

skrub/_agg_pandas.py
LilianBoulard, line 1: A few thoughts about the implementation:
1. Since we have a few files that are of the same utility, they could have their own folder, it would make the root directory more organized. This is a general comment by the way, we could also move the `SimilarityEncoder`, `GapEncoder`, etc. in their own folder "encoding". Note: this does not have to change the way we import them, we would just have to tweak the `__init__` files.
2. I think we need to use appropriate design patterns, the code is too complex and coupled with each implementation. It would greatly benefit from something like an [abstract class](https://refactoring.guru/design-patterns/abstract-factory).

This also applies to `_agg_polars.py`.

skrub/datasets/_fetching.py
```diff
@@ -430,6 +432,70 @@ def _fetch_figshare(
             raise URLError("No internet connection or the website is down.")
 
 
+def _fetch_movielens(dataset_id: str, data_directory: Path | None = None) -> dict[str]:
```
LilianBoulard, line 435: Without getting into the specifics, you might benefit from having a look at #669 even before it is merged, it would make the adaptation simpler :)

skrub/tests/test_agg_joiner.py
```diff
@@ -0,0 +1,429 @@
...
+        suffixes=["_user", "_movie"],
+        agg_ops=["mean", "mode"],
+    )
+    X_string_user_movie = agg_join.fit_transform(X)
+
+    assert_frame_equal_(main_user_movie, X_string_user_movie)
+
+
+def test_polars_unavailable_ops():
+    if POLARS_SETUP:
```
LilianBoulard, line 75: Maybe this could be `pytest.skip`ed instead?

skrub/tests/test_agg_joiner.py
```diff
@@ -0,0 +1,429 @@
...
+def test_agg_t_check_input():
+    agg_target = AggTarget(
+        main_key=["userId"],
+        suffixes=["_user"],
+    )
+    msg = (
+        "Only Pandas or Polars dataframes are currently supported, "
+        "got [<class 'numpy.ndarray'>]."
+    )
+    with pytest.raises(TypeError, match=re.escape(msg)):
```
LilianBoulard, line 404: When testing for error messages, it is often better to check for specific keywords instead of the full sentence.
This is because if it slightly changes (typo, rewording), the test doesn't pass anymore.
For this specific case, it could be:
```suggestion
    # First keyword checks it's an issue of "supported",
    # second one checks that the type of the passed object
    # is explicitly mentioned in the error, which makes it more
    # descriptive, thus easier to debug.
    with pytest.raises(TypeError, match=r"supported|ndarray"):
```
Note: I did not check if it runs

LilianBoulard (commented):
Thanks again for the PR, this is a big one, took me a while to review :)

A lot of the comments I have made below apply to multiple instances, but I've submitted not submitted them all to keep it short.
One more thing: please use type hints! For everything but especially for the public API!

Also, following my comments on #628, I dislike the abbreviations. Now to justify, here's a snippet (from the docstring example) with the current naming:
```py
AggJoiner(
    tables=[
        (main, "userId", ["rating", "genre"]),
        (main, "movieId", ["rating"]),
    ],
    main_key=["userId", "movieId"],
    suffixes=["_user", "_movie"],
    agg_ops=["mean", "mode"],
)
```
and the one I argue for:
```py
AggregationJoiner(
    tables=[
        (main, "userId", ["rating", "genre"]),
        (main, "movieId", ["rating"]),
    ],
    main_key=["userId", "movieId"],
    suffixes=["_user", "_movie"],
    operations=["mean", "mode"],
)
```
- It reads better
- there is no ambiguity on whether `agg_ops` stands for "aggregation options" or "aggregation operations"
- IMO, there is no need to repeat "agg"/"aggregation" in both the name of the class and the parameter.

Finally, one proposal to make the aggregation code simpler would be to use a class with static attributes. It could probably be an `enum.Enum`, although the specifics are to be discussed later if you feel this is a good workflow:

On the surface, it replaces
```py
AggJoiner(
    tables=[
        (main, "userId", ["rating", "genre"]),
        (main, "movieId", ["rating"]),
    ],
    main_key=["userId", "movieId"],
    suffixes=["_user", "_movie"],
    agg_ops=["mean", "mode"],
)
```

with
```py
from skrub.aggregation import Operations

AggJoiner(
    tables=[
        (main, "userId", ["rating", "genre"]),
        (main, "movieId", ["rating"]),
    ],
    main_key=["userId", "movieId"],
    suffixes=["_user", "_movie"],
    agg_ops=[Operations.mean, Operations.mode],
)
```

| Pros | Cons |
| --- | --- |
| Gets rid of the string parsing logic | Requires the class to be imported, or interfaced within the aggregator |
| Simplifies the implementation | Although not a rare pattern, it's not quite as common as simple strings |
| Is less error-prone than strings | Makes the code slightly longer (although it could be named `Ops` instead of `Operations`, but you know my opinion on that :p) |

Linked to that is my comment on the file `_agg_pandas.py`, which I think is very important.

GaelVaroquaux:
>   • there is no ambiguity on whether agg_ops stands for "aggregation options"

I do prefer "operations" to "agg_ops".

> Finally, one proposal to make the aggregation code simpler would be to use a class with static attributes.

:-1: on this. I makes it hard for non advanced coders. In my experience, beginner struggle with imports, they struggle with nested. I've been teaching data-scientists for more than a decade, and it has changed my understanding of which good practices are actually good for people who are not trained coders.

skrub/_agg_joiner.py
Vincent-Maladiere: When dealing with namespaces (`np`, `cupy`, `torch`, etc) [scikit-learn uses the generic `xp`](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_array_api.py#L358). Since we consider pandas (`pd`) and polars (`pl`), `px` seemed relevant to me.

However, we're not returning either `pd` or `pl` namespace, but our own (either the file `_agg_pandas.py` or `_agg_polars.py`). So `agg_px` was what I came up with.

I'll use `px` to simplify things for now, but feel free to suggest anything else.

Vincent-Maladiere:
Thank you both, @GaelVaroquaux and @LilianBoulard, for these extensive reviews. I'm currently iterating on this!

setup.cfg
```diff
@@ -47,6 +47,7 @@ dev =
     pre-commit
 pyarrow =
     pyarrow
+    polars
```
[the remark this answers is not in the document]
Vincent-Maladiere: You're right. Let's define a proper `extras_require` option for polars for now. We can still remove it before merging this PR if we have doubts.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+            f"got {[type(df) for df in dataframes]}."
+        )
+
+
+def check_missing_columns(X, main_key):
+    main_keys = np.atleast_1d(main_key).tolist()
+    missing_cols = set(main_keys) - set(X.columns)
+    if len(missing_cols) > 0:
+        raise ValueError(
+            f"Got main_key={main_key!r}, but column not in X.columns {list(X.columns)}."
```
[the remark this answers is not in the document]
Vincent-Maladiere: Thanks for this useful tip :)

skrub/_agg_joiner.py
Vincent-Maladiere: Yes, but it also forces people to use kwargs, so I'm hesitant.

skrub/_agg_joiner.py
Vincent-Maladiere: Yes, good catch

skrub/_agg_joiner.py
Vincent-Maladiere: scikit-learn, numpy, scipy or pandas objects don't behave like this, so I would weight that against enforcing good practices that might be misleading for new users. 

Since the scope of this is broader than this PR, let's wait for a concensus.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+        if op_root in NUM_OPS:
+            num_ops.append(op_name)
+        if op_root in CATEG_OPS:
+            categ_ops.append(op_name)
+        if op_root not in ALL_OPS:
+            raise ValueError(f"'agg_ops' options are {ALL_OPS}, got: '{op_name}'.")
+    return num_ops, categ_ops
+
+
+def get_namespace(dataframes):
```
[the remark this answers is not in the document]
Vincent-Maladiere: Thanks for the comments on typing. I prefer doing these last because they are subject to change when the API changes. So it's like painting the house after it's fully built ;)

skrub/_agg_joiner.py
Vincent-Maladiere: WDYT?

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+
+    elif _is_module(dataframes, "polars"):
+        import polars as pl
+
+        if all([isinstance(table, pl.DataFrame) for table in dataframes]) or all(
+            [isinstance(table, pl.LazyFrame) for table in dataframes]
+        ):
+            return agg_pl, pl
+        else:
+            # XXX: polars series will raise this error
```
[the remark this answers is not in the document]
Vincent-Maladiere: In scikit-learn, XXX are general comments on stuff we'd like to address, but there's no direct solution.

skrub/_agg_joiner.py
LilianBoulard: You're right, pandas & co don't, that's the reason we didn't enforce that in the `fuzzy_join`, since we want it to be a drop-in replacement of `pandas.merge`. But I don't think this logic applies here, and similar to the [other encoders](https://github.com/skrub-data/skrub/blob/3e3075c5812be45d2e21dc312c8d5f58623dff45/skrub/_gap_encoder.py#L650) and even the [`Joiner`](https://github.com/skrub-data/skrub/blob/2473cd8ffb626a67c9a8275a922021393e3fee80/skrub/_joiner.py#L129), I think this should be the norm. For reference, #514.

I agree that this should not be blocking this PR, but definitely something that needs to be discussed and addressed before the release :)

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+        If set to None, we will use the table index, by looking at their
+        order in 'tables', e.g. for a duplicate columns: price (main table),
+        price_1 (auxiliary table 1), price_2 (auxiliary table 2), etc.
+
+    See Also
+    --------
+    :class:`AggTarget` :
+        Aggregates the target `y` before joining its aggregation
+        on the base dataframe.
+    :class:`FeatureAugmenter` :
```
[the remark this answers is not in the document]
LilianBoulard: Update: should be `Joiner` instead :)

skrub/_agg_joiner.py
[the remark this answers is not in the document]
Vincent-Maladiere: I'll replace these with FIXME or remove them entirely

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,592 @@
...
+    Uses :obj:`~pandas.DataFrame` inputs only.
+
+    Parameters
+    ----------
+    tables : list of tuples
+        List of (dataframe, columns_to_join, columns_to_agg) tuple
+        specifying the auxiliary dataframes and their columns for joining
+        and aggregation operations.
+
+        dataframe : :obj:`~pandas.DataFrame` or str
```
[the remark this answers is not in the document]
Vincent-Maladiere: I like the `DataFrameLike`! Also `SeriesLike`

skrub/datasets/_fetching.py
Vincent-Maladiere: I prefer to adapt the dataset fetcher in a subsequent PR to stay lean on this. The risk is having this PR coupled with #669.

Vincent-Maladiere:
A TODO list for subsequent PRs on `AggJoiner` and `AggTarget`:
- [ ] Support `value_counts` and `hist` operations for Polars
- [ ] Support `polars.LazyFrame`, which misses the `__dataframe__` attribute —A duck-typing like `lazy_df.__dataframe__ = None` is enough to get it working.
- [ ] Add more operations for both polars and pandas:
	- [ ] 	topk
	- [ ] unique-count
	- [ ] quantile(x)
	- [ ] first
	- [ ] last

skrub/tests/test_agg_joiner.py
Vincent-Maladiere: That's a great remark, thanks! FYI, the correct version is `r"(?=.*supported)(?=.*ndarray)"`. See [this guide](https://www.ocpsoft.org/tutorials/regular-expressions/password-regular-expression/).

GaelVaroquaux (changes requested):
I finished my review!

Amazing PR! I'm so excited!! And the code is very clean. I made a lot of minor suggestions. I have three major points:

- [x] I'm wondering about the API: we currently use a "tables" argument that is a list of 2-tuples or 3-tuples. But this is very hard to read and most often we use a list of length one. Should we rather "transpose" this signature and use 3 arguments (one optional) that can be lists optionally?

- [x] We need dedicated unit tests for the agg_pandas and agg_polars files (possibly via the get_namespace function. The reason is that testing directly this lower-level functionality will make it easier to maintain the code long term and have it evolve (for instance adding a backend)
- [ ] We need narrative documentation. It should be an entry in the assembling section of the docs. It does not have to be long, it should come without code and should focus on giving the intuitions of what these objects do and when they are useful.

Congratulations for this functionality!!

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,387 @@
...
+    "timestamp_year",
+    "timestamp_month",
+    "timestamp_hour",
+    "timestamp_dayofweek",
+]
+
+agg_joiner_user = AggJoiner(
+    tables=[
+        (X_date_encoded, "userId", timestamp_cols),
+    ],
```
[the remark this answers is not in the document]
Vincent-Maladiere: Great idea, I like it very much! It feels more natural and is easier to understand.

jovan-stojanovic (commented):
A few comments on my side!

skrub/agg_joiner/_agg_joiner.py
```diff
@@ -0,0 +1,675 @@
...
+        columns_to_join : str or array-like
+            Select the columns from the auxiliary dataframe to use as keys during
+            the join operation.
+
+        columns_to_agg : str or array-like, optional
+            Select the columns from the auxiliary dataframe to use as values during
+            the aggregation operations.
+            If missing, use all columns except `columns_to_join`.
+
+    main_keys : str or array-like
```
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, you're right

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,387 @@
...
+# columns, and doesn't interact with numerical columns.
+from skrub import TableVectorizer, DatetimeEncoder
+
+
+table_vectorizer = TableVectorizer(
+    datetime_transformer=DatetimeEncoder(add_day_of_the_week=True)
+)
+table_vectorizer.set_output(transform="pandas")
+X_date_encoded = table_vectorizer.fit_transform(X)
+print(X_date_encoded)
```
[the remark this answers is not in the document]
Vincent-Maladiere: I used `print` because I had issues with column display. Columns were overlapping on my local rendering of the example. The online doc doesn't suffer from this though. I will follow your recommendation.

examples/07_join_aggregation.py
```diff
@@ -0,0 +1,387 @@
...
+# We can finally create our hyper-parameter search space, and use a
+# |GridSearchCV|. We select the cross validation splitter to be
+# the |TimeSeriesSplit| to prevent leakage, since our data are timestamped
+# logs.
+#
+# Note that you need the name of the pipeline elements to assign them
+# hyper-parameters search.
+#
+# You can lookup the name of the pipeline elements by doing:
+list(pipeline.named_steps)
```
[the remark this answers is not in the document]
Vincent-Maladiere: Because it returns a dict with a lot of elements. If I'm only interested in the name of the steps, I need to fetch the keys. I prefer `list(pipeline.named_steps)` over `pipeline.named_steps.keys()`. WDYT?

GaelVaroquaux:
> Because it returns a dict with a lot of elements. If I'm only interested in the name of the steps, I need to fetch the keys. I prefer list (pipeline.named_steps) over pipeline.named_steps.keys(). WDYT?

OK

GaelVaroquaux:
> The dummy would predict the mean value over the entire y_train so it would be easier to beat.

:+1:

Vincent-Maladiere:
As this PR is getting quite lengthy, let's review https://github.com/skrub-data/skrub/pull/733 first. It only deals with the pandas and polars namespaces and should halve the line changes of this PR after merging.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,597 @@
...
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils.multiclass import type_of_target
+from sklearn.utils.validation import check_is_fitted
+
+from skrub._utils import atleast_1d_or_none, atleast_2d_or_none
+from skrub.dataframe import DataFrameLike, SeriesLike
+from skrub.dataframe._namespace import get_df_namespace
+
+NUM_OPERATIONS = ["sum", "mean", "std", "min", "max", "hist", "value_counts"]
+CATEG_OPERATIONS = ["mode", "count", "value_counts"]
```
jeromedockes, line 22: if we use "count" we end up with the (same) count for all categorical operations? also we can't have the count if our table has only numerical columns?

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,597 @@
...
+            Prediction target. Used to compute correlations between the
+            generated covariates and the target for screening purposes.
+
+        Returns
+        -------
+        AggJoiner
+            Fitted :class:`AggJoiner` instance (self).
+        """
+        self.check_input(X)
+        skrub_px, _ = get_df_namespace(*self.table_)
```
jeromedockes, line 200: `px` is maybe not super obvious and would stop making sense if we add a backend besides pd and pl?

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,597 @@
...
+                raise ValueError(
+                    "The number of keys to join must match, got "
+                    f"main_key={self.main_key_!r} and "
+                    f"{foreign_key=!r} for the table at index {idx}."
+                )
+
+        self.table_ = tables
+
+        # Check tables and list of cols match
+        if len(tables) != len(self.cols_):
```
jeromedockes, line 354: the amount of validation needed and the complexity of the parameter makes me wonder if we should allow only a single auxiliary table? In general I find parameters that can be str, list[str] or list[list[str]] a bit confusing because here if I want a single column per table I might expect to be able to pass them as a list of str, since I am allowed to pass a str for one table

jeromedockes (commented):
just a first batch of small comments

skrub/_agg_joiner.py
jeromedockes: are "hist" and "value_counts" the same?

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,597 @@
...
+
+    Returns
+    -------
+    num_operations, categ_operations : Tuple of List of str
+        List of operator names
+    """
+    num_operations, categ_operations = [], []
+    for operation in operations:
+        # hist(5) -> hist
+        op_root = operation.split("(")[0]
```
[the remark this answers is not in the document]
Vincent-Maladiere: No, we want to have a string and limit the cognitive load. Let's keep things simple.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,597 @@
...
+    foreign_key : str or iterable of str
+        Select the columns from the auxiliary dataframe to use as keys during
+        the join operation.
+
+    main_key : str or iterable of str
+        Select the columns from the main table to use as keys during
+        the join operation.
+        If main_key is a list, we will perform a multi-column join.
+
+    cols : str or iterable of str, default=None
```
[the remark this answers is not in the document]
jeromedockes: also, I think this and operation, in addition to str and iterable of str, can be iterable of iterable of str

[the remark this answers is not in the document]
Vincent-Maladiere: I suggested `agg_cols` previously, @gvaroquaux prefered `cols`. I don't have a strong opinion on this, although cols being shorter is handy for data scientists working on notebooks.

skrub/_agg_joiner.py
Vincent-Maladiere: Not exactly. Value counts does't accept any argument while hist does. Internaly, we use value counts to perform hist.

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,597 @@
...
+    1          2      NY JFK           DL              80.000000
+    """  # noqa: E501
+
+    def __init__(
+        self,
+        table: DataFrameLike | Iterable[DataFrameLike] | str | Iterable[str],
+        *,
+        foreign_key: str | Iterable[str],
+        main_key: str | Iterable[str],
+        cols: str | Iterable[str] | None = None,
```
jeromedockes, line 166: I think Iterable[Iterable[str]] is also allowed to specify a list of columns for each table? this is a more general discussion but it seems the type annotations in skrub in general don't type check: `mypy  skrub`  returns many errors. shall we consider either fixing the annotations and adding mypy to the CI, or removing the annotations?

skrub/_agg_joiner.py
```diff
@@ -0,0 +1,597 @@
...
+    Accepts :obj:`pandas.DataFrame` and :obj:`polars.DataFrame` inputs.
+
+    Parameters
+    ----------
+    table : DataFrameLike or str or iterable
+        Auxiliary dataframe to aggregate then join on the base table.
+        The placeholder string "X" can be provided to perform
+        self-aggregation on the input data.
+
+    foreign_key : str or iterable of str
```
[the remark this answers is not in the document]
Vincent-Maladiere: You have to keep in mind that we first perform an aggregation, then a join. In pandas, aggregation uses the "key" argument, but the merge or join uses "on" indeed. So to me both options are valid.

skrub/_agg_joiner.py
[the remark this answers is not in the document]
Vincent-Maladiere: > also, I think this and operation, in addition to str and iterable of str, can be iterable of iterable of str

This PR is a first iteration, let's keep things simple for now.

jeromedockes:
> Maybe we should define these functions as staticmethod? So that things are close to each other.
> No, staticmethods are to be used only when they are important for inheritance reasons. Classes are not to be confused with modules. Classes serve to define inheritance. Objects serve to associate functions to data (attributes). Modules serve to group code (symbols) together.

just for the sake of argument, having a class, say `skrub.DataFrame` would allow to:
- move around the methods together with the dataframe, without having to call `get_df_namespace` every time we need a specialized function
- describe explicitly the interface that `_polars` and `_pandas` must implement

skrub/_agg_joiner.py
jeromedockes: ok thanks. and IIUC the user passes all the operations name and then skrub splits them into numerical and categorical. for example this means there is no way to ask for value_counts for numerical columns but not categorical ones or vice-versa?

skrub/_agg_joiner.py
Vincent-Maladiere: Yes, so this is in reference to the array API used in scikit-learn and as a standalone project, naming the namespace variable `xp`, for numpy, cupy and pytorch.

Happy to hear your suggestions if you have some :)

skrub/_agg_joiner.py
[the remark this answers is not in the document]
jeromedockes: I meant the code is made to accept iterable of iterable of str, but the docstring  and type annotation don't mention it

skrub/_agg_joiner.py
Vincent-Maladiere: Yes that's a good point. Our perspective is that running agg joiner or joiner on multiple aux tables is probably quite niche, but still something we'd like to have, when working with many tables from a datalake for instance.

So using a list of list allows to makes things less ambiguous as a whole, but I agree that the params validation starts to get complex.

skrub/_agg_joiner.py
jeromedockes: as it's the output of `get_df_namespace`, maybe `df_namespace`. but it doesn't really matter as it's just a local variable we can always change it. I just thought it's a name that will come up often so we may pick one that we use all the time to refer to those modules

skrub/_agg_joiner.py
Vincent-Maladiere: Yes, correct. I agree that this feels limiting. 

One idea could be to ask for list first, but to also enable tweaking operations using a dict. 

Do you have suggestions?

skrub/_agg_joiner.py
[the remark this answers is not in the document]
Vincent-Maladiere: Oh yes correct, sorry read too fast

skrub/_agg_joiner.py
Vincent-Maladiere: > if we use "count" we end up with the (same) count for all categorical operations? also we can't have the count if our table has only numerical columns?

You're right. There's probably room for improvement here.

Would you consider running count only once, as an valid operation for both numerical and categorical columns?

skrub/_agg_joiner.py
jeromedockes: that makes sense. in this case could we have `main_key` and `foreign_key` have a plural name and force them to be a list of str for each table? this way we would have only 2 options: `list[list[str]]` for multiple tables or `list[str]` for a single table, and it would remove any ambiguity of the interpretation of `list[str]`

jeromedockes: also comparing lengths might fail if someone has 2 tables and 2 grouping columns

Vincent-Maladiere:
You're right. But it feels like we're going to reimplement the dataframe API in some way, WDYT?

skrub/_agg_joiner.py
jeromedockes: yes using a dict can be good, or we could have a `numerical_operations` and a `categorical_operations` separate parameters

jeromedockes:
> You're right. But it feels like we're going to reimplement the dataframe API in some way, WDYT?

I agree but isn't this already what we're doing, just with the `dataframe` submodules instead of `DataFrame` subclasses?

it is definitely true it feels like we're doing something too similar to ibis's or the dataframe API's objectives; what I understood from IRL discussions is we want to do it for a much more restricted/specialized set of operations until the Dataframe API covers all we need in skrub but I may have misunderstood

skrub/_agg_joiner.py
Vincent-Maladiere: Probably an even better option!

skrub/_agg_joiner.py
Vincent-Maladiere: > that makes sense. in this case could we have `main_key` and `foreign_key` have a plural name and force them to be a list of str for each table? this way we would have only 2 options: `list[list[str]]` for multiple tables or `list[str]` for a single table, and it would remove any ambiguity of the interpretation of `list[str]`

I would agree with @gvaroquaux here that we want the user to be able to use simple strings. The ambiguity is limited to the case of multiple tables, raising errors when the format is incorrect.

If this is an issue, I'd rather drop the support of multiple tables, and maybe create a specific estimator like MultiAggJoiner.

WDYT?

Vincent-Maladiere: > also comparing lengths might fail if someone has 2 tables and 2 grouping columns

I think we already handle this situation

skrub/_agg_joiner.py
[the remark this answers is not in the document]
Vincent-Maladiere: But I agree that this parsing helper could be improved to raise more user friendly errors

skrub/_agg_joiner.py
jeromedockes: ah yes you're right, with the `atleast_2d_or_none` call

skrub/_agg_joiner.py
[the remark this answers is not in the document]
jeromedockes: we could also implement it only once; atm parsing is done differently here and in the `_pandas` module

skrub/_agg_joiner.py
jeromedockes: > Would you consider running count only once, as an valid operation for both numerical and categorical columns?

yes I guess we could make a special case for count, and include just one count column whenever it appears in numerical ops, categorical ops, or both. or as it is cheap to compute and quite generic we could just add it all the time and remove it from the options. but maybe that's too much of a detail and we can leave that for later

skrub/_agg_joiner.py
jeromedockes: > > that makes sense. in this case could we have `main_key` and `foreign_key` have a plural name and force them to be a list of str for each table? this way we would have only 2 options: `list[list[str]]` for multiple tables or `list[str]` for a single table, and it would remove any ambiguity of the interpretation of `list[str]`
> 
> I would agree with @gvaroquaux here that we want the user to be able to use simple strings. The ambiguity is limited to the case of multiple tables, raising errors when the format is incorrect.
> 
> If this is an issue, I'd rather drop the support of multiple tables, and maybe create a specific estimator like MultiAggJoiner.
> 
> WDYT?

I like the MultiAggJoiner idea! I find it hard when a parameter can take many similar forms with different meanings so splitting the functionality like this might be a good thing. but if you both prefer the current interface it makes sense too!

Vincent-Maladiere:
> I agree but isn't this already what we're doing, just with the `dataframe` submodules instead of `DataFrame` subclasses?

I understand your point, we're indeed already simulating our tiny dataframe API.

> it is definitely true it feels like we're doing something too similar to ibis's or the dataframe API's objectives; what I understood from IRL discussions is we want to do it for a much more restricted/specialized set of operations until the Dataframe API covers all we need in skrub but I may have misunderstood

Yes, you're right, this is where we're heading. So I agree with what you say on the `skrub.DataFrame`, providing it doesn't add too much complexity and cost on our side.

skrub/_agg_joiner.py
Vincent-Maladiere: > this is a more general discussion but it seems the type annotations in skrub in general don't type check.

I agree, this is a discussion we need to have. It feels redundant to have both type hints in the function signatures and parameters in the docstrings.

skrub/_agg_joiner.py
Vincent-Maladiere: I've already created a meta-issue to keep track of our subsequent todos, if you want to add some inputs :) https://github.com/skrub-data/skrub/issues/734

skrub/_agg_joiner.py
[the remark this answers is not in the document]
Vincent-Maladiere: Yes correct, we could refactor this parsing logic

skrub/_agg_joiner.py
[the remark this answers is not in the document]
jeromedockes: ok I guess key is quite a generic term but at least "foreign key" doesn't apply here -- just writing it so we don't forget what we discussed IRL

skrub/_agg_joiner.py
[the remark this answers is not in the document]
jeromedockes: actually do we really need this parameter? specifying a `cols` argument is not less work than selecting those same columns (or dropping the rest) from the aux table before passing it to the AggJoiner

skrub/_agg_joiner.py
jeromedockes: An alternative could be to accept only one table but provide a method to add more:

```python
# constructor used by all users

joiner = AggJoiner(table=aux, foreign_key="from_airport", main_key="airportId")

# method to add more columns, used by more advanced users who need to join many tables

joiner.add_table(table=aux1, foreign_key="to_airport", main_key="airportId")

# or if we really want the list of tables

joiner.add_tables(tables=[...], ...)

# to avoid having to awkwardly pass the first table to the constructor and then
# add the rest in the advanced use case we could allow joiner=AggJoiner(),
# creating an empty joiner that doesn't do anything unless we call add_table
# later
```

or have a classmethod like `from_tables`

skrub/_agg_joiner.py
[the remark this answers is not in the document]
GaelVaroquaux: Yes we do: long term we want to enable covering the 90% cases without writing pandas (or polars) code. One reason is because long term we will enable working directly from data sources (eg directly from the database) and we want to make it easy for the user to switch across ways of working. Another reason is to supervise those choices.

Vincent-Maladiere:
@GaelVaroquaux, does this new version match your requirements?

Vincent-Maladiere:
I guess we can merge this now since we've converged on the design. https://github.com/skrub-data/skrub/issues/734 lists the next TODOs.

examples/08_join_aggregation.py
```diff
@@ -0,0 +1,365 @@
...
+    "timestamp_hour",
+    "timestamp_dayofweek",
+]
+
+agg_joiner_user = AggJoiner(
+    aux_table=X_date_encoded,
+    aux_key="userId",
+    cols=timestamp_cols,
+    main_key="userId",
+    suffix="_user",
```
GaelVaroquaux, line 166: Is specifying the suffix necessary for the example to work (not only here, but all over the example)?

If yes, can we think of a default for the suffix that makes it work here (and is not too strange / magic)

If no, I think that we should remove it from the example, to make the example a bit lighter.

GaelVaroquaux (commented):
Very very cool. 

I left one small inline comment (I hope that it won't be too much to address), and I have a couple of comments on the docs / website:

# Front page:

Replace on the front page:

"Joiner, a transformer for joining multiple tables together."

by

"Joiner, AggJoiner", transformers for joining multiple tables together."

# Assembling docs

Rework a tiny bit the assembling narrative docs to list the AggJoiner and
AggTarget.

I think that the way that I would do this is by add to the section
"Joinning external tables for machine learning. Where the Joiner is
mentioned, I would do a list with Joiner, AggJoiner and AggTarget, giving
quickly the differences between these.

examples/08_join_aggregation.py
Vincent-Maladiere: This is a relatively complete example. In most simple settings, users won't need to set a suffix.

When a suffix is set, the logic is the following:
1. This suffix is added to each column of the `aux` table after aggregating it.
2. Then, the join procedure is performed between the main table and the aggregated aux table, using the default suffix values, e.g., `("_x", "_y")` for pandas.

Therefore, we won't have errors if we don't set the suffix in this example. However, we will have columns suffixed with variations of `_x` and `_y`, which will be hard to decipher.

Note that if we input severable tables without setting suffixes, we will automatically generate suffixes using the index of the auxiliary tables (`_1`, `_2`, ...).

However, if we call several `AggJoiner` like in this example, I'm afraid we can't make good default suffixes.

WDYT?

GaelVaroquaux:
> This is a relatively complete example. In most simple settings, users won't need to set a suffix.

OK, but in the current situation the website will convey the impression that it is very complicated to use and will scare people away. I'm trying to act on this as much as I can by removing every possible element of complexity.

It's a pity that we cannot have heuristics that avoid collisions and that we need to set suffixes in this example.

Vincent-Maladiere:
> This is a relatively complete example. In most simple settings, users won't need to set a suffix.
> OK, but in the current situation the website will convey the impression that it is very complicated to use and will scare people away. I'm trying to act on this as much as I can by removing every possible element of complexity. It's a pity that we cannot have heuristics that avoid collisions and that we need to set suffixes in this example.

We have heuristics that avoid collisions, but these will be very un-informative in this example (what is `_x`, what is `_y`). We could do magic stuff with metaclasses or caching, but I'm not eager to go down that road.

Vincent-Maladiere:
Alternatively, we can simplify the example by removing one of the AggJoiner and AggTarget and hope it does not degrade the already weak performances too much.

examples/08_join_aggregation.py
jeromedockes: I think the meaningful suffixes "_movie" and "_user" actually help understand what we are doing and what we see in the cells' outputs (although that part is a bit hard still because the tables are quite wide)

jeromedockes:
>  degrade the already weak performances too much.

as the baseline have you tried the same estimator but without joining any auxiliary tables?

Vincent-Maladiere:
Here is a quick ablation study.

**Full pipeline**
```python
pipeline = make_pipeline(
    table_vectorizer,
    agg_joiner_user,
    agg_joiner_movie,
    agg_target_user,
    agg_target_movie,
    HistGradientBoostingRegressor(learning_rate=0.1, max_depth=4, max_iter=40),
)
```
</img>

**Without `agg_joiner_movie` and `agg_joiner_user`**
</img>

**Without `agg_joiner_movie` and `agg_joiner_user` and `agg_target_user`**
</img>

**Without `agg_joiner_movie` and `agg_joiner_user` and `agg_target_movie`**
</img>

<br>

**Conclusions**:
- both `AggJoiner` don't improve performances
- both `AggTarget` play a significant role

So, we can remove `AggJoiner` , but we have to keep both `AggTarget`.
Are we happy with this simplification? I know it doesn't showcase `AggJoiner` anymore, but we can find another example who will.

GaelVaroquaux:
> So, we can remove `AggJoiner` , but we have to keep both `AggTarget`.
> Are we happy with this simplification? I know it doesn't showcase `AggJoiner` anymore, but we can find another example who will.

Yes!! Thank you for doing this.

Maybe add a note in the example saying that if we wanted to aggregate on a value of X rather than y we have the AggJoiner.

Vincent-Maladiere:
It's done! LMK what you think :)

Vincent-Maladiere:
@GaelVaroquaux should we merge this now?

GaelVaroquaux (approved):
LGTM. Merging, thanks!

GaelVaroquaux:
:tada:

PR #605, opened 2023-06-15
MAINT bump python version to 3.10

glemaitre:
I saw that in the documentation that you have an extra colon in the docstring:

image

Indeed, this is due to the version of `sphinx` and `docutils` used: https://github.com/sphinx-doc/sphinx/issues/10594

This has been solved in the new sphinx version but I assume that using Python 3.9 is locking the version of `sphinx` used. Just wanted to give it a try to bump from 3.9 to 3.10.

Alternatively, one can overwrite the CSS most probably with something like:

```css
dl.field-list  .colon {
    display: none;
}
```

However, it would be easier to just fix it by installing a more recent sphinx version. FYI, we use sphinx 6.0.0 in scikit-learn.

### DIFF
--- a/build_tools/circle/build_doc.sh
+++ b/build_tools/circle/build_doc.sh
@@ -113,7 +113,7 @@ export PATH="miniconda/bin:$PATH"
 
 # Configure the conda environment and put it in the path using the
 # provided versions
-mamba create -n $CONDA_ENV_NAME --yes --quiet python="${PYTHON_VERSION:-*}" \
+mamba create -n $CONDA_ENV_NAME --yes python="${PYTHON_VERSION:-*}" \
   numpy="${NUMPY_VERSION:-*}" scipy="${SCIPY_VERSION:-*}" \
   pytest coverage matplotlib="${MATPLOTLIB_VERSION:-*}" sphinx \
   seaborn statsmodels pillow cython joblib pandas="${PANDAS_VERSION:-*}"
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
> Be aware that here this is only the python version of the documentation CI. I am forcing anything on the minimum Python version used by the package itself.

Yes. I'm in favor of doing this for the whole package. Let me do a another PR for this in parallel with yours

build_tools/circle/build_doc.sh
```diff
@@ -106,15 +106,14 @@ if [[ `type -t deactivate` ]]; then
...
+wget https://github.com/conda-forge/miniforge/releases/latest/download/Mambaforge-Linux-x86_64.sh \
    -O miniconda.sh
 chmod +x miniconda.sh && bash ./miniconda.sh -b -p "miniconda"
 export PATH="miniconda/bin:$PATH"
-conda update --yes --quiet conda
 
 # Configure the conda environment and put it in the path using the
 # provided versions
-conda create -n $CONDA_ENV_NAME --yes --quiet python="${PYTHON_VERSION:-*}" \
+mamba create -n $CONDA_ENV_NAME --yes python="${PYTHON_VERSION:-*}" \
```
glemaitre, line 116: I remove the `--quiet` option to check which version is installed by `mamba`. `conda` was printing these infos even in quiet mode.

GaelVaroquaux:
Your last push solved the initial formatting problem, no?

https://output.circle-artifacts.com/output/job/e221e210-8368-4339-a07f-4b632ae616d3/artifacts/0/doc/generated/skrub.fuzzy_join.html

GaelVaroquaux:
@glemaitre : this looks good to me. Is it OK to you if I merge?

glemaitre:
Yep.

GaelVaroquaux:
Merged, thanks!!

PR #610, opened 2023-06-15
Add Windows CI workers

### LINKED ISSUE #567: Add Windows workers in the CI
While discussing with folks at JupyterCon a couple weeks ago, a remark I had was that a significant part of our audience use Windows, and we loosely support it at the moment.

We should therefore add Windows workers to our CI.

LilianBoulard:
Fixes #567

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -30,7 +30,7 @@ jobs:
     strategy:
         fail-fast: false
         matrix:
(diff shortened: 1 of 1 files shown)


LilianBoulard:
Of course... of all the issues we could've had, it had to be a stupid path one... I'm looking into it.

jovan-stojanovic (approved):
Thanks!!

PR #612, opened 2023-06-15
Normalize tags

LilianBoulard:
Two things that bug me:
- The tag `No Changelog Needed` is title-cased, while the other tags are lowercase
- When we submit a fix for the CI, no changelog entry is required

This PR fixes that.


LilianBoulard:
The PR is ready, I'm waiting for a review, then I'll merge and fix the tags :)

GaelVaroquaux (approved):
LGTM. Merging

PR #613, opened 2023-06-15
Bump requirements

LilianBoulard:
Bump the minimal versions of all dependencies and removes the requirement on `requests` (we don't use directly).

For numpy, scipy and pandas, versions date back to mid/late 2022, and scikit-learn 1.2.1 was released in January.

PR #620, opened 2023-06-26
Use modern numpy type hints

LilianBoulard:
Following #613, we can now use the modern type hints provided by numpy to document array-likes and n-D arrays!

### DIFF
--- a/skrub/_datetime_encoder.py
+++ b/skrub/_datetime_encoder.py
@@ -2,6 +2,7 @@ from typing import Literal
 
 import numpy as np
 import pandas as pd
+from numpy.typing import ArrayLike, NDArray
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils.validation import check_is_fitted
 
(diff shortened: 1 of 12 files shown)


jovan-stojanovic (approved):
Thanks Lilian! I find this good, I just have a hard time to keep up with all the new type hinting practices. I feel like we/they change this every few months.

LilianBoulard:
> I just have a hard time to keep up with all the new type hinting practices. I feel like we/they change this every few months.

Yes, with the migration we've been bumping the minimum requirements aggressively, which leads us to update our code to the modern practices of the language :)
Recently they have changed because we bumped the minimal Python requirement from 3.8 to 3.10, which lets us use [PEP 585](https://peps.python.org/pep-0585/) annotations (introduced in Python 3.9). We also have bumped numpy to a more recent version, and in-between, they added proper support for type hinting their arrays.
To keep up to date, I simply read the Python major version changelogs (every on or two years), and I try to keep up with pandas, numpy and scikit-learn's adoption of type hints.

PR #630, opened 2023-07-17
MAINT Test against nightly dependencies

jeremiedbb:
Setup a cron job to test against nightly dependencies. It can also be manually triggered with `[deps nightly]` in the commit message.

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -7,6 +7,10 @@ on:
   pull_request:
     branches:
       - '*'
+  schedule:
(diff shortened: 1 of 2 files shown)


jeremiedbb:
I think that it's also useful to turn deprecation warnings into errors now that we test against nightly builds of dependencies. I think it's better to do that as a follow up PR because it does trigger some errors and fixing them would make the scope of this PR too large.

LilianBoulard (approved):
LGTM, thanks!

PR #639, opened 2023-07-17
MAINT make pyarrow optional

glemaitre:
Make `pyarrow` optional such that it can be imported with `pyodide`.


skrub/datasets/_fetching.py
```diff
@@ -371,6 +371,10 @@ def _fetch_figshare(
             UserWarning,
             stacklevel=2,
         )
+        import_optional_dependency(
+            "pyarrow", extra="pyarrow is required for parquet support."
+        )
```
GaelVaroquaux, line 376: I think that I'd prefer to implement the try-except ourselves rather than import from a private pandas module.

GaelVaroquaux (changes requested):
One minor change requested, plus:

- add in the docstring of all the functions skrub.datasets.fetch_ken* that they require the optional dependency pyarrow
- Add a note saying the same thing on top of https://skrub-data.org/stable/auto_examples/06_ken_embeddings_example.html

glemaitre:
I added in the "Notes" section that `pyarrow` is required for some of the functions.
I also added a note in the example as suggested.

GaelVaroquaux:
Can you add a changelog

PR #645, opened 2023-07-18
CI Turn deprecation warnings into errors

jeremiedbb:
or let's say almost all of them.

Unilke what we do in scikit-learn, this will also trigger locally, but I think it's a good thing.

### DIFF
--- a/skrub/_gap_encoder.py
+++ b/skrub/_gap_encoder.py
@@ -1044,7 +1044,7 @@ def get_kmeans_prototypes(
     projected = vectorizer.transform(X)
     if not sparse:
         projected = projected.toarray()
-    kmeans = KMeans(n_clusters=n_prototypes, random_state=random_state)
+    kmeans = KMeans(n_clusters=n_prototypes, n_init=10, random_state=random_state)
     kmeans.fit(projected, sample_weight=sample_weight)
     centers = kmeans.cluster_centers_
     neighbors = NearestNeighbors()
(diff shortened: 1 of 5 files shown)


skrub/_table_vectorizer.py
```diff
@@ -555,16 +555,14 @@ def _apply_cast(self, X: pd.DataFrame) -> pd.DataFrame:
...
                 known_categories = dtype.categories
                 new_categories = pd.unique(X[col])
                 dtype = pd.CategoricalDtype(
                     categories=known_categories.union(new_categories)
                 )
                 self.types_[col] = dtype
-            X[col] = X[col].astype(
-                dtype
-            )  # we dont use .loc (https://github.com/pandas-dev/pandas/issues/53729)
+        X = X.astype(self.types_)
```
jeremiedbb, line 565: This one has been modified at the same time and might not be raising a future warning anymore but I still think this solution is more efficient because we convert the dtypes all at once.

skrub/_gap_encoder.py
```diff
@@ -1044,7 +1044,7 @@ def get_kmeans_prototypes(
     projected = vectorizer.transform(X)
     if not sparse:
         projected = projected.toarray()
-    kmeans = KMeans(n_clusters=n_prototypes, random_state=random_state)
+    kmeans = KMeans(n_clusters=n_prototypes, n_init=10, random_state=random_state)
```
jeremiedbb, line 1047: In the future, the default `n_init` will change to 1. I explicitly set it to the actual default 10 to preserve the existing behavior.

jeremiedbb:
You can see that in [a55bc13](https://github.com/skrub-data/skrub/pull/645/commits/a55bc1339f7622e9b1c9c9f8fa29c3b0d8d6edf4), where I left 1 on purpose, the warning is turned into an error as expected.

GaelVaroquaux (approved):
Thanks!! Merging!

PR #646, opened 2023-07-18
change default sparse_threshold to 0 in TableVectorizer

LeoGrin:
which means that the output is always dense. I don't know which is better between this and a small value (like 0.01 or 0.05). Right now it is 0.3, which means that a simple pipeline TableVectorizer-->HistGradientBoostingClassifier often breaks when experimenting.


GaelVaroquaux (approved):
LGTM, merging

PR #647, opened 2023-07-18
MAINT activate common test sklearn

glemaitre:
Run scikit-learn common test.

### DIFF
--- a/skrub/tests/test_sklearn.py
+++ b/skrub/tests/test_sklearn.py
@@ -1,33 +1,29 @@
-from sklearn.utils.estimator_checks import check_estimator
+from sklearn.utils.estimator_checks import parametrize_with_checks
 
-from skrub import (  # TableVectorizer,
+from skrub import (
     DatetimeEncoder,
+    FeatureAugmenter,
     GapEncoder,
     MinHashEncoder,
     SimilarityEncoder,
+    TableVectorizer,
     TargetEncoder,
 )
 
 
-def test_sklearn_compatible_DatetimeEncoder():
-    check_estimator(DatetimeEncoder())
+def _tested_estimators():
+    for Estimator in [
+        DatetimeEncoder,
+        # FeatureAugmenter,
+        GapEncoder,
+        MinHashEncoder,
+        SimilarityEncoder,
+        TableVectorizer,
+        TargetEncoder,
+    ]:
+        yield Estimator()
 
 
-def test_sklearn_compatible_GapEncoder():
-    check_estimator(GapEncoder())
-
-
-def test_sklearn_compatible_MinHashEncoder():
-    check_estimator(MinHashEncoder())
-
-
-def test_sklearn_compatible_SimilarityEncoder():
-    check_estimator(SimilarityEncoder())
-
-
-# def test_sklearn_compatible_TableVectorizer():
-#     check_estimator(TableVectorizer())
-
-
-def test_sklearn_compatible_TargetEncoder():
-    check_estimator(TargetEncoder())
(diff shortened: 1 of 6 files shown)


GaelVaroquaux:
There is a failing test. It seems that the TableVectorizer no longer accepts 1D inputs.

This is a good thing, but the changelog must be updated to reflect this, and the test updated.

skrub/tests/test_sklearn.py
```diff
@@ -1,33 +1,86 @@
...
+    if "categorical" in _safe_tags(estimator, key="X_types"):
+        X = np.round((X - X.min()))
+        if "string" in _safe_tags(estimator, key="X_types"):
+            # Note: this part is the monkey patch
+            X = X.astype(object)
+            for i in range(X.shape[0]):
+                for j in range(X.shape[1]):
+                    X[i, j] = str(X[i, j])
+        elif _safe_tags(estimator, key="allow_nan"):
+            X = X.astype(np.float64)
```
GaelVaroquaux, line 45: Why float64 and not 32?

skrub/tests/test_sklearn.py
```diff
@@ -1,33 +1,86 @@
...
+        else:
+            X = X.astype(np.int32)
 
-def test_sklearn_compatible_GapEncoder():
-    check_estimator(GapEncoder())
+    if estimator.__class__.__name__ == "SkewedChi2Sampler":
```
GaelVaroquaux, line 49: Aaaaaaaah :rofl:  :rofl:  :rofl:  :rofl:

GaelVaroquaux (commented):
One minor comment and then merge

skrub/tests/test_sklearn.py
```diff
@@ -1,33 +1,86 @@
...
+    elif _safe_tags(estimator, key="pairwise"):
+        X = kernel(X, X)
+    return X
 
-def test_sklearn_compatible_MinHashEncoder():
-    check_estimator(MinHashEncoder())
 
-
-def test_sklearn_compatible_SimilarityEncoder():
-    check_estimator(SimilarityEncoder())
```
GaelVaroquaux, line 25: Please add a todo here

skrub/tests/test_sklearn.py
glemaitre: The scikit-learn conversion will be `np.float64` so it will be better for all common test.

GaelVaroquaux:
You have another test failing. The feature names have changed (I'm a bit surprised, I don't see where the change comes from, but it does not matter terribly).

I don't mind the change of name. We should however update the test and maybe mention it in the changelog.

glemaitre:
The reason is that we now first `check_input(X)` before running the `_check_feature_names` to pass the common test. Before, I assume that we create the `feature_names` potentially on a data frame (before the validation of `X`) while now this is on a numpy array.

GaelVaroquaux:
> However, this is a regression because this is now different from what the scikit-learn encoder are doing.

Right, that might be a problem

skrub/tests/test_sklearn.py
```diff
@@ -1,4 +1,5 @@
 import numpy as np
+import pandas as pd
```
GaelVaroquaux, line 2: Is this still useful?

glemaitre:
```suggestion
```

skrub/tests/test_sklearn.py
GaelVaroquaux: :+1:

GaelVaroquaux:
LGTM. Updating the branch with main and then merging.

Issue #550, opened 2023-04-20
`pkg_ressources` is deprecated


LilianBoulard:
Seen in `_check_dependencies.py`:
![image](https://user-images.githubusercontent.com/33222654/233378673-4168e6af-80fb-4f04-9b87-a546ce644a4a.png)

Vincent-Maladiere:
cc @fcharras, who helped with the setup of `pkg_resources` :)

fcharras:
It looks like it's a bug in latest `setuptools` release ? see https://stackoverflow.com/a/76044568 and https://github.com/pypa/pip/issues/11975

fcharras:
IIRC, `pkg_resources` is indeed deprecated or at least its use is discouraged in projects, see [setuptools doc](https://setuptools.pypa.io/en/latest/pkg_resources.html#package-discovery-and-resource-access-using-pkg-resources) . But what is supposed to be the replacement, [importlib](https://docs.python.org/3/library/importlib.html), does not embed the features that are needed to implement `check_dependencies` nor any other alternatives I'm aware of (or at least it did not at the time it was merged).

PR #648, opened 2023-07-18
MAINT Use importlib + packaging instead of pkg_resources

jeremiedbb:
Fixes https://github.com/skrub-data/skrub/issues/550

usage of `pkg_resources` is deprecated. There's still no way to use only `importlib` to do the same thing, at least without having to manually parse the requirements. This PR uses `packaging` instead that provides tools to parse the requirements.

Like `pkg_resources`, `packaging` is not in the std lib, but is very likely to be installed since it comes with `pip`. And at least this one is not deprecated :)

### DIFF
--- a/skrub/_check_dependencies.py
+++ b/skrub/_check_dependencies.py
@@ -1,30 +1,29 @@
-import pkg_resources
+from importlib.metadata import requires, version, PackageNotFoundError
+from packaging.requirements import Requirement
 
 
 def check_dependencies():
     package_name = "skrub"
-    env = pkg_resources.Environment()
-    package = env[package_name][0]
-    requirements = package.requires()
-    for req in requirements:
-        try:
-            installed_dep = next(
-                iter(
-                    (
-                        installed_dep
-                        for installed_dep in env[req.name]
-                        if installed_dep.project_name == req.name
-                    )
-                )
-            )
-        except StopIteration:
-            raise ImportError(
-                f"{package_name} {package.version} requires {req}, "
-                "which you don't have."
-            )
+    package_version = version(package_name)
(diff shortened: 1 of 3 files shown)


glemaitre (approved):
LGTM.

skrub/_check_dependencies.py
```diff
@@ -1,30 +1,29 @@
...
-                "which you don't have."
-            )
 
-        if installed_dep not in req:
+        except PackageNotFoundError:
             raise ImportError(
-                f"{package_name} {package.version} requires {req} "
-                f"but you have {installed_dep}, which is not compatible."
+                f"{package_name} {package_version} requires {str(req)}, "
+                "which you don't have."
```
LilianBoulard, line 28:
```suggestion
                f"{package_name} {package_version} requires {req!s}, "
                "which you don't have installed."
```

skrub/_check_dependencies.py
```diff
@@ -1,30 +1,29 @@
...
-                    (
-                        installed_dep
-                        for installed_dep in env[req.name]
-                        if installed_dep.project_name == req.name
-                    )
+            installed_dep = version(req.name)
+            if not req.specifier.contains(installed_dep):
+                raise ImportError(
+                    f"{package_name} {package_version} requires {str(req)} "
+                    f"but you have {req.name}={installed_dep}, which is not compatible."
```
LilianBoulard, line 22:
```suggestion
                    f"{package_name} {package_version} requires {req!s} "
                    f"but you have {req.name} {installed_dep} installed, which is incompatible."
```

LilianBoulard (approved):
Couple of minor changes (which I'll commit) and we're good to merge.

PR #649, opened 2023-07-18
TST: parallelism tests

GaelVaroquaux:
Addresses two problems:
- The test was way too long (2mn)
- The test was brittle given that on CI workers there are only 2 possible CPUs


jeremiedbb (approved):
measuring timings in CI tests is usually very brittle. It should be part of a benchmark suite instead :)

LGTM.

GaelVaroquaux:
Thanks @jeremiedbb for the review. Merging

PR #650, opened 2023-07-18
added duplicate column names features

myungkim930:
Added raising an error for duplicate column names and a test.

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -618,7 +618,8 @@ class TableVectorizer(ColumnTransformer):
         # It is checked by comparing the number of unique values
         # to the length of the column names
         if len(set(X.columns)) != len(X.columns):
-            raise AssertionError("Duplicate column names in the dataframe")
+            raise AssertionError("Duplicate column names in the dataframe."
+                                 f"The column names are {X.columns}")
         ###
 
         # If auto_cast is True, we'll find and apply the best possible type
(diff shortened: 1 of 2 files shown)


skrub/_table_vectorizer.py
```diff
@@ -613,6 +613,14 @@ def fit_transform(self, X: ArrayLike, y: ArrayLike = None) -> ArrayLike:
         # in numerical columns for instance.
         X = _replace_false_missing(X)
 
+        ###
+        # We need to check for duplicate column names.
+        # It is checked by comparing the number of unique values
+        # to the length of the column names
+        if len(set(X.columns)) != len(X.columns):
+            raise AssertionError("Duplicate column names in the dataframe")
```
GaelVaroquaux, line 621: Can you add to the error message the values of X.columns, so as to add the user to diagnose the problem. Something like the suggestion below:
```suggestion
            raise AssertionError("Duplicate column names in the dataframe."
                                 f"The column names are {X.columns}")
```

skrub/tests/test_table_vectorizer.py
```diff
@@ -245,6 +245,20 @@ def _test_possibilities(X) -> None:
...
+    """
+    Test to check if the tablevectorizer raises an error with
+    duplicate column names
+    """
+    tablevectorizer = TableVectorizer()
+    # Creates a simple dataframe with duplicate column names
+    data = [(3, "a"), (2, "b"), (1, "c"), (0, "d")]
+    X_dup_col_names = pd.DataFrame.from_records(data, columns=["col_1", "col_1"])
+
+    with pytest.raises(AssertionError):
```
GaelVaroquaux, line 258: Let's check that the error is the one that we think we should be getting:
```suggestion
    with pytest.raises(AssertionError,  match=r"Duplicate column names"):
```

GaelVaroquaux (approved):
A few suggestions that I am applying

PR #655, opened 2023-07-19
Adding a checksum when downloading figshare files

### LINKED ISSUE #642: We should checksum the file to be sure that downloading went fine
A way to be sure that downloading went fine is to check the hash of the file.

This function of scikit-learn could be useful:

https://github.com/scikit-learn/scikit-learn/blob/fa8cd70bf33f4e64fed17d2952ffc7b900381e40/sklearn/datasets/_base.py#L1365-L1396

flefebv:
Closes #642
Supersedes #654

Modifies the `_fetch_figshare` function to checksum the files when downloading.

Remark: For openml, the checksum is already done within the sklearn function `fetch_openml`.

Warning: does not deal with world_bank, because world_bank does not provide checksum, and we do not have a fixed list of datasets that the user can download from world_bank.


skrub/datasets/_fetching.py
```diff
@@ -376,6 +391,20 @@ def _fetch_figshare(
...
+            checksum = _sha256(filehandle)
+            if figshare_id in figshare_id_to_hash:
+                expected_checksum = figshare_id_to_hash[figshare_id]
+                if checksum != expected_checksum:
+                    raise OSError(
+                        "{} has an SHA256 checksum ({}) "
+                        "differing from expected ({}), "
+                        "file may be corrupted.".format(
+                            filehandle, checksum, expected_checksum
+                        )
```
LilianBoulard, line 405:
```suggestion
                        f"{filehandle!r} SHA256 checksum differs from "
                        f"expected ({checksum}!={expected_checksum}) ; "
                        f"file is probably corrupted."
```

LilianBoulard (approved):

GaelVaroquaux:
LGTM. Merging

PR #658, opened 2023-07-20
[WIP] Vendor TargetEncoder from scikit-learn

jeremiedbb:
The long-term goal is to remove the `TargetEncoder` from `skrub` and use the one from `scikit-learn`. But since it's only available in 1.3 and `multiclass` support will only be available in 1.4, this PR proposes to vendor it in the mean time.

If the version of the installed `scikit-learn` is recent enough, i.e. 1.4 (which is not possible right now as it does not exist), `skrub` directly uses the `scikit-learn` version.

The vendored version reimplements the compiled parts in pure python and thus is slower than the `scikit-learn` version (approx 2x slower). This acceptable since as soon as `sklearn` 1.4 is release we can advise them to upgrade `sklearn` to get the best performance.

This PR proposes to warn the user who is importing the `TargetEncoder` from `skrub` when the version of `scikit-learn` is >= 1.4, saying that it should be imported from `sklearn`. When the `scikit-learn` version is lower, I chose to not raise a warning otherwise there would be no way to use it without getting a warning because `scikit-learn` 1.4 does not exist yet. I think it's better to completely deprecate the class after `scikit-learn` 1.4 is released.

There's still one thing missing in this PR: the support for multiclass. There's a PR in `scikit-learn` that is not merged yet (https://github.com/scikit-learn/scikit-learn/pull/26674).

TODO:
- [ ] synchronize with https://github.com/scikit-learn/scikit-learn/pull/26674 when it's merged.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,6 +49,11 @@ Major changes
 
 * Removed `requests` from the requirements. :pr:`613` by :user:`Lilian Boulard <LilianBoulard>`
 
+* skrub now vendors the scikit-learn implementation of the :class:`TargetEncoder` until
+  the minimum required version of scikit-learn is 1.4, after what it will be removed and
+  should be imported from scikit-learn instead.
+  :pr:`??` by :user:`Jérémie du Boisberranger <jeremiedbb>`.
+
 Minor changes
 -------------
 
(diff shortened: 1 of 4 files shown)


Vincent-Maladiere (commented):
Hey, @jeremiedbb thanks for this PR! What is missing from this WIP to make it a proper PR?

jeromedockes:
note the multiclass target encoder pr in scikit-learn has been merged now: https://github.com/scikit-learn/scikit-learn/pull/26674

Vincent-Maladiere:
I'll resolve the conflicts and merge this PR then

jeromedockes:
closing in favor of #829

PR #667, opened 2023-07-20
FIX check_dependencies takes prereleases into account

jeremiedbb:
currently check_dependencies fails when scikit-learn=1.4.0.dev0 for instance.

The fix is to include prereleases in the version check, as documented here
https://packaging.pypa.io/en/stable/specifiers.html#packaging.specifiers.Specifier.contains


GaelVaroquaux (approved):
LGTM. Thanks

PR #669, opened 2023-07-20
Refactor fetching

LilianBoulard:
The goal of this PR is to refactor the fetching system to:
- Rely on the sklearn cache implementation of `fetch_openml`, which makes a lot of our code outdated
- Remove function `_fetch_as_dataclass`, which has grown too big
- Separate the public from the private API, since the file is about 1K lines long, making it hard to work with
- Simplify the Dataset API, so that we don't return different objects depending on the value of `load_dataframe`
- Align with the API of sklearn's fetchers

### DIFF
--- a/skrub/datasets/_fetchers.py
+++ b/skrub/datasets/_fetchers.py
@@ -1,0 +1,429 @@
+"""
+Implements the public fetchers from the private API available in `_fetching.py`
+"""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+from typing import Literal
+
+import pandas as pd
+
+from skrub.datasets._fetching import (
+    _fetch_figshare,
+    _fetch_openml_dataset,
+    _fetch_world_bank_data,
+    _resolve_path,
+)
+
+
+@dataclass(unsafe_hash=True)
+class Dataset:
+    """Represents a dataset and its information.
+
+    Parameters
+    ----------
+
+    name : str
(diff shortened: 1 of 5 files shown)


GaelVaroquaux:
Hold your horses :)

We have too many things happening right now to undertake a refactor. Let's wait a few days to merge a few things and have less ongoing modifications in the code base.

skrub/datasets/_fetchers.py
```diff
@@ -0,0 +1,419 @@
+"""
```
GaelVaroquaux, line 1: I'm not super enthousiastic about having a file called "_fetching.py" and one called "_fetcher.py". It's very unclear from the name of the files in which file should go what.

LilianBoulard: What about `_fetching_backend.py` and `_fetching_functions.py`? And to further clarify, add docstrings to the file, like we do for others?

GaelVaroquaux:
> What about _fetching_backend.py and _fetching_functions.py?

+1

> And to further clarify, add docstrings to the file, like we do for others?

+1

LilianBoulard:
From my quick analysis of the test error, it seems there is a difference somewhere in `fetch_world_bank_indicator`, which causes the indices of the two tables to mismatch. However, doing `self.X.reset_index(drop=True).equals(other.X.reset_index(drop=True))` in `Dataset.__eq__` doesn't work either, and I don't know why yet.

PR #672, opened 2023-07-20
Fix randomly failing deduplicate test

LeoGrin:
`test_backend_respected` from `test_deduplicate` is randomly failing when `make_deduplication_data` doesn't generate enough variation (#673), which breaks the CI. This PR sets the random_state of `make_deduplication_data` to prevent the failure.


GaelVaroquaux (approved):
That's good practice. Thanks!

PR #675, opened 2023-07-21
MAINT use composition in TableVectorizer

### LINKED ISSUE #660: Use composition in the `TableVectorizer`
### Problem Description

Several issues were found with the `TableVectorizer` recently, namely:
- Some of the internal mechanism clash with sklearn's principles, and thus some of the tests don't pass (see #647)
- We have some confusion between the attributes `transformers` and `transformers_`
- It seems kind of dirty to have a boiled down version of `fit_transform` in `transform`: I think we would be better implementing a proper and clean `fit`, and a clean `transform`, and a standard `fit_transform` that would just call `self.fit(X).transform(X)`.
  - And we often had issues in the past where we would implement something new in `fit_transform` and forget to implement it in `transform`, since the pattern is error-prone.

### Feature Description

The obvious solution to me would be to use composition (i.e. having a `column_transformer_` attribute that would be manipulated within the `TableVectorizer`) and drop the `ColumnTransformer` inheritance altogether.

### Alternative Solutions

None that I see

### Additional Context

_No response_

glemaitre:
closes #660

This is a POC for #660 that uses internally a `ColumnTransformer` instead of inheriting from it.


LilianBoulard (commented):
Thanks for the PoC, the implementation looks good!

skrub/_table_vectorizer.py
```diff
@@ -812,6 +836,18 @@ def get_feature_names_out(self, input_features=None) -> list[str]:
...
+        return self._column_transformer.named_transformers_
+
+    @property
+    def sparse_output_(self):
+        return self._column_transformer.sparse_output_
+
+    @property
+    def output_indices_(self):
+        return self._column_transformer.output_indices_
+
```
glemaitre, line 850: We need tests to check those attributes.

LilianBoulard:
I think you need to merge with main for the tests to run :)

skrub/_table_vectorizer.py
Vincent-Maladiere: What do we need these attributes for specifically?

skrub/_table_vectorizer.py
```diff
@@ -715,9 +708,44 @@ def _check_X(self, X, reset):
...
+        # whether to raise an error instead.
+        #
+        # Note that when fitting on a dataframe and transforming on
+        # the same dataframe with different column names,
+        # _check_feature_names will raise an error.
+        self._check_feature_names(X, reset=reset)
+        feature_names = _get_feature_names(X)
+        feature_names_in = getattr(self, "feature_names_in_", None)
+        if feature_names is None and feature_names_in is not None:
+            X.columns = feature_names_in
```
Vincent-Maladiere, line 723: Following our IRL discussion, @glemaitre

skrub/_table_vectorizer.py
Vincent-Maladiere, line 1: @LeoGrin and @glemaitre this should fix #709! This is very similar to what is done in `ColumnTransformer`

jovan-stojanovic (approved):

Vincent-Maladiere:
Should we close this PR @glemaitre?

glemaitre:
It will be automatically close when #761 will be merged.

Issue #585, opened 2023-06-09
Handle id columns differently


LeoGrin:
### Problem Description

Trying to understand better why the GapEncoder can very slow (#342), I found that it is at its slowest when dealing with "id" columns, which contain a lot of different ngrams. 
For instance for `traffic_violations` (restricted to 5000 rows), here is the time for each column:
```
{'seqid': 17.375709772109985,
 'description': 4.1079487800598145,
 'location': 5.461570978164673,
 'search_reason_for_stop': 1.0070619583129883,
 'state': 0.1596517562866211,
 'make': 1.005993127822876,
 'model': 1.8683619499206543,
 'charge': 1.4155240058898926,
 'driver_city': 1.2626869678497314,
 'driver_state': 0.11691594123840332,
 'dl_state': 0.125899076461792,
 'geolocation': 8.70003890991211}
```
(geolocation is also a weird column because it's tuples of floats with a lot of decimals, but that's another topic)
and for `drug_directory`
```
{'PRODUCTID': 23.933102130889893,
 'PRODUCTNDC': 9.601576089859009,
 'PROPRIETARYNAME': 5.157470941543579,
 'PROPRIETARYNAMESUFFIX': 0.5035130977630615,
 'NONPROPRIETARYNAME': 5.09909725189209,
 'DOSAGEFORMNAME': 0.6999397277832031,
 'ROUTENAME': 0.47968101501464844,
 'APPLICATIONNUMBER': 2.047283887863159,
 'LABELERNAME': 0.847294807434082,
 'SUBSTANCENAME': 4.429254055023193,
 'ACTIVE_NUMERATOR_STRENGTH': 1.4205100536346436,
 'ACTIVE_INGRED_UNIT': 0.8683860301971436,
 'PHARM_CLASSES': 3.5505928993225098}
 ```
 It is a shame, as they're probably the columns which are least useful to encode with the `GapEncoder`. Maybe we shouldn't count these columns like the other high-cardinality columns

### Detect these id columns
Different non-exclusive possibility:
- filter for "id" in column name
- compute the number of different ngrams for each column with sklearn's CountVectorizer (eventually normalize by string length). On the two examples above, the id columns stand out.

### Deal with these columns
Several possibilities:
- Drop these columns. I think people rarely want to use id columns for prediction: if the ids are completely random, they're not useful for prediction. If there is some structure, they can be useful, but also very misleading (for instance if a date is contained in the id, and the same date is in another column).
[...]

PR #680, opened 2023-07-28
Gap encoder speedups

LeoGrin:
Speedup the GapEncoder.

- Implement a better early stopping for the GapEncoder, based on the score. Benchmarked in #681. Based on #663 and #593.
    - Use an Exponentially weighted average of the batch score instead of computing the full score, based on sklearn's [MinibatchNMF](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.MiniBatchNMF.html#sklearn.decomposition.MiniBatchNMF)'s code.
     - Remove the early stopping logic based on `W_change`, as it was never reached. Thus remove the parameters `tol`.
     - Remove the `min_iter` parameter.

- Speedup each loop by taking profit of the sparsity of the count matrix. More precisely, compute H @ W only when X is non-zero (this is adapted from sklearn's MinibatchNMF). I sped up the W update which was the bottleneck. The H update was already better optimized for sparsity, but maybe it can also be sped-up using `_special_sparse_dot` (I won't have time to do it in the near future.)

- Benchmark hyperparameters `batch_size`, `max_iter_e_step` and `max_no_improvements` after these changes. They now default to 1024, 1 and 5. *Note: the benchmark should be rerun when we deal with id columns (#585), because the hp are chosen to mostly speed up the `GapEncoder` on these columns, as they are currently the bottleneck. `max_iter_e_step` should also perhaps be increased if we manage to speedup the H update.*

Quick benchmark done on my laptop comparing the speed of this implementation to the current main branch, on the columns of `traffic_violations` (restricted to 100K rows):

image


benchmarks/bench_gap_encoder_hp.py
LilianBoulard, line 1: You might want to exclude this file and the parquet results file from this PR, since it's already part of #681.

skrub/_gap_encoder.py
```diff
@@ -566,8 +629,13 @@ class GapEncoder(TransformerMixin, BaseEstimator):
...
         If `True`, the weight matrix `W` is rescaled at each iteration
         to have a l1 norm equal to 1 for each row.
-    max_iter_e_step : int, default=20
+    max_iter_e_step : int, default=1
         Maximum number of iterations to adjust the activations h at each step.
+    max_no_improvement : int, default=5
+        Control early stopping based on the consecutive number of mini batches
+        that does not yield an improvement on the smoothed cost function.
+        To disable convergence detection based on cost function,
+        set max_no_improvement to None.
```
LilianBoulard, line 638: It's not be obvious what happens then when `max_no_improvement=None`. Maybe this is clearer:
```suggestion
        To disable early stopping and run the process fully,
        set ``max_no_improvement=None``.
```

skrub/_gap_encoder.py
```diff
@@ -198,6 +199,57 @@ def _init_w(self, V: NDArray, X) -> tuple[NDArray, NDArray, NDArray]:
...
+
+        # Ignore first iteration because H is not updated yet.
+        if step == 1:
+            if self.verbose:
+                print(f"Minibatch step {step}/{n_steps}: mean batch cost: {batch_cost}")
+            return False
+
+        # Compute an Exponentially Weighted Average of the cost function to
+        # monitor the convergence while discarding minibatch-local stochastic
+        # variability: https://en.wikipedia.org/wiki/Moving_average
```
LilianBoulard, line 217: Same comment as #681, keeping the naming the same as Wikipedia could help with maintenance:
```suggestion
        # Compute an Exponentially Moving Average (EMA) of the cost function to
        # monitor the convergence while discarding minibatch-local stochastic
        # variability: https://en.wikipedia.org/wiki/Moving_average#Exponential_moving_average
```
If accepted, the variables need to be renamed accordingly (e.g. `_ewa_cost` to `_ema_cost`).

LilianBoulard (commented):
Nice implem, thanks!
On top of the below comments, it also needs a proper test :)

benchmarks/bench_gap_encoder_hp.py
LeoGrin: It is not the same benchmark, here it's a benchmark to find the best default hyperparameters. In #681 it's to check that the early stopping logic doesn't modify the gap behavior.

LeoGrin:
> Nice implem, thanks! On top of the below comments, it also needs a proper test :)

Thanks for the review! What kind of test do you have in mind? A test to check that `update_multiplicative_w` is the same function before and after the change?

skrub/_gap_encoder.py
LeoGrin: Hum I think I prefer to keep the same name than in sklearn's MinibatchNMF, WDYT? The full name is Exponential Weighted Moving Average so both ewa and ema seem fine to me.

LeoGrin:
Quick benchmark done on my laptop comparing the speed of this implementation to the current main branch:

image

benchmarks/bench_gap_encoder_hp.py
LilianBoulard: Okay, my bad! I glanced over it

LilianBoulard:
Maybe test that passing `None` runs the maximum number of iterations, and with an early stopping value, that it doesn't (if possible).

GaelVaroquaux:
Very nice. I just gave it a quick try:

On main:
```
In [1]: from skrub import datasets, TableVectorizer
   ...: 

In [2]: data = datasets.fetch_employee_salaries()

In [3]: tab_vec = TableVectorizer()

In [4]: %time X = tab_vec.fit_transform(data.X)
CPU times: user 36.5 s, sys: 2min 9s, total: 2min 46s
Wall time: 22.3 s
```
On this branch:
```
CPU times: user 11 s, sys: 30.9 s, total: 41.9 s
Wall time: 6.22 s
```
6 seconds has a very very different feel than 22 seconds

GaelVaroquaux (approved):
LGTM. Good for merge as far as I am concerned. Thanks!!

LeoGrin:
> Maybe test that passing `None` runs the maximum number of iterations, and with an early stopping value, that it doesn't (if possible).

I added a test to check that it doesn't break with max_no_improvement=None, but it seems hard to check the number of iteration (we don't have a `n_iter_` attribute). Do you have an idea how it could be done simply? Otherwise, are you fine with merging without this test?

skrub/_gap_encoder.py
```diff
@@ -251,13 +306,25 @@ def fit(self, X: ArrayLike, y=None) -> "GapEncoderColumn":
...
+                    batch_cost=batch_cost,
+                    n_samples=n_samples,
+                    step=n_iter_ * n_batch + i,
+                    n_steps=self.max_iter * n_batch,
+                ):
+                    break
+            else:
+                # only continue if no break occurred
+                continue
+            break
```
Vincent-Maladiere, line 327: This is pretty hard to read. Should we put the second for-loop in a function and use `return` instead? Alternatively, we could use a boolean flag, but the current function will look cluttered. WDYT?

skrub/_gap_encoder.py
```diff
@@ -945,6 +1012,29 @@ def _rescale_W(W: NDArray, A: NDArray) -> None:
...
+        for start in range(0, n_vals, batch_size):
+            batch = slice(start, start + batch_size)
+            dot_vals[batch] = np.multiply(H[ii[batch], :], W.T[jj[batch], :]).sum(
+                axis=1
+            )
+
+        HW = sp.coo_matrix((dot_vals, (ii, jj)), shape=X.shape)
+        # in sklearn, it was return WH.tocsr(), but it breaks the code in our case
+        # I'm not sure why
+        return HW
```
Vincent-Maladiere, line 1033: This will need further investigation in another PR

Vincent-Maladiere (approved):
A few comments and this is good to go!

LeoGrin:
> Also, why is the parquet file necessary?

It contains the result of the hyperparameters benchmark

Vincent-Maladiere:
Yes but still why would we need this file? It's usually not recommended to commit data in repository, the point of having benchmark files is also to reproduce this data.

LeoGrin:
Yeah I'm not sure. We've been doing this for the previous benchmarks: it's
pretty small files, and it makes it straightforward for other people to
analyse the results. But I see your point, I'm kinda pro pushing the
results but I don't have a strong opinion.

Vincent-Maladiere:
Ok, let's keep this file and move fast. We'll discuss these benchmark files later :)

skrub/_gap_encoder.py
LeoGrin: I like the boolean solution, it's easier to read

skrub/_gap_encoder.py
```diff
@@ -958,7 +1048,12 @@ def _multiplicative_update_w(
     Multiplicative update step for the topics `W`.
     """
     A *= rho
-    A += W * safe_sparse_dot(Ht.T, Vt.multiply(1 / (np.dot(Ht, W) + 1e-10)))
+    HtW = _special_sparse_dot(Ht, W, Vt)
+    Vt_data = Vt.data
+    HtW_data = HtW.data
```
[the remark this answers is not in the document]
LeoGrin: We want to compute `Vt / (elementwise) (np.dot(Ht, W)`, with `Vt` being sparse. We first compute `np.dot(Ht, W)` for indices corresponding to non-zero values of Vt. This gives us `HtW`, which has the same sparsity pattern as `Vt`. Then we use the `data` attribute to divide the non-null elements.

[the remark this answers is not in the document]
Vincent-Maladiere: My bad, I forgot the data in csr is actually stored in the data attribute. I thought `data` here was a reference to the buffer API of arrays, as in scikit-learn.

Vincent-Maladiere:
Hey @LeoGrin, could you add a test with `verbose=True`? That would help with the coverage! We'll merge then.

jovan-stojanovic (approved):
Hey @LeoGrin thanks, I think this is ready to be merged after you merge with main.

Vincent-Maladiere:
There's an error that seems unrelated to your PR. I fixed it, could you rebase on main to see if this is still there?

benchmarks/bench_gap_encoder_hp.py
```diff
@@ -0,0 +1,175 @@
...
+        "roc_auc_hgb_train": roc_auc_hgb_train,
+        "roc_auc_hgb_test": roc_auc_hgb_test,
+        "balanced_accuracy_hgb_train": balanced_accuracy_hgb_train,
+        "balanced_accuracy_hgb_test": balanced_accuracy_hgb_test,
+        "train_size": X_train.shape[0],
+        "test_size": X_test.shape[0],
+    }
+
+    print("Done")
+    print(res_dic)
```
LilianBoulard, line 111: Seems like debug
```suggestion
```

skrub/_gap_encoder.py
```diff
@@ -198,6 +199,77 @@ def _init_w(self, V: NDArray, X) -> tuple[NDArray, NDArray, NDArray]:
         B = A.copy()
         return W, A, B
 
+    def _minibatch_convergence(self, batch_size, batch_cost, n_samples, step, n_steps):
```
LilianBoulard, line 202: Useful for readability and IDE support, even on private API.
```suggestion
    def _minibatch_convergence(self, batch_size: int, batch_cost: float, n_samples: int, step: int, n_steps: int):
```

skrub/_gap_encoder.py
```diff
@@ -566,8 +648,13 @@ class GapEncoder(TransformerMixin, BaseEstimator):
     rescale_W : bool, default=True
         If `True`, the weight matrix `W` is rescaled at each iteration
         to have a l1 norm equal to 1 for each row.
-    max_iter_e_step : int, default=20
+    max_iter_e_step : int, default=1
         Maximum number of iterations to adjust the activations h at each step.
+    max_no_improvement : int, default=5
+        Control early stopping based on the consecutive number of mini batches
+        that does not yield an improvement on the smoothed cost function.
```
LilianBoulard, line 655:
```suggestion
        that do not yield an improvement on the smoothed cost function.
```

CHANGES.rst
```diff
@@ -65,6 +65,20 @@ Major changes
...
   :pr:`678` by :user:`Guillaume Lemaitre <glemaitre>`
 
+* :class:`GapEncoder` speedup :pr:`680` by :user:`Leo Grinsztajn <LeoGrin>`
+
+  - Improved :class:`GapEncoder`'s early stopping logic. The parameters `tol` and `min_iter`
+  have been removed. The parameter `max_no_improvement` can now be used to control the
+  early stopping.
+  :pr:`663` by :user:`Simona Maggio <simonamaggio>`
+  :pr:`593` by  :user:`Lilian Boulard <LilianBoulard>`
+  :pr:`681` by  :user:`Leo Grinsztajn <LeoGrin>`
```
LilianBoulard, line 75: Almost there
```suggestion
  - Improved :class:`GapEncoder`'s early stopping logic. The parameters `tol` and `min_iter`
    have been removed. The parameter `max_no_improvement` can now be used to control the
    early stopping.
    :pr:`663` by :user:`Simona Maggio <simonamaggio>`
    :pr:`593` by  :user:`Lilian Boulard <LilianBoulard>`
    :pr:`681` by  :user:`Leo Grinsztajn <LeoGrin>`
```

benchmarks/bench_gap_encoder_hp.py
```diff
@@ -13,10 +13,11 @@
 from skrub import GapEncoder
 import seaborn as sns
 import matplotlib.pyplot as plt
+import loguru
```
LilianBoulard, line 16: The standard usage is rather
```suggestion
from loguru import logger
```

benchmarks/bench_gap_encoder_hp.py
```diff
@@ -165,7 +161,7 @@ def plot(df: pd.DataFrame):
     ds = fetch_traffic_violations()
 
     if _args.run:
-        print("Running benchmark")
+        loguru.logger.info("Running benchmark")
```
LilianBoulard, line 164:
```suggestion
        logger.info("Running benchmark")
```

jovan-stojanovic:
Thanks @LeoGrin !! This will be such an improvement ..

GaelVaroquaux:
The doc building is indeed markedly faster:
image
image

Congratulations!

Vincent-Maladiere:
where do you see this @GaelVaroquaux ?

GaelVaroquaux:
I go to the history of commit (from the main github page), and I click on the green "tick" for a given commit, then follow the circleci link.

PR #681, opened 2023-07-28
Benchmark gap encoder early stopping

LeoGrin:
Benchmark to evaluate the early stopping change made in #680, compared to the previous version of the gap encoder.

Based on #663 by @simonamaggio and #593 by @LilianBoulard 

Compared to #663 , we don't compute the full score regularly, as this operation is very slow. Instead, we use an exponentially weighted average of each batch score. This is based on sklearn's [MinibatchNMF](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.MiniBatchNMF.html#sklearn.decomposition.MiniBatchNMF)'s code.

## Results
It seems that the new version is faster, while the KL score and balanced accuracy don't change. I think the speed can be improved further by tuning hyperparameters, this is the subject of another benchmark in #680. Furthermore, id columns like `seqid` take a long time to fit even with early stopping, as the score keeps decreasing, so #585 is still relevant even after this speedup. 

image
image
image


benchmarks/bench_gap_es_score.py
```diff
@@ -0,0 +1,344 @@
...
+            rescale_W=self.rescale_W,
+            max_iter_e_step=self.max_iter_e_step,
+            max_no_improvement=10,
+            verbose=True,
+        )
+
+
+#########################################################
+# Benchmarking accuracy and speed on actual traffic_violations dataset
+#########################################################
```
LilianBoulard, line 194:
```suggestion
###############################################################
# Benchmarking accuracy and speed on traffic_violations dataset
###############################################################
```

benchmarks/bench_gap_es_score.py
```diff
@@ -0,0 +1,344 @@
...
+    # split the data into train and test
+    X_train, X_test, y_train, y_test = train_test_split(
+        X[:max_rows],
+        y[:max_rows],
+        test_size=0.2,
+    )
+    if not modif:
+        gap = GapEncoder()
+    else:
+        gap = ModifiedGapEncoder(verbose=True, batch_size=512)
```
LilianBoulard, line 239: Maybe, to compare them equally, we should have the same batch size on both? Or am I missing something?
```suggestion
    if not modif:
        gap = GapEncoder(batch_size=512)
    else:
        gap = ModifiedGapEncoder(verbose=True, batch_size=512)
```

LilianBoulard (commented):
Benchmark looks good, thanks for the contribution!
Maybe one minor modif before merging (ping me when you think this is ready): remove the print statements (and pass `verbose=False`).

LeoGrin:
Running with the same batch size, the speedup is smaller, but I think still useful.

image

image

image

LeoGrin:
@LilianBoulard I think it's ready

jovan-stojanovic (approved):
LGTM, thanks!

PR #683, opened 2023-07-31
Rework `DatetimeEncoder` example

LilianBoulard:
Extracted from #546.

### DIFF
--- a/examples/03_datetime_encoder.py
+++ b/examples/03_datetime_encoder.py
@@ -1,22 +1,42 @@
 """
+.. _example_datetime_encoder
+
+===================================================
 Handling datetime features with the DatetimeEncoder
 ===================================================
 
-We illustrate here how to handle datetime features with the
-DatetimeEncoder.
+In this example, we illustrate how to better integrate datetime features
+in machine learning models with the |DatetimeEncoder|.
 
-The |DtE| breaks down each datetime
-features into several numerical features, by extracting relevant
-information from the datetime features, such as the month, the day
-of the week, the hour of the day, etc.
-Used in the |TV|, which automatically detects
-the datetime features, the |DtE| allows to
-handle datetime features easily.
+This encoder breaks down passed datetime features into relevant numerical
+features, such as the month, the day of the week, the hour of the day, etc.
+
+It is used by default in the |TableVectorizer|.
 
 
-.. |DtE| replace:: :class:`~skrub.DatetimeEncoder`
+.. |DatetimeEncoder| replace::
+    :class:`~skrub.DatetimeEncoder`
 
-.. |TV| replace:: :class:`~skrub.TableVectorizer`
+.. |TableVectorizer| replace::
+    :class:`~skrub.TableVectorizer`
+
+.. |OneHotEncoder| replace::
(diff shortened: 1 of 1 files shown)


examples/03_datetime_encoder.py
```diff
@@ -1,89 +1,112 @@
...
 X
 
 ###############################################################################
-# Encoding the data to numerical representations
-# ----------------------------------------------
+# Encoding the features
+# .....................
+#
+# We will construct a |ColumnTransformer| with |make_column_transformer|.
+# We'll want to encode the city names with a |OneHotEncoder|, and the date
```
jovan-stojanovic, line 71:
```suggestion
# We will construct a |ColumnTransformer| in which we will encode
# the city names with a |OneHotEncoder|, and the date
```

examples/03_datetime_encoder.py
```diff
@@ -1,89 +1,112 @@
...
-# been replaced by features for the month, day, hour, and day of the week.
-# Note that the year and minute features have been removed by the encoder
-# because they are constant.
-X_ = encoder.fit_transform(X)
+X_enc = encoder.fit_transform(X)
 encoder.get_feature_names_out()
+
 ###############################################################################
-X_
+# We can see that the encoder is working as expected: the "date.utc" column has
```
jovan-stojanovic, line 95:
```suggestion
# We see that the encoder is working as expected: the "date.utc" column has
```

examples/03_datetime_encoder.py
```diff
@@ -1,89 +1,112 @@
...
+# been replaced by features extracting the month, day, hour, and day of the
+# week information.
+#
+# Noting the year and minute features are not present, this is because they
+# have been removed by the encoder as they are the same for the whole feature.
```
jovan-stojanovic, line 100:
```suggestion
# Note the year and minute features are not present, this is because they
# have been removed by the encoder as they are constant the whole period.
```

examples/03_datetime_encoder.py
```diff
@@ -93,42 +116,63 @@
 pprint(table_vec.get_feature_names_out())
 
 ###############################################################################
-# If we want the day of the week, we can just replace |TV|'s default parameter:
+# If we want to customize the |DatetimeEncoder| inside the |TableVectorizer|,
+# we can replace its default parameter with a new, custom instance:
+#
+# Here, for example, we want to extract the day of the week (which it doesn't
+# do by default).
```
jovan-stojanovic, line 123:
```suggestion
# Here, for example, we want it to extract the day of the week.
```
Obvious from above, less is more :)

examples/03_datetime_encoder.py
```diff
@@ -219,9 +267,10 @@
 ###############################################################################
 # Feature importances
```
jovan-stojanovic, line 268:
```suggestion
# Features importance
```
I think this is better

jovan-stojanovic (changes requested):
Thanks for splitting these examples, it will be easier to review. Some small changes here and this is good for me.

jovan-stojanovic (approved):

PR #685, opened 2023-07-31
[WIP] Add compare encoders example

### LINKED ISSUE #139: Comparison to factorization methods (e.g. SVD)

LilianBoulard:
Extracted from #546.

Fixes #139


examples/07_compare_encoders.py
```diff
@@ -0,0 +1,289 @@
+"""
+.. _example_compare_encoders
+
+============================================
+Understanding and comparing encoding methods
+============================================
+
+In this example, we will compare the mechanisms and use-cases of
+a classical encoding method, the |OneHotEncoder|,
```
GaelVaroquaux, line 9:
```suggestion
In this example, we compare various categorical encoders:
a classical encoding method, the |OneHotEncoder|,
```

examples/07_compare_encoders.py
```diff
@@ -0,0 +1,289 @@
+"""
+.. _example_compare_encoders
+
+============================================
+Understanding and comparing encoding methods
+============================================
```
GaelVaroquaux, line 6:
```suggestion
====================================================
Understanding and comparing categorical encoding methods
====================================================
```

examples/07_compare_encoders.py
```diff
@@ -0,0 +1,289 @@
...
+
+In this example, we will compare the mechanisms and use-cases of
+a classical encoding method, the |OneHotEncoder|,
+and those implemented in *skrub*, that is, the |SimilarityEncoder|,
+the |GapEncoder| and the |MinHashEncoder|.
+
+.. note:
```
GaelVaroquaux, line 13: Maybe a seealso block rather than a note?

examples/07_compare_encoders.py
```diff
@@ -0,0 +1,289 @@
...
+
+.. |MinHashEncoder| replace::
+    :class:`~skrub.MinHashEncoder`
+"""
+
+###############################################################################
+# Encoding dirty categorical variables
+# ------------------------------------
+#
+# As usual, let's first retrieve the dataset:
```
GaelVaroquaux, line 39:
```suggestion
# Let's first retrieve the dataset:
```

examples/07_compare_encoders.py
```diff
@@ -0,0 +1,289 @@
...
+###############################################################################
+# Next up: import the |OneHotEncoder| and apply it to our sample:
+
+from sklearn.preprocessing import OneHotEncoder
+import pandas as pd
+
+one_hot = OneHotEncoder(sparse_output=False)
+sample_enc_ohe = one_hot.fit_transform(sample)
+
+# (make it look nice in Jupyter by wrapping the resulting array in a DataFrame)
```
GaelVaroquaux, line 97: I would rather we use the "set_output" method of scikit-learn transformer. Just call "one_hot.set_output(transform="pandas")" above

examples/07_compare_encoders.py
```diff
@@ -0,0 +1,289 @@
...
+# The |SimilarityEncoder| is a generalization of the |OneHotEncoder|, which
+# uses a string similarity metric to encode the similarity between categories.
+# As we did for the one-hot, let's visualize what that means!
+
+from skrub import SimilarityEncoder
+
+sim_enc = SimilarityEncoder()
+sample_enc_sim = sim_enc.fit_transform(sample)
+
+# (make it look nice in Jupyter by wrapping the resulting array in a DataFrame)
```
GaelVaroquaux, line 134: Same thing. Let's use "set_output"

examples/07_compare_encoders.py
```diff
@@ -0,0 +1,289 @@
...
+heatmap(
+    sample_enc_sim,
+    xticklabels=sim_enc.categories_[0],
+    yticklabels=sample["employee_position_title"],
+    ax=ax,
+)
+plt.tight_layout()
+
+###############################################################################
+# While this is a cool result, it suffers from the high computational cost
```
GaelVaroquaux, line 165: Let's be slightly less casual and avoid the word "cool" :)

GaelVaroquaux (commented):
An early set of comments (I note that there still is the MinHashEcoder to implement, and also maybe a comparison of the performance of all encoders, as we have here https://dirty-cat.github.io/stable/auto_examples/01_dirty_categories.html#sphx-glr-auto-examples-01-dirty-categories-py )

Also, the documentation is not building. I am not sure why

jovan-stojanovic:
> I can continue this work if that's ok, @jovan-stojanovic ?

Yes of course!

jovan-stojanovic:
I will close this due to inactivity but also because the example comparing encoders is less important now that we have new features and that part of it was implemented in the first example. Thanks Lilian for your work!

PR #687, opened 2023-07-31
DOC Rework fuzzy join example

LilianBoulard:
Extracted from #546.


jovan-stojanovic:
I think improvements in the wording proposed here are good, but I still think that the examples should not be separated.

Looking at it again, I think this example's structure should stay as it was: joining with fuzzy_join on 3 tables and then showing that this can be done easily in one cell with the FA. Also the grid search part at the end helped reach a better R2 score.

If we want to make it shorter/simpler, we can do improvements in the text. WDYT?

jovan-stojanovic:
Should fix #696 as well :)

PR #688, opened 2023-07-31
Add `SimilarityEncoder` explanation example

LilianBoulard:
Extracted from #546.


examples/02_explaining_similarity_encoding.py
```diff
@@ -0,0 +1,148 @@
...
+# The problem becomes easier if we can capture relationships between
+# entries.
+#
+# To simplify understanding, we will focus on the column describing the
+# employee's position title:
+
+values = X[["employee_position_title", "gender"]]
+values.insert(0, "current_annual_salary", dataset.y)
+
+sorted_values = values["employee_position_title"].sort_values().unique()
```
jovan-stojanovic, line 63: Do a small print of `sorted_values` so the user does not need to scroll up to see what is the column we are working on

jovan-stojanovic, line 63: Do a small print of `sorted_values` here so the user does not need to scroll up to see what is the column we are working on.

examples/02_explaining_similarity_encoding.py
```diff
@@ -0,0 +1,148 @@
...
+
+###############################################################################
+# In order to robustly embed dirty semantic data, the |SimilarityEncoder|
+# creates a similarity matrix based on an n-gram representation of the data.
+
+from skrub import SimilarityEncoder
+
+similarity_encoder = SimilarityEncoder()
+transformed_values = similarity_encoder.fit_transform(sorted_values.reshape(-1, 1))
```
jovan-stojanovic, line 72: Print a small subset of the similarity matrix to show what it looks like?
You can quickly state below that the values are in between 0 and 1, based on the level of n-gram similarity.

examples/02_explaining_similarity_encoding.py
```diff
@@ -0,0 +1,148 @@
...
+# representation to get an intuition of what the |SimilarityEncoder| is doing:
+
+from sklearn.manifold import MDS
+
+mds = MDS(dissimilarity="precomputed", n_init=10, random_state=42)
+two_dim_data = mds.fit_transform(1 - transformed_values)
+# transformed values lie in the 0-1 range,
+# so 1-transformed_value yields a positive dissimilarity matrix
+print(two_dim_data.shape)
+print(sorted_values.shape)
```
jovan-stojanovic, line 88: If you insert the above explanations, I feel you don't need this, and you can directly do the plot with the points.

examples/02_explaining_similarity_encoding.py
```diff
@@ -0,0 +1,148 @@
...
+ax2.xaxis.tick_bottom()
+ax2.set_title("Similarities across categories")
+f2.colorbar(cax2)
+f2.tight_layout()
+
+###############################################################################
+# As shown in the previous plot, we see that the nearest neighbor of
+# "Communication Equipment Technician"
+# is "Telecommunication Technician", although it is also
+# very close to "Senior Supply Technician": therefore, we grasp the
```
jovan-stojanovic, line 146:
```suggestion
# very close to "Senior Supply Technician": therefore, the |SimilarityEncoder| grasps the
```

examples/02_explaining_similarity_encoding.py
```diff
@@ -0,0 +1,148 @@
...
+# "Communication" part (not initially present in the category as a unique word)
+# as well as the "Technician" part of this category.
```
jovan-stojanovic, line 148:
```suggestion
# as well as the "Technician" part of this category.
#
# In conclusion, the |SimilarityEncoder| works as a generalized upgrade of the |OneHotEncoder|, that takes into account string similarities.
```

Propose a final conclusion (feel free to improve it).

jovan-stojanovic (changes requested):
Very nice! A few small comments.

GaelVaroquaux (commented):
I think that this example should be merged in the one comparing categorical encoders in https://github.com/skrub-data/skrub/pull/685

jovan-stojanovic:
This will merged with #685 for simplicity.

Issue #706, opened 2023-08-18
Adding a frequency encoder


coconutattitude:
### Problem Description

Sometimes I have a feature in my dataset that has 20 or more unique values. I noticed that some of those values could be grouped together with their frequencies, so I would end with 4-6 values instead of 20. This would help later while doing one-hot encoding and clustering. There are the cases when similiar number of frequencies happens because the categories share some kind of similiarity. 

### Feature Description

The chosen column will be processed in the following manner:
1. we will take the value counts of the column
2. we will run pandas.cut() method on the value counts of the column with the bins provided by the user ( in the future those might be quantiles)
3. then we will map the result to the original column which will result with a column with values grouped together basing on their frequency. 

### Additional Context

At work and during kaggle competitions I often end up working with highly cardinal data. Some of the variables have high number of unique values and high variance. I realized that in order to perform further operations and not increase too much the dimensionality of my data, I could group the data based on the frequency of its values before I perform one-hot encoding and clustering. In my context doing syntactic comparison doesn't make sense because I use machine addresses.

Vincent-Maladiere:
Hey @coconutattitude, thanks for this issue and PR! Could you link the Kaggle competitions you mentioned? It's always great to see new use cases!

coconutattitude:
Sure, it was this Kaggle Playground Series competition: https://www.kaggle.com/competitions/playground-series-s3e19/overview

PR #707, opened 2023-08-18
FEAT add the Frequency Encoder

coconutattitude:
Addresses #706 issue. 
Here's a first iteration of the class and it's a proof of concept made during EuroSciPy 2023. Documentation and tests are needed. I'll add quantile strategies in the future. What do you think about this proof of concept?

### DIFF
--- a/examples/08_frequence_encoder.ipynb
+++ b/examples/08_frequence_encoder.ipynb
@@ -1,0 +1,408 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 11,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from skrub.datasets import fetch_road_safety"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "road_safety_dataset = fetch_road_safety()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 34,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import pandas as pd"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 35,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "Vehicle_Type\n",
+       "9.0     271120\n",
+       "1.0      20281\n",
+       "19.0     19612\n",
+       "3.0       9966\n",
+       "5.0       8268\n",
+       "11.0      7758\n",
+       "8.0       7474\n",
+       "21.0      6785\n",
+       "4.0       2446\n",
+       "2.0       2392\n",
+       "20.0      2318\n",
+       "90.0      1740\n",
+       "10.0       906\n",
+       "98.0       790\n",
+       "17.0       650\n",
+       "97.0       301\n",
+       "22.0       226\n",
+       "16.0       119\n",
+       "18.0        20\n",
+       "23.0         9\n",
+       "Name: count, dtype: int64"
+      ]
+     },
+     "execution_count": 35,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "road_safety_dataset.X.Vehicle_Type.value_counts()"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": 15,
+   "metadata": {},
+   "outputs": [
+    {
+     "data": {
+      "text/plain": [
+       "<Axes: xlabel='Vehicle_Type'>"
+      ]
+     },
+     "execution_count": 15,
+     "metadata": {},
+     "output_type": "execute_result"
+    },
+    {
+     "data": {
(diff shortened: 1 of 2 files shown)


Vincent-Maladiere:
Thanks for this feature, @coconutattitude
Here is a quick roadmap to improve this PR before engaging in more formal reviews and discussions.

1. Use pre-commit to linter and standardize your code (it is usually run during `git commit`).
To do so:
	```shell
	pip install -e '.[dev]'
	```
	to install the dependencies used for development, then:
	```shell
	pre-commit init
	```
	to initialize pre-commit.

2. Add a `get_feature_name_out()` method.
	This is useful for inspecting the data used to fit transformers.
	You can take inspiration from the [GapEncoder method](https://github.com/skrub-data/skrub/blob/main/skrub/_gap_encoder.py#L843) or [DatetimeEncoder method](https://github.com/skrub-data/skrub/blob/main/skrub/_datetime_encoder.py#L260)

3. Check for input types, i.e. what happens when X is a numpy array? What happens when it's a pandas dataframe?
	Most encoders/estimators across skrub use the `check_input()` utils function.
	https://github.com/skrub-data/skrub/blob/9d2e64f22f745f83fb95ace168ec1652b883b511/skrub/_datetime_encoder.py#L192-L196
	
4. Add basic tests in a dedicated `_test_frequency_encoder.py` file test. This will help discover edge cases and situations in which the estimator would fail.

5. Convert your example notebook into a Python file using [jupytext](https://jupytext.readthedocs.io/en/latest/using-cli.html)
	```
	pip install jupytext
    ```
	then:
	```
	jupytext --set-formats ipynb,py:percent example/08_frequence_encoder.ipynb
    jupytext --sync example/08_frequence_encoder.ipynb
    ```
    We prefer Python files over notebooks because we can review them in PR more easily, and they are more compact.

6. Enrich your example by adding a regression/classification or even clustering task. Reproducing a similar use case to the one you mentioned IRL with IP addresses could bring value.

Vincent-Maladiere:
Hi @coconutattitude, are you planning to continue this PR?

jeromedockes:
shall we try to revive this? :)

Vincent-Maladiere:
@rcap107 are you interested in continuing this PR once StringEncoder is merged?

rcap107:
I can work on it, yes

rcap107:
This feature was implemented in #1235

Vincent-Maladiere:
> This feature was implemented in #1235

I don't think it was because this PR aims to encode categories using their frequencies; hence, I don't see how the periodic encoder solves this. WDYT?

GaelVaroquaux:
> I don't think it was because this PR aims to encode categories using their frequencies; hence, I don't see how the periodic encoder solves this. WDYT?

I agree. I might have misunderstood this PR.

Given that such encoder is quite popular, I am +1 with the proposed feature.

rcap107:
> > This feature was implemented in #1235
> 
> I don't think it was because this PR aims to encode categories using their frequencies; hence, I don't see how the periodic encoder solves this. WDYT?

My bad, I mixed up this PR and the old discussion on the periodic encoder. This is absolutely unrelated from #1235

rcap107:
Hey, I'm picking up this one. Looking at the possible implementation, there are some points that we should consider:

- Pandas accepts the number of bins, the bin edges or the exact bins to be used, while polars only accepts the bin edges. 
- What should we use as the interface for the user? The number of bins? The bin edges? If it's the number of bins, we need to add the computation of the edges. 
- We need to name the bins ourselves because the two libraries have different conventions
- pl.cut is unstable according to the polars docs 
- Do we want to have both cut and qcut in the same transformer? I think it makes more sense like that

Vincent-Maladiere:
I think that identifying a dataset where this transformer performs well would strongly motivate the inclusion of this feature. It would also help us craft a narrative around it and clarify its role among the other encoders. If I recall correctly, OP was using it on IP addresses.

Vincent-Maladiere:
> What should we use as the interface for the user? The number of bins? The bin edges? If it's the number of bins, we need to add the computation of the edges.

I would go for the number of bins, which is easier to set and more intuitive than bin edges

> Do we want to have both cut and qcut in the same transformer? I think it makes more sense like that

Do you mean cut for pandas and qcut polars?

rcap107:
> I think that identifying a dataset where this transformer performs well would strongly motivate the inclusion of this feature. It would also help us craft a narrative around it and clarify its role among the other encoders. If I recall correctly, OP was using it on IP addresses.

Good point, I do not have good suggestions for that unfortunately :sweat_smile: 

> > What should we use as the interface for the user? The number of bins? The bin edges? If it's the number of bins, we need to add the computation of the edges.
> 
> I would go for the number of bins, which is easier to set and more intuitive than bin edges

I agree

> 
> > Do we want to have both cut and qcut in the same transformer? I think it makes more sense like that
> 
> Do you mean cut for pandas and qcut polars?

No, I mean having both uniform and quantile bins in the same transformer, rather than having a transformer that does uniform binning and one that does quantile binning. There would be a flag that selects the transformer to use. Both pandas and polars have a qcut function, so it's just a matter of deciding where to put it.

rcap107:
Actually, doesn't the KBinsDiscretizer already do most of what this feature needs?

Vincent-Maladiere:
> Actually, doesn't the KBinsDiscretizer already do most of what this feature needs?

Yes indeed! The only drawback I see is that the KBinsDiscretizer converts Pandas and Polars dataframes into NumPy arrays, which creates a copy and has a larger memory footprint. @GaelVaroquaux, should we make our tiny KBinsDiscretizer with skrub's sauce?

GaelVaroquaux:
> Yes indeed! The only drawback I see is that the KBinsDiscretizer converts Pandas and Polars dataframes into NumPy arrays, which creates a copy and has a larger memory footprint. @GaelVaroquaux, should we make our tiny KBinsDiscretizer with skrub's sauce?

Let's say not for now, because we are not sure how important the usecase is. But if the usecase pops up in datasets and examples, why not.

rcap107:
Closing this because it's not been worked on or talked about for a very long time

Issue #673, opened 2023-07-20
Deduplicate fails when there is not enough variations


LeoGrin:
### Describe the bug

Using `deduplicate` on a list with not enough unique values makes it fail in an uninformative way.

### Steps/Code to Reproduce

```
from skrub import deduplicate

deduplicate(["black", "white", "black", "black", "blac"])

deduplicate(["black", "white", "black", "black", "blac"])
```

### Expected Results

No error is thrown or an informative error is thrown

### Actual Results

for the first example
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[19], line 3
      1 from skrub import deduplicate
----> 3 deduplicate(["black", "white", "black", "black", "blac"])

File [~/VSCProjects/skrub/skrub/_deduplicate.py:248](https://file+.vscode-resource.vscode-cdn.net/Users/leo/VSCProjects/skrub/benchmarks/~/VSCProjects/skrub/skrub/_deduplicate.py:248), in deduplicate(data, n_clusters, ngram_range, analyzer, method, n_jobs)
    246 Z = linkage(distance_mat, method=method, optimal_ordering=True)
    247 if n_clusters is None:
--> 248     n_clusters = _guess_clusters(Z, distance_mat, n_jobs)
    249 clusters = fcluster(Z, n_clusters, criterion="maxclust")
    251 translation_table = _create_spelling_correction(unique_words, counts, clusters)

File [~/VSCProjects/skrub/skrub/_deduplicate.py:87](https://file+.vscode-resource.vscode-cdn.net/Users/leo/VSCProjects/skrub/benchmarks/~/VSCProjects/skrub/skrub/_deduplicate.py:87), in _guess_clusters(Z, distance_mat, n_jobs)
     85 # silhouette score needs a redundant distance matrix
     86 redundant_dist = squareform(distance_mat)
---> 87 silhouette_scores = Parallel(n_jobs=n_jobs, prefer="processes")(
     88     delayed(_get_silhouette_avg)(Z, n_clust, redundant_dist)
     89     for n_clust in n_clusters
     90 )
     91 return n_clusters[np.argmax(silhouette_scores)]

[...]

PR #715, opened 2023-08-21
Use RandomState rng instead of np.random in `make_duplicate()`

Vincent-Maladiere:
***What does this PR implement?***
PR #700 experienced what seems to be [a rare error](https://github.com/skrub-data/skrub/actions/runs/5928122276/job/16073058088?pr=700) from a non-deterministic function, making the test unstable.

***What changes***
Replace np.random.choice with the random state provided in input to make the test stable.

***Additional comments***
This rare error might mean the deduplicate function returns unexpected results when the problem is ill-conditioned. This instability will need further investigation.

### DIFF
--- a/skrub/datasets/_generating.py
+++ b/skrub/datasets/_generating.py
@@ -52,7 +52,7 @@ def make_deduplication_data(
         # Randomly pick with which character to replace
         replacements = [
             string.ascii_lowercase[i]
-            for i in np.random.choice(np.arange(26), len(mis_idx)).astype(int)
+            for i in rng.choice(np.arange(26), len(mis_idx)).astype(int)
         ]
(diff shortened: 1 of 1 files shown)


jovan-stojanovic:
LGTM, I agree. Will merge this so that the test don't fail. 

There is issue #673 for deduplicate, which is still open and will need more investigation.

jovan-stojanovic (approved):
LGTM, I agree. Will merge this so that the test don't fail.

There is issue https://github.com/skrub-data/skrub/issues/673 for deduplicate, which is still open and will need more investigation.

Vincent-Maladiere:
Ok, thanks for pointing out this issue!

Issue #624, opened 2023-07-03
Readme shields display is broken


LilianBoulard:
### Describe the bug

Shields are not displayed on the README rendering on GitHub.

### Steps/Code to Reproduce

Check out the [README rendering](https://github.com/skrub-data/skrub/blob/main/README.rst).

### Expected Results

The shields should be displayed, like [dirty_cat](https://github.com/dirty-cat/dirty_cat).

![image](https://github.com/skrub-data/skrub/assets/33222654/c55b712c-32df-4c5a-8e69-f9cd3c239dd6)

### Actual Results

Shields are not visible:
![image](https://github.com/skrub-data/skrub/assets/33222654/98aea0f2-9015-48af-be8e-c4e21020c5e1)

### Versions

```shell
Main branch
```

PR #721, opened 2023-08-24
Fix Badges on README.rst

Vincent-Maladiere:
***What does this PR implement?***
Fix https://github.com/skrub-data/skrub/issues/624

***What changes***
The centering directive needed a line break with the badges to render. After doing that, the rendering wasn't even centered, so I removed the directive entirely.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -6,8 +6,8 @@
    :width: 50 %
    :alt: skrub logo
 
-.. class:: center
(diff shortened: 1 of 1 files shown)


jovan-stojanovic:
Looks good except that the python versions are now missing :thinking:

Vincent-Maladiere:
[It is now fixed](https://img.shields.io/pypi/pyversions/skrub), by adding classifiers metadata to the project on Pypi :)

Issue #704, opened 2023-08-18
Remove warnings from 01_dirty_categories.py


kazmierczakP:
### Describe the issue linked to the documentation

1. In the first notebook (Example 1) i get warnings for one-hot encoder:
```
/home/circleci/project/miniconda/envs/testenv/lib/python3.10/site-packages/sklearn/preprocessing/_encoders.py:972: FutureWarning: `sparse` was renamed to `sparse_output` in version 1.2 and will be removed in 1.4. `sparse_output` is ignored unless you leave `sparse` to its default value.
```

and additionally for the 'minhash' encoder:
```
/home/circleci/project/miniconda/envs/testenv/lib/python3.10/site-packages/sklearn/base.py:464: UserWarning: X does not have valid feature names, but MinHashEncoder was fitted with feature names
```

![image](https://github.com/skrub-data/skrub/assets/61516624/0957eac3-3594-414c-9147-6335924e2398)

2. Why do we need this in describing 'Employee Position Title' column:
```python
import numpy as np
np.unique(y)
```

Vincent-Maladiere:
Congrats, @kazmierczakP for this PR! 🎉

PR #725, opened 2023-08-29
MAINT Fix `feature_name` warning during `transform` for `MinHashEncoder`

Vincent-Maladiere:
**References**
Addresses https://github.com/skrub-data/skrub/issues/704

**What changes**
https://github.com/skrub-data/skrub/pull/647 introduced a warning upon running the `transform` method from `MinHashEncoder`. This is due to `check_input(X)` returning a numpy array, which we pass to `self._check_feature_names`. This latter method raises a warning because it's expecting a Pandas DataFrame.

**Alternative solutions**
Remove `_check_feature_names` entirely, but I wouldn't recommend that.

**Additionnal comment**
Overall, I'm concerned that skrub is becoming increasingly more coupled with scikit-learn private methods. Changes made to these private methods (like https://github.com/skrub-data/skrub/pull/722, even though https://github.com/skrub-data/skrub/pull/675 is an attempt to decouple stuff) take time to fix in skrub, and we just don't have enough maintainers as of today, IMHO.

WDYT @GaelVaroquaux and @glemaitre ?

### DIFF
--- a/skrub/_minhash_encoder.py
+++ b/skrub/_minhash_encoder.py
@@ -275,8 +275,8 @@ class MinHashEncoder(TransformerMixin, BaseEstimator):
             Transformed input.
         """
         check_is_fitted(self, "hash_dict_")
-        X = check_input(X)
         self._check_feature_names(X, reset=False)
+        X = check_input(X)
         self._check_n_features(X, reset=False)
         if self.minmax_hash:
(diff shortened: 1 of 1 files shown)


LeoGrin (approved):

Vincent-Maladiere:
Merging after IRL discussion with @glemaitre

PR #727, opened 2023-08-31
Faster example 07_multiple_key_join

GaelVaroquaux:
- Fixes rst
- Makes 07_multiple_key_join faster (addresses  #729 )

### DIFF
--- a/examples/07_multiple_key_join.py
+++ b/examples/07_multiple_key_join.py
@@ -48,7 +48,7 @@ from skrub.datasets import fetch_figshare
 
 flights = fetch_figshare("41771418").X
 # Sampling for faster computation.
-flights = flights.sample(50_000, random_state=1, ignore_index=True)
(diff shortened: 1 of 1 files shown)


Vincent-Maladiere (approved):

GaelVaroquaux:
The exemple now takes 11mn. It's still long, but better. I think that we should accept the loss in prediction accuracy.

Vincent-Maladiere:
What's taking so long in this example? Is this the quadratic time for nearest neighbors?

GaelVaroquaux:
> What's taking so long in this example? Is this the quadratic time for nearest neighbors?

Yes, I think so. I hope that when we have the InterpJoin this will be much faster.

jovan-stojanovic (approved):
LGTM. Actually it took 40min with 50 000 before #680 was merged so this is much better

Issue #691, opened 2023-08-03
Pandas dtype of nan error in fuzzy_join


jovan-stojanovic:
### Describe the bug

Pandas nightly dependencies check error: 
```
FAILED skrub/tests/test_fuzzy_join.py::test_drop_unmatched - FutureWarning: Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value 'nan' has dtype incompatible with int64, please explicitly cast to a compatible dtype first.
```
We should check what the drop_unmatched parameter is doing.

### Steps/Code to Reproduce

Use `test_drop_unmatched` from [skrub/tests/test_fuzzy_join.py](https://github.com/skrub-data/skrub/blob/main/skrub/tests/test_fuzzy_join.py#L143)

### Expected Results

No error is thrown

### Actual Results

```
FAILED skrub/tests/test_fuzzy_join.py::test_drop_unmatched - FutureWarning: Setting an item of incompatible dtype is deprecated and will raise in a future error of pandas. Value 'nan' has dtype incompatible with int64, please explicitly cast to a compatible dtype first.
```

### Versions

```shell
Nightly dependencies
```

PR #728, opened 2023-08-31
MAINT Fix pandas nightly

Vincent-Maladiere:
**Reference**
Fix https://github.com/skrub-data/skrub/issues/691

**What does this PR implement?***
Pandas is deprecated setting values to columns of different types. In particular, setting np.nan across columns that are not float is deprecated.

We now need to [convert the dataframe to nullable dtypes](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.convert_dtypes.html#pandas.DataFrame.convert_dtypes), accepting pd.NA as values.

### DIFF
--- a/skrub/_fuzzy_join.py
+++ b/skrub/_fuzzy_join.py
@@ -506,9 +506,11 @@ def fuzzy_join(
     if drop_unmatched:
         df_joined.drop(columns=["fj_idx"], inplace=True)
     else:
-        idx = df_joined.index[df_joined["fj_nan"] == 1]
-        if len(idx) != 0:
-            df_joined.iloc[idx, df_joined.columns.get_loc("fj_idx") :] = np.NaN
+        mask_na = df_joined["fj_nan"] == 1
+        if mask_na.sum() > 0:
+            right_cols = df_joined.columns[df_joined.columns.get_loc("fj_idx") :]
+            df_joined[right_cols] = pd.DataFrame.convert_dtypes(df_joined[right_cols])
+            df_joined.loc[mask_na, right_cols] = pd.NA
         df_joined.drop(columns=["fj_idx", "fj_nan"], inplace=True)
 
     if return_score:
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
I don't think that we test on pandas nightly. Is this a thing that we could do?

Vincent-Maladiere:
I think we do, and [last run failed](https://github.com/skrub-data/skrub/actions/runs/6044774661).

This is defined here:
https://github.com/skrub-data/skrub/blob/97c459761a5e5e08398c38b1f6542d115e3524c6/.github/workflows/testing.yml#L92-L111

skrub/_fuzzy_join.py
```diff
@@ -506,9 +506,11 @@ def fuzzy_join(
     if drop_unmatched:
         df_joined.drop(columns=["fj_idx"], inplace=True)
     else:
-        idx = df_joined.index[df_joined["fj_nan"] == 1]
-        if len(idx) != 0:
-            df_joined.iloc[idx, df_joined.columns.get_loc("fj_idx") :] = np.NaN
+        mask_na = df_joined["fj_nan"] == 1
+        if mask_na.sum() > 0:
```
GaelVaroquaux, line 510:
```suggestion
        if mask_na.any():
```
I would rather use "any", rather than "sum". It's more readable, and it's also faster (though we are in fast operations):
```
In [2]: a = np.zeros(10_000, dtype=bool)

In [3]: a[100] = 1

In [4]: %timeit np.sum(a) > 0
7.13 µs ± 58 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [5]: %timeit np.any(a)
2.7 µs ± 30.6 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)

In [6]: %timeit a.any()
1.03 µs ± 7.16 ns per loop (mean ± std. dev. of 7 runs, 1,000,000 loops each)

In [7]: %timeit a.sum() > 0
5.8 µs ± 376 ns per loop (mean ± std. dev. of 7 runs, 100,000 loops each)
```

GaelVaroquaux (approved):
LGTM with a small suggestion.

I'll commit the suggestion and merge once the tests have run

PR #735, opened 2023-09-06
MAINT Remove `sklearn.utils.parse_version` from `skrub.utils`

Vincent-Maladiere:
**Context**
For some reason, we import the `parse_version` function from `sklearn.utils` in `skrub.utils`, without using it. Then we import it from skrub instead of from scikit-learn across the project, which seems odd.

- Replace the skrub import by the sklearn import across the project
- Remove the import from sklearn in `skrub.utils`


Vincent-Maladiere:
Let's merge this since it is a very small fix

Issue #679, opened 2023-07-27
`TableVectorizer` fails on sparse Dataframes


LeoGrin:
### Describe the bug

`TableVectorizer` fails when fitting on dataframes with sparse dtypes.

### Steps/Code to Reproduce

```
from skrub import TableVectorizer
import pandas as pd

df = pd.DataFrame({
    'a': [1, 2, 3, 4, 5]
})
# convert to sparse
df['a'] = pd.arrays.SparseArray(df['a'])
tbl_vec = TableVectorizer()
tbl_vec.fit(df)
```

### Expected Results

No error is thrown.

### Actual Results

```
Unexpected exception formatting exception. Falling back to standard exception
Traceback (most recent call last):
  File "/Users/leo/mambaforge/envs/skrub/lib/python3.10/site-packages/IPython/core/interactiveshell.py", line 3508, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "/var/folders/fy/__8z8cpn6gs04465sq9g1nq80000gn/T/ipykernel_7311/2936428833.py", line 10, in <module>
    tbl_vec.fit(df)
  File "/Users/leo/mambaforge/envs/skrub/lib/python3.10/site-packages/sklearn/compose/_column_transformer.py", line 694, in fit
    self.fit_transform(X, y=y)
  File "/Users/leo/mambaforge/envs/skrub/lib/python3.10/site-packages/sklearn/utils/_set_output.py", line 140, in wrapped
    data_to_wrap = f(self, X, *args, **kwargs)
  File "/Users/leo/VSCProjects/skrub/skrub/_table_vectorizer.py", line 607, in fit_transform
  File "/Users/leo/VSCProjects/skrub/skrub/_table_vectorizer.py", line 130, in _replace_false_missing
    "NaN",
  File "/Users/leo/mambaforge/envs/skrub/lib/python3.10/site-packages/pandas/core/frame.py", line 5582, in replace
    return super().replace(
  File "/Users/leo/mambaforge/envs/skrub/lib/python3.10/site-packages/pandas/core/generic.py", line 7383, in replace
    new_data = self._mgr.replace_regex(
  File "/Users/leo/mambaforge/envs/skrub/lib/python3.10/site-packages/pandas/core/internals/managers.py", line 484, in replace_regex
    return self.apply("_replace_regex", **kwargs, using_cow=using_copy_on_write())
[...]

PR #737, opened 2023-09-08
MAIN Improve the error message for Pandas sparse arrays in TableVectorizer

Vincent-Maladiere:
**References**
Addresses https://github.com/skrub-data/skrub/issues/679

`TableVectorizer` currently doesn't support sparse matrices. Until then, we should also prevent users from using Pandas sparse arrays.

This PR suggests that, during the `_check_input` call, we iterate on dataframe columns to look for the "sparse" accessor. We raise an error when at least one column is sparse.

**Alternative solution**
AFAIK, there is no way to identify a sparse column within a Pandas dataframe without iterating on the columns.

cc @LeoGrin

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -679,6 +679,16 @@ class TableVectorizer(ColumnTransformer):
                 "X.toarray() to convert to a dense numpy array."
             )
 
+        # Check Pandas sparse arrays
+        is_sparse_col = [hasattr(X[col], "sparse") for col in X.columns]
+        if any(is_sparse_col):
+            sparse_cols = X.columns[is_sparse_col]
+            raise TypeError(
+                f"Columns {sparse_cols!r} are sparse Pandas series, but dense "
+                "data is required. Use df[col].sparse.to_dense() to convert "
+                "a series from sparse to dense."
+            )
+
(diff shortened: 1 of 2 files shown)


skrub/_table_vectorizer.py
```diff
@@ -679,6 +679,16 @@ def _check_X(self, X, reset):
...
+        is_sparse_col = [hasattr(X[col], "sparse") for col in X.columns]
+        if any(is_sparse_col):
+            sparse_cols = X.columns[is_sparse_col]
+            raise TypeError(
+                f"Columns {sparse_cols!r} are sparse Pandas series, but dense "
+                "data is required. Use df[col].sparse.to_dense() to convert "
+                "a series from sparse to dense."
+            )
+
         if not isinstance(X, pd.DataFrame):
```
LeoGrin, line 692: I think moving your code after the conversion to pandas would fix the errors you have.

LeoGrin (changes requested):
Thanks!
> AFAIK, there is no way to identify a sparse column within a Pandas dataframe without iterating on the columns.

I don't see any either.

LeoGrin (approved):

jovan-stojanovic (approved):

PR #738, opened 2023-09-11
[DONT MERGE] add a package data file

jeromedockes:
this is a dummy PR to run the CI

I have the impression that at the moment:

- the way we run tests in the CI invokes pytest from the repo root with `python -m pytest`, which adds the cwd to the PYTHONPATH, resulting in pytest importing skrub from the repo rather than the installed package
- we have no mechanism in place to copy package data to the installation directory, the VERSION.txt gets included as a special case

This is what I see locally: if I add a package data file without adding it to a MANIFEST.in or listing it in setup.py, and install skrub, `build_tools/test.sh` runs without errors even though the data file is not copied to the installation directory. However, if I edit `test.sh` to replace `python -m pytest` with `pytest`, the tests fail

I'm opening this dummy PR to see if the same thing happens in the CI. If it does, the tests would pass now, then fail when I push another commit with the modified pytest invokation

### DIFF
--- a/build_tools/github/test.sh
+++ b/build_tools/github/test.sh
@@ -1,3 +1,3 @@
 #!/bin/bash -x
 
-python -m pytest --pyargs skrub --cov=skrub -n auto
+pytest --pyargs skrub --cov=skrub -n auto
--- a/skrub/__init__.py
+++ b/skrub/__init__.py
@@ -32,4 +32,16 @@ __all__ = [
     "TargetEncoder",
     "deduplicate",
     "compute_ngram_distance",
+    "load_package_data",
 ]
+
+
+def load_package_data() -> str:
+    """Dummy function that loads some package data.
+
+    Returns
+    -------
+    data: str
+        Some data.
+    """
+    return _Path(__file__).parent.joinpath("data_file.txt").read_text("utf-8")
--- a/skrub/data_file.txt
+++ b/skrub/data_file.txt
@@ -1,0 +1,1 @@
+some package data
--- a/skrub/tests/test_package_data.py
+++ b/skrub/tests/test_package_data.py
@@ -1,0 +1,5 @@
+import skrub
+
+
+def test_load_package_data():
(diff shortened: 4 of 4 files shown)


jeromedockes:
As suspected, the current CI configuration would not catch errors in the build of the installable package because it tests the code in the repository rather than the installed one -- the tests pass in the 1st commit and fail in the 2nd one.

we can fix the pytest call, and we may also consider:

moving to an temporary empty directory before running the tests in the CI, to be sure results are not polluted by files or directories that happen to be in the cwd

using a so-called "src layout" which adds a buffer src/ directory so that the package doesn't sit at the root of the repository:
```
skrub
├── pyproject.toml
└── src
    └── skrub
        └── __init__.py
```


see also

https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code


re. data files, I did not find a clear explanation in setuptool's [docs on package data](https://setuptools.pypa.io/en/latest/userguide/datafiles.html#include-package-data) guaranteeing that our VERSION.txt would be packaged, so we might want to take one of the steps mentioned in the documentation to ensure that it is, even in future versions of setuptools

Vincent-Maladiere (commented):
Thanks for investigating this issue! Which way to include data files would you advocate for? And what name would you give to the "buffer" folder? `python`, `src`?

jeromedockes:
> Which way to include data files would you advocate for?

as we have just the one VERSION.txt ATM maybe just a manifest, and we
can always revisit if it becomes inconvenient later

> And what name would you give to the "buffer" folder? `python`, `src`?

src seems to be the typical name according to the python packaging user
guide:

https://packaging.python.org/en/latest/discussions/src-layout-vs-flat-layout/

but I'm not saying we should definitely do that, maybe we can discuss it
briefly at the next scrub meeting?

in the meanwhile we can make sure tests run in an empty directory and
import the installed package in the CI; it's probably a good thing to
do anyway. I can open a PR if it helps

PR #739, opened 2023-09-11
Misc details

jeromedockes:
this PR is to keep track of small details that surprise me as I look around the documentation. I suspect many are misunderstandings on my part so please LMK which changes can be kept and which should be removed!


CONTRIBUTING.rst
```diff
@@ -103,8 +103,7 @@ Let maintainers know whether:
   is needed! To go further, refer to the section
   :ref:`writing-your-first-pull-request`.
 - **You won't write the code**, in which case a
-  developer can start working on it. Note however that maintainers
-  are **volunteers**, and therefore cannot guarantee how much time
+  developer can start working on it. Note however that we cannot guarantee how much time
```
jeromedockes, line 106: I wouldn't hide the fact that the library also has full-time maintainers as it's probably a good thing

skrub/_gap_encoder.py
```diff
@@ -605,6 +605,10 @@ class GapEncoder(TransformerMixin, BaseEstimator):
        with the Kullback-Leibler divergence as loss, and a Gamma prior on `H`.
        We thus optimize `H` and `W` with the multiplicative update method.
 
+    "Gap" stands for "Gamma-Poisson", the families of distributions used in the
```
jeromedockes, line 608: save following an extra link for the mildly curious

skrub/datasets/_fetching.py
```diff
@@ -190,15 +190,6 @@ def _fetch_openml_dataset(
     if not details_gz_path.is_file() or not features_gz_path.is_file():
         # If the details file or the features file don't exist,
         # download the dataset.
-        warnings.warn(
```
jeromedockes, line 193: I believe scikit-learn's openml fetcher makes the download in a temporary directory, then renames it if the checksum is correct to avoid corrupted downloads, so this warning may be superfluous.

there is a similar warning in another function that does not follow the same strategy, I'll open another issue about fixing the fetcher code before removing that second warning

doc/vision.rst
```diff
@@ -52,9 +52,7 @@ From shorter term to longer term:
 - Add a Join-aggregator object, to do feature augmentation on one-to-many
   correspondances
```
glemaitre, line 53:
```suggestion
  correspondences
```

doc/vision.rst
```diff
@@ -52,9 +52,7 @@ From shorter term to longer term:
...
 
-- fuzzy joining on datetime values
-
-- Support polars?
+- Support polars
 
 - Support time series (eg in the aggregations)
```
glemaitre, line 58: Just below "correspondances" -> "correspondences

examples/01_dirty_categories.py
```diff
@@ -384,9 +384,12 @@
 plt.show()
 
 ###############################################################################
-# We can deduce from this data that the three factors that define the
-# most the salary are: being hired for a long time, being a manager, and
-# having a permanent, full-time job :)
+# We can see that features such the time elapsed since being hired, having a
+# full-time employment, and the position, seem to be the most informative for
+# prediction. However, feature importances must not be over-interpreted -- they
+# capture statistical associations `rather than causal effects <https://en.wikipedia.org/wiki/Correlation_does_not_imply_causation>`_.
```
glemaitre, line 390:
```suggestion
# capture statistical associations `rather than causal effects
# <https://en.wikipedia.org/wiki/Correlation_does_not_imply_causation>`_.
```

examples/01_dirty_categories.py
```diff
@@ -384,9 +384,12 @@
...
+# Moreover, the fast feature importance method used here is less reliable than
```
glemaitre, line 391: Instead of reliable, I would use already mentioned that "it suffers from biases favouring features with larger cardinality".

I would also link to the scikit-learn example:

https://scikit-learn.org/dev/auto_examples/inspection/plot_permutation_importance.html#sphx-glr-auto-examples-inspection-plot-permutation-importance-py

skrub/_gap_encoder.py
```diff
@@ -605,6 +605,10 @@ class GapEncoder(TransformerMixin, BaseEstimator):
...
+    topic model to model the importance of topics in a document (Gamma), and
```
glemaitre, line 609: "in the topic model to model" is a bit weird. Maybe "used in topic modelling to model"

skrub/_joiner.py
```diff
@@ -20,9 +20,9 @@ class Joiner(TransformerMixin, BaseEstimator):
 
     The principle is as follows:
 
-    1. The main table and the key column name are provided at initialisation.
-    2. The auxiliary tables are provided for fitting, and will be joined
-       sequentially when Joiner.transform is called.
+    1. The auxiliary tables and the key column names are provided at initialisation.
+    2. The main table is provided for fitting, and the auxiliary tables will be joined
+       to the main table sequentially when Joiner.transform is called.
```
glemaitre, line 25:
```suggestion
       to the main table sequentially when `Joiner.transform` is called.
```

glemaitre (commented):
Just nitpicking.

glemaitre (approved):

Issue #547, opened 2023-04-17
Fuzzy joining on mixed typed columns is with unequal weights


jovan-stojanovic:
When doing [mixed joins (on numerical + string columns) in the fuzzy_join](https://github.com/dirty-cat/dirty_cat/pull/530) our columns are not equally weighted. The string columns values will have to be re-adapted.

Suppose we want to join on two columns - one is string (e.g. person name) and the second numerical (e.g. person age).
What we currently do is the following:

   - we encode the string column using a HashingVectorizer+TF-IDF. We get an array with 2 ** 20 columns of encoded values.
   - we scale the numerical column using a StandardScaler + MinMaxScaler and we place it in a second array.

Then, we stack the two arrays into one and then do the usual nearest neighbor search to find the best match.

However, it is clear that the columns are heterogeneous, and that 2*20 columns outputted for strings may weight more than the one column for nums.

Potential solutions discussed with @GaelVaroquaux. @ogrisel and @Vincent-Maladiere:
- if array is dense, subtract the column-wise mean (to ensure stability) and take Frobenius norm
- if array is sparse, take directly the Frobenius norm (substracting the mean would make it dense)
- maybe compute the average pairwise distance using [sklearn.pairwise](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise_distances.html)?

Vincent-Maladiere:
Hi @jovan-stojanovic, @GaelVaroquaux, @ogrisel, after further diving into this issue, we might not need to use the Frobenius norm or pairwise distance at all!

The concern of having equal representativity for each column's type (str, num) stems from the downstream kNN tasks with Euclidean distance. For example, let's say we run a fuzzy join on one string column and one numerical column. Below is the Euclidean distance between two mixed embeddings of index $i$ and $j$:

$$d_{ij} = \sqrt{\sum_{k\in \mathcal{S}_{ij}} (x_{ik} - x_{jk})^2 + (x_{ik_{num}} - x_{jk_{num}})^2}$$

Where $S_{ij}$ is the set of string embedding columns where $x_i$  and  $x_j$ are not both zero and $k_{num}$ is the column with numerical values.

Ideally, to equally take into account both numerical and string columns, we would like to ensure: 

[...]

PR #742, opened 2023-09-20
[MRG] FEA Add interpolation join

jeromedockes:
For now it is implemented with pandas. I have a polars version that I can integrate once #733 is merged. It is fairly straightforward to implement with the current specification of the dataframe api, but it seems polars and pandas implement an older specification


Vincent-Maladiere (commented):
Hey @jeromedockes, thank you for this PR! I have two main points to address:

1. `InterpolationJoin` is a generalization of `Joiner`. Should we expose both methods, or restrain the choice to `InterpolationJoin` only? This would make our API less confusing for the user.
2. If we keep the two classes separated, should `Joiner` inherit from `InterpolationJoin`? Note that, in Joiner, we use l2 regularization for both numerical columns (via `StandardScaler`) and categorical (via `TfIdfTransformer`), so that each encoded feature has a similar weight for the 1-nearest-neighbor (see [this discussion](https://github.com/skrub-data/skrub/issues/547) for more details). Let's also note that we won't have this scaling issue with boosting trees.

GaelVaroquaux:
> InterpolationJoin is a generalization of Joiner. Should we expose both methods

I think that we should expose the two methods. "InterpolationJoin" is further away to a join then the typical definition in DB (if there are only exact matched, Joiner boils down to a standard join).

> should Joiner inherit from InterpolationJoin?

I would say: only if it simplifies the codebase.

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,173 @@
...
+# Load weather data
+# -----------------
+# We join the table containing the measurements to the table that contains the weather stations’ latitude and longitude.
+# We subsample these large tables for the example to run faster.
+
+from skrub.datasets import fetch_figshare
+import pandas as pd
+
+weather = fetch_figshare("41771457").X
+weather = weather.sample(100_000, random_state=0, ignore_index=True)
```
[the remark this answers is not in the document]
jeromedockes: I used `sample` because that is what is done in the "spatial join" [example](https://skrub-data.org/stable/auto_examples/07_multiple_key_join.html#sphx-glr-auto-examples-07-multiple-key-join-py). I think it's a good idea to take a random sample rather than the first rows in case the dataset is sorted by date or weather station. But I can check whether that is the case

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,173 @@
...
+stations = fetch_figshare("41710524").X
+weather = pd.merge(stations, weather, on="ID").loc[
```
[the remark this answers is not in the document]
jeromedockes: 1. no reason, I just have the habit to always use either `.loc` or `.iloc` with pandas to reduce the number of pandas' surprising quirks -- but in this case I agree it's not necessary so I removed it

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,173 @@
...
+# So we will only add the temperature to our flights table.
+
+
+######################################################################
+# Loading the flights table
+# -------------------------
+# We load the flights table and join it to the airports table using the flights’ "Origin" which refers to the departure airport’s IATA code.
+# We use only a subset to speed up the example.
+
+flights = fetch_figshare("41771418").X.loc[
```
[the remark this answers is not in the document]
jeromedockes: here also, I followed the other example and took a random sample, I can check if it's necessary

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,173 @@
...
+join = joiner.transform(flights)
+join.iloc[0]
+
+######################################################################
+# Sanity checks
+# -------------
+# This time we do not have a ground truth for the temperatures.
+# We can perform a few basic sanity checks.
+
+state_temperatures = join.groupby("state")["TMAX"].mean().sort_values()
```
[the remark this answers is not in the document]
jeromedockes: when I render the docs it shows the whole column and I have to scroll down to see the tail (and the display does not make it obvious that I *can* scroll down) so I think it's still worth printing the head and tail separately, WDYT?

Vincent-Maladiere:
> I think that we should expose the two methods. "InterpolationJoin" is further away to a join then the typical definition in DB (if there are only exact matches, Joiner boils down to a standard join).

Oh right, I see the nuance now, you couldn't reproduce `Joiner` results based on `InterpolateJoiner` directly. How could we make this simpler for the user though? I'm not super comfortable with exposing these two transformers because it creates complexity.

> I would say: only if it simplifies the codebase.

With Jerome's implementation, it could be worth making a POC of `Joiner` using `InterpolateJoiner`. My intuition is that we can make the codebase of `Joiner` simpler while keeping the same logic. WDYT?

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+        The columns in ``right_table`` used for joining. Their number and types
+        must match those of the `left_on` columns in the left table. These
+        columns provide the features for the estimators to be fitted. As for
+        ``left_on``, it is possible to pass a string when using a single column.
+
+    on : list of str, or str
+        Column names to use both `left_on` and `right_on`, when they are the
+        same. Provide either `on` (only) or both `left_on` and `right_on`.
+
+    suffix : str
```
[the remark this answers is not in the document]
jeromedockes: So we should accept suffixes for both the left and right dataframe, as in pandas? btw note `fuzzy_join` has `suffixes`, but `Joiner` doesn't have such a parameter. the AggJoiner PR atm uses `suffix` and applies suffixes to the right tables only. polars uses `suffix`, only for the right table, but only applies it if not applying it would result in a duplicate name (it doesn't check that applying it doesn't cause a duplicate name)

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+        self,
+        right_table,
+        *,
+        left_on=None,
+        right_on=None,
+        on=None,
+        suffix="",
+        regressor=ensemble.HistGradientBoostingRegressor(),
+        classifier=ensemble.HistGradientBoostingClassifier(),
+        vectorizer=TableVectorizer(),
```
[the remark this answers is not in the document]
jeromedockes: in any case we have to make sure to clone them before modifying them, because we don't want to modify the user-provided estimators (if any), so mutability shouldn't be an issue. for clarity, I would say putting them here directly avoids the need to look up what `None` means, so I'm not sure `None` is clearer?

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+        self : InterpolationJoin
+            Returns self.
+        """
+        del data, targets
+        self._check_inputs()
+        X_values = self.vectorizer_.fit_transform(
+            self.right_table.loc[:, self._right_on]
+        )
+        estimators = self._get_estimator_assignments()
+        self.estimators_ = joblib.Parallel(self.n_jobs, verbose=3)(
```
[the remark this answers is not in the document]
jeromedockes: I left joblib pick the default for now, we can see later if we really need to control this parameter

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+        )
+        return assignments
+
+
+def _get_assignments_for_estimator(table, estimator):
+    """Get the groups of columns assigned to a single estimator.
+
+    (which is either the regressor or the classifier)."""
+    if not table.shape[1]:
+        return []
```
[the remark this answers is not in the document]
jeromedockes: I added a comment, LMK if it helps

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+        {"columns": [col], "estimator": estimator} for col in columns_with_nulls
+    ]
+    columns_without_nulls = list(set(table.columns).difference(columns_with_nulls))
+    if columns_without_nulls:
+        assignments.append({"columns": columns_without_nulls, "estimator": estimator})
+    return assignments
+
+
+def _handles_multioutput(estimator):
+    return getattr(estimator, "_get_tags")().get("multioutput", False)
```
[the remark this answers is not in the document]
jeromedockes: I changed it to use `_safe_tags`. I wonder if this function will ever be made public in scikit-learn? it seems like querying a model's capabilities would be useful for users (like skrub), in addition to scikit-learn itself

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+
+def _fit(X_values, right_table, target_columns, estimator):
+    estimator = base.clone(estimator)
+    kept_rows = right_table.loc[:, target_columns].notnull().all(axis=1).to_numpy()
+    X_values = X_values[kept_rows]
+    right_table = right_table[kept_rows]
+    Y = right_table.loc[:, target_columns]
+    Y_values = Y.to_numpy()
+    if Y_values.shape[-1] == 1:
+        Y_values = Y_values.ravel()
```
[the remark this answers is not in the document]
jeromedockes: I added a comment to explain -- the gradient boosting estimator doesn't like a (n, 1)-shaped Y

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+
+
+def _handles_multioutput(estimator):
+    return getattr(estimator, "_get_tags")().get("multioutput", False)
+
+
+def _fit(X_values, right_table, target_columns, estimator):
+    estimator = base.clone(estimator)
+    kept_rows = right_table.loc[:, target_columns].notnull().all(axis=1).to_numpy()
+    X_values = X_values[kept_rows]
```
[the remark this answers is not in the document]
jeromedockes: here the X is a numpy array. the reason is that in case to_numpy() needs to make a copy (for example if the data is not contiguous in the pandas dataframe, or later if it is a polars dataframe), I prefer to do it only once in the parent process

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,328 @@
...
+            joblib.delayed(_predict)(
+                X_values, assignment["columns"], assignment["estimator"]
+            )
+            for assignment in self.estimators_
+        )
+        interpolated_parts = _add_column_name_suffix(interpolated_parts, self.suffix)
+        original_index = left_table.index
+        return pd.concat(
+            [left_table.reset_index(drop=True)] + interpolated_parts, axis=1
+        ).set_index(original_index)
```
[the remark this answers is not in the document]
jeromedockes: that seems like we're reimplementing pd.concat? also it's not strictly the same because when we append many columns 1 by 1 we end up with a very fragmented dataframe:
```python
import pandas as pd

df1 = pd.DataFrame({c: range(3) for c in "ABC"})
df2 = pd.DataFrame({c: range(3) for c in "DEF"})
df3 = pd.concat([df1, df2])
print(df3._mgr.nblocks) # prints 2

df4 = df1.copy()
for col in df2:
    df4[col] = df2[col].values
print(df4._mgr.nblocks) # prints 4
```

(also it will consolidate the blocks if we reach 100 columns, though that probably wouldn't happen a lot)

that probably wouldn't be a problem though. still why do you think it's less readable to use concat and then restore the index?

skrub/_interpolation_join.py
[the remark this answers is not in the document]
jeromedockes: from the skrub meeting: we keep `suffix` that is applied to the right table column names (see #751 )

jeromedockes:
The errors seem unrelated to this PR; rather, I have the impression that numpydoc is choking on non-ascii characters in the SimilarityEncoder's docstring (“”) on Windows

skrub/_interpolation_join.py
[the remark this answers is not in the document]
Vincent-Maladiere: Indeed, it would be nice to create an issue in scikit-learn to ask them.

skrub/_interpolation_join.py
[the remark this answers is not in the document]
Vincent-Maladiere: Oh right sorry I overlooked the `to_numpy()`

skrub/_interpolation_join.py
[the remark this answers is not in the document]
Vincent-Maladiere: Nicely done on the fragmented analysis, I didn't have that in mind. I think that, in this situation, changing the index is error-prone and harder to interpret.
Could we at least agree on having `[left_table.reset_index(drop=True)] + interpolated_parts` on a separate line, for readability and debuggability?

examples/08_interpolation_join.py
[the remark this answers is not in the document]
Vincent-Maladiere: Ok, fair enough

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
...
+######################################################################
+# states with the highest predicted temperatures: Puerto Rico, Virgin Islands, Hawaii, Florida, Louisiana
+state_temperatures.tail()
+
+######################################################################
+# Higher latitudes (farther up north) are colder -- the airports in this dataset are in the United States.
+fig, ax = plt.subplots()
+ax.scatter(join["lat"], join["TMAX"])
+ax.set_xlabel("Latitude (higher is farther north)")
+ax.set_ylabel("TMAX")
```
Vincent-Maladiere, line 157: To remove the useless matplotlib print
```suggestion
ax.set_ylabel("TMAX");
```

skrub/_interpolation_join.py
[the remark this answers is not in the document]
Vincent-Maladiere: I see, then `if table.empty` would be clearer, WDYT?

skrub/tests/test_interpolation_join.py
```diff
@@ -0,0 +1,117 @@
...
+
+@pytest.fixture
+def annual_avg_temp():
+    return pd.DataFrame(
+        {
+            "latitude": [1.2, 0.9, 1.9, 1.7, 5.0, 5.0],
+            "longitude": [0.8, 1.1, 1.8, 1.8, 5.0, 5.0],
+            "avg_temp": [10.0, 11.0, 15.0, 16.0, 20.0, None],
+        }
+    )
```
Vincent-Maladiere, line 23: Do we need the fixtures? We could define the dataframes at the top of the file and use them in the functions without passing them as parameters. scikit-learn does this a lot.

skrub/tests/test_interpolation_join.py
```diff
@@ -0,0 +1,117 @@
...
+        aux, key="A", regressor=KNeighborsRegressor(1)
+    ).fit_transform(main)
+    assert (join["C"].values == [10, 11, 12]).all()
+
+    join = InterpolationJoin(
+        aux, main_key="A", aux_key="rB", regressor=KNeighborsRegressor(1)
+    ).fit_transform(main)
+    assert (join["C"].values == [11, 12, 10]).all()
+
+    with pytest.raises(ValueError):
```
Vincent-Maladiere, line 66: Could you also provide the expected error message as a regex with `pytest.raises(ValueError, match=...)`?

skrub/tests/test_interpolation_join.py
```diff
@@ -0,0 +1,117 @@
...
+    if not with_nulls:
+        annual_avg_temp = annual_avg_temp.fillna(0.0)
+    params = {} if with_vectorizer else {"vectorizer": None}
+    transformed = InterpolationJoin(
+        annual_avg_temp,
+        key=key,
+        regressor=KNeighborsRegressor(2),
+        **params,
+    ).fit_transform(buildings)
+    assert (transformed["avg_temp"].values == [10.5, 15.5]).all()
```
Vincent-Maladiere, line 41: Nitpick: this looks close to `numpy.testing.assert_array_equal`, WDYT?

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,341 @@
...
+        The columns in the main table used for joining. The main table is the
+        argument of ``transform``, to which we add information inferred using
+        ``aux_table``. The column names listed in ``main_key`` will provide the
+        inputs (features) of the interpolators at prediction (joining) time. In
+        the example above, ``main_key`` is ``["latitude", "longitude"]``, which
+        refer to columns in the ``buildings`` table. When joining on a single
+        column, we can pass its name rather than a list: ``"latitude"`` is
+        equivalent to ``["latitude"]``.
+
+    aux_key : list of str, or str
```
Vincent-Maladiere, line 57: Is this a necessary list, or could it be any iterable?

Vincent-Maladiere:
> The errors seem unrelated to this PR; rather, I have the impression that numpydoc is choking on non-ascii characters in the SimilarityEncoder's docstring (“”) on Windows

You can now rebase on main to remove this error.

skrub/_interpolation_join.py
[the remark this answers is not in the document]
jeromedockes: sure, or do you think it is easier to set the index in the dataframes we concatenate before concatenating? it avoids the reset_index and having to store the original index: https://github.com/skrub-data/skrub/pull/742/commits/31cf759c82e7ce0b8f7d148c3534f88c56d5a9ab

skrub/tests/test_interpolation_join.py
jeromedockes: sure, we can do that! one advantage of a fixture is we get a new dataframe each time. with a global variable all tests need to be careful not to modify it otherwise it can cause test failures that show up or not depending on the (unpredictable) order tests are executed in

examples/08_interpolation_join.py
jeromedockes: I like it; the pre-commit formatting style check doesn't

jeromedockes:
as we don't have a consensus on `Join` vs `Joiner` (vs  **`:Joine.`**), shall I rename it `InterpolationJoiner` for now?

skrub/_interpolation_join.py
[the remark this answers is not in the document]
jeromedockes: or always vectorize the key, as is done in the Joiner

Vincent-Maladiere:
You have a docstring error on your estimator:
```
 - GL03: Double line break found; please use only one blank line to separate sections or paragraphs, and do not leave blank lines at the end of docstrings
```

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
+"""
+Interpolation join: infer missing rows when joining two tables
+==============================================================
+
+We illustrate the ``InterpolationJoiner``, which is a type of join where values from the second table are inferred with machine-learning, rather than looked up in the table.
```
jovan-stojanovic, line 5: Best is to link the name to the function (see other examples):
```suggestion
We illustrate the :class:`~skrub.InterpolationJoin`, which is a type of join where values from the second table are inferred with machine-learning, rather than looked up in the table.
```

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
...
+It is useful when exact matches are not available but we have rows that are close enough to make an educated guess -- in this sense it is a generalization of a ``fuzzy_join``.
```
jovan-stojanovic, line 6: Same for fuzzy_join. If the links need to be repeatedly added (as it seems to me) look at the other examples for a way to link them all to one.
```suggestion
It is useful when exact matches are not available but we have rows that are close enough to make an educated guess -- in this sense it is a generalization of a :func:`skrub.fuzzy_join`.
```

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
...
+
+weather = fetch_figshare("41771457").X
+weather = weather.sample(100_000, random_state=0, ignore_index=True)
+stations = fetch_figshare("41710524").X
+weather = stations.merge(weather, on="ID")[
+    ["LATITUDE", "LONGITUDE", "YEAR/MONTH/DAY", "TMAX", "PRCP", "SNOW"]
+]
+
+######################################################################
+# The TMAX is in tenths of degree Celsius -- a TMAX of 297 means the maximum temperature that day was 29.7℃.
```
jovan-stojanovic, line 39: I think this is how we refer to column names in examples (btw we should have a standardized way)
```suggestion
# The `'TMAX'` is in tenths of degree Celsius -- a TMAX of 297 means the maximum temperature that day was 29.7℃.
```

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
...
+######################################################################
+# Sanity checks
+# -------------
+# This time we do not have a ground truth for the temperatures.
+# We can perform a few basic sanity checks.
+
+state_temperatures = join.groupby("state")["TMAX"].mean().sort_values()
+
+######################################################################
+# states with the lowest average predicted temperatures: Alaska, Montana, North Dakota, Washington, Minnesota
```
jovan-stojanovic, line 145: I think this should be capital (see rendering)
```suggestion
# States with the lowest average predicted temperatures: Alaska, Montana, North Dakota, Washington, Minnesota.
```

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
...
+state_temperatures.head()
+
+######################################################################
+# states with the highest predicted temperatures: Puerto Rico, Virgin Islands, Hawaii, Florida, Louisiana
```
jovan-stojanovic, line 149:
```suggestion
# States with the highest predicted temperatures: Puerto Rico, Virgin Islands, Hawaii, Florida, Louisiana.
```

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
...
+
+######################################################################
+# Winter months are colder than spring -- in the north hemisphere January is colder than April
+#
+
+import seaborn as sns
+
+join["month"] = join["Year_Month_DayofMonth"].dt.strftime("%m %B")
+plt.figure()
+sns.barplot(data=join.sort_values(by="month"), y="month", x="TMAX")
```
jovan-stojanovic, line 167: [We have a rendering issue here](https://output.circle-artifacts.com/output/job/289f832a-cccf-41de-85f2-5a9b8b15bdc2/artifacts/0/doc/auto_examples/08_interpolation_join.html#sphx-glr-auto-examples-08-interpolation-join-py). Can we move the plot a bit to the right so that the y-axis labels are seen?

examples/08_interpolation_join.py
```diff
@@ -0,0 +1,179 @@
...
+
+######################################################################
+# Of course these checks do not guarantee that the inferred values in our ``join`` table’s ``TMAX`` column are accurate.
+# But at least the ``InterpolationJoiner`` seems to have learned a few reasonable trends from its training table.
+
```
jovan-stojanovic, line 172: Did you try predicting the flights delay with the new interpolated table? How do the results compare to a simple fuzzy join ?

doc/api.rst
```diff
@@ -33,6 +33,13 @@ This page lists all available functions and classes of `skrub`.
 
    Joiner
 
+.. autosummary::
+   :toctree: generated/
+   :template: class.rst
+   :nosignatures:
+
+   InterpolationJoin
```
jovan-stojanovic, line 41: Does not appear in the API generated docs (because you renamed it in the meantime)
```suggestion
   InterpolationJoiner
```

skrub/_interpolation_join.py
```diff
@@ -0,0 +1,343 @@
...
+    the data points it does contain. Using ``annual_avg_temp``, we train a
+    model to predict the temperature, given the latitude and longitude. Then,
+    we use this model to estimate the values we want to add to our
+    ``buildings`` table. In a way we are joining ``buildings`` to a virtual
+    table, in which rows for any (latitude, longitude) location are inferred,
+    rather than retrieved, when requested. This is done with::
+
+        InterpolationJoiner(
+            annual_avg_temp, on=["latitude", "longitude"]
+        ).fit_transform(buildings)
```
jovan-stojanovic, line 34: This is somewhat linked to what you do in the simple reproducible example below. Why not just add these explanations below?

jovan-stojanovic (changes requested):
Hey @jeromedockes, thanks for this new feature! I have a few comments, mostly on docs for now.

examples/08_interpolation_join.py
Vincent-Maladiere: Yes, we should write a standard somewhere, feel free to open an issue for it.
Btw I think it's  \``'TMAX'``

examples/08_interpolation_join.py
jeromedockes: I didn't because I felt the example was quite long already, but I can if you think it's useful! I'd be quite surprised if it noticeably improved delay predictions though

skrub/_interpolation_join.py
jeromedockes: I thought having the explanations sooner would help understand the list of parameters which comes just below. the Examples section appears towards the end (both in the docstring and the rendered doc) and ideally we would already have a good understanding of what the function does before reaching the detailed parameter list. But I can move that paragraph if you think it belongs in the examples section

examples/08_interpolation_join.py
Vincent-Maladiere: Ah, annoying. Should we remove this check from flake8 (or black) for the example section?

skrub/tests/test_interpolation_join.py
Vincent-Maladiere: Ok got it, let's keep the fixtures then

skrub/_interpolation_join.py
Vincent-Maladiere: Right, well spotted, thanks. 

For the sake of argument, since we already behave slightly differently than Pandas, should we be more flexible in terms of parameters and accept any iterable? This discussion is also useful for the `AggJoiner` PR.

examples/08_interpolation_join.py
jovan-stojanovic: Ok, I guess it's pertinent to add only if there are noticeable improvements in prediction to motivate the usage, otherwise no need to prolong the example..

skrub/_interpolation_join.py
jeromedockes: sure -- we do accept any iterable in practice (it gets converted to a [list](https://github.com/skrub-data/skrub/pull/742/files#diff-5988773cfa49ef9e6c03e0039fa0456a9df0d2d53ba6e38d13bb6bfee9572be4R227) ). so you mean change the docstring to say Iterable of str instead of list of str?

examples/08_interpolation_join.py
jeromedockes: +1 for removing the check -- I think your suggestion makes the rendered example nicer so it would be disappointing to forgo it to make a linter happy. In general I wonder if the style checks aren't a touch too aggressive; for example your fix #754 had to work around a limitation of the linter, not fix a problem in skrub

skrub/_interpolation_join.py
jovan-stojanovic: I agree that we need a paragraph about the mains parameters here, but I find this part redundant with the example below. 
Also, I don't find the simple example is towards the end (it comes before methods and attributes) and I often look at it, for instance in the pandas or polars docs, just after having a glimpse at the parameters.
How about:
```We have a main and auxiliary table that we join on imperfect correspondences. Instead of matching on the nearest neighbor in the auxiliary table, the objective of the InterpolationJoiner is to infer and estimate from close data points the corresponding value(s). This is particularly useful in temporal or spatial joins (see example with latitude and longitude below).```
and incorporating your text into the example?

File: examples/08_interpolation_join.py (hunk 13 above)
Comment (Vincent-Maladiere) [reply examples/08_interpolation_join.py:156]: Yes, there is a discussion we need to have around linting. https://github.com/skrub-data/skrub/issues/755 is a first step in this direction.

File: skrub/_interpolation_joiner.py [hunk 27]
```diff
@@ -0,0 +1,343 @@
...
+        equivalent to ``["latitude"]``.
+
+    aux_key : list of str, or str
+        The columns in ``aux_table`` used for joining. Their number and types
+        must match those of the ``main_key`` columns in the main table. These
+        columns provide the features for the estimators to be fitted. As for
+        ``main_key``, it is possible to pass a string when using a single
+        column.
+
+    key : list of str, or str
```
Vincent-Maladiere, line 64: Good addition!

skrub/_interpolation_joiner.py
```diff
@@ -0,0 +1,343 @@
...
+                key_values,
+                self.aux_table[assignment["columns"]],
+                assignment["estimator"],
+            )
+            for assignment in estimators
+        )
+        return self
+
+    def _check_inputs(self):
+        self.vectorizer_ = TableVectorizer()
```
Vincent-Maladiere, line 190: I can't remember our discussion about the vectorizer as an input. Why remove it? This would allow building the `Joiner` from the `InterpolationJoiner` rather than duplicating the logic.

jeromedockes: Now that I think about it I'm not sure why I removed it actually 😂. I added it back now

examples/08_interpolation_join.py
jeromedockes: I compared the r2 scores by adding this to the end of the example:

```python
from sklearn.model_selection import cross_val_score

flights = flights.dropna(subset="ArrDelay")
join = join.dropna(subset="ArrDelay")
X = flights.drop("ArrDelay", axis=1)
y = flights["ArrDelay"].values
regressor = make_pipeline(TableVectorizer(), HistGradientBoostingRegressor())
score = cross_val_score(regressor, X, y)
print(score)
X = join.drop("ArrDelay", axis=1)
score = cross_val_score(regressor, X, y)
print(score)
```

in both cases the prediction is at chance level (probably because I kept very few features from the flights dataset). it makes the example run slower. I guess we can keep it out for now and maybe search for a dataset where the benefit of the interpolation joiner on the downstream task is more obvious?

GaelVaroquaux:
> in both cases the prediction is at chance level (probably because I kept very few features from the flights dataset). it makes the example run slower.

That's not good. We do not want unconvincing examples.

On that specific dataset, I have in mind that the spatial join was helpful in the example that we had. Am I wrong?

jeromedockes:
> On that specific dataset, I have in mind that the spatial join was helpful in the example that we had. Am I wrong?

ok let me rerun it on the full dataset with all features then

skrub/_interpolation_join.py
Vincent-Maladiere: This is a way of documenting a class I've found clearer than what I'm used to. There is a slight redundancy, but it doesn't hurt the user experience, IMHO. I especially appreciate using the example in the parameter section as well.
What do the other @skrub-data/devs think?

examples/08_interpolation_join.py
jovan-stojanovic: Yes, let's keep it out then. Thanks for checking.

jeromedockes:
One thing we haven't discussed is what to do when fitting or querying an estimator raises an exception. This can happen for a variety of reasons we may not be able to fully anticipate, as the user is allowed to choose the estimators. I added a `on_estimator_failure` parameter which can be "warn" (default) "raise" or "pass" and has the meaning described [here](https://github.com/skrub-data/skrub/pull/742/files#diff-5988773cfa49ef9e6c03e0039fa0456a9df0d2d53ba6e38d13bb6bfee9572be4R100) . LMK if that seems reasonable or if you would like it handled some other way

jeromedockes:
I also wonder if for this one it would make sense to use the min hash rather than gap encoder in the TableVectorizer. I have in mind min hash works well with decision trees (thus the default predictors we use here) and is faster. But I don't have much experience using either. WDYT?

GaelVaroquaux:
> One thing we haven't discussed is what to do when fitting or querying an estimator raises an exception. This can happen for a variety of reasons we may not be able to fully anticipate, as the user is allowed to choose the estimators. I added a on_estimator_failure parameter which can be "warn" (default) "raise" or "pass" and has the meaning described here .

That's interesting and important.

I think that we should consider a "skrub.set_config" setting for this.

skrub/_interpolation_joiner.py
```diff
@@ -0,0 +1,451 @@
...
+        fit_results = self._check_fit_results(fit_results)
+        for res in fit_results:
+            del res["failed"]
+        self.estimators_ = fit_results
+        return self
+
+    def _check_inputs(self):
+        if self.vectorizer is None:
+            self.vectorizer_ = TableVectorizer(
+                high_card_cat_transformer=MinHashEncoder()
```
Vincent-Maladiere, line 223: I probably missed some discussions; why choose the MinHashEncoder as default?

jeromedockes:
> I probably missed some discussions; why choose the MinHashEncoder as default?

Ah yes sorry, we talked about it in yesterday's meeting. As the default estimator is the gradient boosting, minhash might be a good choice because it is faster than GAPEncoder and is supposed to work well with models based on decision trees. So we decided to use it for now. We could always change it if we see that it performs worse on some example datasets. I don't have much experience using minhash nor the gap encoder so I don't really have an opinion.

examples/08_interpolation_join.py
jeromedockes: I saw other examples call plt.show after making figures; I did this too and therefore the cell's output is now None

examples/09_interpolation_join.py
```diff
@@ -0,0 +1,181 @@
...
+join = joiner.transform(main_table)
+join.head()
+
+######################################################################
+# Comparing the estimated values to the ground truth
+# --------------------------------------------------
+
+from matplotlib import pyplot as plt
+
+join = join.sample(2000, random_state=0, ignore_index=True)
```
Vincent-Maladiere, line 81: This entire section is so helpful for discovering what features to join that I wonder if we should make a plot utils based on this in a subsequent PR. WDYT?

examples/09_interpolation_join.py
```diff
@@ -0,0 +1,181 @@
...
+# Loading the flights table
+# -------------------------
+# We load the flights table and join it to the airports table using the flights’ ``'Origin'`` which refers to the departure airport’s IATA code.
+# We use only a subset to speed up the example.
+
+flights = fetch_figshare("41771418").X[["Year_Month_DayofMonth", "Origin", "ArrDelay"]]
+flights = flights.sample(20_000, random_state=0, ignore_index=True)
+airports = fetch_figshare("41710257").X.loc[
+    :, ["iata", "airport", "state", "lat", "long"]
+]
```
Vincent-Maladiere, line 116: Nitpick for consistency with L112.
```suggestion
airports = fetch_figshare("41710257").X[["iata", "airport", "state", "lat", "long"]]
```

examples/09_interpolation_join.py
```diff
@@ -0,0 +1,181 @@
...
+join = joiner.fit_transform(flights)
+join.head()
+
+######################################################################
+# Sanity checks
+# -------------
+# This time we do not have a ground truth for the temperatures.
+# We can perform a few basic sanity checks.
+
+state_temperatures = join.groupby("state")["TMAX"].mean().sort_values()
```
Vincent-Maladiere, line 142: These sanity checks are simple but also very helpful, cf my suggestion of plot utils above.

examples/09_interpolation_join.py
```diff
@@ -0,0 +1,181 @@
...
+
+import seaborn as sns
+
+join["month"] = join["Year_Month_DayofMonth"].dt.strftime("%m %B")
+plt.figure(layout="constrained")
+sns.barplot(data=join.sort_values(by="month"), y="month", x="TMAX")
+plt.show()
+
+######################################################################
+# Of course these checks do not guarantee that the inferred values in our ``join`` table’s ````'TMAX'```` column are accurate.
```
Vincent-Maladiere, line 172:
```suggestion
# Of course these checks do not guarantee that the inferred values in our ``join`` table’s ``'TMAX'`` column are accurate.
```

Vincent-Maladiere (commented):
A couple of additional remarks!

examples/09_interpolation_join.py
jeromedockes: thanks, yes I think that's a good idea. skrub is likely to be used in some rather complex pipelines so it would be very useful for users if we could help with inspection, debugging etc

examples/09_interpolation_join.py
```diff
@@ -0,0 +1,181 @@
...
+# But at least the :class:`~skrub.InterpolationJoiner` seems to have learned a few reasonable trends from its training table.
+
+
+######################################################################
+# Conclusion
+# ----------
+# We have seen how to fit an :class:`~skrub.InterpolationJoiner` transformer: we give it a table (the weather data) and a set of matching columns (here date, latitude, longitude) and it learns to predict the other columns’ values  (such as the max daily temperature).
+# Then, it transforms tables by *predicting* values that a matching row would contain, rather than by searching for an actual match.
+# It is a generalization of the :func:`~skrub.fuzzy_join`, as :func:`~skrub.fuzzy_join` is the same thing as an :class:`~skrub.InterpolationJoiner` where the estimators are 1-nearest-neighbor estimators.
```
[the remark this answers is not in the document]
jeromedockes: yes I think that would be great! the first step will be finding a slightly easier dataset, where we see a bigger benefit of joining an extra table on a downstream task -- otherwise we won't see any difference between the different classifiers and regressors

skrub/_interpolation_joiner.py
```diff
@@ -0,0 +1,451 @@
...
+        self,
+        aux_table,
+        *,
+        main_key=None,
+        aux_key=None,
+        key=None,
+        suffix="",
+        regressor=None,
+        classifier=None,
+        vectorizer=None,
```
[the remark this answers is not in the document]
jeromedockes: see #796

Vincent-Maladiere:
We need to address @jovan-stojanovic's remark before merging this PR

jeromedockes:
> We need to address @jovan-stojanovic's remark before merging this PR

I agree but I'm not sure what we decided in the end?

Vincent-Maladiere:
Let's fix the conflict and merge then :)

jeromedockes:
I also changed the way default sub-estimators are managed to be consistent with TableVectorizer and #814 , ie the default is now an estimator instead of `None`. This makes the parameter list a bit more scary in the doc :D but probably not a big deal

screenshot_2023-11-10T14:45:33+01:00

see also [this note](https://github.com/skrub-data/skrub/discussions/796#discussioncomment-7533197)

jeromedockes:
but maybe I should use the default DatetimeEncoder, WDYT? I figured it might be a bit surprising that the closest date representations do not correspond to the closest dates, but maybe it's not important for the InterpolationJoiner and users who want a more predictable behavior would just use the Joiner

jeromedockes:
> but maybe I should use the default DatetimeEncoder, WDYT? 

I changed it to use the default DatetimeEncoder, now the only TableVectorizer parameter we set is the high cardinality transformer (to use the hashing vectorizer)

I guess users can always override datetime encoding (as is done in one of the tests) with 

```python
InterpolationJoiner(aux, key="key").set_params(vectorizer__datetime_transformer__resolution=None)
```

although the fact that you can't set it in `__init__` but only with set_params and that it's 2 levels of sub-estimator deep + the `set_params` double underscore syntax makes it unlikely that they will ...

Vincent-Maladiere (approved):
> I changed it to use the default DatetimeEncoder, now the only TableVectorizer parameter we set is the high cardinality transformer (to use the hashing vectorizer)

I think this is fine —and more readable! Maybe it would be worth mentioning why we choose `MinHash` as the default transformer in the vectorizer (we care more about speed than explainability here)

You have a weird MacOS error that seems unrelated to this PR (let's try to run the CI one more time with an empty commit?)

Apart from that, LGTM! Thank you, Jérome!

Vincent-Maladiere:
Merging, thank you @jeromedockes!

PR #743, opened 2023-09-20
DatetimeEncoder fixes

### LINKED ISSUE #741: BUG example 03 breaks with `exact_until='microsecond'`
### Describe the bug

example 03 breaks with `exact_until='microsecond'`

### Steps/Code to Reproduce

modify "examples/03_datetime_encoder.py" to have `extract_until='nanosecond'` instead of `exact_until='minute'`

```python
import pandas as pd

data = pd.read_csv(
    "https://raw.githubusercontent.com/pandas-dev/pandas"
    "/main/doc/data/air_quality_no2_long.csv"
)
# Extract our input data (X) and the target column (y)
y = data["value"]
X = data[["city", "date.utc"]]

X

# Encoding the features
# .....................
#
# We will construct a |ColumnTransformer| in which we will encode
# the city names with a |OneHotEncoder|, and the date
# with a |DatetimeEncoder|.
#
# During the instantiation of the |DatetimeEncoder|, we specify that we want
# to extract the day of the week, and that we don't want to extract anything
# finer than minutes. This is because we don't want to extract seconds and
# lower units, as they are probably unimportant.

from sklearn.preprocessing import OneHotEncoder

encoder = make_column_transformer(
    (OneHotEncoder(handle_unknown="ignore"), ["city"]),
    (DatetimeEncoder(add_day_of_the_week=True, extract_until="nanosecond"), ["date.utc"]),
    remainder="drop",
)

```

### Expected Results

no error is thrown

### Actual Results

  File "/home/marcogorelli/skrub-dev/skrub/_datetime_encoder.py", line 149, in _extract_from_date
    return pd.DatetimeIndex(date_series).millisecond.to_numpy()
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'DatetimeIndex' object has no attribute 'millisecond'. Did you mean: 'microsecond'?

### Versions

```shell
System:
    python: 3.11.4 (main, Jun  7 2023, 12:45:48) [GCC 11.3.0]
executable: /home/marcogorelli/skrub-dev/.venv/bin/python
   machine: Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
      sklearn: 1.3.0
          pip: 23.1.2
   setuptools: 65.5.0
        numpy: 1.25.2
        scipy: 1.11.2
       Cython: None
       pandas: 2.0.3
   matplotlib: None
       joblib: 1.3.2
threadpoolctl: 3.2.0

Built with OpenMP: True

threadpoolctl info:
       user_api: openmp
   internal_api: openmp
    num_threads: 16
         prefix: libgomp
       filepath: /home/marcogorelli/skrub-dev/.venv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-a34b3233.so.1.0.0
        version: None

       user_api: blas
   internal_api: openblas
    num_threads: 16
         prefix: libopenblas
       filepath: /home/marcogorelli/skrub-dev/.venv/lib/python3.11/site-packages/numpy.libs/libopenblas64_p-r0-5007b62f.3.23.dev.so
        version: 0.3.23.dev
threading_layer: pthreads
   architecture: SkylakeX

       user_api: blas
   internal_api: openblas
    num_threads: 16
         prefix: libopenblas
       filepath: /home/marcogorelli/skrub-dev/.venv/lib/python3.11/site-packages/scipy.libs/libopenblasp-r0-23e5df77.3.21.dev.so
        version: 0.3.21.dev
threading_layer: pthreads
   architecture: SkylakeX
0.0.1.dev0
```

### LINKED ISSUE #744: FEAT Getting only the time since epoch from the DateTimeEncoder
### Problem Description

Is there a way to use the TableVectorizer and the DatetimeEncoder to transform datetime columns into the time elapsed since epoch (only)? The features such as `year` can be useful in many cases but they mean the closest dates (in terms of time difference) don't necessarily have the closest transformed representations so in some cases we may want to get a single numeric column that represents the absolute time

```
>>> a
         date  A
0  2023-01-01  1
>>> b
         date  B
0  2022-12-31  1
1  2023-12-01  2
>>> fuzzy_join(a, b, on="date")
       date_x  A      date_y  B
0  2023-01-01  1  2023-12-01  2
>>> 
```

### Feature Description

I would like a way to get just `total_time` and not the other features; maybe by passing `extract_until=None` or similar

### Alternative Solutions

_No response_

### Additional Context

_No response_

### LINKED ISSUE #745: BUG DateTimeEncoder fails when `extract_until` is "year" or "month"
### Describe the bug

raises an error during `fit` when those values are used

### Steps/Code to Reproduce

```python

DatetimeEncoder(extract_until="year").fit([["2023-09-20"]])
```

### Expected Results

No error

### Actual Results

```
ValueError: <YearEnd: month=12> is a non-fixed frequency
```

### Versions

```shell
System:
    python: 3.11.5 (main, Aug 25 2023, 13:19:50) [GCC 11.4.0]
executable: /home/jerome/.virtualenvs/df/bin/python
   machine: Linux-6.2.0-32-generic-x86_64-with-glibc2.35

Python dependencies:
      sklearn: 1.3.0
          pip: 23.2.1
   setuptools: 65.5.0
        numpy: 1.25.2
        scipy: 1.11.2
       Cython: None
       pandas: 2.1.0
   matplotlib: 3.7.3
       joblib: 1.3.2
threadpoolctl: 3.2.0

Built with OpenMP: True

threadpoolctl info:
       user_api: openmp
   internal_api: openmp
    num_threads: 4
         prefix: libgomp
       filepath: /home/jerome/.virtualenvs/df/lib/python3.11/site-packages/scikit_learn.libs/libgomp-a34b3233.so.1.0.0
        version: None

       user_api: blas
   internal_api: openblas
    num_threads: 4
         prefix: libopenblas
       filepath: /home/jerome/.virtualenvs/df/lib/python3.11/site-packages/numpy.libs/libopenblas64_p-r0-5007b62f.3.23.dev.so
        version: 0.3.23.dev
threading_layer: pthreads
   architecture: Prescott

       user_api: blas
   internal_api: openblas
    num_threads: 4
         prefix: libopenblas
       filepath: /home/jerome/.virtualenvs/df/lib/python3.11/site-packages/scipy.libs/libopenblasp-r0-23e5df77.3.21.dev.so
        version: 0.3.21.dev
threading_layer: pthreads
   architecture: Prescott
0.0.1.dev0
```
```

### PR DESCRIPTION
**Pull request by LeoGrin**
- Fix #741 Remove millisecond extraction in DatetimeEncoder (Turns out you pandas' `DatetimeIndex` don't have a `millisecond` attribute.)
- Fix #745 Stop using the `floor` function and simplify the logic to check if we need `total_time`
- Close #744 Adds a `None` option for `extract_until` to only extract `total_time`.

@jeromedockes

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -137,8 +137,9 @@ Minor changes
   which provides some more information about the job title.
   :pr:`581` by :user:`Lilian Boulard <LilianBoulard>`
 
-* Fix bug which was triggered when `extract_until` was smaller than `seconds`
-  in :class:`DatetimeEncoder`. :pr:`743` by :user:`Leo Grinsztajn <LeoGrin>`
+* Fix bugs which was triggered when `extract_until` was "year", "month", "microseconds"
+  or "nanoseconds", and add the option to set it to `None` to only extract `total_time`,
+  the time from epoch. :class:`DatetimeEncoder`. :pr:`743` by :user:`Leo Grinsztajn <LeoGrin>`
 
 Before skrub: dirty_cat
 ========================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -138,7 +138,7 @@ Minor changes
   :pr:`581` by :user:`Lilian Boulard <LilianBoulard>`
 
 * Fix bug which was triggered when `extract_until` was smaller than `seconds`
-  in :class:`DatetimeEncoder`. :pr:`1526` by :user:`Leo Grinsztajn <LeoGrin>`
+  in :class:`DatetimeEncoder`. :pr:`743` by :user:`Leo Grinsztajn <LeoGrin>`
 
 Before skrub: dirty_cat
 ========================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -137,6 +137,9 @@ Minor changes
   which provides some more information about the job title.
   :pr:`581` by :user:`Lilian Boulard <LilianBoulard>`
 
+* Fix bug which was triggered when `extract_until` was smaller than `seconds`
+  in :class:`DatetimeEncoder`. :pr:`1526` by :user:`Leo Grinsztajn <LeoGrin>`
+
 Before skrub: dirty_cat
 ========================
 
--- a/skrub/_datetime_encoder.py
+++ b/skrub/_datetime_encoder.py
@@ -19,7 +19,6 @@ WORD_TO_ALIAS: dict[str, str] = {
     "hour": "H",
     "minute": "min",
     "second": "S",
-    "millisecond": "ms",
     "microsecond": "us",
     "nanosecond": "N",
 }
@@ -31,7 +30,6 @@ AcceptedTimeValues = Literal[
     "hour",
     "minute",
     "second",
-    "millisecond",
     "microsecond",
     "nanosecond",
 ]
@@ -47,7 +45,7 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
 
     Parameters
     ----------
-    extract_until : {'year', 'month', 'day', 'hour', 'minute', 'second', 'millisecond', 'microsecond', 'nanosecond'}, default='hour'
+    extract_until : {'year', 'month', 'day', 'hour', 'minute', 'second', 'microsecond', 'nanosecond'}, default='hour'
         Extract up to this granularity.
         If all features have not been extracted, add the 'total_time' feature,
         which contains the time to epoch (in seconds).
@@ -145,8 +143,6 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
             return pd.DatetimeIndex(date_series).minute.to_numpy()
         elif feature == "second":
             return pd.DatetimeIndex(date_series).second.to_numpy()
-        elif feature == "millisecond":
-            return pd.DatetimeIndex(date_series).millisecond.to_numpy()
         elif feature == "microsecond":
             return pd.DatetimeIndex(date_series).microsecond.to_numpy()
         elif feature == "nanosecond":
@@ -263,7 +259,7 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
         Feature names are formatted like: "<column_name>_<new_feature>"
         if the original data has column names, otherwise with format
         "<column_index>_<new_feature>" where `<new_feature>` is one of
-        {"year", "month", "day", "hour", "minute", "second", "millisecond",
+        {"year", "month", "day", "hour", "minute", "second",
         "microsecond", "nanosecond", "dayofweek"}.
 
         Parameters
--- a/skrub/_datetime_encoder.py
+++ b/skrub/_datetime_encoder.py
@@ -124,7 +124,7 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
         }
 
     def _validate_keywords(self):
-        if self.extract_until not in TIME_LEVELS and not self.extract_until is None:
+        if self.extract_until not in TIME_LEVELS and self.extract_until is not None:
             raise ValueError(
                 f'"extract_until" should be one of {TIME_LEVELS}, '
                 f"got {self.extract_until}. "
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Review by Vincent-Maladiere — APPROVED** [review]
Thanks @LeoGrin for this bug fix. This might help fix https://github.com/skrub-data/skrub/issues/741 afterward.
Overall, I'm concerned with the quality of this estimator as we have other bugs like https://github.com/skrub-data/skrub/issues/746. This means that we didn't test it enough.

**Comment by Vincent-Maladiere** [comment]
Hey @LeoGrin, I'd advocate not to change the scope of PRs once started because it's confusing to review. To fix new issues, let's open new PRs instead :)

File: skrub/_datetime_encoder.py [hunk 1]
```diff
@@ -125,7 +124,7 @@ def _more_tags(self):
         }
 
     def _validate_keywords(self):
-        if self.extract_until not in TIME_LEVELS:
+        if self.extract_until not in TIME_LEVELS and not self.extract_until is None:
```
jeromedockes, line 127: it's an unimportant style detail but most people would prefer `x is not None` to `not x is None`

Vincent-Maladiere (commented):
I think returning the total number of seconds since Epoch when `self.extract_until=None` is misleading for the user.
I'd rather have `self.extract_until=None` as a no-op (not returning anything), and have an additional keyword init parameter like `self.seconds_since_epoch`.
What do you both think?

Otherwise, I haven't specific remarks on the PR itself, but I plan to refactorize a lot of the DatetimeEncoder implementation that I'm not super happy with in a subsequence PR.

jeromedockes:
> I think returning the total number of seconds since Epoch when `self.extract_until=None` is misleading for the user.
> I'd rather have `self.extract_until=None` as a no-op (not returning anything), and have an additional keyword init parameter like `self.seconds_since_epoch`.
> What do you both think?
>

I agree, that seems better. If we don't want to add a parameter we could
even always return the seconds since epoch in addition to any other
features the user requests

> Otherwise, I haven't specific remarks on the PR itself, but I plan to refactorize a lot of the DatetimeEncoder implementation that I'm not super happy with in a subsequence PR.
>

I agree the DateTimeEncoder could use some cleanup. For example it re-parses the whole date column to extract each feature.

Also the DateTimeEncoder and the TableVectorizer both parse date strings, and they do it in different ways so it may be worth thinking about their respective responsibilities and doing some refactoring.

jeromedockes:
> Thinking about it a bit more, it would better for the shape of the output to not depend on the data, for instance if we want to use this inside a grid search, so it may be better to get rid of the logic we have now (check if columns are constant).

I think that's a good idea. Dropping constant columns can be done later
in the pipeline with a VarianceThreshold transformer from scikit-learn.

I guess the idea was to avoid adding eg an hour column full of 0 when
the column we are transforming contains dates not times. But for that it
may be better to rely on the date format detected when parsing the dates
than to check the variance a posteriori

Vincent-Maladiere:
> I guess the idea was to avoid adding eg an hour column full of 0 when
the column we are transforming contains dates not times. But for that it
may be better to rely on the date format detected when parsing the dates
than to check the variance a posterior

Alternatively, we could let the user switch between these two behaviors with an additional `drop_if_constant` keyword argument.

> I think that's a good idea. Dropping constant columns can be done later
in the pipeline with a VarianceThreshold transformer from scikit-learn.

We definitely need to add the `VarianceThreshold` estimator to the example!

CHANGES.rst
```diff
@@ -137,6 +137,10 @@ Minor changes
   which provides some more information about the job title.
   :pr:`581` by :user:`Lilian Boulard <LilianBoulard>`
 
+* Fix bugs which was triggered when `extract_until` was "year", "month", "microseconds"
```
Vincent-Maladiere, line 140:
```suggestion
* Fix bugs which where triggered when `extract_until` was "year", "month", "microseconds"
```

Vincent-Maladiere (changes requested):
Let's change the behavior of `extract_until=None` to be a no-op and add an extra `add_total_seconds_since_epoch` or `add_seconds_since_epoch` keyword parameter (default True) as discussed :)

Vincent-Maladiere (approved):
Let's merge so that we can do the refactoring soon.

PR #747, opened 2023-09-21
use pytest rather than python -m pytest

jeromedockes:
part of changes discussed in #738


Vincent-Maladiere (approved):
Do we also need to discuss adding a "python" or "src" folder to avoid importing the local files rather than using the installed package?

jeromedockes:
> Do we also need to discuss adding a "python" or "src" folder to avoid importing the local files rather than using the installed package?

this PR already prevents importing the local files when testing in the
CI. adding a src folder would be an extra safety net to catch this kind
of mistake in case we mess up the CI in the future, so it is not
strictly necessary if the CI remains configured correctly. I still think
it could be marginally useful because a misconfigured CI that passes
when it should fail can be hard to detect before users report defects
after a release. However, Gaël was not a fan because it would make the
layout of skrub more different from scikit-learn. I don't have a very
strong opinion, happy to discuss it some more.

Locally during development, the package is likely to be installed in
editable mode so the files in the repo will be imported during tests
anyway, unless we use something like tox. It could be useful to add a
tox config file for contributors who wish to run the tests in isolated
virtual envs (with different python and dependencies versions) and empty
test directories locally (contributors who don't install or run tox
wouldn't be affected). Here also I don't think it is strictly necessary
nor have a strong opinion about it.

Vincent-Maladiere:
Sounds fair! Thank you for these insights. I guess we can safely merge this.

PR #757, opened 2023-09-26
[MRG] harmonizing the Joiner parameters

jeromedockes:
start applying changes decided in https://github.com/skrub-data/skrub/discussions/751


examples/04_fuzzy_joining.py
```diff
@@ -416,7 +419,13 @@
 from sklearn.model_selection import GridSearchCV
 
 # We will test four possible values of match_score:
-params = {"joiner__match_score": [0.2, 0.3, 0.4, 0.5]}
```
jeromedockes, line 419: here AFAICT all the considered hyperparameters result in the same transformation (`match_score` is low and no fuzzy match is rejected). The selected hyperparameter is the first we specify in the list because they all get the same score. If we increase the threshold so that some rows don't match, we get some [`pd.NA`](https://github.com/skrub-data/skrub/blob/a8af7d5869d7a8084c46b013efb4d6a1bd88182c/skrub/_fuzzy_join.py#L513) in the transform output and the gradient boosting raises an exception.

on the `main` branch, the same problem exists, but when we increase the `match_score`,  the join crashes with a different error before we get to the gradient boosting, because the `Joiner` uses the same suffix for all tables, which results in duplicate column names in this example, which causes a cryptic pandas error when trying to assign values to a part of the dataframe [here](https://github.com/skrub-data/skrub/blob/a8af7d5869d7a8084c46b013efb4d6a1bd88182c/skrub/_fuzzy_join.py#L512)

skrub/_joiner.py
```diff
@@ -155,30 +144,23 @@ def fit(self, X: pd.DataFrame, y=None) -> "Joiner":
...
-                    raise ValueError(
-                        f"Column key {col!r} not found in columns of "
-                        f"table index {table_idx}: {df.columns.tolist()}. "
-                    )
+        for col in self._aux_key:
+            if col not in self.aux_table.columns:
+                raise ValueError(
+                    f"Column key {col!r} not found in columns of "
+                    f"auxiliary table: {self.aux_table.columns.tolist()}. "
+                )
```
glemaitre, line 163: This would not make sense to have it in `check_key`? Or is it really specific to this class?

skrub/_joiner.py
```diff
@@ -197,16 +179,14 @@ def transform(self, X: pd.DataFrame, y=None) -> pd.DataFrame:
...
+            X,
+            self.aux_table,
+            left_on=self._main_key,
+            right_on=self._aux_key,
+            match_score=self.match_score,
+            analyzer=self.analyzer,
+            ngram_range=self.ngram_range,
+            suffixes=("", self.suffix),
+        )
         return X
```
glemaitre, line 192:
```suggestion
        return fuzzy_join(
            X,
            self.aux_table,
            left_on=self._main_key,
            right_on=self._aux_key,
            match_score=self.match_score,
            analyzer=self.analyzer,
            ngram_range=self.ngram_range,
            suffixes=("", self.suffix),
        )
```

skrub/_joiner.py
glemaitre: Actually the `check_key` expect `aux_key`, `main_key` so I would expect to make such check as well.

skrub/tests/test_joiner.py
```diff
@@ -14,77 +15,32 @@ def test_joiner() -> None:
...
-    )[1]
-
     big_table = joiner.transform(main_table)
-    assert big_table.shape == (main_table.shape[0], number_of_cols)
-
-    big_table = joiner.fit_transform(main_table)
-    assert big_table.shape == (main_table.shape[0], number_of_cols)
+    assert big_table.shape == (main_table.shape[0], 3)
+    assert_array_equal(
+        big_table["Population"].values, aux_table["Population"].values[[1, 0, 2]]
```
glemaitre, line 32: I would use `.to_numpy` instead of `values`

examples/04_fuzzy_joining.py
```diff
@@ -416,7 +419,13 @@
...
+param_grid = []
+for match_score in [0.2, 0.3, 0.4, 0.5]:
+    params = {
+        f"{joiner_name}__match_score": [match_score]
+        for (joiner_name, joiner) in all_joiners
```
glemaitre, line 426:
```suggestion
        for (joiner_name, _) in all_joiners
```

glemaitre (commented):
LGTM. I would expect only a couple of changes in the documentation mainly.

skrub/_joiner.py
jeromedockes: actually I'm not sure this check is super useful, if we don't do it we get a more accurate KeyError when trying to use the key

glemaitre: So maybe we should think of removing it then.

glemaitre (approved):
LGTM. I am undecided regarding removing or not the check.

skrub/_joiner.py
Vincent-Maladiere: It's nice to raise an error early on (during `fit` instead of during `predict`), IMHO. I'm +1 on keeping it or putting it into `check_key` as @glemaitre suggests since both checks will be performed in practice.

Vincent-Maladiere (commented):
A few additional comments!

examples/04_fuzzy_joining.py
```diff
@@ -406,7 +407,9 @@
     remainder="drop",
 )
 
-pipeline = make_pipeline(joiner, encoder, HistGradientBoostingRegressor())
+pipeline = Pipeline(
```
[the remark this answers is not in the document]
jeromedockes: sorry the example was outdated after the merge of the SelectCols pr, should be a bit better now

examples/04_fuzzy_joining.py
```diff
@@ -362,27 +362,28 @@
...
 #################################################################
 # Fitting and transforming into the final table
 # .............................................
 # To get our final joined table we will fit and transform the main table (df)
 # with our create instance of the |joiner|:
-df_final = joiner.fit_transform(df)
+from skrub import Joiner
+
+all_joiners = []
+for aux_table, aux_name in [(gdppc, "GDP"), (life_exp, "DYN"), (legal_rights, "LGL")]:
```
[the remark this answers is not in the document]
jeromedockes: here also the example was outdated, with `make_pipeline` chaining joiners should look a bit simpler now. But I'm ok with keeping only 1 joiner, as you prefer

skrub/_joiner.py
```diff
@@ -78,65 +88,43 @@ class Joiner(TransformerMixin, BaseEstimator):
...
+    2    Italy       italy    59000000
     """
 
     def __init__(
         self,
-        tables: tuple[pd.DataFrame, str] | list[tuple[pd.DataFrame, str]],
-        main_key: str | list[str],
+        aux_table: pd.DataFrame,
         *,
+        main_key: str | list[str] | None = None,
```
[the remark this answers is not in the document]
jeromedockes: I guess not!

skrub/_join_utils.py
```diff
@@ -0,0 +1,74 @@
+"""Utilities specific to the JOIN operations."""
+
+from skrub import _utils
+
+
+def add_column_name_suffix(dataframes, suffix):
```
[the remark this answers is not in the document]
jeromedockes: as we discussed IRL, I removed it from this PR and the behavior of the suffix will be handled in a future PR that will refactor fuzzy_join to rely on the Joiner

skrub/_joiner.py
jeromedockes: I put it in a different function as check_key is rather for figuring out what key to use from main_key, aux_key and key, and because this check happens in fit and transform whereas check_key is only in fit

examples/04_fuzzy_joining.py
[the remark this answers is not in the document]
jeromedockes: after merging `main` this PR makes very few changes to the example. So I suggest we discuss and apply any other changes in this example in #802  (or a later PR) to avoid duplicating the review work and having  too many merge conflicts. WDYT?

Vincent-Maladiere (approved):
LGTM! Let's move forward with https://github.com/skrub-data/skrub/pull/802

Issue #755, opened 2023-09-26
DOC Remove file-level flake8 noqa


Vincent-Maladiere:
### Describe the issue linked to the documentation

The following files have a file-level noqa for flake8 E501, meaning that we don't check the length of lines in the docstrings —this is not a desired behavior, as we prefer having well-formatted docstrings.

- [ ] `skrub/_datetime_encoder.py`
- [x] `skrub/_deduplicate.py`
- [x] `skrub/_fuzzy_join.py`
- [x] `skrub/_minhash_encoder.py`
- [x] `skrub/_table_vectorizer.py`

To contribute to this meta-issue, you have to:

1. Select one of the files from the list above.
2. Remove the lines:
    ```python
    # Required for ignoring lines too long in the docstrings
	# flake8: noqa: E501
    ```
     and run `flake8 <name_of_the_file>` to see errors like:
    
    ```shell
    skrub/_datetime_encoder.py:40:89: E501 line too long (116 > 88 characters)
    ```
3. Fix the errors by formatting the docstring (e.g. breaking long lines into smaller lines).
4. Create a pull request for this file only

PR #765, opened 2023-09-29
Fix flake8 noqa

juanitorduz:
Closes https://github.com/skrub-data/skrub/issues/755

- There was a table I was not able to fix, so I kept a single `# noqa: E501`

(BTW: have you thought about [`ruff`](https://github.com/astral-sh/ruff) ? 🙂 )

OMG: I just realized you wanted one PR per file 🤦 . Let me know if you want me to split this PR 🙈

PR #771, opened 2023-09-29
Fix flake8 noqa: `_deduplicate`

juanitorduz:
Partially closes https://github.com/skrub-data/skrub/issues/755

See https://github.com/skrub-data/skrub/pull/765


Vincent-Maladiere (approved):
Thanks, @juanitorduz!

PR #774, opened 2023-09-29
Fix flake8 noqa: `_table_vectorizer`

juanitorduz:
Partially closes https://github.com/skrub-data/skrub/issues/755

- There was a table I was not able to fix, so I kept a single # noqa: E501

See https://github.com/skrub-data/skrub/pull/765


Vincent-Maladiere (approved):
LGTM, we can't do much for this table indeed, unless we remove it.

PR #775, opened 2023-09-29
DOC Ignore unrecognized types in docstrings

TheooJ:
When building the doc, many terms are highlighted, even though they don't link to anything in particular. This happens in the parameters and returns type hinting of docstrings

For example below, you can see that "of", "or", and "optional" are in purple:
doc_before
I'm proposing to disable this like so:
doc_after

I've added a line in `doc/conf.py` that tells numpydoc to ignore all unrecognized terms. More info here: https://numpydoc.readthedocs.io/en/latest/install.html

---------

Another before/after example:
doc_before_bis
doc_after_bis


Vincent-Maladiere (approved):
Thanks, @TheooJ, for this neat improvement of the doc!

Issue #770, opened 2023-09-29
Missing values support is not consistent


Vincent-Maladiere:
`GapEncoder` and `deduplicate` raise different errors when dealing with None or np.nan values, unlike `MinHashEncoder` and `SimilarityEncoder` which run successfully.

Interestingly, the errors differ when the column to encode is of high cardinality, like "department" or low cardinality/binary like "gender", from the employee dataset.

In the table below, we replace values in the columns "department" and "gender" with either `np.nan` or a `None` values, e.g. "department" with None corresponds to:

```python
from skrub import GapEncoder
from skrub.datasets import fetch_employee_salaries

df = fetch_employee_salaries().X
df["department"].replace("POL", None, inplace=True)

GapEncoder().fit_transform(df[["department"]])
# AssertionError: Input data is not string. 
```

|             | "department" with np.nan                                                      | "department" with None                                                        | "gender" with np.nan                                                        | "gender" with None                                                     |
|-------------|-------------------------------------------------------------------------------|-------------------------------------------------------------------------------|-----------------------------------------------------------------------------|------------------------------------------------------------------------|
| GapEncoder  | Success                                                                       | AssertionError: Input data is not string                                      | ValueError: empty vocabulary; perhaps the documents only contain stop words | TypeError: '<' not supported between instances of 'NoneType' and 'str' |
| deduplicate | # TypeError: '<' not supported between instances of 'NoneType' and 'NoneType' | # TypeError: '<' not supported between instances of 'NoneType' and 'NoneType' | TypeError: '<' not supported between instances of 'float' and 'str'         | TypeError: '<' not supported between instances of 'NoneType' and 'str' |

jeromedockes:
For `GapEncoder` and "department":


`GapEncoder` converts to numpy array, then finds and handles missing values by calling `sklearn.utils.fixes._object_dtype_isnan`

https://github.com/skrub-data/skrub/blob/fade2006aa6a57255ac77e170b2516e2b41f48f2/skrub/_gap_encoder.py#L860

This in turn finds null values by comparing `X != X`.
`np.nan != np.nan` is `True`, but `None != None` is `False`, which is why this method does not find `None` entries as being missing values, they are not imputed (replaced with `""`), and later the check which asserts the first value in the series is a string fails.

Note using `_object_dtype_isnan` before extracting the dataframe values into a numpy array, or simply using `pd.isnull` or `pd.isna`, would correctly find the `None` entries.

jeromedockes:
[...]

PR #779, opened 2023-10-04
handling `None` in `GapEncoder`

jeromedockes:
currently `GapEncoder` does not treat `None` as a missing value, part of #770.

this small patch fixes that issue by relying on `pd.isna` rather than scikit-learn's  private `_object_dtype_isnan` to find missing values

### DIFF
--- a/skrub/_gap_encoder.py
+++ b/skrub/_gap_encoder.py
@@ -671,6 +671,8 @@ class GapEncoder(TransformerMixin, BaseEstimator):
         values (NaN) are present during GapEncoder.fit (default is to impute).
         In GapEncoder.inverse_transform, the missing categories will
         be denoted as `None`.
+        "Missing values" are any value for which ``pandas.isna`` returns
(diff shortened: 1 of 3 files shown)


LeoGrin (approved):

skrub/_gap_encoder.py
```diff
@@ -857,7 +856,7 @@ def _handle_missing(self, X):
                 f"'zero_impute', got {self.handle_missing!r}. "
             )
 
-        missing_mask = _object_dtype_isnan(X)
+        missing_mask = pd.isna(X)
```
glemaitre, line 859: I assume that we can make the same changes in `SimilarityEncoder` and `TargetEncoder`.
I would also expect to see a mention of the values considered as missing values in the docstring of the encoder since it is not straightforward.

jeromedockes: good point, thanks!

glemaitre (approved):
LGTM

Issue #768, opened 2023-09-29
ENH Improve `DatetimeEncoder`


Vincent-Maladiere:
Recently, several bugs have been discovered with `DatetimeEncoder` (https://github.com/skrub-data/skrub/issues/741, https://github.com/skrub-data/skrub/issues/744, https://github.com/skrub-data/skrub/issues/745, https://github.com/skrub-data/skrub/issues/746), so this estimator is a good candidate for thorough refactoring.

https://github.com/skrub-data/skrub/pull/743 raised the following points to be addressed:
- Parse date column only once (when `DatetimeEncoder` is used within the `TableVectorizer`)
- Exposes the parsing result with `TableVectorizer` to unify the logic
- Avoid removing constant columns, since the shape of the dataframe varies (having it as a keyword argument)
	- Should we add the `VarianceThreshold` estimator to the example to remove constant features?

jeromedockes:
One other thing we may want to fix is that in `pd.DateTimeIndex` (and thus `skrub.DateTimeEncoder`) day first vs month first is inferred separately for each row, which seems suboptimal because most tables probably don't mix the 2:
```python
>>> pd.DatetimeIndex(["28/01/2023", "04/07/2024"])
DatetimeIndex(['2023-01-28', '2024-04-07'], dtype='datetime64[ns]', freq=None)
>>> 
```

jeromedockes:
Also if we decide to look at the whole column (or at least a chunk of it) as the TableVectorizer does ATM to determine the format, we would probably want to store that state during fit so that the same format is used for transform. this could lead to an organization like:
- datetimeencoder fit: find datetime columns and their format, expose the list of found columns in a public attribute
- tablevectorizer: use this list to decide which transformer to apply to which columns
- datetimeencoder transform: extract features from datetimecolumns

or even have separate estimators for parsing string columns that contain dates and for extracting features from datetime columns

Vincent-Maladiere:
> One other thing we may want to fix is that in pd.DateTimeIndex (and thus skrub.DateTimeEncoder) day first vs month first is inferred separately for each row, which seems suboptimal because most tables probably don't mix the 2:
>  ```python
>  pd.DatetimeIndex(["28/01/2023", "04/07/2024"])
>  ```

Yes indeed, good catch. `to_datetime()` behaves as expected (with a warning `dayfirst` in this instance)
```python
>>> pd.to_datetime(["28/01/2023", "04/07/2024"])
[...]

Issue #746, opened 2023-09-20
BUG DateTimeEncoder fails when a column mixes formats


jeromedockes:
### Describe the bug

The example shown in the DateTimeEncoder's docstring raises an exception

### Steps/Code to Reproduce

```
from skrub import DatetimeEncoder

enc = DatetimeEncoder()
X = [['2022-10-15'], ['2021-12-25'], ['2020-05-18'], ['2019-10-15 12:00:00']]

enc.fit(X)
enc.transform(X)
```

### Expected Results

the example runs

### Actual Results

```
ValueError: unconverted data remains when parsing with format "%Y-%m-%d": " 12:00:00", at position 3. You might want to try:
    - passing `format` if your strings have a consistent format;
    - passing `format='ISO8601'` if your strings are all ISO8601 but not necessarily in exactly the same format;
    - passing `format='mixed'`, and the format will be inferred for each element individually. You might want to use `dayfirst` alongside this.

```

### Versions

```shell
System:
    python: 3.11.5 (main, Aug 25 2023, 13:19:50) [GCC 11.4.0]
executable: /home/jerome/.virtualenvs/df/bin/python
   machine: Linux-6.2.0-32-generic-x86_64-with-glibc2.35

Python dependencies:
      sklearn: 1.3.0
          pip: 23.2.1
   setuptools: 65.5.0
        numpy: 1.25.2
        scipy: 1.11.2
       Cython: None
       pandas: 2.1.0
   matplotlib: 3.7.3
       joblib: 1.3.2
threadpoolctl: 3.2.0

Built with OpenMP: True

threadpoolctl info:
       user_api: openmp
   internal_api: openmp
    num_threads: 4
         prefix: libgomp
       filepath: /home/jerome/.virtualenvs/df/lib/python3.11/site-packages/scikit_learn.libs/libgomp-a34b3233.so.1.0.0
        version: None

       user_api: blas
   internal_api: openblas
    num_threads: 4
         prefix: libopenblas
       filepath: /home/jerome/.virtualenvs/df/lib/python3.11/site-packages/numpy.libs/libopenblas64_p-r0-5007b62f.3.23.dev.so
        version: 0.3.23.dev
threading_layer: pthreads
   architecture: Prescott

       user_api: blas
   internal_api: openblas
    num_threads: 4
         prefix: libopenblas
[...]

PR #784, opened 2023-10-05
MAIN Improve `DatetimeEncoder`

Vincent-Maladiere:
**References**
Addresses issues raised in https://github.com/skrub-data/skrub/pull/743 and detailed in https://github.com/skrub-data/skrub/issues/768.

**What changes?**
1. `DatetimeEncoder` exposes `format_per_column_`, which maps datetime parsable columns (seen during `fit`) to their first non-null entry. We aim to use this column selection in the `TableVectorizer` to unify the datetime column selection logic.
2. We now keep constant features, as removing them dynamically changes the shape of the output. This previous behavior might break pipelines requiring a specific shape and is surprising for the user.
3. On the other hand, we detect date columns (in opposition to datetime columns), e.g. formatted like "2023-01-01", and we only extract features up to the "day" level for those.
4. `extract_until = None` doesn't create an extra column `"total_second"`. It simply doesn't extract features among `{"year", ..., "nanoseconds"}`.
5. The parameter `add_total_second` creates the "total_second" feature. The default is True.
6. The parameter `errors` either raises errors (`errors="raise"`) or outputs `NaT` (`errors="coerce"`) when non-datetime-parsable values are given during `transform`. The default is `"coerce"`.

This PR required additional tests.

cc @LeoGrin @jeromedockes @GaelVaroquaux


skrub/_datetime_encoder.py
```diff
@@ -18,22 +16,56 @@
...
+    Returns
+    -------
+    is_dt_parsable : bool
+    """
+    if len(X.shape) > 1:
+        raise ValueError(f"X must be 1d, got shape: {X.shape}.")
+    np_dtypes_candidates = [np.object_, np.str_, np.datetime64]
+    if any(np.issubdtype(X.dtype, np_dtype) for np_dtype in np_dtypes_candidates):
+        try:
+            _ = pd.to_datetime(X)
```
jeromedockes, line 39: I think `to_datetime` is stricter than `DateTimeIndex` which was used before (@LeoGrin  also mentioned that in #768 [here](https://github.com/skrub-data/skrub/issues/746#issuecomment-1731417491)), as a result the docstring [example](https://github.com/skrub-data/skrub/pull/784/files#diff-af731509aaef29b0ef5b34bb9bd06c8956a0f1112e0b40218c5ea7c460b4c6a9L94) prints an empty vectorized dataframe on my machine. 

being a bit more strict is probably a good idea especially at first; using `to_datetime` instead of `DateTimeIndex` avoids issues such as [this one](https://github.com/skrub-data/skrub/issues/768#issuecomment-1742753260), but the example needs to be updated

skrub/_datetime_encoder.py
```diff
@@ -45,35 +77,44 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
     extract_until : {"year", "month", "day", "hour", "minute", "second",
         "microsecond", "nanosecond", None}, default="hour"
         Extract up to this granularity.
```
jeromedockes, line 79: if it is a "granularity" maybe "resolution" or "granularity" could be a better parameter name? "until" sounds more like we stop at a certain date

skrub/_datetime_encoder.py
```diff
@@ -45,35 +77,44 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
...
-        If all non-constant features have not been extracted,
-        add the "total_time" feature, which contains the time to epoch (in seconds).
         For instance, if you specify "day", only "year", "month", "day" and
-        "total_time" features will be created.
-        If None, only the "total_time" feature will be created.
+        features will be created.
+        If ``None``, no feature will be created.
+
     add_day_of_the_week : bool, default=False
         Add day of the week feature (if day is extracted).
```
jeromedockes, line 85: why can't we have the day of the week without the day number?

skrub/_datetime_encoder.py
```diff
@@ -45,35 +77,44 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
...
         This is a numerical feature from 0 (Monday) to 6 (Sunday).
 
+    add_total_second : bool, default=True
```
jeromedockes, line 88: second**S**?

skrub/_datetime_encoder.py
```diff
@@ -98,79 +139,28 @@ class DatetimeEncoder(BaseEstimator, TransformerMixin):
...
-                    pd.DatetimeIndex(date_series).tz_convert("utc")
-                    - pd.Timestamp("1970-01-01", tz="utc")
-                ) // pd.Timedelta("1s")
-
-    def fit(self, X: ArrayLike, y=None) -> "DatetimeEncoder":
-        """Fit the instance to ``X``.
+    def fit(self, X, y=None):
+        """Fit the instance to X.
 
         In practice, just check keywords and input validity,
```
jeromedockes, line 158: this is now outdated

skrub/_datetime_encoder.py
```diff
@@ -180,95 +170,145 @@ def fit(self, X: ArrayLike, y=None) -> "DatetimeEncoder":
...
+        """Select datetime-like columns and infer features to be parsed.
+
+        If the input only contains dates (and no datetimes), only the features
+        ["year", "month", "day"] will be filtered with extract_until.
+
+        Parameters
+        ----------
+        X : array-like of shape (n_samples, n_features)
+        """
+        # Features to extract for each column, after removing constant features
```
jeromedockes, line 203: the "constant features" part is outdated?

skrub/_datetime_encoder.py
```diff
@@ -180,95 +170,145 @@ def fit(self, X: ArrayLike, y=None) -> "DatetimeEncoder":
...
+                features = self.features_per_column_[col]
+                for feat_idx, feature in enumerate(features):
+                    if feature == "total_second":
+                        if X_col.tz is not None:
+                            X_col = X_col.tz_convert("utc")
+                        # Total seconds since epoch
+                        mask_notnull = X_col == X_col
+                        X_feature = np.where(
+                            mask_notnull,
+                            X_col.astype("int64") // 1e9,
```
jeromedockes, line 293: why use floor division?

jeromedockes:
in a later PR the TableVectorizer will start using this module's `is_datetime_parsable` to identify datetime columns? Note there is some logic in the TableVectorizer to handle month first vs day first that may need to be moved into the DateTimeEncoder module

jeromedockes:
I wonder if we could have a transformer that just parses string columns that contain dates and replaces them with columns with datetime64 dtype? This is one thing that is annoying to do in pandas if we don't know in advance which columns are dates

```python
import pandas as pd

df = pd.DataFrame({"date": ["2023-10-06", "2023-10-07"]})
df["date"] = pd.to_datetime(df["date"]) # avoid having to do this
```

skrub/_datetime_encoder.py
Vincent-Maladiere: Useless here indeed, old reflex to get integers but `1e9` is a float and `X_out` dtype is also float

skrub/_datetime_encoder.py
Vincent-Maladiere: If you mean why "`(if day is extracted)`", this is also outdated :)

jeromedockes: yes that's what I meant

skrub/_datetime_encoder.py
```diff
@@ -180,95 +170,145 @@ def fit(self, X: ArrayLike, y=None) -> "DatetimeEncoder":
...
+        X_out : ndarray of shape ``(n_samples, n_features_out_)``
+        """
+        columns = getattr(self, "feature_names_in_", list(range(X.shape[1])))
+        # X_out must be of dtype float64 to handle np.nan
+        X_out = np.empty((X.shape[0], self.n_features_out_), dtype=np.float64)
+        offset_idx = 0
+        for col_idx, col in enumerate(columns):
+            if col in self.features_per_column_:
+                # X_col is a DatetimeIndex
+                X_col = pd.to_datetime(X[:, col_idx], errors=self.errors)
```
Vincent-Maladiere, line 282: This also needs `format=self.format_per_column[col]`, after adding `_table_vectorizer._infer_date_format` logic.

Vincent-Maladiere:
> his is one thing that is annoying to do in pandas if we don't know in advance which columns are dates

I like this suggestion a lot, this is indeed very annoying. Does it need to be a transformer? I feel we could use a "switch-bait" function in the sense of `fuzzy_join`, like `skrub.to_datetime()`.  WDYT? cc @GaelVaroquaux

Vincent-Maladiere:
@jeromedockes I went for the switch-bait `to_datetime` with a rework on the backend side to decouple `DatetimeEncoder` from the column selection, the format guessing, and the parsing logic. LMKWYT :)

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+
+    Returns
+    -------
+    X_split : list of 1d array of length n_features
+    """
+    if indices is None:
+        indices, indice_to_format = _get_datetime_column_indices(X_split)
+
+    # format overwrite indices_to_format
+    if format is not None or indice_to_format is None:
```
jeromedockes, line 292: it seems `indice_to_format` can never be `None` -- values in this dict may be `None` though

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+
+    Returns
+    -------
+    is_dt_parsable : bool
+    """
+    # Remove columns of int, float or bool casted as object.
+    try:
+        if np.array_equal(X_col, X_col.astype(np.float64)):
+            return False
+    except ValueError:
```
jeromedockes, line 348: this can raise a TypeError too
```python
import pandas as pd
from skrub import to_datetime

df = pd.DataFrame({"date": ["2023-10-12T17:36:50+00:00"]})
df["date"] = pd.to_datetime(df["date"])

to_datetime(df) # TypeError
```

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+
+    Returns
+    -------
+    format : str
+    """
+    X_col = X_col.astype(np.object_)
+    vfunc = np.vectorize(guess_datetime_format)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+        month_first_formats = np.unique(vfunc(X_col, dayfirst=False))
```
jeromedockes, line 388: this will raise a TypeError on datetime inputs:
```python
import pandas as pd
from skrub import to_datetime

df = pd.DataFrame({"date": ["2023-10-12T17:36:50"]})
df["date"] = pd.to_datetime(df["date"])

to_datetime(df)
```

(Note this is a different error than the one  in `is_datetime_parseable` mentioned above)

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+    X : Pandas or Polars dataframe
+    """
+    _, px = get_df_namespace(X)
+    index = getattr(X, "index", None)
+    X_split = [X[col].to_numpy() for col in X.columns]
+    X_split = _to_datetime_2d(X_split, **kwargs)
+    X_split = {col: X_split[col_idx] for col_idx, col in enumerate(X.columns)}
+    X = pd.DataFrame(X_split, index=index)
+    # conversion is px is Polars, no-op if Pandas
+    return px.DataFrame(X)
```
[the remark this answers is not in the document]
jeromedockes: also it is not due to the skrub code but the conversion fails when timezones are involved:

```python
import polars as pl
from skrub import to_datetime

df = pl.DataFrame({"date": ["2023-10-12T17:36:50+01:00"]})
to_datetime(df) # pl.exceptions.ComputeError
```

but this seems to be more of a polars & pandas compatibility issue:

```python
import pandas as pd
import polars as pl

df = pd.DataFrame({"date": ["2023-10-12T17:36:50+02:00"]})
df["date"] = pd.to_datetime(df["date"])
pl.DataFrame(df)
```

I'll open an issue on the polars repo

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+    X_split : list of array, of shape ``n_features``
+    """
+    X_split = np.hsplit(X, X.shape[1])
+    X_split = [X_col.ravel() for X_col in X_split]
+    return _to_datetime_2d(X_split, **kwargs)
+
+
+def _to_datetime_2d(
+    X_split,
+    indices=None,
```
jeromedockes, line 250: it seems `indices` is always `None`? at least if only the documented **kwargs of `to_datetime` are used

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+    indice_to_format=None,
```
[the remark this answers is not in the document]
jeromedockes: and `indice_to_format` also seems to be always none?

jeromedockes (commented):
Thanks, it looks great! I wonder if we should make it for dataframes only (and possibly 2d arrays), and maybe give it a different names because not all columns are converted. I get your "switchbait" point but it may make the function more complex, 1d or less inputs are well handled by polars and pandas functions, and the skrub use case is DataFrames

I found a couple of issues, anything related to datetimes, timezones, and the interactions between standard library datetimes, pandas and polars is always a bit tricky :D

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
-
-    Constant extracted features are dropped; for instance, if the year is
-    always the same in a feature, the extracted "year" column won't be added.
+TIME_LEVELS = list(WORD_TO_ALIAS)
+
+
+def to_datetime(
+    X,
+    errors="coerce",
+    **kwargs,
```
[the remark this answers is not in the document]
Vincent-Maladiere: I removed the Pandas docstring entirely with several references to pandas. My point is we don't want to maintain a docstring that is 95% written by pandas and subject to changes and limit the confusion for the user.

skrub/_datetime_encoder.py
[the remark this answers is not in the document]
Vincent-Maladiere: > could we build a polars dataframe directly instead of pandas first and then convert? AFAIK at the moment initializing a polars dataframe with a pandas one will make a copy

Polars is still an optional dependency, so I guess we can't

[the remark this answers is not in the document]
Vincent-Maladiere: > I'll open an issue on the polars repo

Thank you for this! Note that I haven't written tests for Polars yet, maybe in a subsequent PR to avoid obfuscating this one.

skrub/_datetime_encoder.py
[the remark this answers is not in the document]
Vincent-Maladiere: It's None when `_to_datetime_2d` is called by `to_datetime`, but is not None when `_to_datetime_2d` is called by `DatetimeEncoder.transform`.

skrub/_datetime_encoder.py
Vincent-Maladiere: I guess it's now fixed, I can't reproduce it.

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+    Parameters
+    ----------
+    X : ndarray of shape ``(n_samples, n_features)``
+
+    Returns
+    -------
+    X_split : list of array, of shape ``n_features``
+    """
+    X_split = np.hsplit(X, X.shape[1])
+    X_split = [X_col.ravel() for X_col in X_split]
```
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, better. Thanks.

Vincent-Maladiere:
> Thanks, it looks great! I wonder if we should make it for dataframes only (and possibly 2d arrays), and maybe give it a different names because not all columns are converted. I get your "switchbait" point but it may make the function more complex, 1d or less inputs are well handled by polars and pandas functions, and the skrub use case is DataFrames
 
Thanks! I get your idea, but we might end up with different behavior for series (`pd.to_datetime`) vs dataframes (`skrub.to_datetime`). For consistency, I'd advocate for handling both 1d and 2d outputs, so that the user can always rely on `skrub.to_datetime`, WDYT? I also simplified the logic for series, to be the same as dataframes.

> I found a couple of issues, anything related to datetimes, timezones, and the interactions between standard library datetimes, pandas and polars is always a bit tricky :D

Thanks for pointing out the issues! I haven't tested it extensively with Polars yet.

skrub/_datetime_encoder.py
Vincent-Maladiere: Yes indeed, thanks

skrub/_datetime_encoder.py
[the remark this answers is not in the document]
jeromedockes: thanks I missed it as it is called indirectly through `_to_datetime_2d_array`

skrub/_datetime_encoder.py
```diff
@@ -18,62 +21,494 @@
...
+    ----------
+    X : Pandas or Polars dataframe
+
+    Returns
+    -------
+    X : Pandas or Polars dataframe
+    """
+    _, px = get_df_namespace(X)
+    index = getattr(X, "index", None)
+    X_split = [X[col].to_numpy() for col in X.columns]
```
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, that's a good suggestion.

skrub/_datetime_encoder.py
jeromedockes: yep sorry for the noise!

skrub/_datetime_encoder.py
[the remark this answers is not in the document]
Vincent-Maladiere: Yes indeed!

Vincent-Maladiere:
The CI errors with the old Pandas version. This is linked to the missing `"format=mixing"` argument, recently introduced in https://github.com/pandas-dev/pandas/pull/50939.

jeromedockes:
> Thanks! I get your idea, but we might end up with different behavior for series (pd.to_datetime) vs dataframes (skrub.to_datetime). For consistency, I'd advocate for handling both 1d and 2d outputs, so that the user can always rely on skrub.to_datetime, WDYT? I also simplified the logic for series, to be the same as dataframes.

yes that makes sense. `skrub.to_datetime` will be different than `pd.to_datetime` and handle only a subset of usecases, so we should be careful to document that and about how we handle parameters that don't apply (such as "unit")

for example pandas can convert to datetime a dataframe with separate columns for year, month, etc. whereas skrub looks at each column separately

```python
import pandas as pd
from skrub import to_datetime

df = pd.DataFrame({"year": [2023, 2024], "month": [1, 2], "day": [12, 13]})
print("pandas:")
dt = pd.to_datetime(df)
print(dt)
print("""
------------
skrub:""")
sdt = to_datetime(df)
print(sdt)
```

```
pandas:
0   2023-01-12
1   2024-02-13
dtype: datetime64[ns]

------------
skrub:
   year  month  day
0  2023      1   12
1  2024      2   13
```

jeromedockes:
also I like your choice to handle series basically as dataframes with one column and I wonder if we should have this harmonization for arrays and scalars too?

```python
import pandas as pd
from skrub import to_datetime

s = pd.Series(["a", "b"])
print(to_datetime(s))
print("""
-----------------
""")
print(to_datetime(s.values))
```

```
0    a
1    b
dtype: object

-----------------

/home/jerome/workspace/backedup_repositories/skrub/skrub/_datetime_encoder.py:85: UserWarning: Could not infer format, so each element will be parsed individually, falling back to `dateutil`. To ensure parsing is consistent and as-expected, please specify a format.
  return pd.to_datetime(X, **kwargs)
DatetimeIndex(['NaT', 'NaT'], dtype='datetime64[ns]', freq=None)

```

jeromedockes:
> We also have doctest errors everywhere. It seems very picky.

yes doctests compares the output as strings, so from this point of view "2022" and "2.022e+03" are not the same. I'll look into setting numpy and pandas formatting options so we get more sensible outputs

Vincent-Maladiere:
> yes that makes sense. `skrub.to_datetime` will be different than `pd.to_datetime` and handle only a subset of usecases, so we should be careful to document that and about how we handle parameters that don't apply (such as "unit")

Yes, there are efforts to be made on the documentation.

> for example pandas can convert to datetime a dataframe with separate columns for year, month, etc. whereas skrub looks at each column separately

I wasn't even aware of this feature 😅 but I got your point

Vincent-Maladiere:
> also I like your choice to handle series basically as dataframes with one column and I wonder if we should have this harmonization for arrays and scalars too?

Maybe we should, to bring consistency in skrub even further?

jeromedockes:
We may need to discuss it a bit more before setting options for the whole
codebase, because the numpy default is actually quite good in most cases. It
looks weird in this particular examples because the array mixes numbers like
2022 which should be formatted with a decimal notation, and very large numbers
which should be formatted with a scientific notation.

so maybe for now you can set the printoptions just for this module


```diff
diff --git a/pyproject.toml b/pyproject.toml
index 3afcafd3..3a182a71 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,3 +39,4 @@ filterwarnings = [
     'ignore:elementwise\ comparison\ failed:FutureWarning',
 ]
 addopts = "--doctest-modules"
+doctest_optionflags = "NORMALIZE_WHITESPACE ELLIPSIS"
diff --git a/skrub/_datetime_encoder.py b/skrub/_datetime_encoder.py
index a94c89e1..9e530e07 100644
--- a/skrub/_datetime_encoder.py
+++ b/skrub/_datetime_encoder.py
@@ -1,3 +1,4 @@
+""">>> np.set_printoptions(formatter={"float": "{:g}".format})"""
 import warnings
 from collections import defaultdict
 from typing import Iterable
@@ -445,10 +446,10 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
     with five columns ("year", "month", "day", "hour" and "total_seconds"):

     >>> enc.transform(X)
-    array([[2022.,   10.,   15.,    0.,    1.6657920e+09],
-           [2021.,   12.,   25.,    0.,    1.6403904e+09],
-           [2020.,    5.,   18.,    0.,    1.5897600e+09],
-           [2019.,   10.,   15.,   12.,    1.5711408e+09]])
+    array([[2022,   10,   15,    0,    1.66579e+09],
+           [2021,   12,   25,    0,    1.64039e+09],
+           [2020,    5,   18,    0,    1.58976e+09],
+           [2019,   10,   15,   12,    1.57114e+09]])
     """

     def __init__(

```

- note the added line in the pyproject.toml -- I can add this one in a small separate PR if you think it makes more sense
- The `set_printoptions` can be done in the module's docstring if you want to hide it from readers of the classe's docstring, or in the classe's docstring if you want to show it. Note as this is a private module we don't mind stuffing it in there as no one reads it.
- you can use any other format you prefer, here are a couple more options

```
'{}':
[2022.0 166579456.0]

'{:g}':
[2022 1.66579e+08]

'{:#.0f}':
[2022. 166579456.]
```

skrub/_datetime_encoder.py
```diff
@@ -18,62 +22,411 @@
...
+    return False
+
+
+def _guess_datetime_format(X_col, require_dayfirst=False):
+    """
+    Parameters
+    ----------
+    X_col : ndarray of shape ``(n_samples,)``
+
+    require_dayfirst : bool, default True
```
LeoGrin, line 288:
```suggestion
    require_dayfirst : bool, default False
```

skrub/_datetime_encoder.py
```diff
@@ -18,62 +22,411 @@
...
+    X_col = X_col.astype(np.object_)
+    vfunc = np.vectorize(guess_datetime_format)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+        # pd.unique handles None
+        month_first_formats = pd.unique(vfunc(X_col, dayfirst=False))
+        day_first_formats = pd.unique(vfunc(X_col, dayfirst=True))
+
+    if None in month_first_formats or None in day_first_formats:
+        return None
```
LeoGrin, line 310: should this be an "and"?

Vincent-Maladiere:
@jeromedockes can you elaborate on the choices of flag? I fail to see how [`NORMALIZE_WHITESPACE`](https://docs.python.org/3/library/doctest.html?highlight=normalize_whitespace#doctest.NORMALIZE_WHITESPACE) or [`ELIPSIS`](https://docs.python.org/3/library/doctest.html?highlight=normalize_whitespace#doctest.ELLIPSIS) would help.

jeromedockes:
> @jeromedockes can you elaborate on the choices of flag? I fail to see how NORMALIZE_WHITESPACE or ELIPSIS would help.

- NORMALIZE_WHITESPACE is because with the `numpy.set_printoptions` call, the printed numpy array will not have its columns aligned anymore. this flag makes differences in whitespace unimportant so it allows you to display in the doctest output a "prettier" array where you use the custom formatting but still add whitespace so columns line up. another option would be to use alignment options in the format eg "{: >4g}" but the NORMALIZE_WHITESPACE is probably good to have anyway
- the ELLIPSIS is not actually needed and is actually the default but is just in the direction of your point that doctest seems picky
- I opened a separate PR for the doctest options: #794
 
Also, actually my suggestion to use `set_printoptions` is not a good one because  it might affect subsequent tests. So we would have to use the context manager:

```diff
diff --git a/pyproject.toml b/pyproject.toml
index 3afcafd3..3a182a71 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,3 +39,4 @@ filterwarnings = [
     'ignore:elementwise\ comparison\ failed:FutureWarning',
 ]
 addopts = "--doctest-modules"
+doctest_optionflags = "NORMALIZE_WHITESPACE ELLIPSIS"
diff --git a/skrub/_datetime_encoder.py b/skrub/_datetime_encoder.py
index a94c89e1..03a83408 100644
--- a/skrub/_datetime_encoder.py
+++ b/skrub/_datetime_encoder.py
@@ -444,11 +444,12 @@ class DatetimeEncoder(TransformerMixin, BaseEstimator):
     The encoder will output a transformed array
     with five columns ("year", "month", "day", "hour" and "total_seconds"):
 
-    >>> enc.transform(X)
-    array([[2022.,   10.,   15.,    0.,    1.6657920e+09],
-           [2021.,   12.,   25.,    0.,    1.6403904e+09],
-           [2020.,    5.,   18.,    0.,    1.5897600e+09],
-           [2019.,   10.,   15.,   12.,    1.5711408e+09]])
+    >>> with np.printoptions(formatter={"float": "{:g}".format}):
+    ...     enc.transform(X)
+    array([[2022,   10,   15,    0,    1.66579e+09],
+           [2021,   12,   25,    0,    1.64039e+09],
+           [2020,    5,   18,    0,    1.58976e+09],
+           [2019,   10,   15,   12,    1.57114e+09]])
     """
 
     def __init__(
```

but that adds noise to the example :/

jeromedockes:
another possible compromise would be to use `total_seconds=False` in the example

Vincent-Maladiere:
> another possible compromise would be to use `total_seconds=False` in the example

This feels like a good compromise 😄

skrub/_datetime_encoder.py
```diff
@@ -18,62 +22,411 @@
...
+        # for columns that are already of type datetime64.
+        return None
+
+    X_col = X_col.astype(np.object_)
+    vfunc = np.vectorize(guess_datetime_format)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+        # pd.unique handles None
+        month_first_formats = pd.unique(vfunc(X_col, dayfirst=False))
+        day_first_formats = pd.unique(vfunc(X_col, dayfirst=True))
```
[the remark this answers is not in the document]
LeoGrin: For instance, this parses the two dates with different formats while one format would work for both:
```
dt_arr = np.array(['04-01-2018', '13-01-2014', np.nan], dtype='str')
dt_enc = DatetimeEncoder()
dt_enc.fit_transform(dt_arr.reshape(-1, 1))
```

skrub/_datetime_encoder.py
Vincent-Maladiere: I don't see a situation where we would return a format when one of the two is `None`. Do you?

skrub/_datetime_encoder.py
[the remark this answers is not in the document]
LeoGrin: My bad, the issue with the above example is actually due to using np.nan in a string array. Maybe the fact that the function doesn't work with null values could be added to the doctring?

skrub/_datetime_encoder.py
LeoGrin: No actually I think you're right.

Vincent-Maladiere:
Remaining questions:
- How should we handle timestamps? By default, passing a regular float timestamp to pandas will result in an incorrect result unless specifying `unit`:
  ```python
  >>> pd.to_datetime(1697471201.980721)  # (equivalent to setting unit="ns")
  Timestamp('1970-01-01 00:00:01.697471201')

  >>> pd.to_datetime(1697471201.980721, unit="s")
  Timestamp('2023-10-16 15:47:12')
  ```
  This is also quite challenging because it involves parsing float or int as timestamps.

LeoGrin:
> * How should we handle timestamps? By default, passing a regular float timestamp to pandas will result in an incorrect result unless specifying `unit`:

It seems really error-prone to try to parse these automatically, and it seems okay to parse actual timestamps as numerical features for most use cases, while the opposite is really annoying and surprising. So I would let users handle it on their own. Another option would be to accept a list of columns which should be parsed as dates, but I don't think the benefit would be worth the cost.

Vincent-Maladiere:
> It seems really error-prone to try to parse these automatically, and it seems okay to parse actual timestamps as numerical features for most use cases, while the opposite is really annoying and surprising. So I would let users handle it on their own. Another option would be to accept a list of columns which should be parsed as dates, but I don't think the benefit would be worth the cost.

Yes, and I'm also thinking about the case when we would run the `DatetimeEncoder` multiple times on the same dataset and end-up transforming the `total_seconds` column each time, which would be extremely confusing.

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,457 @@
...
+
+    # scalar or unknown type
+    elif np.asarray(X).ndim == 0:
+        return _to_datetime_scalar(X, **kwargs)
+
+    else:
+        raise TypeError(
+            "X must be a string, datetime, list, tuple, 1-d array, Series, "
+            f"2-d array or dataframe. Got {X=!r}."
+        )
```
Vincent-Maladiere, line 117: @jeromedockes support for 1d array and scalar, LMKWYT

skrub/_datetime_encoder.py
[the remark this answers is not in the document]
jeromedockes: > I'll open an issue on the polars repo

I opened https://github.com/pola-rs/polars/issues/11774

jeromedockes:
> @jeromedockes I think we lost the codecov report in the CI?

yes I guess that must have been introduced in #2615 and we missed it because codecov upload still  exits with status 0 even if it fails to upload anything. I'll try to fix it in #797

LeoGrin (approved):
Looks great to me!

For the example, the prediction plot looks a bit weird with the total_seconds features added, I'm not sure why, but it seems a bit nicer with a cutoff on 06-05, I'll let you decide if you want to change it.

Vincent-Maladiere:
> For the example, the prediction plot looks a bit weird with the total_seconds features added, I'm not sure why, but it seems a bit nicer with a cutoff on 06-05, I'll let you decide if you want to change it.

Thanks @LeoGrin. I realize I've neglected the example in my revamp of `DatetimeEncoder`; I need to focus on it. Interestingly, notice that adding `total_seconds` has improved the NMSE:
from:
```
array([-120.29551054, -192.65273375, -170.69606296, -225.30467065,
       -214.21981371])
```
to:
```
array([-108.89923208, -189.50744903, -181.15543923, -143.86698948,
       -207.90131453])
```
Also, `total_seconds` is now the most important feature.

skrub/_datetime_encoder.py
jeromedockes: it looks great!

dcor01:
I have just changed a few of the formatting of the x-labels of the graphs in example 3 in #809

Vincent-Maladiere:
@jeromedockes I think this is now reviewable :)

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+MIXED_FORMAT = "mixed" if _is_pandas_format_mixed_available() else None
+
+
+def to_datetime(
+    X,
+    errors="coerce",
+    **kwargs,
+):
+    """
+    Convert argument to datetime. Return the input if not datetime-parsable.
```
jeromedockes, line 43: maybe describe the main use case first eg "convert columns in X that store dates as strings to proper datetimes"

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+
+    Augment :func:`pandas.to_datetime` by supporting dataframes
+    and 2d arrays inputs. It converts compatible columns to datetime, and
+    pass incompatible columns unchanged.
```
jeromedockes, line 47:
```suggestion
    passes incompatible columns unchanged.
```

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+
+    int, float, str, datetime, list, tuple, 1d array, and Series are defered to
```
jeromedockes, line 49:
```suggestion
    int, float, str, datetime, list, tuple, 1d array, and Series are deferred to
```

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+    :func:`pandas.to_datetime` directly.
+
+    Parameters
+    ----------
+    X : int, float, str, datetime, list, tuple, nd array, Series, DataFrame/dict-like
+        The object to convert to a datetime.
```
jeromedockes, line 55: this gives the impression the output will be a single datetime, maybe "the object to convert" and above I would start with the main use case: DataFrame or any valid argument for pd.to_datetime

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+
+    errors : {'ignore', 'raise', 'coerce'}, default 'coerce'
```
jeromedockes, line 57:
```suggestion
    errors : {'coerce', 'raise'}, default 'coerce'
```

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+        - If ``'raise'``, then invalid parsing will raise an exception.
+        - If ``'coerce'``, then invalid parsing will be set as ``NaT``.
+        Note that ``'ignore'`` is not used and will raise an error.
+
+    **kwargs : key, value mappings
+        Other keyword arguments are passed down to :func:`pandas.to_datetime`.
+        Raise an error if 'unit' is set to any value. This is because, in
+        `pandas.to_datetime`, unit is specific to timestamps, whereas in
+        `skru`.to_datetime` we don't attempt to parse numeric columns.
```
jeromedockes, line 66:
```suggestion
        `skrub.to_datetime` we don't attempt to parse numeric columns.
```

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+    --------
+    :func:`pandas.to_datetime`
+        Convert argument to datetime.
+
+    Examples
+    --------
+    >>> X = pd.DataFrame(dict(a=[1, 2], b=["2021-01-01", "2021-02-02"]))
+    >>> X = to_datetime(X)
+    >>> X.dtypes.to_list()
+    [dtype('int64'), dtype('<M8[ns]')]
```
jeromedockes, line 85: maybe showing X would help
```python
>>> X = pd.DataFrame(dict(a=[1, 2], b=["31/01/2021", "01/02/2022"]))
>>> X
   a           b
0  1  31/01/2021
1  2  01/02/2022
>>> to_datetime(X)
   a          b
0  1 2021-01-31
1  2 2022-02-01
```

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+    """
+    errors_options = ["coerce", "raise"]
+    if errors not in errors_options:
+        raise ValueError(f"errors options are {errors_options!r}, got {errors!r}.")
+    kwargs["errors"] = errors
+
+    if kwargs.get("unit", None) is not None:
```
jeromedockes, line 92:
```suggestion
    if "unit" in kwargs:
```

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,456 @@
...
+        raise ValueError("unit ")
```
jeromedockes, line 93:
```suggestion
        raise ValueError("'unit' is not a parameter of skrub.to_datetime; it is only meaningful when applying pandas.to_datetime to a numerical column")
```

jeromedockes (commented):
first batch of comments

skrub/_datetime_encoder.py
Vincent-Maladiere: haha it seems that something is off with my last commit

Vincent-Maladiere:
One caveat of this first implementation of `skrub.to_datetime` is that when we pass a numpy array of object dtypes, with mixed-typed entries, we might end up with some numeric representations of the date that are not helpful:
```python
import numpy as np
import pandas as pd
from skrub import to_datetime

df = pd.DataFrame(
    dict(
        a=["2020-01-01", "2021-01-01"],
        b=[2020, 2021],
    )
)

# OK
to_datetime(df)
# a	b
# 0	2020-01-01	2020
# 1	2021-01-01	2021

# Not very helpful
to_datetime(df.values)
# array([[1609459200000000000, 2021],
#       [1609545600000000000, 1021]], dtype=object)

# Better, but needs all columns to be datetime
to_datetime(df.values[:, [0]])
# array([['2020-01-01T00:00:00.000000000'],
#       ['2021-01-01T00:00:00.000000000']], dtype='datetime64[ns]')

# OK, by converting to pydatetime
X_split = list(df.values.T)
X_split[0] = pd.to_datetime(X_split[0]).to_pydatetime()
out = np.vstack(X_split).T
out
# array([[datetime.datetime(2020, 1, 1, 0, 0), 2020],
#      [datetime.datetime(2021, 1, 1, 0, 0), 2021]], dtype=object)

# Notice how we still lose the integer representation of the second column.
pd.DataFrame(out).dtypes
# 0    datetime64[ns]
# 1            object
```

Also, we convert dataframes to a set of numpy 1d array, which might be problematic for Polars datetime representation for example.

Therefore, we should consider improving the `to_datetime` logic in a subsequent PR to specialize the ndarray, Polars, and Pandas dataframes / series cases, instead of reducing everything to a set of 1d numpy array.

jeromedockes:
```
# Not very helpful
to_datetime(df.values)
# array([[1609459200000000000, 2020],
#       [1609545600000000000, 2021]], dtype=object)
```

for this one I see a different result:

```
[[1577836800000000000 2020]
 [1609459200000000000 2021]]
```

which corresponds to the timestamps of the 2 dates. Why do you think that is not helpful?

jeromedockes:
in the case where we have to output a numerical numpy array, the timestamp is probably the best we can do. but as you said in most cases we probably want dataframes as input and outputs, with proper datetime columns in the outputs

skrub/_datetime_encoder.py
```diff
@@ -18,78 +23,473 @@
...
+    """
+    # Remove columns of int, float or bool casted as object.
+    # Pandas < 2.0.0 raise a deprecation warning instead of an error.
+    with (
+        warnings.catch_warnings()
+        if not _is_pandas_format_mixed_available()
+        else nullcontext()
+    ):
+        if not _is_pandas_format_mixed_available():
+            warnings.simplefilter("ignore", category=DeprecationWarning)
```
jeromedockes, line 320: I guess we could just filter warnings for all pandas versions

skrub/_datetime_encoder.py
```diff
@@ -18,78 +23,473 @@
...
+        # We don't need to specify a parsing format
+        # for columns that are already of type datetime64.
+        return None
+
+    X_col = X_col.astype(np.object_)
+    vfunc = np.vectorize(guess_datetime_format)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+        # pd.unique handles None
+        month_first_formats = pd.unique(vfunc(X_col, dayfirst=False))
```
jeromedockes, line 370: Not sure if it could be a problem but note here we might end up with the string `'None'` in the result, eg for `_guess_datetime_format(np.asarray(["01/23/2021", ""]))` we get `array(['%m/%d/%Y', 'None'], dtype='<U8')` in `month_first_formats`

skrub/_datetime_encoder.py
```diff
@@ -18,78 +23,473 @@
...
+
+    require_dayfirst : bool, default False
+        Whether to return the dayfirst format when both dayfirst
+        and monthfirst are valid.
+
+    Returns
+    -------
+    format : str
+    """
+    if np.issubdtype(X_col.dtype, np.datetime64):
```
jeromedockes, line 360: maybe in that case we should instead only pass string columns to this function?

skrub/_datetime_encoder.py
```diff
@@ -18,78 +23,473 @@
...
-        Extract up to this granularity.
-        If all non-constant features have not been extracted,
-        add the "total_time" feature, which contains the time to epoch (in seconds).
-        For instance, if you specify "day", only "year", "month", "day" and
-        "total_time" features will be created.
-        If None, only the "total_time" feature will be created.
+        Extract up to this resolution.
+        E.g., ``resolution="day"`` generates the features "year", "month",
+        "day" only.
+        If ``None``, no feature will be created.
```
jeromedockes, line 443:
```suggestion
        If ``None``, no such feature will be created (but day of the week and total seconds may still be extracted, see below).
```

skrub/_datetime_encoder.py
```diff
@@ -180,95 +531,145 @@ def fit(self, X: ArrayLike, y=None) -> "DatetimeEncoder":
...
+            levels = []
+        else:
+            idx_level = TIME_LEVELS.index(self.resolution)
+            levels = TIME_LEVELS[: idx_level + 1]
+
+        X_split = np.hsplit(X, X.shape[1])
+        self.column_indices_, self.index_to_format_ = _get_datetime_column_indices(
+            X_split
+        )
+        del X_split
```
jeromedockes, line 576: why do we need to delete it explicitly?

skrub/_datetime_encoder.py
Vincent-Maladiere: Yes, but this might be informative for Pandas > 2.0.0 as we are not expecting this warning, WDYT? The logic is not pleasant to read, though.

skrub/_datetime_encoder.py
```diff
@@ -18,78 +23,473 @@
...
+    for col_idx, X_col in enumerate(X_split):
+        X_col = X_col[pd.notnull(X_col)]
+
+        # convert pd.TimeStamp to np.datetime64
+        if all(isinstance(val, pd.Timestamp) for val in X_col):
+            X_col = X_col.astype("datetime64")
+
+        if _is_column_datetime_parsable(X_col):
+            indices.append(col_idx)
+            # TODO: pass require_dayfirst to _guess_datetime_format
```
Vincent-Maladiere, line 294: This TODO is outdated

skrub/_datetime_encoder.py
```diff
@@ -18,78 +23,473 @@
...
+
+    X_col = X_col.astype(np.object_)
+    vfunc = np.vectorize(guess_datetime_format)
+    with warnings.catch_warnings():
+        warnings.simplefilter("ignore", category=UserWarning)
+        # pd.unique handles None
+        month_first_formats = pd.unique(vfunc(X_col, dayfirst=False))
+        day_first_formats = pd.unique(vfunc(X_col, dayfirst=True))
+
+    if len(month_first_formats) == 1 and month_first_formats[0] is not None:
```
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, let's document this here and in the `to_datetime` function.

skrub/_datetime_encoder.py
Vincent-Maladiere: It's not as useful as I thought, let's remove it

Vincent-Maladiere:
> for this one I see a different result:

Yes, the results are just for illustration don't mind them.

> which corresponds to the timestamps of the 2 dates. Why do you think that is not helpful?

More specifically, not helpful in the context of working with dataframes because we lose the datetime representation (using `pd.to_datetime` with timestamps is always tricky because of the `unit` kwarg, and this is not something we can do with `skrub.to_datetime`), but helpful for ML downstream applications.

skrub/_datetime_encoder.py
jeromedockes: I'm wondering if it's worth it just to avoid masking a potential different DeprecationWarning raised by astype with a recent pandas. but it's not a big deal either way

Vincent-Maladiere: Okay, let's apply it for all pandas versions

jeromedockes (approved):
LGTM!

jeromedockes:
@LeoGrin  it has changed a bit since your approval, do you want to have another look? If not I will merge it

LeoGrin:
> @LeoGrin it has changed a bit since your approval, do you want to have another look? If not I will merge it

I'll have a look tonight, thanks!

skrub/tests/test_datetime_encoder.py
```diff
@@ -1,494 +1,461 @@
...
+        )
+    )
+    assert_frame_equal(df_dt, expected_df_dt)
+
+    series_dt = to_datetime(df["a"])
+    expected_series_dt = expected_df_dt["a"]
+    assert_array_equal(series_dt, expected_series_dt)
+
+
+def test_mix_of_unambiguous():
```
LeoGrin, line 441: I think that the fact that this is a no-op is because YDM formats don't exist, not because of the mix of unambiguous.
For instance:
```
to_datetime(["10/15/2021", "13/01/2021"])
```
returns
```
array(['2021-10-15T00:00:00.000000000',                           'NaT'],
      dtype='datetime64[ns]')
```
for pandas > 2 (mixed otherwise).

skrub/_datetime_encoder.py
```diff
@@ -18,78 +23,473 @@
...
+        except (pd.errors.ParserError, ValueError):
+            pass
+    return False
+
+
+def _guess_datetime_format(X_col):
+    """
+    Parameters
+    ----------
+    X_col : ndarray of shape ``(n_samples,)``
```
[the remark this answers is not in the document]
LeoGrin: I think this is still missing

skrub/_datetime_encoder.py
```diff
@@ -18,78 +22,489 @@
...
+    When both dayfirst and monthfirst format are possible, we select
+    monthfirst by default.
+
+    You can overwrite this behaviour by setting a format of the caller function.
+    Setting a format always take precedence over infering it using
+    ``_guess_datetime_format``.
+
+    Parameters
+    ----------
+    X_col : ndarray of shape ``(n_samples,)``
```
LeoGrin, line 373: Upping Jerome's comment "If missing values are not allowed in X_col I think we should mention it in the docstring"

LeoGrin:
Just a few comments but otherwise LGTM, thanks!

skrub/tests/test_datetime_encoder.py
Vincent-Maladiere: You're right, this is not what we want to test

Vincent-Maladiere:
Let's merge then!

PR #785, opened 2023-10-06
TST polars support for deduplicate

TheooJ:
Closing issue #789  

Checking if `deduplicate` works with polars input in `test_deduplicate.py`

I propose to catch any exceptions, let me know if you want to be more specific

### DIFF
--- a/skrub/tests/test_deduplicate.py
+++ b/skrub/tests/test_deduplicate.py
@@ -22,7 +22,7 @@ from skrub.datasets import make_deduplication_data
 def test_polars_input():
     import polars as pl
 
-    data = [
+    duplicated_data = [
         "blacn",
         "black",
         "black",
@@ -35,15 +35,10 @@ def test_polars_input():
         "white",
     ]
 
-    try:
-        pl_output = deduplicate(pl.Series(data))
-    except Exception as exc_info:
-        pytest.xfail(f"Polars not supported yet, deduplicate raised {exc_info}.")
+    pl_output = deduplicate(pl.Series(duplicated_data))
+    pd_output = deduplicate(pd.Series(duplicated_data))
 
-    pd_output = deduplicate(pd.Series(data))
-
-    if not all(pl_output == pd_output):
-        pytest.xfail("Polars and Pandas outputs don't match.")
+    assert_array_equal(pl_output, pd_output)
 
 
 @pytest.mark.parametrize(


Vincent-Maladiere:
@TheooJ you have to create a specific issue for the meta-issue you're addressing.
- This PR should mention the sub-issue only (you can edit your message)
- The sub-issue should mention the meta-issue only

Otherwise, the meta-issue will be automatically closed when we merge a single PR that refers to it.

TheooJ:
Thanks, done.

Vincent-Maladiere (changes requested):
Thanks for this PR. Good job using utils from `test_polars`! In addition to checking errors, we also want to check the results. In that sense, check that identical Pandas and Polars series inputs return the same output.

If the function you are testing already works fine with Polars, you don't need to catch a potential error. However, if this function doesn't work with Polars yet, we need to:
1. Add it to a list of functions that don't support Polars yet (create one in `test_polars.py`)
2. Use `pytest.mark.xfail` if the function you are testing belongs to the list. This will skip the test by displaying a soft fail (meaning: we acknowledge that this is failing, but this needs to be fixed). [Example in scikit-learn](https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/preprocessing/tests/test_polynomial.py#L1215). Note that scikit-learn uses a more complex scheme for xfailing tests by using `_xfail_checks` tags in estimators. We can keep it simple for now and use a list, without using tags.

TheooJ:
I've compared pandas and polars outputs in the same test function. If polars is missing, the test will be skipped. If the function either returns an error or if the pandas and outputs don't match, the test will xfail.

Let me know what you think

TheooJ:
As discussed, I've removed the xfail completely if the polars test is successful.

If the test was failing, I would add 

```
# in test_polars
XFAIL_POLARS = ["deduplicate"]
```

```
# in test_deduplicate
from test_polars import XFAIL_POLARS

@pytest.mark.xfail("deduplicate" in XFAIL_POLARS, reason=Polars not supported for deduplicate yet.")
@pytest.mark.skipif(not POLARS_SETUP, reason=POLARS_MISSING_MSG)
def test_polars_input():
    [...]
```

Vincent-Maladiere (approved):
Hey @TheooJ, thank you for this PR! LGTM

PR #798, opened 2023-10-16
FEAT Add make_series and make_dataframe

Vincent-Maladiere:
Addresses https://github.com/skrub-data/skrub/pull/784/files#r1358253388

This simple PR implements, for both Pandas and Polars:
- A function to create a dataframe from a dictionary of columns
- A function to create a series from a 1d array

When merged, this can be used directly in https://github.com/skrub-data/skrub/pull/784 for example.


skrub/dataframe/_pandas.py
```diff
@@ -12,6 +12,52 @@
 from skrub._utils import atleast_1d_or_none
 
 
+def make_dataframe(X, index):
```
jeromedockes, line 15:
```suggestion
def make_dataframe(X, index=None):
```

skrub/dataframe/_polars.py
```diff
@@ -14,6 +16,54 @@
...
+    index : 1d array-like, default=None
+        Unused since polars doesn't use index.
+        Only here for compatibility with Pandas.
+
+    Returns
+    -------
+    X : Polars dataframe
+        Converted output.
+    """
+    if not isinstance(X, dict) or not all(
```
jeromedockes, line 36: would it make sense to raise an error when `index` is not `None`? having an index when we are building a polars dataframe might point to a bug. If not, maybe we could add a `del index` statement to make it clear we don't use it

skrub/dataframe/_pandas.py
```diff
@@ -12,6 +12,52 @@
...
+
+    index : 1d array-like, default=None
+        The index of the dataframe.
+
+    Returns
+    -------
+    X : Pandas dataframe
+        Converted output.
+    """
+    if not isinstance(X, dict) or not all(
```
jeromedockes, line 31: what kind of mistake do you foresee to make this check necessary?

Vincent-Maladiere: It may be slightly overkill, but I think about dictionaries of 2d arrays or dataframes.
For instance:
```python
pd.DataFrame(
	dict(
    	a=[[1, 2, 1]],
	    b=[1, 2, 3],
	)
)
```
Will raise `ValueError: All arrays must be of the same length`, but this error is not informative enough IMO.
I suspect this kind of error might happen more than we'd expect. WDYT?

skrub/dataframe/_polars.py
Vincent-Maladiere: Yes, I think it's reasonable to raise an error when `index` is not `None` for Polars.

skrub/dataframe/_pandas.py
jeromedockes: but this is a private function, right? that will be used only by skrub code not directly by users. so if skrub passes it incorrect inputs there isn't too much the user can do about it except report the problem

btw WDYT about renaming `dataframe` to `_dataframe`? we want skrub estimators to support both pandas and polars, but I'm not sure we want to publicly expose the machinery that makes it possible

Vincent-Maladiere: Ok for the renaming, I'll remove the check

Vincent-Maladiere:
I adjusted the API to make the `dataframe` module private (it seems I still have some work on the examples).
I also did some housekeeping by removing the `DataFrameLike` and `SeriesLike` type hints, LMKWYT :)

jeromedockes:
LGTM, I think once you remove the merge conflict we can merge it

jeromedockes:
I think you also need to update `test_select_cols`

jeromedockes:
and the docstring of `get_df_namespace` also mentions `skrub.dataframe`

jeromedockes (approved):
LGTM, thanks!

Issue #764, opened 2023-09-28
division by 0 in fuzzy_join when there are only perfect matches


jeromedockes:
### Describe the bug

if we join 2 tables on columns for which matches do exist, the rescaling of nearest neighbor distances does a division by 0; we get a warning and the matching scores are NaN

### Steps/Code to Reproduce

```python
>>> warnings.filterwarnings("default")
>>> import pandas as pd
... from skrub import fuzzy_join
... 
... df1 = pd.DataFrame({"A": [0, 1]})
... 
... fuzzy_join(df1, df1, on="A", return_score=True)
/home/jerome/workspace/backedup_repositories/skrub/skrub/_fuzzy_join.py:359: UserWarning: This feature is still experimental.
  warnings.warn("This feature is still experimental.")
/home/jerome/workspace/backedup_repositories/skrub/skrub/_fuzzy_join.py:189: RuntimeWarning: invalid value encountered in divide
  distance = distance / np.max(distance)
   A_x  A_y  matching_score
0    0    0             NaN
1    1    1             NaN
>>> 
```

### Expected Results

no division by 0, scores are numbers (prob. equal to 1)

### Actual Results

division by 0, scores are NaN

### Versions

```shell
main branch
```

GaelVaroquaux:
Good catch, this is an important special case.

Ideally, when there are only exact matches, we shouldn't even go down that code path and only use standard matching from the backend (pandas, polars...), as it will be much faster.

jovan-stojanovic:
Yes, indeed, you are right. This is IMO equivalent to #587 if we agree that we should rely on the backend for these matches..

jeromedockes:
this is not the same as when the user requires perfect matches by setting `match_score=1`. This division by 0 is triggered when all rows happen to have a perfect match in the joined dataframe. it will be fixed by #802

jeromedockes:
fixed by #802

PR #802, opened 2023-10-19
[MRG] Fix the match score scaling

### LINKED ISSUE #763: fuzzy_join's match_score starts at 0.5 not 0.0
### Describe the bug

The fuzzy join documentation gives the impression the worst match will get a score of 0.0, but it will get 0.5:

https://github.com/skrub-data/skrub/blob/404275c73299aaa70cadbf31b34dc34509f58da7/skrub/_fuzzy_join.py#L191

### Steps/Code to Reproduce

any call to fuzzy_join

### Expected Results

the worst score is 0

### Actual Results

the worst score is 0.5

### Versions

```shell
main branch
```

jeromedockes:
closes #763 , #764

note example 04 will fail due to an issue with duplicate column names; this will have to wait until #757 is merged


jeromedockes:
this is ready for review but I made it a draft so it is not merged by mistake, indeed #757 must be merged before to fix the example 04

skrub/_fuzzy_join.py
```diff
@@ -183,9 +183,10 @@ def _nearest_matches(
     neigh.fit(aux_array)
     distance, neighbors = neigh.kneighbors(main_array, return_distance=True)
     idx_closest = np.ravel(neighbors)
-    distance = distance / np.max(distance)
-    # Normalizing distance between 0 and 1:
-    matching_score = 1 - (distance / 2)
+    max_dist = distance.max()
+    if max_dist != 0:
```
Tialo, line 187: Is it safe to compare floats without any tolerance?
Wouldn’t something like this be more preferable https://numpy.org/doc/stable/reference/generated/numpy.isclose.html

jeromedockes: I think an exact comparison is ok here because it is just to avoid a division by 0

examples/04_fuzzy_joining.py
```diff
@@ -397,9 +397,9 @@
 
 # We will test four possible values of match_score:
 params = {
-    "joiner-1__match_score": [0.2, 0.9],
-    "joiner-2__match_score": [0.2, 0.9],
-    "joiner-3__match_score": [0.2, 0.9],
+    "joiner-1__match_score": [0.1, 0.9],
+    "joiner-2__match_score": [0.1, 0.9],
```
Vincent-Maladiere, line 401: I can't comment on it but you need to update the description regarding the best `match_score` parameter here and in other places of the example.
"The grid searching gave us the best value of 0.5 for the parameter"

examples/04_fuzzy_joining.py
```diff
@@ -392,33 +392,24 @@
...
 
-print(grid.best_params_)
+print("Best parameters:", grid.best_params_)
 
-##########################################################################
-# The grid searching gave us the best value of 0.5 for the parameter
-# ``match_score``. Let's use this value in our regression:
-#
-
-print(f"Mean R2 score with pipeline is {grid.score(df, y):.2f}")
```
jeromedockes, line 410: @Vincent-Maladiere having a closer look, this part does not make sense: we are scoring on the training data. if we cross-validate the grid search correctly the score does not improve

jeromedockes: depending on the `cv` random state the grid search score can be the same or worse than the one without grid search

Vincent-Maladiere (approved):
LGTM!

jeromedockes:
@jovan-stojanovic  when you have time would you mind having a look at this one? (Note the example might still be improved in a later PR eg to highlight the gain of fuzzy_join vs a regular join, and we will think about other score rescaling strategies, but the goal of this PR is just to fix the scale so it starts at 0 and to remove the division by 0)

examples/04_fuzzy_joining.py
jovan-stojanovic: Are you improving the results while doing the grid search? If yes, I think this should be kept.

PR #804, opened 2023-10-23
[MRG] Add `SelectCols(cols)` and `DropCols(cols)` transformers

### LINKED ISSUE #670: API Idea: add a "drop" argument
Dropping a column is a very common need that we might need to facilitate. Here are a few ideas to do this:

1. We could add a "drop=None" argument to the TableVectorizer. It would be "None" by default and could take either a column name or a list of columns

2. A more radical point of view would be that all our major transformers should have the drop argument. This way it is easy to do feature 

3. We could have a "Drop" transformer

4. We could even a more general transformer that changes columns (mostly renames and drop, I suspect). The rename might be very useful to deal with input tables with multi-index as columns: these are definitely going to lead to convoluted code to support. It could be called "ColSelect". This would be using a verb rather than a noun, as opposed to "ColSelector" but I am more and more thinking that verbs are better: the code looks like a phrase. The drawback is that it contrasts with scikit-learn

jeromedockes:
closes #670 

a decision on the API has actually not been made in the linked issue, this is to help make the discussion more concrete

### DIFF
--- a/examples/04_fuzzy_joining.py
+++ b/examples/04_fuzzy_joining.py
@@ -368,7 +368,7 @@ from skrub import Joiner, SelectCols, DropCols
 from sklearn.pipeline import make_pipeline
 
 # We create a selector that we will insert at the end of our pipeline, to
-# project on the relevant attributes before fitting the regressor
+# select the relevant columns before fitting the regressor
 
 selector = SelectCols(
     [
--- a/examples/04_fuzzy_joining.py
+++ b/examples/04_fuzzy_joining.py
@@ -362,7 +362,7 @@ print(f"Mean R2 score is {cv_r2_t.mean():.2f} +- {cv_r2_t.std():.2f}")
 # .............................
 
 y = df["Happiness score"]
-df = df.drop("Happines score")
+df = df.drop("Happiness score", axis=1)
 
 from skrub import Joiner, SelectCols, DropCols
 from sklearn.pipeline import Pipeline
--- a/examples/04_fuzzy_joining.py
+++ b/examples/04_fuzzy_joining.py
@@ -365,31 +365,28 @@ y = df["Happiness score"]
 df = df.drop("Happiness score", axis=1)
 
 from skrub import Joiner, SelectCols, DropCols
-from sklearn.pipeline import Pipeline
+from sklearn.pipeline import make_pipeline
 
-steps = [
-    ("gdppc", Joiner((gdppc, "Country Name"), "Country")),
-    ("gdppc-drop", DropCols(["Country Name"])),
-    ("life_exp", Joiner((life_exp, "Country Name"), "Country")),
-    ("life_exp_drop", DropCols(["Country Name"])),
-    ("legal_rights", Joiner((legal_rights, "Country Name"), "Country")),
-]
+# We create a selector that we will insert at the end of our pipeline, to
+# project on the relevant attributes before fitting the regressor
 
-# At the end of our pipeline, we project on the relevant attributes and finally
-# fit our regressor
-
-select = SelectCols(
+selector = SelectCols(
     [
         "GDP per capita (current US$)",
         "Life expectancy at birth, total (years)",
         "Strength of legal rights index (0=weak to 12=strong)",
     ]
 )
-steps = steps + [
-    ("select", select),
-    ("regressor", HistGradientBoostingRegressor()),
-]
-pipeline = Pipeline(steps)
+pipeline = make_pipeline(
+    Joiner((gdppc, "Country Name"), "Country"),
+    DropCols(["Country Name"]),
+    Joiner((life_exp, "Country Name"), "Country"),
+    DropCols(["Country Name"]),
+    Joiner((legal_rights, "Country Name"), "Country"),
+    selector,
+    HistGradientBoostingRegressor(),
+)
+
 
 ##########################################################################
 # And the best part is that we are now able to evaluate the parameters of the |fj|.
@@ -400,14 +397,15 @@ from sklearn.model_selection import GridSearchCV
 
 # We will test four possible values of match_score:
 params = {
-    f"{joiner}__match_score": [0.2, 0.9]
-    for joiner in ["gdppc", "life_exp", "legal_rights"]
+    "joiner-1__match_score": [0.2, 0.9],
+    "joiner-2__match_score": [0.2, 0.9],
+    "joiner-3__match_score": [0.2, 0.9],
 }
 
-grid = GridSearchCV(pipeline, param_grid=params, error_score="raise")
+grid = GridSearchCV(pipeline, param_grid=params)
 grid.fit(df, y)
 
-print(grid.best_params_)
+print(grid.cv_results_)
 ##########################################################################
 # The grid searching gave us the best value of 0.5 for the parameter
 # ``match_score``. Let's use this value in our regression:
--- a/examples/04_fuzzy_joining.py
+++ b/examples/04_fuzzy_joining.py
@@ -362,51 +362,34 @@ print(f"Mean R2 score is {cv_r2_t.mean():.2f} +- {cv_r2_t.std():.2f}")
 # .............................
 
 y = df["Happiness score"]
-#######################################################################
-# We gather the auxilliary tables into a
-# list of (tables, keys) for the `tables` parameter.
-# An instance of the transformer with the necessary information is:
-from skrub import Joiner
+df = df.drop("Happines score")
 
-joiner = Joiner(
-    tables=[
-        (gdppc, "Country Name"),
-        (life_exp, "Country Name"),
-        (legal_rights, "Country Name"),
-    ],
-    main_key="Country",
+from skrub import Joiner, SelectCols, DropCols
+from sklearn.pipeline import Pipeline
+
+steps = [
+    ("gdppc", Joiner((gdppc, "Country Name"), "Country")),
+    ("gdppc-drop", DropCols(["Country Name"])),
+    ("life_exp", Joiner((life_exp, "Country Name"), "Country")),
+    ("life_exp_drop", DropCols(["Country Name"])),
+    ("legal_rights", Joiner((legal_rights, "Country Name"), "Country")),
+]
+
+# At the end of our pipeline, we project on the relevant attributes and finally
+# fit our regressor
+
+select = SelectCols(
+    [
+        "GDP per capita (current US$)",
+        "Life expectancy at birth, total (years)",
+        "Strength of legal rights index (0=weak to 12=strong)",
+    ]
 )
-
-#################################################################
-# Fitting and transforming into the final table
-# .............................................
-# To get our final joined table we will fit and transform the main table (df)
-# with our create instance of the |joiner|:
-df_final = joiner.fit_transform(df)
-
-df_final.head(10)
-
-##########################################################################
-# And that's it! As previously, we now have a big table
-# ready for machine learning.
-# Let's create our machine learning pipeline:
-from sklearn.compose import make_column_transformer
-from sklearn.pipeline import make_pipeline
-
-# We include only the columns that will be pertinent for our regression:
-encoder = make_column_transformer(
-    (
-        "passthrough",
-        [
-            "GDP per capita (current US$)",
-            "Life expectancy at birth, total (years)",
-            "Strength of legal rights index (0=weak to 12=strong)",
-        ],
-    ),
-    remainder="drop",
-)
-
-pipeline = make_pipeline(joiner, encoder, HistGradientBoostingRegressor())
+steps = steps + [
+    ("select", select),
+    ("regressor", HistGradientBoostingRegressor()),
+]
+pipeline = Pipeline(steps)
 
 ##########################################################################
 # And the best part is that we are now able to evaluate the parameters of the |fj|.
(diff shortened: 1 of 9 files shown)


examples/04_fuzzy_joining.py
```diff
@@ -362,51 +362,34 @@
...
+        "Life expectancy at birth, total (years)",
+        "Strength of legal rights index (0=weak to 12=strong)",
+    ]
 )
-
-pipeline = make_pipeline(joiner, encoder, HistGradientBoostingRegressor())
+steps = steps + [
+    ("select", select),
+    ("regressor", HistGradientBoostingRegressor()),
+]
```
GaelVaroquaux, line 391: Why not build the "steps" list in one go.

Also, people typically find "make_pipeline" more readable. I would argue for using it here.

examples/04_fuzzy_joining.py
```diff
@@ -416,9 +399,12 @@
 from sklearn.model_selection import GridSearchCV
 
 # We will test four possible values of match_score:
-params = {"joiner__match_score": [0.2, 0.3, 0.4, 0.5]}
+params = {
+    f"{joiner}__match_score": [0.2, 0.9]
+    for joiner in ["gdppc", "life_exp", "legal_rights"]
```
GaelVaroquaux, line 404: Can we have a simpler (more tedious) syntax here, expanding the dict comprehension manually. It will be more readable for people who are not well versed in Python

examples/04_fuzzy_joining.py
```diff
@@ -416,9 +399,12 @@
...
+}
 
-grid = GridSearchCV(pipeline, param_grid=params)
+grid = GridSearchCV(pipeline, param_grid=params, error_score="raise")
```
GaelVaroquaux, line 407: Why this change?

skrub/_select_cols.py
```diff
@@ -0,0 +1,156 @@
+from sklearn.base import BaseEstimator, TransformerMixin
+
+from .dataframe import get_df_namespace
+
+
+def _check_columns(df, columns):
```
GaelVaroquaux, line 6: Maybe a small docstring here, just to tell us the intent of this function?

GaelVaroquaux (commented):
I had a look at this PR, and I'm quite convinced that this is the right way to go. Great job @jeromedockes !

Can you add the two objects in the API documentation?

And maybe add a tiny section on https://skrub-data.org/stable/assembling.html, between "joining" and "going further". We should stress here that the benefit is to be able to build pipelines.

jeromedockes:
> My original view was to rather have only a single object that merged both SelectCols and DropCols.

That's definitely an option, too. A few small things added up and made me lean towards the version with 2 classes:

- They only have 1 parameter each, so it is natural to pass it as positional argument. Instantiations are a bit less verbose and users don't need to remember the parameter name.

- The classes are easy to understand almost without reading the documentation because they do 1 simple thing and hold 1 list of column names.

- "I can drop columns with DropCols" feels easier to discover than "I can drop columns with the 'drop' parameter of SelectCols" (or the other way around "select with SelectCols" vs "select with 'keep' parameter of DropCols")

- If both drop and select were in one class, we would need a default argument for the "select" parameter to mean "select all" (otherwise using the "drop" parameter would be hard). I didn't see an option I found 100% satisfying for that default value: "None" means the opposite of what we want (ie "all"), a string like "*" would be ok but still it's a bit weird as column names are strings as well, sentinel values like a singleton or enum with a nice repr are too "magic" and we don't know how to import them if we want to pass the default explicitly, ellipsis seems arbitrary, we probably don't want to support slices and they're rarely used outside "[]", etc.

- We would need to decide and document what happens when both parameters are used, and when a column name ends up in both "drop" and "select". It will almost never happen in practice because it doesn't make much sense, but it still needs to be handled so it adds uninteresting cruft both in the code and the docstring. In general having 2 parameters such that it only makes sense to use 1 at a time sometimes results in annoying docstrings and parameter validation code IMHO.

- It's likely that we will add other simple manipulations of columns, like renaming. With the current option we could add another simple transformer "RenameCols". If we stuff all operations in one class we would add one more parameter and need to think of its interactions with the other parameters (does the selection happen before or after the renaming?, ...) and guess what users will find surprising or not. So it might become cumbersome to use, describe & maintain.


Of course that's only a first impression and we'll have to see how it plays out in practice and in the examples. They're simple things so we can prototype several approaches and see what we prefer.

The issue also suggested adding a "drop" parameter to most transformers. I thought it would be more work (especially if we want to add "rename", "select" parameters), add parameters to already complex classes with other interactions to think about (do we drop before or after vectorizing?), duplicate some of the work (or at least parameter documentation) across the existing transformers.

OTOH, it would result in shorter pipelines and fewer classes for users to discover, so it is still certainly worth discussing.

examples/04_fuzzy_joining.py
```diff
@@ -362,51 +362,34 @@
...
-        [
-            "GDP per capita (current US$)",
-            "Life expectancy at birth, total (years)",
-            "Strength of legal rights index (0=weak to 12=strong)",
-        ],
-    ),
-    remainder="drop",
+df = df.drop("Happiness score", axis=1)
+
+from skrub import Joiner, SelectCols, DropCols
```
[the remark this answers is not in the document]
jeromedockes: Note after https://github.com/skrub-data/skrub/pull/757 Joiner will have the "suffix" parameter so if we choose to use it here we won't need the intermediate drops and the pipeline could look more like this (not sure which is better):

```python
df = df.rename(cols={"Country": "Country Name"})
steps = [
    ("gdppc", Joiner(gdppc, on="Country Name", suffix=" gdppc")),
    ("life_exp", Joiner(life_exp, on="Country Name", suffix=" life_exp")),
    ("legal_rights", Joiner(legal_rights, on="Country Name", suffix=" legal_rights")),
    (
        "select",
        SelectCols(
            [
                "GDP per capita (current US$) gdppc",
                "Life expectancy at birth, total (years) life_exp",
                "Strength of legal rights index (0=weak to 12=strong) legal_rights",
            ]
        ),
    ),
    ("regressor", HistGradientBoostingRegressor()),
]
```

examples/04_fuzzy_joining.py
jeromedockes: I felt like putting the `SelectCols` in a separate statement because it adds several lines and levels of indentation due to the long column names and black formatting, but I can put it before and make the whole pipeline in one statement.

I also prefer "make_pipeline" so I'll change it, the drawback is as we have several joiners we have to refer to them with "Joiner-1", "Joiner-2" in the gridsearch

jeromedockes: > I also prefer "make_pipeline" so I'll change it, the drawback is as we have several joiners we have to refer to them with "Joiner-1", "Joiner-2" in the gridsearch

actually that's not so bad even if it's less obvious where the names came from

examples/04_fuzzy_joining.py
jeromedockes: good catch, I'll remove it. I added it because otherwise the error that happened with the columntransformer we discussed this morning would get silently swallowed, so I added the `error_score` to make sure I could trigger it

jeromedockes:
also pandas and polars both have separate select / getitem and drop

jeromedockes:
> Can you add the two objects in the API documentation?
> 
> And maybe add a tiny section on https://skrub-data.org/stable/assembling.html, between "joining" and "going further". We should stress here that the benefit is to be able to build pipelines.

I added the api entries and the paragraph in the user guide. I put the api section after "joining" to keep the order similar to the user guide

jeromedockes:
do we want to allow passing a string instead of a list of (1) string if we want to select or drop a single column?

GaelVaroquaux:
> do we want to allow passing a string instead of a list of (1) string if we want to select or drop a single column?

I think that it would be great!

jeromedockes:
ok it is ready to review

jeromedockes:
Also, I named it `SelectCols` but maybe to be consistent with other transformers it  should be `ColsSelector` and `ColsDropper` 😐
we still don't have a decision on nouns vs verbs in #751

jeromedockes:
> Question: could it, should it, be used in https://skrub-data.org/stable/auto_examples/08_join_aggregation.html#sphx-glr-auto-examples-08-join-aggregation-py ?

I'm not sure; the only projection I see is to separate X and y from the dataset that contains both at the beginning. And then X only has 3 columns, all of which are used. Looking at the features added by the transformers in the pipeline, I don't see any that should be dropped

examples/04_fuzzy_joining.py
```diff
@@ -362,51 +362,31 @@
...
+selector = SelectCols(
+    [
+        "GDP per capita (current US$)",
+        "Life expectancy at birth, total (years)",
+        "Strength of legal rights index (0=weak to 12=strong)",
+    ]
+)
+pipeline = make_pipeline(
+    Joiner((gdppc, "Country Name"), "Country"),
+    DropCols(["Country Name"]),
```
GaelVaroquaux, line 382: Can we use the shorter syntax here?
```suggestion
    DropCols("Country Name"),
```

examples/04_fuzzy_joining.py
```diff
@@ -362,51 +362,31 @@
...
+    Joiner((life_exp, "Country Name"), "Country"),
+    DropCols(["Country Name"]),
```
GaelVaroquaux, line 384: Shorter syntax here?
```suggestion
    DropCols("Country Name"),
```

examples/04_fuzzy_joining.py
```diff
@@ -416,12 +396,16 @@
...
+    "joiner-1__match_score": [0.2, 0.9],
+    "joiner-2__match_score": [0.2, 0.9],
+    "joiner-3__match_score": [0.2, 0.9],
+}
 
 grid = GridSearchCV(pipeline, param_grid=params)
 grid.fit(df, y)
 
-print(grid.best_params_)
+print(grid.cv_results_)
```
GaelVaroquaux, line 408: I think that the following will render better on the web page, as sphinx gallery will pretty print the list.
```suggestion
grid.cv_results_
```

skrub/_select_cols.py
```diff
@@ -0,0 +1,175 @@
...
+    raise ValueError(
+        f"The following columns were not found in the input DataFrame: {diff}"
+    )
+
+
+class SelectCols(TransformerMixin, BaseEstimator):
+    """Select a subset of a DataFrame's columns.
+
+    A ``ValueError`` is raised if any of the provided column names are not in
+    the dataframe.
```
GaelVaroquaux, line 38: Should we say that this object works on both pandas and polars dataframe?

skrub/_select_cols.py
```diff
@@ -0,0 +1,175 @@
...
+        self.cols = cols
+
+    def fit(self, X, y=None):
+        """Fit the transformer.
+
+        Parameters
+        ----------
+        X : DataFrame or None
+            If `X` is a DataFrame, the transformer checks that all the column
+            names provided in ``self.cols`` can be found in `X`.
```
GaelVaroquaux, line 74: Should we be explicit that it supports both pandas and polars dataframe?
I don't think that currently skrub has uniform support for both across the whole library, hence it is useful, for now, to be explicit, IMHO.

skrub/_select_cols.py
```diff
@@ -0,0 +1,175 @@
...
+        cols = _check_columns(X, self.cols)
+        namespace, _ = get_df_namespace(X)
+        return namespace.select(X, cols)
+
+
+class DropCols(TransformerMixin, BaseEstimator):
+    """Drop a subset of a DataFrame's columns.
+
+    The other columns are kept in their original order. A ``ValueError`` is
+    raised if any of the provided column names are not in the dataframe.
```
GaelVaroquaux, line 110: Should we say that this object supports both pandas and polars

GaelVaroquaux (commented):
A few comments.

I am not sure if we have polars support. Is it easy to add? In all cases, we should be explicit

examples/04_fuzzy_joining.py
jeromedockes: oops it is actuall the best_params_ we want; I removed the print as well

jeromedockes:
> I am not sure if we have polars support

yes you can see in the [tests](https://github.com/skrub-data/skrub/pull/804/files#diff-ad85606c13d3df253cc61453e08804bf230385eb7a865b2f6f70e1d7b574de26R15)

GaelVaroquaux:
> > I am not sure if we have polars support
> 
> yes you can see in the [tests](https://github.com/skrub-data/skrub/pull/804/files#diff-ad85606c13d3df253cc61453e08804bf230385eb7a865b2f6f70e1d7b574de26R15)

Awesome!!

examples/04_fuzzy_joining.py
```diff
@@ -362,51 +362,31 @@
...
-        "passthrough",
-        [
-            "GDP per capita (current US$)",
-            "Life expectancy at birth, total (years)",
-            "Strength of legal rights index (0=weak to 12=strong)",
-        ],
-    ),
-    remainder="drop",
+# We create a selector that we will insert at the end of our pipeline, to
+# project on the relevant attributes before fitting the regressor
```
GaelVaroquaux, line 371:
```suggestion
# select on the relevant attributes before fitting the regressor
```
IMHO the terminology "project" is unusual to many users.

GaelVaroquaux (approved):
One tiny comment, but to me, this is good to go.

jovan-stojanovic (approved):
Very nice and useful features, thanks @jeromedockes!

GaelVaroquaux:
OK, merging then. Thanks @jeromedockes !

PR #806, opened 2023-10-25
DOC: Add polars to intersphinx

GaelVaroquaux:
This will enable resolution of the polars objects in the docs and examples, eg on
https://skrub-data.org/stable/generated/skrub.AggJoiner.html

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+#  -*- coding: utf-8 -*-
 #
 # skrub documentation build configuration file, created by
 # sphinx-quickstart on Tue Mar 13 14:34:47 2018.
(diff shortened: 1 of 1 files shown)


jeromedockes:
after uncompressing the `objects.inv` it does not seem to contain a class entry for DataFrame nor Series

```python
from urllib.request import urlopen
import zlib

url = "https://pola-rs.github.io/polars/py-polars/html/objects.inv"
with urlopen(url) as resp:
    data = resp.read().partition(b"zlib.\n")[2]
text = zlib.decompress(data).decode("utf-8")
print("polars.Array py:class" in text)
print("polars.DataFrame py:class" in text)
```

```
True
False
```

I'll open an issue on the polars repo

skrub/_agg_joiner.py
```diff
@@ -82,7 +82,7 @@ class AggJoiner(BaseEstimator, TransformerMixin):
     The grouping columns used during the aggregation are the columns used
     as keys for joining.
 
-    Accepts :obj:`pandas.DataFrame` and :obj:`polars.DataFrame` inputs.
+    Accepts :obj:`pandas.DataFrame` and :class:`polars.DataFrame` inputs.
```
jeromedockes, line 85: I guess the same change could be done in aggtarget

btw if we want to specify in all skrub docstrings if polars inputs are accepted, that info is missing from many of them

jeromedockes:
actually there is already an issue: https://github.com/pola-rs/polars/issues/7027

jeromedockes (approved):

jeromedockes:
the rendered docs look good, merging! thanks!

GaelVaroquaux:
Thanks Jérôme!!

Issue #766, opened 2023-09-29
DOC Remove `filterwarnings` from some examples


Vincent-Maladiere:
### Describe the issue linked to the documentation

Some examples filter all warnings, but we want those for debugging, catching deprecation, and getting closer to the user reality.

We need to remove `warnings.filterwarnings("ignore")` —and fix the potential warnings— from the following files:

- [x] `03_datetime_encoder.py`
- [x] `04_fuzzy_joining.py`

Tialo:
https://github.com/skrub-data/skrub/blob/main/skrub/_fuzzy_join.py#L359

If it warns whenever used, probably it's bad idea to remove `filterwarnings`.

Tialo:
There is also warning in https://skrub-data.org/stable/auto_examples/02_investigating_dirty_categories.html

Vincent-Maladiere:
> [`main`/skrub/_fuzzy_join.py#L359](https://github.com/skrub-data/skrub/blob/main/skrub/_fuzzy_join.py?rgh-link-date=2023-09-29T20%3A51%3A01Z#L359)
> 
> If it warns whenever used, probably it's bad idea to remove `filterwarnings`.

Indeed, we plan to remove `warnings.warn("This feature is still experimental.")` since the entire package is experimental.

Vincent-Maladiere:
> There is also warning in [skrub-data.org/stable/auto_examples/02_investigating_dirty_categories.html](https://skrub-data.org/stable/auto_examples/02_investigating_dirty_categories.html)
> 
> I can fix it if necessary.

That would be great, you can go ahead and create a PR!

dcor01:
Hi guys, I'm hoping to try and fix these warnings in example 3 and 4

Vincent-Maladiere:
Hey @dcor01, example 3 is already solved, but you can focus on example 4.

dcor01:
I have made a pull request let me know your thoughts. #807

dcor01:
Changed the filter warnings in example 4 in #808

dcor01:
Changed the filter warnings in example 4 in #808

Vincent-Maladiere:
Thank you @dcor01 and @Tialo, closing the issue :)

PR #807, opened 2023-10-26
PR - issue766 - Fixed the ignore warnings for example 04_fuzzy_joining

dcor01:
Working towards issue #766 

I have removed the warning thrown everytime fuzzy_join() is called. I have also got rid of the warnings.filterwarnings("ignore") in example 4. 

I have also changed a few of the titles and capitalisation in example 4 that I noticed. Happy to remove them and resubmit the pull request if you guys don't like them.

Let me know your thoughts - this is my first PR. Thanks guys

### DIFF
--- a/examples/04_fuzzy_joining.py
+++ b/examples/04_fuzzy_joining.py
@@ -110,11 +110,6 @@ gdppc.sort_values(by="Country Name").tail(7)
 # we have extracted.
 #
 
-# We will ignore the warnings:
-import warnings
-
-warnings.filterwarnings("ignore")
-
 ###############################################################################
 # .. _example_fuzzy_join:
 #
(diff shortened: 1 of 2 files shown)


examples/04_fuzzy_joining.py
```diff
@@ -253,8 +248,8 @@
     y="Happiness score",
     lowess=True,
 )
-fig.set_ylabel("Happiness index")
```
jeromedockes, line 256: thanks for the fix! the figure title changes don't seem related to filtering warnings, right?
I think the old title might easier to understand for some readers, because "effect" could be understood to imply that life expectancy has a causal effect on happiness

examples/04_fuzzy_joining.py
```diff
@@ -289,8 +284,8 @@
...
-fig.set_title("Does a country's legal rights strength lead to happiness?")
+fig.set_ylabel("Happiness Index")
+fig.set_title("Effect of a Country's Strength of Legal Rights on Happiness")
```
jeromedockes, line 288: if we're changing the title I rather like the "is x linked to y" formulation above because it is more clear that we are only talking about correlations

dcor01: Yeah fair enough. I will delete the wording changes. I just thought a question as a graph title looked a bit strange but happy to just leave them as is.

dcor01:
I will close this PR. Instead using #808 without the wording changes

PR #812, opened 2023-11-01
DOC - fix layout of search results page 

Badr-MOUFAD:
## Context of the PR

The search result page is currently justified on the right as shown in the figure below.
Reproduce that by typing in text on the search bar.

- Before
Screenshot from 2023-11-01 18-33-54

- After
Screenshot from 2023-11-01 18-39-51

## Contributions of the PR

- remove justify-content property (checked it has no side-effects)
- fix the format and syntax of ``custom.css``

### DIFF
--- a/doc/_static/css/custom.css
+++ b/doc/_static/css/custom.css
@@ -17,7 +17,6 @@
 .bd-main .bd-content {
     display: flex;
     height: 100%;
-    justify-content: flex-end;
     gap: 60px;
     /* margin: 0 20px; */
 }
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
Thanks a lot for the PR.

I'm not very good at flex css, but it seems like something important.

Would you mind doing a blame to figure out who added it, so that we can ask this person why it was important (the odds are that it will be me 😆)

Badr-MOUFAD:
> Would you mind doing a blame to figure out who added it, so that we can ask this person why it was important (the odds are that it will be me 😆)

It seems that this change emanates from https://github.com/skrub-data/skrub/pull/568.
I can't see where it is needed. But I agree, perhaps I'm lacking context.

Vincent-Maladiere (approved):
Hey, @Badr-MOUFAD, thanks for the PR. LGTM.

GaelVaroquaux:
OK, I tested things a bit, and I didn't see any adverse effects. I am merging this. Thanks!!

PR #814, opened 2023-11-02
[ENH] Enable Grid-Search for `TableVectorizer`

Vincent-Maladiere:
**What does this PR fix/address?**

Apply Gaël's suggestions and the outputs of discussion #796 to make grid-search possible.

```python

dataset = fetch_employee_salaries()
X, y = dataset.X.head(500), dataset.y.head(500)

pipe = make_pipeline(
    TableVectorizer(),
    HistGradientBoostingRegressor()
)

param_grid = {"tablevectorizer__high_cardinality_transformer__n_components": [10, 20]}

cv = GridSearchCV(pipe, param_grid)
cv.fit(X, y)
```

**What does it change?**
- Replace the default None of the transformers with their global default
- When the user actively sets transformers to `None`, they are turned into "passthrough" during fit (e.g. `high_cardinality_encoder = None` will result in `high_cardinality_encoder_ = passthrough`
- Some revamp to make cloning during init and during fit more readable


Vincent-Maladiere:
I need to address a small docstring error

skrub/_table_vectorizer.py
```diff
@@ -146,10 +150,72 @@ def _replace_missing_in_cat_col(ser: pd.Series, value: str = "missing") -> pd.Se
     return ser
 
 
-Transformer = TransformerMixin | Literal["drop", "remainder", "passthrough"]
+def _clone_if_default(transformer, DEFAULT_TRANSFORMER):
+    return clone(transformer) if transformer is DEFAULT_TRANSFORMER else transformer
```
jeromedockes, line 154:
```suggestion
def _clone_if_default(transformer, default_transformer):
    return clone(transformer) if transformer is default_transformer else transformer
```

skrub/_table_vectorizer.py
```diff
@@ -146,10 +150,72 @@ def _replace_missing_in_cat_col(ser: pd.Series, value: str = "missing") -> pd.Se
...
+
+
+def _clone_during_fit(transformer, remainder, n_jobs):
```
jeromedockes, line 157: as `_clone_during_fit`, `_check_specific_transformers` and `_propagate_n_jobs` are called by `TableVectorizer` methods, with attributes of `TableVectorizer` as their arguments, why did you choose to make them module-level functions rather than methods?

skrub/_table_vectorizer.py
```diff
@@ -146,10 +150,72 @@ def _replace_missing_in_cat_col(ser: pd.Series, value: str = "missing") -> pd.Se
...
+    if isinstance(transformer, sklearn.base.TransformerMixin):
+        return _propagate_n_jobs(clone(transformer), n_jobs)
+    elif transformer is None:
```
jeromedockes, line 160: I think we might want to avoid this conversion of `None` to `"passthrough"`:
- it is not super obvious whether `None` should mean "passthrough" or "drop"
- "passthrough" is already a convenient and more explicit way to mean "passthrough"
- if we don't give a special meaning to None now (making it an invalid input ATM) we keep the possibility to give it one later (eg if we decide to revisit our way of specifying the default parameter, for example if one day we decide we want the `TableVectorizer` to pass the `check_estimator` tests)

Vincent-Maladiere: Ok, sounds good!

skrub/_table_vectorizer.py
jeromedockes: ok I was just curious I don't have a strong opinion

skrub/_table_vectorizer.py
```diff
@@ -6,27 +6,31 @@
...
-from sklearn.utils.deprecation import deprecated
-from sklearn.utils.metaestimators import _BaseComposition
 from sklearn.utils.validation import check_is_fitted
 
 from skrub import DatetimeEncoder, GapEncoder
 from skrub._utils import parse_astype_error_message
 
+HIGH_CARDINALITY_TRANSFORMER = GapEncoder(n_components=30)
+LOW_CARDINALITY_TRANSFORMER = OneHotEncoder(
+    sparse_output=False,
```
jeromedockes, line 28: we expose the `ColumnTransformer`'s `sparse_threshold` parameter but with our default transformers the default will always be dense (even if toe onehot encoder yields many zeros)

we could consider
- pointing out in the doc that users need to change the transformers if they want sparse output
- not exposing the `sparse_threshold` and always returning dense data
- making the onehot encoder sparse by default

(not in this PR)

skrub/_table_vectorizer.py
```diff
@@ -849,11 +805,15 @@ def fit_transform(self, X: ArrayLike, y: ArrayLike = None) -> ArrayLike:
 
         # Next part: construct the transformers
         # Create the list of all the transformers.
-        all_transformers: list[tuple[str, Transformer, list[str]]] = [
+        all_transformers = [
             ("numeric", self.numerical_transformer_, numeric_columns),
             ("datetime", self.datetime_transformer_, datetime_columns),
-            ("low_card_cat", self.low_card_cat_transformer_, low_card_cat_columns),
-            ("high_card_cat", self.high_card_cat_transformer_, high_card_cat_columns),
+            ("low_card_cat", self.low_cardinality_transformer_, low_card_cat_columns),
```
jeromedockes, line 811: not super important but you could propagate the change 'card_cat' -> 'cardinality' to local variables

along that line it would be nice if we picked one of "numeric" or "numerical" and used it all the time :)

jeromedockes (approved):
LGTM, thanks!

skrub/_table_vectorizer.py
Vincent-Maladiere: numeric, since it's shorter?

jeromedockes:
I guess this one is ready to merge?

Vincent-Maladiere:
I think so!

GaelVaroquaux:
Very nice. Congratulations!

PR #817, opened 2023-11-09
fix column name in example

jeromedockes:


jeromedockes:
@Vincent-Maladiere  once the CI runs could you check & merge this 1-liner to fix the example failure introduced by #784 784 (I think)

jeromedockes:
the CI passed and the example looks good now so I'll merge

PR #819, opened 2023-11-09
ENH Using `to_datetime` within the `TableVectorizer`

Vincent-Maladiere:
**References**
Following https://github.com/skrub-data/skrub/pull/784 and https://github.com/skrub-data/skrub/pull/814

- Replace the datetime logic detection and parsing from the `TableVectorizer` with `to_datetime`.
- Revamp `_auto_cast` and `_apply_cast` which duplicate some logic and are not very straightforward.
- Update tests (lot of them)

**Additional remarks**
- We could implement a `to_numeric` method the same way we implemented `to_datetime`. The aim is to handle Polars dataframes as well.


Vincent-Maladiere:
There is some logic around dtypes in the TableVectorizer that seems dangerous.

https://github.com/skrub-data/skrub/blob/90d978c85784cd9fd8065c6a90e6e0e6ebc9bbfb/skrub/_table_vectorizer.py#L626-L632

with its test:

https://github.com/skrub-data/skrub/blob/90d978c85784cd9fd8065c6a90e6e0e6ebc9bbfb/skrub/tests/test_table_vectorizer.py#L852-L857

AFAIU, this logic suggests that when the column dtype in the test set doesn't match its dtype in the train set, the value should be omitted.

Example:
```python
import numpy as np
import pandas as pd

X_train = pd.DataFrame(dict(a=[np.nan, np.nan, np.nan]))
X_test = pd.DataFrame(dict(a=[np.nan, np.nan, "a"]))

tv = TableVectorizer()
tv.fit_transform(X_train)
# array([[nan],
#       [nan],
#       [nan]])

tv.transform(X_test)
# /Users/vincentmaladiere/INRIA/skrub/skrub/_table_vectorizer.py:885: 
# UserWarning: Value 'a' could not be converted to inferred type float64 in column 'a'.
# Such values will be replaced by NaN.
  X = self._apply_cast(X)
# array([[nan],
#       [nan],
#       [nan]])
```

I'd rather avoid imputing NaNs during `transform`, as we don't want to lose data. Having an entire column filled with nan during training is problematic anyway.

What do the @skrub-data/devs think?

Vincent-Maladiere:
I need to add some tests.

Vincent-Maladiere:
I have some additional tests to fix

Vincent-Maladiere:
I now have errors with Windows using 32-bit integers by default. I'm not sure how to debug that without a Windows computer, though.

Vincent-Maladiere:
@jeromedockes this is now reviewable :)

skrub/_table_vectorizer.py
```diff
@@ -294,18 +252,24 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
...
+        - pandas extension dtypes conversion to numpy dtype
+        - datetime conversion using ``skrub.to_datetime``
+        - numeric conversion using ``pandas.to_numeric``
+        - numeric columns with missing values are converted to float to input np.nan
+        - categorical columns with missing values are imputed with 'missing' if
+          ``impute_missing_categories`` is ``True``.
+        - categorical columns dtypes are updated with the new entries (if any)
+          during transform.
+
+    impute_missing_categories : bool, default=True
```
jeromedockes, line 267: I'm not sure I understand the behavior around missing values. for categories we replace missing values with "missing", and for strings (which get handled as categories when low cardinality) we do the opposite, replacing "false missing" values such as the empty string with np.nan

also missing values in columns with the `string` dtype raise an error
```
>>> df = pd.DataFrame({"astring": ["Aa", "Bb", None]}).convert_dtypes()
>>> TableVectorizer().fit_transform(df) #TypeError
```

skrub/_table_vectorizer.py
```diff
@@ -32,79 +29,41 @@
...
-        return date_format_monthfirst.iloc[0]
-    elif date_format_dayfirst.nunique() == 1:
-        # only this format works for all the rows
-        return date_format_dayfirst.iloc[0]
-    else:
-        # more than two different formats were found
-        # TODO: maybe we could deal with this case
-        return
+    columns = X.columns
+    X_out = X.copy()
```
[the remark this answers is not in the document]
Vincent-Maladiere: I'm sure to understand your point, could you elaborate on "because we want to replace some columns but we don't need to copy the values"?

skrub/_table_vectorizer.py
```diff
@@ -32,79 +29,41 @@
...
+                X_out[column] = pd.to_numeric(X_col, errors="raise")
+            except (ValueError, TypeError):
+                pass
+    return X_out
 
 
-def _has_missing_values(df: pd.DataFrame | pd.Series) -> bool:
+def _has_missing_values(df):
     """
     Returns True if `array` contains missing values, False otherwise.
```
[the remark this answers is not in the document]
Vincent-Maladiere: Nice catch. This is indeed for a series. Do you prefer:
a. reverting to `any` to keep it generalistic
b. removing the function since it's a one-liner, and we won't need to document it

skrub/_table_vectorizer.py
Vincent-Maladiere: Yes, thank you for asking. It doesn't make much sense to me either. I simplified the logic without altering it as much as possible. We have to discuss with the rest of the time some of the choices made for impute such as this one.

skrub/_table_vectorizer.py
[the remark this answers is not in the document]
jeromedockes: I mean that copy() is for copying the data in the dataframe, so we might use it if we were going to modify some values in place inside its columns. here we want to build a new dataframe while reusing some columns (without modification) and assigning some new columns so I thought something like 
```
new_columns = {}
for column in dataframe:
    new_columns[column name] = to_numeric(column) if possible else column
return DataFrame(new_columns)
```

might express that more clearly and avoid copying (i) columns that are already numeric and (ii) columns that cannot be converted to numeric

but fee free to ignore if you prefer making a copy

skrub/_table_vectorizer.py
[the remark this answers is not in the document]
jeromedockes: I think b because it is only used in 2 places, and adds an indirection without making the code shorter, encapsulating something that might change or anything complex

skrub/_table_vectorizer.py
jeromedockes: that makes sense, we can open another issue

skrub/_table_vectorizer.py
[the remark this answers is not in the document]
Vincent-Maladiere: Got it, it makes sense for both memory management and readability

skrub/_table_vectorizer.py
Vincent-Maladiere: Here it is: https://github.com/skrub-data/skrub/issues/837

Vincent-Maladiere:
TODO:
- [ ] remove the imputing logic entirely
- [ ] keep the np.nan imputing when the dtype goes from numeric to category / str / object

skrub/_table_vectorizer.py
```diff
@@ -32,79 +35,34 @@
...
+            # We don't use errors="ignore" because it casts string
+            # and categories to object dtype in Pandas < 2.0.
+            # TODO: replace 'raise' by 'ignore' and remove the exception
+            # catching when the minimum pandas version of skrub is 2.0.
+            try:
+                X_out[col] = pd.to_numeric(X[col], errors="raise")
+                continue
+            except (ValueError, TypeError):
+                pass
+        X_out[col] = X[col]
```
jeromedockes, line 61: are we missing an `else`?

skrub/_table_vectorizer.py
```diff
@@ -32,79 +35,34 @@
...
+    return pd.DataFrame(X_out, index=X.index)
 
 
-def _replace_false_missing(df: pd.DataFrame | pd.Series) -> pd.DataFrame | pd.Series:
+def _replace_false_missing(df):
```
jeromedockes, line 65: in practice it seems to be only used with series as input. can we change the docstring to say it only takes series? I  think it would reduce a bit the cognitive load when reading it and it will ease porting to polars or the dataframe api too

skrub/_table_vectorizer.py
```diff
@@ -495,147 +435,101 @@ def _clone_transformers(self):
...
+        X : :obj:`~pandas.DataFrame`
+            The same :obj:`~pandas.DataFrame`, with its columns cast.
         """
-        # Handle missing values
         for col in X.columns:
-            # Convert pandas' NaN value (pd.NA) to numpy NaN value (np.nan)
-            # because the former tends to raise all kind of issues when dealing
-            # with scikit-learn (as of version 0.24).
-            if _has_missing_values(X[col]):
+            # Convert pd.NA to np.nan because the former tends to raise all
```
jeromedockes, line 467: The position of this comment makes it seem like `_replace_false_missing` makes the pd.NA -> np.nan conversion, but that actually happens later. could we move it or reward it ?

jeromedockes (commented):
wip

skrub/_table_vectorizer.py
```diff
@@ -240,8 +198,6 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
         'passthrough' to return the unencoded columns.
         The default transformer is \
             ``OneHotEncoder(handle_unknown="ignore", drop="if_binary")``.
```
jeromedockes, line 200: here we could add a short sentence saying that when using a tree-based model on the resulting features we may want to use a different transformer

skrub/_table_vectorizer.py
```diff
@@ -293,40 +243,35 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
...
-        (we are using an encoder that does not support missing values and/or
-        specific versions of pandas, numpy and scikit-learn).
-        'force' will impute missing values in all categorical columns.
-        'skip' will not impute at all.
-        When imputed, missing values are replaced by the string 'missing'
-        before being encoded.
-        As imputation logic for numerical features can be quite intricate,
-        it is left to the user to manage.
-        See also attribute :attr:`~skrub.TableVectorizer.imputed_columns_`.
+    activate_auto_cast : bool, default=True
```
jeromedockes, line 246: I'm not sure we need to add "activate"

skrub/_table_vectorizer.py
```diff
@@ -293,40 +243,35 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
...
+        estimator must support ``fit`` and ``transform``.
         Note that using this feature requires that the DataFrame columns
-        input at :term:`fit` and :term:`transform` have identical order.
+        input at ``fit`` and ``transform`` have identical order.
 
     sparse_threshold : float, default=0.0
         If the output of the different transformers contains sparse matrices,
         these will be stacked as a sparse matrix if the overall density is
-        lower than this value. Use `sparse_threshold=0` to always return dense.
+        lower than this value. Use ``sparse_threshold=0`` to always return dense.
```
jeromedockes, line 274: here we should probably note that with the default encoders output will never be sparse

skrub/_table_vectorizer.py
```diff
@@ -293,40 +243,35 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
...
-        By specifying `remainder='drop'`, only the specified columns
+        By specifying ``remainder='drop'``, only the specified columns
         in `transformers` are transformed and combined in the output, and the
         non-specified columns are dropped.
         By setting `remainder` to be an estimator, the remaining
         non-specified columns will use the `remainder` estimator. The
-        estimator must support :term:`fit` and :term:`transform`.
+        estimator must support ``fit`` and ``transform``.
         Note that using this feature requires that the DataFrame columns
-        input at :term:`fit` and :term:`transform` have identical order.
```
jeromedockes, line 324: in the line above I don't know what "this feature" means -- passing a Transformer as the remainder?
In general I think we may want to impose always having the same column names (maybe in the same order?) in fit and transform for all estimators

skrub/_table_vectorizer.py
```diff
@@ -368,13 +313,10 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
         ``len(transformers_)==len(transformers)+1``, otherwise
         ``len(transformers_)==len(transformers)``.
 
-    types_ : dict mapping of int to type
+    type_per_column_ : dict mapping of int to type
```
jeromedockes, line 316: I think we can come up with a better name but I don't have one :sweat_smile:  maybe something like `inferred_column_types_`, `autocast_dtypes_`, `detected_types_`, :thinking:

skrub/_table_vectorizer.py
```diff
@@ -368,13 +313,10 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
...
         A mapping of inferred types per column.
         Key is the index of a column, value is the inferred dtype.
-        Exists only if `auto_cast=True`.
-
-    imputed_columns_ : list of str
-        The list of columns in which we imputed the missing values.
+        Exists only if ``activate_auto_cast=True``.
```
jeromedockes, line 319: it should exist in all cases, when autocast is disabled then it will just be the input's dtypes

skrub/_table_vectorizer.py
```diff
@@ -495,147 +435,101 @@ def _clone_transformers(self):
...
                 # Some numerical dtypes like Int64 or Float64 only support
-                # pd.NA, so they must be converted to np.float64 before.
-                if pd.api.types.is_numeric_dtype(X[col]):
+                # pd.NA, so they must be converted to np.float64 before imputing
+                # with np.nan.
+                if is_numeric_dtype(X[col]):
                     X[col] = X[col].astype(np.float64)
+
+                # We impute with np.nan in all columns regardless of their type,
+                # as we might have some false missing in numerical columns for instance.
```
jeromedockes, line 479: I'm not sure I understand the comment -- what would be false missing in numerical columns?

jeromedockes (commented):
a few more

skrub/_table_vectorizer.py
```diff
@@ -495,147 +435,101 @@ def _clone_transformers(self):
             self.n_jobs,
         )
 
-    def _auto_cast(self, X: pd.DataFrame) -> pd.DataFrame:
-        """Takes a dataframe and tries to convert its columns to their best possible
-        data type.
+    def auto_cast(self, X, reset=True):
```
jeromedockes, line 438: why make it public?

skrub/_table_vectorizer.py
```diff
@@ -495,147 +435,101 @@ def _clone_transformers(self):
...
+            X = _to_numeric(X)
+            self.type_per_column_ = X.dtypes.to_dict()
+
+        else:
+            # Update categorical dtype.
+            error_msg = (
+                "This %(name)s instance is not fitted yet. "
+                "Please run auto_cast(X, reset=True) first."
+            )
+            check_is_fitted(self, ["type_per_column_"], msg=error_msg)
```
jeromedockes, line 510: I think we should keep `_auto_cast` private and remove this check

skrub/_table_vectorizer.py
```diff
@@ -495,147 +435,101 @@ def _clone_transformers(self):
...
+            # As this behavior is more aggressive than skrub's to_datetime
+            # or _to_numeric and only makes sense for the TableVectorizer, we
+            # define it here rather than within these two functions.
+            # See: https://github.com/skrub-data/skrub/issues/837
+            for column, dtype in self.type_per_column_.items():
+                if is_numeric_dtype(dtype):
+                    X[column] = pd.to_numeric(X[column], errors="coerce")
+
+                elif is_datetime64_any_dtype(dtype):
+                    X[column] = pd.to_datetime(X[column], errors="coerce")
```
jeromedockes, line 528: IIRC we wanted to fit a DatetimeEncoder during train and use it to parse dates (with the same format) in transform?

skrub/_table_vectorizer.py
Vincent-Maladiere: `inferred_column_types_` sounds better

skrub/_table_vectorizer.py
Vincent-Maladiere: While writing it, I thought it might be helpful for the user. Now I realize it will just bring confusion, so I'll revert.

skrub/_table_vectorizer.py
```diff
@@ -671,10 +565,18 @@ def _check_X(self, X):
             # Create a copy to avoid altering the original data.
             X = X.copy()
 
+        # Check for duplicate column names.
+        duplicate_columns = {k for k, v in Counter(X.columns).items() if v > 1}
```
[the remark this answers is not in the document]
Vincent-Maladiere: Absolutely

skrub/_table_vectorizer.py
Vincent-Maladiere: Not 100% sure either, let's tackle this issue later.

skrub/_table_vectorizer.py
Vincent-Maladiere: That doesn't make much sense to me either. After looking at the tests, I think we can safely remove it.

skrub/_table_vectorizer.py
Vincent-Maladiere: Addressed in the follow-up PR: https://github.com/skrub-data/skrub/pull/848

skrub/_table_vectorizer.py
```diff
@@ -495,147 +449,95 @@ def _clone_transformers(self):
...
 
-    def _apply_cast(self, X: pd.DataFrame) -> pd.DataFrame:
-        """Takes a dataframe, and applies the best data types learnt during fitting.
+        else:
+            # Update categorical dtype.
+            error_msg = (
+                "This %(name)s instance is not fitted yet. "
+                "Please run auto_cast(X, reset=True) first."
+            )
+            check_is_fitted(self, ["inferred_column_types_"], msg=error_msg)
```
jeromedockes, line 518: we can remove this check; this is already checked by `transform`. the users don't call this private method themselves so an error message telling them what should be done before calling it doesn't help them

skrub/_table_vectorizer.py
```diff
@@ -495,147 +449,95 @@ def _clone_transformers(self):
...
+            category_columns = X.select_dtypes("category").columns
+            for col in category_columns:
+                dtype = self.inferred_column_types_[col]
+                dtype = _union_category(X[col], dtype)
```
jeromedockes, line 522: I'm not sure we should update the type to add categories in transform but let's deal with it in #848

skrub/_table_vectorizer.py
```diff
@@ -671,10 +573,18 @@ def _check_X(self, X):
             # Create a copy to avoid altering the original data.
             X = X.copy()
 
+        # Check for duplicate column names.
+        duplicate_columns = {k for k, v in Counter(X.columns).items() if v > 1}
+        if len(duplicate_columns) > 0:
```
jeromedockes, line 578: I can't comment on it but in this function it seems we copy X if it is a dataframe but not if it is an array.
but we will remove support for arrays anyway in #831.
also we already copy all columns in replace_missing_indicators so we should avoid the `X.copy()`; we can deal with that in one of the other PRs

jeromedockes (approved):
another batch of small comments but otherwise LGTM and we can deal with them in #848

skrub/_table_vectorizer.py
```diff
@@ -233,15 +193,26 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
...
-        'passthrough' to return the unencoded columns.
-        The default transformer is \
-            ``OneHotEncoder(handle_unknown="ignore", drop="if_binary")``.
-        Features classified under this category are imputed based on the
-        strategy defined with `impute_missing`.
+        Can either be a:
+        - transformer object instance (e.g. OneHotEncoder)
+        - a Pipeline containing the preprocessing steps
+        - 'drop' for dropping the columns
+        - 'remainder' for applying `remainder`
```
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, I agree. We should simplify this in a subsequent PR.

skrub/_table_vectorizer.py
```diff
@@ -419,10 +375,11 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
     >>> tv.transformers_
     [('numeric', 'passthrough', ['year_first_hired']), \
 ('datetime', DatetimeEncoder(), ['date_first_hired']), \
-('low_card_cat', OneHotEncoder(drop='if_binary', handle_unknown='ignore', \
+('low_cardinality', OneHotEncoder(drop='if_binary', handle_unknown='ignore', \
```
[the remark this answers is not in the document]
Vincent-Maladiere: I agree the transformer names should be displayed.
Also, let's remove `transformer_weights` in a subsequent PR, as I don't think this parameter is worth its cognitive load.

skrub/_table_vectorizer.py
Vincent-Maladiere: Yes, I forgot to update that.

skrub/_table_vectorizer.py
Vincent-Maladiere: I'm not 100% sure of the rational of doing it, we should discuss it

skrub/_table_vectorizer.py
```diff
@@ -495,147 +449,95 @@ def _clone_transformers(self):
...
+            the mapping between columns of the training dataframe and their types.
+            If set to ``False`` (during transform), updates ``inferred_column_types_``
+            for the categorical columns with the new categories seen during transform.
+
         Returns
         -------
-        :obj:`~pandas.DataFrame`
-            The same :obj:`~pandas.DataFrame`, with its columns cast to their
-            best possible data type.
+        X : :obj:`~pandas.DataFrame`
```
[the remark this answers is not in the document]
Vincent-Maladiere: Regarding that, if we want to avoid using numpy array in the future, we should avoid copying the input with scikit-learn check_array. One option would be to copy the dataframe in `auto_cast` and return a copy. IMHO, parsers shouldn't modify their inputs in place.
Let's discuss this in #848.

PR #823, opened 2023-11-13
[ENH] make `to_datetime` more robust to bad inputs

Vincent-Maladiere:
**Reference**
Follows-up on https://github.com/skrub-data/skrub/pull/784

- Heterogeneous arrays containing arbitrary objects like sets raise TypeErrors. This PR fixes these by catching TypeError, which allows returning the input.
- Heterogeneous arrays containing datetimes objects like `pd.Timestamp` and `datetime.datetime` also raise errors, because Pandas' `guess_datetime_format` only accepts string values. We now skip the call to `guess_datetime_format`  by ensuring the input array is invariant under `astype(str)` transformation.

Before:
```python
import pandas as pd
from skrub import to_datetime

to_datetime([pd.Timestamp("2019-01-01"), "yo"])
# TypeError: float() argument must be a string or a real number, not 'Timestamp'

to_datetime([pd.Timestamp("2019-01-01"), "2020-01-01"])
# TypeError: float() argument must be a string or a real number, not 'Timestamp'

to_datetime(['2020-01-01', {'hello'}])
# TypeError: <class 'set'> is not convertible to datetime, at position 1

```

After:
```python
import pandas as pd
from skrub import to_datetime

to_datetime([pd.Timestamp("2019-01-01"), "yo"])
# no-op
#  array([Timestamp('2019-01-01 00:00:00'), 'yo'], dtype=object)

to_datetime([pd.Timestamp("2019-01-01"), "2020-01-01"])
# parsing
# array(['2019-01-01T00:00:00.000000000', '2020-01-01T00:00:00.000000000'],
#      dtype='datetime64[ns]')

to_datetime(['2020-01-01', {'hello'}])
# no-op
# array(['2020-01-01', {'hello'}], dtype=object)
```

### DIFF
--- a/skrub/_datetime_encoder.py
+++ b/skrub/_datetime_encoder.py
@@ -292,21 +292,21 @@ def _get_datetime_column_indices(X_split, dayfirst=True):
     index_to_format = {}
 
     for col_idx, X_col in enumerate(X_split):
-        X_col = X_col[pd.notnull(X_col)]
-
-        # convert pd.TimeStamp to np.datetime64
-        if all(isinstance(val, pd.Timestamp) for val in X_col):
-            X_col = X_col.astype("datetime64")
+        X_col = X_col[pd.notnull(X_col)]  # X_col is a numpy array
 
         if _is_column_datetime_parsable(X_col):
             indices.append(col_idx)
 
-            if np.issubdtype(X_col.dtype, np.datetime64):
+            # _guess_datetime_format only accept string columns.
+            # We need to filter out columns of object dtype that
+            # contains e.g., datetime.datetime or pd.Timestamp.
+            X_col_str = X_col.astype(str)
+            if np.array_equal(X_col, X_col_str):
+                datetime_format = _guess_datetime_format(X_col)
+            else:
                 # We don't need to specify a parsing format
                 # for columns that are already of type datetime64.
                 datetime_format = None
-            else:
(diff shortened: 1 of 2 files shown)


skrub/_datetime_encoder.py
```diff
@@ -292,21 +292,21 @@ def _get_datetime_column_indices(X_split, dayfirst=True):
...
 
     for col_idx, X_col in enumerate(X_split):
-        X_col = X_col[pd.notnull(X_col)]
-
-        # convert pd.TimeStamp to np.datetime64
-        if all(isinstance(val, pd.Timestamp) for val in X_col):
-            X_col = X_col.astype("datetime64")
+        X_col = X_col[pd.notnull(X_col)]  # X_col is a numpy array
 
         if _is_column_datetime_parsable(X_col):
```
jeromedockes, line 297: I guess if the col is already datetime or Timestamp or numeric -- basically anything else than str or object we can skip this and set datetime_format to None

Vincent-Maladiere: We still want columns with datetime dtype to be added in the datetime index list

jeromedockes: I meant for datetime columns we don't need to cast to string but anyway it doesn't matter

jeromedockes (approved):
LGTM, thanks!

PR #828, opened 2023-11-15
[MAIN] Preserve extension dtypes and categorical dtypes in `to_datetime`

Vincent-Maladiere:
**What does this PR address?**

`to_datetime` converts every dataframe's column into a list of vectors using `to_numpy()` for every column. This results in the loss of extension dtypes (e.g.,`pd.Int64Dtype`) and categorical dtypes since we recreate the original dataframe from this mapping of numpy 1d arrays.

- This PR suggests a simple **temporary** workaround that casts back to their original types all columns that are not parsed as datetime. In the future, we need to stop converting series with `to_numpy()` and specialize the datetime parsing logic for Pandas and Polars. We will remove this workaround then.
- It also extends the `make_series` and `make_dataframe` methods to accept dtypes argument.

Before
```python
import pandas as pd
from skrub import to_datetime

to_datetime(pd.Series(["1", "2", "3"], dtype="category"))
# 0    1
# 1    2
# 2    3
# dtype: object
```

After
```python
import pandas as pd
from skrub import to_datetime

to_datetime(pd.Series(["1", "2", "3"], dtype="category"))
# 0    1
# 1    2
# 2    3
# dtype: category
# Categories (3, object): ['1', '2', '3']
```

### DIFF
--- a/skrub/_dataframe/_pandas.py
+++ b/skrub/_dataframe/_pandas.py
@@ -35,6 +35,8 @@ def make_dataframe(X, index=None, dtypes=None):
     """
     df = pd.DataFrame(X, index=index)
     if dtypes is not None:
+        # 'df.astype(None)' might raise a ValueError because
+        # it tries to cast all columns to floats.
         df = df.astype(dtypes)
     return df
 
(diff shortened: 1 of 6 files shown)


jeromedockes (approved):
LGTM as a temporary fix

indeed the better solution, as you say, is to avoid calling to_numpy on those columns

the dtypes is a useful addition to make_dataframe and make_series. in the near future we may want to replace those with the dataframe api's dataframe_from_columns and column_from_sequence or column_from_1d_array

PR #829, opened 2023-11-17
[FEAT] Drop the TargetEncoder entirely

Vincent-Maladiere:
`TargetEncoder` has been available in scikit-learn since version 1.3, with multiclass support in 1.4.
In the meantime, PR https://github.com/skrub-data/skrub/pull/658 suggests replacing the skrub implementation with a scikit-learn vendor.

However, this creates more confusion than it solves problems and forces us to start a deprecation cycle at the beginning of the release. I suggest dropping the support for the TargetEncoder entirely before the release to make our lives easier.


GaelVaroquaux:
Works for me.

Did you git grep to check that we had no leftover references to TargetEncoder?

Vincent-Maladiere:
Yes, I did. We didn't reference it much in the doc or the code.
Not directly related, I wonder if we should add a link to the [preprocessing section of scikit-learn](https://scikit-learn.org/1.3/modules/preprocessing.html) in the [encoding section of skrub](https://skrub-data.org/stable/encoding.html), or at least mention the popular alternatives, like TF-IDF, TargetEncoder, WDYT?

jeromedockes (approved):

jeromedockes:
great, thanks!

Issue #835, opened 2023-11-22
datetimeencoder is very slow


MarcoGorelli:
### Describe the bug

Looks like the format is being guessed for every single element, twice (once with day first, once with month first)

`np.vectorize` doesn't speed things up, it's just syntactic sugar

The code below has just 70 thousand rows, but it 14 seconds to execute on my laptop

### Steps/Code to Reproduce
```python
from pprint import pprint
import pandas as pd

data = pd.DataFrame({
    'date.utc': pd.date_range('1900-01-01', '2100-01-01', freq='1D').strftime('%Y-%m-%d'),
    'city': 'Paris',
    'value': 3.
}
)
print('data shape: ', data.shape)
# Extract our input data (X) and the target column (y)
y = data["value"]
X = data[["city", "date.utc"]]

X

from skrub import to_datetime

X = to_datetime(X)
X.dtypes

from sklearn.preprocessing import OneHotEncoder
from sklearn.compose import make_column_transformer
from skrub import DatetimeEncoder

encoder = make_column_transformer(
    (OneHotEncoder(handle_unknown="ignore"), ["city"]),
    (DatetimeEncoder(add_day_of_the_week=True, resolution="minute"), ["date.utc"]),
    remainder="drop",
)

X_enc = encoder.fit_transform(X)
pprint(encoder.get_feature_names_out())
```

### Expected Results

No more than 1 second, probably 😄 

### Actual Results

the results are correct, just too slow

### Versions

```shell
System:
    python: 3.11.6 (main, Oct 23 2023, 22:48:54) [GCC 11.4.0]
executable: /home/marcogorelli/skrub-dev/.venv/bin/python
   machine: Linux-5.10.102.1-microsoft-standard-WSL2-x86_64-with-glibc2.35

Python dependencies:
      sklearn: 1.3.0
          pip: 23.1.2
   setuptools: 65.5.0
        numpy: 1.25.2
        scipy: 1.11.2
       Cython: None
       pandas: 2.1.1
   matplotlib: None
       joblib: 1.3.2
threadpoolctl: 3.2.0

Built with OpenMP: True

threadpoolctl info:
       user_api: openmp
   internal_api: openmp
    num_threads: 16
         prefix: libgomp
       filepath: /home/marcogorelli/skrub-dev/.venv/lib/python3.11/site-packages/scikit_learn.libs/libgomp-a34b3233.so.1.0.0
        version: None

       user_api: blas
   internal_api: openblas
    num_threads: 16
[...]

PR #836, opened 2023-11-22
[MAIN] Accelerate guess datetime format

Vincent-Maladiere:
**Reference**

This PR addresses https://github.com/skrub-data/skrub/issues/835

- Sample 10 rows to infer the datetime format instead of using the whole array, yielding a 500x speed up on a 1d array of size 100k.
- Add some tests using 20 different `random_state`

### DIFF
--- a/skrub/_datetime_encoder.py
+++ b/skrub/_datetime_encoder.py
@@ -9,7 +9,7 @@ from pandas.api.types import is_datetime64_any_dtype, is_numeric_dtype
 from sklearn.base import BaseEstimator, TransformerMixin
 from sklearn.utils import check_array
 from sklearn.utils.fixes import parse_version
-from sklearn.utils.validation import check_is_fitted
+from sklearn.utils.validation import check_is_fitted, check_random_state
 
 from ._dataframe._namespace import get_df_namespace
 
@@ -38,6 +38,7 @@ MIXED_FORMAT = "mixed" if _is_pandas_format_mixed_available() else None
 def to_datetime(
     X,
     errors="coerce",
+    random_state=None,
     **kwargs,
 ):
     """Convert the columns of a dataframe or 2d array into a datetime representation.
@@ -75,6 +76,10 @@ def to_datetime(
         match this format to ``NaT``.
 
         Note that the ``'ignore'`` option is not used and will raise an error.
+
+    random_state : int, RandomState instance or None, default=None
+        Determines random number generation for the subsampling during
+        datetime guessing. Use an int to make the randomness deterministic.
 
     **kwargs : key, value mappings
         Other keyword arguments are passed down to :func:`pandas.to_datetime`.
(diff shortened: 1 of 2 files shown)


skrub/_datetime_encoder.py
```diff
@@ -38,6 +38,7 @@ def _is_pandas_format_mixed_available():
 def to_datetime(
     X,
     errors="coerce",
+    random_state=None,
```
jeromedockes, line 41: i'm wondering if we should make it seeded by default, and even maybe not expose the random state

skrub/_datetime_encoder.py
```diff
@@ -403,6 +422,12 @@ def _guess_datetime_format(X_col):
     # raises a TypeError.
     # We have to convert these to the object dtype first.
     X_col = X_col.astype("object")
+
+    # Subsample samples for fast format estimation
+    size = min(X_col.shape[0], 10)
```
jeromedockes, line 427: i guess we could expose the number of subsamples as a parameter. although if days are roughly uniformly distributed the probability of getting one < 12 10 times is very low so that's probably ok. we might want to give the constant a name or mention the value in the docstring

skrub/_datetime_encoder.py
Vincent-Maladiere: In scikit-learn, you always have a `random_state` parameter for testing purposes and reproducibility. But we might not always want to be consistent with scikit-learn; I don't have a strong opinion here.

LeoGrin:
Should we use a bit more than 10 rows? For 10 rows the proba of having all days below 12 is ~1/1000 for uniformly spreaded dates, which is small but still large enough for the format detected to change when doing things like hyperparameter optimisation. For 20 days the proba drops to 1e-8.

Vincent-Maladiere:
Let's make it 30

GaelVaroquaux:
> Should we use a bit more than 10 rows? For 10 rows the proba of having all days below 12 is ~1/1000 for uniformly spreaded dates, which is small but still large enough for the format detected to change when doing things like hyperparameter optimisation. For 20 days the proba drops to 1e-8.

Sounds good!

Vincent-Maladiere:
I have weird and seemingly unrelated errors with the GapEncoder

Vincent-Maladiere:
The last version of Polars (0.19.16 at the time of testing) was the source of errors, it has been yanked from Pypi since (replaced with version 0.19.18).

Polars seems to release very frequently and yank packages quite often ([see Pypi release history](https://pypi.org/project/polars/#history)). Maybe we should fix and update the Polars version regularly for testing instead of using the latest. This would limit this kind of error on our CI, WDYT?

skrub/_datetime_encoder.py
jeromedockes: I don't either. we can discuss at the next meeting which generators should be seeded by default. In the meanwhile LGTM!

jeromedockes (approved):

Vincent-Maladiere:
Merging since the changes are quite small

Issue #832, opened 2023-11-21
`get_feature_names_out` returns lists instead of numpy arrays


Vincent-Maladiere:
### Problem Description

This issue is very minor. In scikit-learn, calling `get_feature_names_out` from any estimator returns a numpy array, while in skrub it returns a list. It might be worth bringing consistency between the two libraries by outputting numpy arrays in skrub and to limit the potential confusion for our users.
What do @skrub-data/devs think?

### Feature Description

This issue suggests returning a numpy array instead of a list in `get_feature_names_out` of all skrub estimators that implement it.

GaelVaroquaux:
> This issue is very minor. In scikit-learn, calling get_feature_names_out from any estimator returns a numpy array, while in skrub it returns a list. It might be worth bringing consistency between the two libraries by outputting numpy arrays in skrub and to limit the potential confusion for our users.

Sounds good to me

jeromedockes:
fixed by #838

PR #838, opened 2023-11-24
[FEAT] `get_feature_names_out` returns an array instead of a list

Vincent-Maladiere:
Addresses https://github.com/skrub-data/skrub/issues/832


jeromedockes:
cool, thanks @Vincent-Maladiere ! LGTM

jeromedockes:
as it is a small change I don't think we need a second review

PR #861, opened 2023-12-14
Fix the homepage example links

Vincent-Maladiere:
The encoding and gap_encoder example links on the homepage are currently broken.


jeromedockes:
thanks for fixing it! I guess those were easier to miss because they are manually inserted in the html template, rather than sphinx references

jeromedockes (approved):

PR #867, opened 2023-12-15
[DOC] Multiple small improvements

TheooJ:
I've added some backslashes for newlines, as well as a code block

I've renamed the joined colums from `_capitals` to `_right` in the `fuzzy_join` example because I think it's clearer, but I'm open to other names

### DIFF
--- a/skrub/_fuzzy_join.py
+++ b/skrub/_fuzzy_join.py
@@ -148,11 +148,11 @@ def fuzzy_join(
     ...     left_table,
     ...     right_table,
     ...     on="Country",
-    ...     suffix="_capitals",
+    ...     suffix="_right",
     ...     max_dist=1.0,
     ...     add_match_info=False,
     ... )
-      Country Country_capitals Capital_capitals
+      Country Country_right Capital_right
     0  France           France            Paris
     1  Italia            Italy             Rome
     2   Spain              NaN              NaN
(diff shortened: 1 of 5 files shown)


skrub/_fuzzy_join.py
```diff
@@ -148,35 +148,35 @@ def fuzzy_join(
     ...     left_table,
     ...     right_table,
     ...     on="Country",
-    ...     suffix="_capitals",
+    ...     suffix="_right",
     ...     max_dist=1.0,
     ...     add_match_info=False,
     ... )
-      Country Country_capitals Capital_capitals
+      Country Country_right Capital_right
```
jeromedockes, line 155: This is an example where having a prefix or format string would be better than a suffix: 
`Country Capitals.Country Capitals.Capital` would be even clearer

jeromedockes (approved):
thanks a lot @TheooJ !

skrub/_fuzzy_join.py
jeromedockes: I guess we could make the same change in the `Joiner` docstring?
:thinking:  another option would be to change the column names so that we don't need the suffix in the toy example but it's probably useful to show it

TheooJ: That's true it could be changed in the `Joiner` docstring too -> i'll do it
Changing the col names could also work. I'll just vouch for removing `Country_capitals` and `Capitals_capitals` because I find it misleading

TheooJ: In the `fuzzy_join` example I'll suggest the suffix `_right`, and in the `Joiner` the suffix `_aux` just to match the class/function arguments

Vincent-Maladiere: Great, thanks

Vincent-Maladiere (approved):
Thanks, @TheooJ; let's change the docstring of Joiner and this LGTM!

jeromedockes:
@TheooJ please merge the small PR I opened on your branch (just whitespace in the displayed output dataframes) and then we can merge this one. Thanks!

Vincent-Maladiere:
You have errors that seems unrelated to this PR. Could you run the CI another time before we merge this?

```python
 git commit -m "running ci to reproduce the error" --allow-empty
```

Vincent-Maladiere (commented):
Awesome! Thank you, @TheooJ

PR #868, opened 2023-12-18
MAINT Support Python 3.8+

### LINKED ISSUE #815: support python 3.8 & 3.9
Why is the oldest python version supported by skrub 3.10?
end of support for 3.8 is only [scheduled](https://www.python.org/downloads/) for October 2024 so unless there is a compelling reason not to I think skrub should work on python 3.8
I see that would require changing some of the type annotations, are there other reasons why we chose 3.10?

jeromedockes:
closes #815 

apparently the only thing that prevents support for python3.8 and python3.9 is the use of operators on built-in types (like `list[str]`, `list | dict` ) for type annotations.

the type annotations as they are are not usable anyway: (i) many functions don't have them and more importantly (ii) a large part of those that exist are incorrect; `mypy ./skrub` yields over a hundred errors.
So for users I think supporting all python version (3.8 EOL is Oct 2024 and 3.9 is Oct 2025) is more important.

This is what we agreed on in the skrub IRL meeting

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -35,7 +35,7 @@ jobs:
         fail-fast: false
         matrix:
             os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-            python-version: ["3.10", "3.11", "3.12"]
+            python-version: ["3.8", "3.11", "3.12"]
             # Install the min or latest dependencies for skrub
             # as defined in setup.cfg at [options.extra_require].
             #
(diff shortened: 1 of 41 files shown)


jeromedockes:
TODO
- [ ] reduce number of CI builds

glemaitre:
Nice. I can review this one once you settle on it.

jeromedockes:
thanks for reviving this @parkervg ! I guess haven't really made a decision on this one as we were focusing on other things; I'll update the PR and we'll make a decision before the next release

glemaitre:
I would be happy to remove the type annotations.

GaelVaroquaux:
> I would be happy to remove the type annotations.

Fine with me

jeromedockes:
ok opening a new PR seemed easier than fixing the conflicts so this is superseded by https://github.com/skrub-data/skrub/pull/939

jeromedockes:
@parkervg the next release of skrub will support python 3.9 (not 3.8 which is not supported by scikit-learn and will be EOL in October)

PR #875, opened 2024-01-08
Add imports in examples

TheooJ:
Some packages / modules were not imported in examples, so they couldn't be used out of the box

### DIFF
--- a/skrub/_agg_joiner.py
+++ b/skrub/_agg_joiner.py
@@ -134,6 +134,7 @@ class AggJoiner(BaseEstimator, TransformerMixin):
     Examples
     --------
     >>> import pandas as pd
+    >>> from skrub import AggJoiner
     >>> main = pd.DataFrame({
     ...     "airportId": [1, 2],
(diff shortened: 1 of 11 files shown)


jeromedockes:
the test failure is failure but not related (we should not rely on datasets we have to fetch from the internet in tests)

jeromedockes (approved):
LGTM, thanks!

jeromedockes:
@TheooJ we had to fix a few things after the latest pandas and scikit-learn releases so this pr now has a couple of conflicts.
Could you fix those and then we can merge it? thanks

jeromedockes:
maybe we can empty the doctest execution context before runnig the doctests to  make sure all the necessary imports are in the doctest

https://docs.python.org/3/library/doctest.html#what-s-the-execution-context
https://docs.pytest.org/en/7.1.x/how-to/doctest.html#doctest-namespace-fixture

glemaitre (approved):

glemaitre:
Thanks @TheooJ

Issue #810, opened 2023-10-30
FEAT Add the MultiJoiner and MultiAggJoiner


Vincent-Maladiere:
### Problem Description

As discussed in https://github.com/skrub-data/skrub/discussions/751, we aim at clarifying Joiner and AggJoiner API by only allowing a single auxiliary table to be passed as `__init__` argument. The multi-table scenario then has to be handled by a specific transformer.

### Feature Description

- Both classes check that the auxiliary input is a list of dataframe.
- They iteratively fit Joiner instances.
- By default, it should add suffixes to the tables to be joined based on their position in the auxiliary table list.

- [ ] `MultiJoiner`
- [x] `MultiAggJoiner` https://github.com/skrub-data/skrub/pull/876

### Alternative Solutions

We can reproduce the MultiJoiner behavior by creating a list of `Joiner` and putting them into a `scikit-learn Pipeline`.

rcap107:
Closed by #876

PR #876, opened 2024-01-16
[WIP, FEAT] Adding MultiJoiner and MultiAggJoiner

TheooJ:
Sharing my WIP for issue https://github.com/skrub-data/skrub/issues/810

Adding the `MultiAggJoiner` also means refactoring the `AggJoiner` to match the `Joiner`, in particular:

- Adding a `key` argument
- Supporting polars in the test suite
- Only accepting a single table as an input
- Performing multiple small checks instead of one big check

Notes: 

- For now, the check concerning duplicate column names happens *after* the aggregation in `AggJoiner`. Later we / I could move it before aggregation, I didn’t want to touch it for now since it is linked to the `_polars.aggregate` and `_pandas.aggregate` functions, let me know how you feel about this
- Should I change the landing page to show the multijoiners, or keep `Joiner` and `AggJoiner` ?

Also, thanks @jeromedockes & @Vincent-Maladiere for the discussions !


TheooJ:
I'm unsure about the default suffixes in `MultiAggJoiner`. If we add suffixes to the tables to be joined based on their position in the auxiliary table list, it will have a different behavior than`AggJoiner`. I think `AggJoiner` should have the same behavior than `MultiAggJoiner` when given a list of 1 dataframe

WDYT ?

skrub/_agg_joiner.py
```diff
@@ -49,87 +51,68 @@ def split_num_categ_operations(operations: list[str]) -> tuple[list[str], list[s
...
-        If set to None (the default), ['mean', 'mode'] will be used.
+        - If set to `None` (the default), ["mean", "mode"] will be used.
 
-    suffix : str or iterable of str, default=None
-        The suffixes that will be added to each table columns in case of
-        duplicate column names.
-        If set to None, the table index in 'aux_table' are used,
-        e.g. for a duplicate columns: price (main table),
-        price_1 (auxiliary table 1), price_2 (auxiliary table 2), etc.
+    suffix : str or iterable of str, default=""
```
jeromedockes, line 102: why iterable of str?

skrub/_agg_joiner.py
```diff
@@ -145,44 +128,166 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+        Parameters
+        ----------
+        X : DataframeLike
+            The main table to augment.
+        aux_table : DataframeLike or "X"
+            The auxiliary table.
+        """
+        if type(aux_table) == str:
+            if aux_table == "X":
+                return X, deepcopy(X)
```
jeromedockes, line 178: why do we need the copy?

skrub/_agg_joiner.py
```diff
@@ -145,44 +128,166 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+        or if there is a Polars lazyframe. Alternatively, allows `aux_table` to be "X".
+
+        Parameters
+        ----------
+        X : DataframeLike
+            The main table to augment.
+        aux_table : DataframeLike or "X"
+            The auxiliary table.
+        """
+        if type(aux_table) == str:
```
jeromedockes, line 176:
```suggestion
        if isinstance(aux_table, str):
```

skrub/_agg_joiner.py
```diff
@@ -145,44 +128,166 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+                raise ValueError("'aux_table' must be a dataframe or 'X'.")
+        # Polars lazyframes will raise an error here.
+        if not hasattr(X, "__dataframe__"):
+            raise TypeError(f"'X' must be a dataframe, got {type(X)}.")
+        if not hasattr(aux_table, "__dataframe__"):
+            raise TypeError(
+                f"'aux_table' must be a dataframe or 'X', got {type(aux_table)}."
+            )
+
+        if (is_pandas(X) and not is_pandas(aux_table)) or (
```
jeromedockes, line 189: maybe this should be shared with the Joiner and Interpolation Joiner (in a later PR)

skrub/_agg_joiner.py
```diff
@@ -145,44 +128,166 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+        -------
+        cols
+            1-dimensional array of columns on which to perform aggregation.
+        """
+        cols = atleast_1d_or_none(self.cols)
+        # If no cols provided, all columns but `aux_key` are used.
+        if len(cols) == 0:
+            cols = list(set(self._aux_table.columns) - set(self._aux_key))
+        if not all([col in self._aux_table.columns for col in cols]):
+            raise ValueError("All 'cols' must be present in 'aux_table'.")
```
jeromedockes, line 214: do you think with some rewording of the error message we could use `_join_utils.check_missing_columns`?

skrub/_agg_joiner.py
```diff
@@ -145,44 +128,166 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+
+        Returns
+        -------
+        operation
+            1-dimensional array of operations to perform on columns.
+        """
+        operation = atleast_1d_or_none(self.operation)
+        if len(operation) == 0:
+            operation = ["mean", "mode"]
+        return operation
```
jeromedockes, line 228: do we want to check the chosen operations are valid here?

skrub/_agg_joiner.py
```diff
@@ -145,44 +128,166 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+
+    def _check_suffix(self):
```
jeromedockes, line 230: I'm not sure we need this check, do you forsee this being a frequent user error?

skrub/_agg_joiner.py
```diff
@@ -145,44 +128,166 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+        if not isinstance(self.suffix, str):
+            raise ValueError(f"'suffix' must be a string. Got {self.suffix}")
+
+    def _check_column_name_duplicates_after_aggregation(
```
jeromedockes, line 234: I think we could adapt `_join_utils.check_column_name_duplicates` (maybe we need to add a parameter to indicate if the suffix has already been added or not), and use it to avoid code duplication

jeromedockes (commented):
thanks a lot @TheooJ ! here is a first small batch of comments. I think some functions in `_join_utils` could be generalized a bit to be useful here.

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,440 @@
...
+        aux_tables : iterator of DataframeLike or "X"
+            The auxiliary tables.
+
+        Raises
+        ------
+        TypeError
+            Raises an error if all the frames don't have the same type,
+            or if there is a Polars lazyframe.
+        """
+        if not _is_array_like(aux_tables):
```
jeromedockes, line 162: for this check and the other ones such as missing columns, couldn't we let the individual joiners do their checks to avoid duplicating the logic here?

skrub/_agg_joiner.py
TheooJ: Corrected, thanks

skrub/_agg_joiner.py
TheooJ: It was used in the original `AggJoiner`, but we don't need it -> removed

TheooJ: Also removed the one in `MultiAggJoiner` at the same time

skrub/_multi_agg_joiner.py
TheooJ: I agree that we could, the idea was that `MultiAggJoiner` is potentially going to be used to merge a large number of tables, possibly with many operations. I thought it was better to perform checks here directly rather than waiting the fits of the individual joiners

If you don't think it's going to be an issue I can remove them from here. It's true that this estimator becomes much more streamlined with letting some checks be performed in the individual joiners

skrub/_agg_joiner.py
TheooJ: For the len check I like it, I think it could be a frequent mistake when using many tables
For the type check I could remove it. It actually works with more than strings since we use `f"{col}{suffix}"` in `px.aggregate`

skrub/_agg_joiner.py
TheooJ: Ah yes that's right, it's done in the `AggJoiner` but not here
I think we should check it here if we decide to perform checks in the `MultiAggJoiner` depending on where we decide to perform checks

skrub/_agg_joiner.py
TheooJ: Yes I agree, when we have polars support for the `Joiner`

skrub/_agg_joiner.py
TheooJ: Yes, let me change that
It could take parameters `main_keys`, `aux_keys` that default to `None` so that it still works the same when not accounting for keys

skrub/_agg_joiner.py
TheooJ: Yes, by handling the 
```
cols = atleast_1d_or_none(self.cols)
if len(cols) == 0:
    cols = list(set(self._aux_table.columns) - set(self._aux_key))
```
outside of the function I can use the utils

TheooJ: It's true that it's the same function at one error msg diff, and I'd rather not duplicate code -> let me change it

skrub/_multi_agg_joiner.py
jeromedockes: that's a good point. still I'd like to avoid duplication as much as possible to help us maintain consistent behavior between the joiner and multi joiner wherever it makes sense.

maybe the individual joiners could separate their checks from their fit and expose a `check_inputs` method to the multi joiner? it could then ask each joiner to perform checks before actually investing time to perform the joins, but the single joiner would keep the responsibility of saying whether it's able to perform a join or not

also, do we want to allow using a column from the first aux table to join the second aux table? if yes, it may not be possible to perform all checks ahead of time. either way we should state explicitly in the documentation whether that's allowed or not

skrub/_agg_joiner.py
TheooJ: That's true, as you were saying IRL we could use something like `[{"table": df, "aux_key":..., "main_key": ...}, {"table: df2, "aux_key": ..., ...]`, we'll explore this for a future PR !

skrub/_multi_agg_joiner.py
TheooJ: It's a good point. I'm rewriting the code so that individual joiners expose a `check_inputs` method, that I use in the multi joiner before fitting

For the second point it's a good question, for now it's not possible but could be explored for a later PR. I can open an issue on this point, as it's probably a case that we'll encounter in other estimators too

TheooJ: Also, I think it makes sense to have some checks in the `MultiAggJoiner` too because the parameter names and input types are different, and this will not reflect if we mostly keep checks in the individual joiners

TheooJ:
I want to be a bit more consistent with parameter names. In the `MultiAggJoiner`, I think we want to keep all plurals i.e. `aux_tables`, `keys`, `main_keys`, `aux_keys`, `cols`, `operations` and `suffixes`.

But in the `AggJoiner`, there are currently `aux_table`, `key`, `main_key`, `aux_key`, `cols` (notice the plural), `operation` (no plural), `suffix`. 
- `aux_table` and `suffix` need to stay singular, but I think we could go either way for the others
- To be consistent with `aux_table` and `suffix` (and to contrast with the `MultiAggJoiner`), we could go singular for all the others, but I think `cols` and `operations` are going to be lists in most cases and so plural makes sense for them. We could also have a mix, and just change `operation` for `operations`

Wondering how you feel about this @jeromedockes

skrub/_agg_joiner.py
TheooJ: I was debating whether this function is actually very useful for the `AggJoiner`. The aggregations already add suffixes like `_mode` or `_mean`, and in the case of multiple tables in the `MultiAggJoiner` we also add suffixes based on the position of the auxiliary tables in the list. Furthermore, we don't check for duplicate names across tables in the `MultiAggJoiner` since we let every joiner performs the check

To adapt `_join_utils.check_column_name_duplicates`, we'd need to define cols for which it would be okay to overlap (since we could aggregate a table on itself, and thus the key would be the same in `X` and `aux_table`)

Don't know whether to keep it or not @jeromedockes

jeromedockes: as we need to handle duplicate column names in other estimators as well I'd be ok to leave this topic for a follow-up pr

jeromedockes: esp since as you say name clashes are less likely in the case of the aggjoiner than eg the Joiner

skrub/_multi_agg_joiner.py
jeromedockes: > , for now it's not possible but could be explored for a later PR

that sounds good, let's state explicitly in the docstring that it is not (yet) supported

> Also, I think it makes sense to have some checks in the MultiAggJoiner too because the parameter names and input types are different, and this will not reflect if we mostly keep checks in the individual joiners

yes, the checks that are the same for both can be done in the individual joiners or in a module-level function, and the multijoiner can add other checks that are specific to the multi- case

skrub/_agg_joiner.py
```diff
@@ -229,17 +336,17 @@ def transform(self, X):
...
+        X, _ = self._check_dataframes(X, self.aux_table_)
+        _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
+
+        skrub_px, _ = get_df_namespace(self.aux_table_)
+        X = skrub_px.join(
+            left=X,
+            right=self.aux_table_,
+            left_on=self._main_key,
+            right_on=self._aux_key,
+        )
```
jeromedockes, line 349: I can't comment on it below but let's add the function `_screen` when we actually have something to put in there, we can remove it for now

skrub/_agg_joiner.py
```diff
@@ -145,44 +127,176 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+        # If no cols provided, all columns but `aux_key` are used.
+        if len(self._cols) == 0:
+            self._cols = list(set(self._aux_table.columns) - set(self._aux_key))
+        _join_utils.check_missing_columns(self._aux_table, self._cols, "'aux_table'")
+
+        self._operation = self._check_operation()
+        self.num_operations, self.categ_operations = split_num_categ_operations(
+            self._operation
+        )
+        self._check_suffix()
```
jeromedockes, line 290: I have the impression some of those checks could benefit other estimators such as Joiner, InterpolationJoiner. in a follow-up PR we may want to move them to join_utils and use them in those other estimators

skrub/_agg_joiner.py
```diff
@@ -193,31 +307,24 @@ def fit(self, X, y=None):
...
+        skrub_px, _ = get_df_namespace(self._aux_table)
+        aux_table = skrub_px.aggregate(
+            self._aux_table,
+            self._aux_key,
+            self._cols,
+            self.num_operations,
+            self.categ_operations,
+            suffix=self.suffix,
+        )
+        self.aux_table_ = self._screen(aux_table, y)
```
jeromedockes, line 320: you can drop this line as it doesn't do anything

skrub/_agg_joiner.py
```diff
@@ -229,17 +336,17 @@ def transform(self, X):
...
-                left=X,
-                right=aux_table,
-                left_on=self.main_key_,
-                right_on=aux_key,
-            )
+        X, _ = self._check_dataframes(X, self.aux_table_)
+        _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
+
+        skrub_px, _ = get_df_namespace(self.aux_table_)
+        X = skrub_px.join(
```
jeromedockes, line 344: let's not deal with this in this PR but it seems in `join` the pandas default suffixes are used

skrub/_agg_joiner.py
```diff
@@ -49,87 +50,68 @@ def split_num_categ_operations(operations: list[str]) -> tuple[list[str], list[s
...
-        Key used to perform join on X.
-    """
-    missing_cols = set(columns) - set(X.columns)
-    if len(missing_cols) > 0:
-        raise ValueError(error_msg)
-    return
-
-
 class AggJoiner(BaseEstimator, TransformerMixin):
-    """Aggregate auxiliary dataframes before joining them on a base dataframe.
```
jeromedockes, line 74: while you're doing some cleanup could you change `BaseEstimator, TransformerMixin` to `TransformerMixin, BaseEstimator` here and in aggtarget?

jeromedockes (commented):
a few comments on the aggjoiner

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,417 @@
...
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils.validation import check_is_fitted
+
+from skrub import _join_utils
+from skrub._agg_joiner import AggJoiner
+from skrub._dataframe._namespace import is_pandas, is_polars
+from skrub._utils import _is_array_like, atleast_2d_or_none
+
+
+class MultiAggJoiner(BaseEstimator, TransformerMixin):
```
jeromedockes, line 14:
```suggestion
class MultiAggJoiner(TransformerMixin, BaseEstimator):
```

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,417 @@
...
+    """Extension of the AggJoiner to multiple auxiliary tables.
+
+    Apply numerical and categorical aggregation operations on the `cols`
+    to aggregate, selected by dtypes. See the list of supported operations
+    at the parameter `operations`.
+
+    If `cols` is not provided, `cols` are all columns from `aux_tables`,
```
jeromedockes, line 21: could you reword this a little bit to make it clear that there is one list of columns, one aux key for each aux table?

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,417 @@
...
+    except `aux_keys`.
+
+    As opposed to AggJoiner, all parameters here must be iterable of str,
```
jeromedockes, line 24: maybe we can be a bit more explicit re. how it differs from the joiner.

As opposed to the AggJoiner, here `aux_tables` is a list of tables, each of which will be joined to the main table. therefore `aux_keys` is now a list of keys, of the same length as `aux_tables`, and each entry in `aux_key` is used to join the corresponding aux table. in the same way, each entry in `cols` is a list of columns to aggregate in the corresponding aux table

Therefore if we have a single table we could either use the AggJoiner
AggJoiner(aux_table, key="ID")

or the MultiAggJoiner
AggJoiner([aux_table], keys=["ID"])

maybe in the examples we can also show how using a multiaggjoiner compares to chaining aggjoiners

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,417 @@
...
+                raise ValueError(
+                    "Must pass EITHER `keys`, OR (`main_keys` AND `aux_keys`)."
+                )
+        main_keys = atleast_2d_or_none(main_keys)
+        aux_keys = atleast_2d_or_none(aux_keys)
+        # Check 2d shape
+        if len(main_keys) != len(aux_keys):
+            raise ValueError(
+                "`main_keys` and `aux_keys` have different lengths"
+                f" ({len(main_keys)} and {len(aux_keys)}). Cannot join on different"
```
jeromedockes, line 231: maybe this could be a bit clearer if we check that each key list has the same length as aux_tables and reflect that in the error message

btw I think it's quite unlikely that using key would be possible in most cases for the multiaggjoiner, aux_key + main_key will probably always be used

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,417 @@
...
+                " numbers of tables."
+            )
+        # Check each element
+        for main_key, aux_key in zip(main_keys, aux_keys):
+            if len(main_key) != len(aux_key):
```
jeromedockes, line 236: it might be nice to indicate at which index we are in the list of keys here

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,417 @@
...
+        the join operation. Note that ``[["a"], ["b"], ["c", "d"]]`` is a valid input
+        while ``["a", "b", ["c", "d"]]`` is not.
+
+    cols : iterable of str, or iterable of iterable of str, default=None
+        Select the columns from the auxiliary dataframes to use as values during
+        the aggregation operations.
+
+        If set to `None`, `cols` are all columns from table, except `aux_keys`.
+
+    operations : iterable of str, or iterable of iterable of str, default=None
```
jeromedockes, line 64: I wonder if we could say that the same operations are applied to all tables, ie operations here has the same type as in aggjoiner.

I don't see why we would want to specify operations specific to each table but not specific to each column within a table

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,417 @@
...
+
+        # Polars lazyframes will raise an error here.
+        if not hasattr(X, "__dataframe__"):
+            raise TypeError(f"`X` must be a dataframe, got {type(X)}.")
+        if not all(hasattr(aux_table, "__dataframe__") for aux_table in aux_tables):
+            raise TypeError("`aux_tables` must all be dataframes.")
+
+        if is_pandas(X):
+            if not all(is_pandas(aux_table) for aux_table in aux_tables):
+                raise TypeError("All `aux_tables` must be Pandas dataframes.")
```
jeromedockes, line 182: maybe say in the error message that this is the case because X is pandas

skrub/_agg_joiner.py
```diff
@@ -145,44 +127,176 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+        aux_table : DataframeLike or "X"
+            The auxiliary table.
+        """
+        if isinstance(aux_table, str):
+            if aux_table == "X":
+                return X, X
+            else:
+                raise ValueError("'aux_table' must be a dataframe or 'X'.")
+        # Polars lazyframes will raise an error here.
+        if not hasattr(X, "__dataframe__"):
```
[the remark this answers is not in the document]
TheooJ: Ah yes great catch ! 

Just switched the check order, thanks !

skrub/_multi_agg_joiner.py
TheooJ: I agree, I now check that they have the same length as `aux_tables`

For the second point, I'll leave it as an option for the user even though it's unlikely -- it's still possible that all tables share the same `key`

skrub/_multi_agg_joiner.py
TheooJ: Added :)

skrub/_multi_agg_joiner.py
TheooJ: I agree that we could select columns to apply operations on, and we would need to rfc the agg joiner for this

jeromedockes: maybe keep things simple for now and do the same operation for everybody (besides the continuous vs categorical split), and then we can think if we want to allow specifying operations by column or if that use case is too specific / covered well enough by chaining several aggjoiners

skrub/_utils.py
```diff
@@ -12,9 +12,10 @@
 
 
 class LRUDict:
-    """dict with limited capacity
```
jeromedockes, line 15: it doesn't matter this time but in future PRs try not to add changes that are not directly related to the PR's stated goal as it makes review harder

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,491 @@
...
+    keys : iterable of str, or iterable of iterable of str, default=None
+        The column names to use for both `main_keys` and `aux_key` when they
+        are the same. Provide either `key` or both `main_keys` and `aux_keys`.
+        If there are multiple auxiliary tables, `keys` will be used to join all
+        of them.
+
+        If not `None`, there must be an iterable of `keys` for each table
+        in `aux_tables`.
+
+    main_keys : iterable of str, or iterable of iterable of str, default=None
```
jeromedockes, line 65: can you say that they must be in the main table, ie we cannot use columns from the first joined table to join the second

jeromedockes (approved):
LGTM, thanks a lot @TheooJ !

skrub/_utils.py
TheooJ: I will ! Sorry about this, I opened another PR for cosmetic changes but was left with a tiny overlap in files related to this PR

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,499 @@
...
+    aux_tables : iterable of DataFrameLike or "X"
+        Auxiliary dataframes to aggregate then join on the base table.
+        The placeholder string "X" can be provided to perform
+        self-aggregation on the input data. To provide a single auxiliary table,
+        ``aux_tables = [table]`` is supported, but not ``aux_tables = table``.
+        It's possible to provide both the placeholder "X" and auxiliary tables,
+        as in ``aux_tables = [table, "X"]``. If that's the case, the second table will
+        be replaced by the input data.
+
+    keys : iterable of str, or iterable of iterable of str, default=None
```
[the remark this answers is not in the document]
glemaitre: In some way, I wondering if by allowing a single iterable, a user could start to think that he can do single agg-join.

skrub/tests/test_multi_agg_joiner.py
```diff
@@ -0,0 +1,436 @@
+import pandas as pd
+import pytest
+from pandas.testing import assert_frame_equal
+
+from skrub._dataframe._polars import POLARS_SETUP
+from skrub._multi_agg_joiner import MultiAggJoiner
+
+
+@pytest.fixture
+def main():
```
glemaitre, line 10: So for this file, I will have the exact same comment than for the `test_agg_joiner`.

skrub/_agg_joiner.py
```diff
@@ -49,87 +49,68 @@ def split_num_categ_operations(operations: list[str]) -> tuple[list[str], list[s
...
     to aggregate, selected by dtypes. See the list of supported operations
-    at the parameter `operation`.
+    at the parameter `operations`.
 
-    The grouping columns used during the aggregation are the columns used
-    as keys for joining.
+    If `cols` is not provided, `cols` are all columns from `aux_table`,
+    except `aux_key`.
 
     Accepts :obj:`pandas.DataFrame` and :class:`polars.DataFrame` inputs.
```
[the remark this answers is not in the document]
jeromedockes: IIRC the issue is on polars' side, I'll try to find and link the polars issue

skrub/_multi_agg_joiner.py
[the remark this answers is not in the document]
jeromedockes: yes I would prefer having only one option. but if others want to avoid typing the [] for single-item lists I'm ok with that, too. at least now that AggJoiner and MultiAggJoiner are separate things we have 2 options not 3 :sweat_smile:

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,499 @@
...
+            if aux_keys is not None or main_keys is not None:
+                raise ValueError(
+                    "Can only pass argument `keys` OR `main_keys` and "
+                    "`aux_keys`, not a combination of both."
+                )
+            main_keys, aux_keys = keys, keys
+        else:
+            if aux_keys is None or main_keys is None:
+                raise ValueError(
+                    "Must pass EITHER `keys`, OR (`main_keys` AND `aux_keys`)."
```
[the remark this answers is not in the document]
jeromedockes: here also I think this was adapted from the Joiner so if you edit this message you can also reflect the change in the Joiner or fuzzy join

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,499 @@
...
+        X, self._aux_tables = self._check_dataframes(X, self.aux_tables)
+        self._main_keys, self._aux_keys = self._check_keys(
+            self.main_keys, self.aux_keys, self.keys, self._aux_tables
+        )
+        self._cols = self._check_cols()
+        self._operations = self._check_operations()
+        self._suffixes = self._check_suffixes()
+
+        self.agg_joiners_ = []
+        for aux_table, main_key, aux_key, cols, operations, suffix in zip(
```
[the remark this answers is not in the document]
jeromedockes: let's leave that for another PR :)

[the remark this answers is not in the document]
glemaitre: yes of course, but I was wondering if the `fit_transform` would benefit from parallel processing or the overhead will kill the performance? Do you have any intuition?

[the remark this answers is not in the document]
jeromedockes: I think it could make sense, depending on whether most time is spent in the agg or the join

GaelVaroquaux:
You need a changelog entry for this (adding an entry in doc/CHANGES.rst)

skrub/_agg_joiner.py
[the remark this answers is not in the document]
TheooJ: Ah we've spotted this too, there's an issue with polars references https://github.com/pola-rs/polars/issues/7027

skrub/_agg_joiner.py
```diff
@@ -49,87 +49,68 @@ def split_num_categ_operations(operations: list[str]) -> tuple[list[str], list[s
...
+    aux_table : DataFrameLike or "X"
         Auxiliary dataframe to aggregate then join on the base table.
         The placeholder string "X" can be provided to perform
         self-aggregation on the input data.
 
-    aux_key : str, or iterable of str, or iterable of iterable of str
-        Select the columns from the auxiliary dataframe to use as keys during
-        the join operation.
+    key : str, default=None
+        The column names to use for both `main_key` and `aux_key` when they
```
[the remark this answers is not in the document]
TheooJ: I need to change it to `str or iterable of str`, good catch !

skrub/_agg_joiner.py
```diff
@@ -145,225 +126,176 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
+        """Check `operations` input.
 
         Returns
         -------
-        X_transformed : DataFrameLike
-            The augmented input.
+        operations
+            1-dimensional array of operations to perform on columns.
         """
+        operations = atleast_1d_or_none(self.operations)
```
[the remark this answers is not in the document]
TheooJ: Yes it was inherited from what existed previously, I'll remove it in the `AggJoiner`. It can be removed later as it's just used in the `AggTarget` and in some utils that will be rewritten or replaced by dispatch

skrub/_agg_joiner.py
```diff
@@ -145,225 +126,176 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
-        for aux_table, aux_key in self.aux_table_:
-            X = skrub_px.join(
-                left=X,
-                right=aux_table,
-                left_on=self.main_key_,
-                right_on=aux_key,
-            )
+    def _check_suffix(self):
+        if not isinstance(self.suffix, str):
+            raise ValueError(f"'suffix' must be a string. Got {self.suffix}")
```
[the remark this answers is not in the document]
TheooJ: You're right, I've removed all functions except for `_check_dataframes`, and wrote the simple checks in the fit directly

skrub/_agg_joiner.py
```diff
@@ -145,225 +126,176 @@ class AggJoiner(BaseEstimator, TransformerMixin):
...
-            The input data to transform.
+    def _check_operations(self):
+        """Check `operations` input.
 
         Returns
         -------
-        X_transformed : DataFrameLike
-            The augmented input.
+        operations
+            1-dimensional array of operations to perform on columns.
```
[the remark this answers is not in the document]
TheooJ: I've removed this function and directly wrote it in `fit`

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,499 @@
...
+                main_key=main_key,
+                aux_key=aux_key,
+                cols=cols,
+                operations=operations,
+                suffix=suffix,
+            )
+            self.agg_joiners_.append(agg_joiner)
+
+        for agg_joiner in self.agg_joiners_:
+            agg_joiner._check_inputs(X)
```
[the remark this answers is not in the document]
TheooJ: It is, but I wanted to performed the checks separately in case there were a large number of operations done. My rationale was that I didn't want to perform a lot of operations without checking that all input data is correct, so it is indeed called here then in the fit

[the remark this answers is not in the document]
TheooJ: If you think this is not necessary now / will be handled another way, I can remove it :)

skrub/_multi_agg_joiner.py
```diff
@@ -0,0 +1,499 @@
...
+                    "Can only pass argument `keys` OR `main_keys` and "
+                    "`aux_keys`, not a combination of both."
+                )
+            main_keys, aux_keys = keys, keys
+        else:
+            if aux_keys is None or main_keys is None:
+                raise ValueError(
+                    "Must pass EITHER `keys`, OR (`main_keys` AND `aux_keys`)."
+                )
+        main_keys = atleast_2d_or_none(main_keys)
```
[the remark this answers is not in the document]
TheooJ: A cleanup of the utils will be useful after Jérôme's PRs

skrub/_multi_agg_joiner.py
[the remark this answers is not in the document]
TheooJ: I've adapted `_join_utils.py` & `test_interpolation_join.py` to remove the screaming 😄

skrub/_multi_agg_joiner.py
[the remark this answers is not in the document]
TheooJ: The downside is that only accepting `iterable of iterable of str` makes a clunky API for single table joins. I guess in that case we enforce that people use the right estimator i.e. the `AggJoiner`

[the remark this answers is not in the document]
TheooJ: If that's the case, we could remove the example in the docstring where we convert the `AggJoiner` example into a `MultiAggJoiner` example make make it more explicit in the docstring

skrub/tests/test_agg_joiner.py
```diff
@@ -1,302 +1,331 @@
...
 
 
 @pytest.mark.parametrize("use_X_placeholder", [False, True])
 @pytest.mark.parametrize(
-    "X, px, assert_frame_equal_",
+    "px, assert_frame_equal_",
     ASSERT_TUPLES,
 )
-def test_simple_fit_transform(use_X_placeholder, X, px, assert_frame_equal_):
+def test_simple_fit_transform(main, use_X_placeholder, px, assert_frame_equal_):
```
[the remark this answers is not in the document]
TheooJ: The ending underscore is also used in `test_datetime_encoder`, `test_fuzzy_join` and `test_joiner`, should I change them there too or do I leave that for another PR ?

skrub/_utils.py
```diff
@@ -132,17 +131,22 @@ def atleast_1d_or_none(x):
...
+        isinstance(x, Iterable)
+        and not isinstance(x, (str, bytes))
+        and not hasattr(x, "__dataframe__")
+    )
 
 
 def atleast_2d_or_none(x):
     """``np.atleast_2d`` helper returning an empty list when x is None.
 
     Note that we don't use ``np.atleast_2d`` because x could be a jagged array.
```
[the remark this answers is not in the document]
TheooJ: doesn't show as outdated but it is :)

skrub/tests/test_agg_joiner.py
```diff
@@ -1,302 +1,331 @@
...
+    main = px.DataFrame(main)
+
+    # Check too many aux_table
+    agg_joiner = AggJoiner(
+        aux_table=[main, main],
+        main_key="userId",
+        aux_key=["userId", "userId"],
+        cols=[["rating"], ["rating"]],
+    )
+    error_msg = r"(?=.*must be a dataframe or 'X', got <class 'list'>)"
```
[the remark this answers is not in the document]
TheooJ: Oh yes, good idea !

skrub/_multi_agg_joiner.py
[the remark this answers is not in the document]
jeromedockes: in any case 1 list of 10 strings would mean join 10 tables, each on 1 column right? and not join 1 table on 10 columns

but to avoid that ambiguity it might be good to force one type only , ie if I want to join 1 table on 1 column with the multiaggjoiner i need to pass `[ [ "my_column" ] ]`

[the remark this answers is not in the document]
TheooJ: The `iterable of str` case only works for joining a single table, to join multiple tables, an `iterable of iterable of str` is needed, i.e.:
- `["a", "b", "c", "d"]` would join 1 table on 4 columns
- `[["a"], ["b"], ["c"], ["d"]]` would join 4 tables on 1 column each

This was done to avoid having to pass `[ [ "my_column" ] ]`, but in this case the user might as well use the `AggJoiner` -> changing it to enforce one type and avoid any confusion

skrub/tests/test_agg_joiner.py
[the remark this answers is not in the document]
TheooJ: I'm leaving them in the other files, we'll clean them up after https://github.com/skrub-data/skrub/pull/888

jeromedockes (approved):
LGTM, thanks a lot @TheooJ !

@glemaitre , I think the main outstanding comment from your previous review was whether passing a list of strings (instead of list of list of strings) for the keys should be allowed, and that has been addressed.

glemaitre (approved):
LGTM. Thanks @TheooJ

PR #878, opened 2024-01-23
MAINT make sure pandas doesn't insert a "key_0" column in merge result

jeromedockes:
started appearing after latest pandas release


glemaitre (approved):
LGTM. Thanks @jeromedockes

PR #881, opened 2024-01-23
DOC more consistent formatting in docstring

TheooJ:
I uniformized the keys order in `InterpolationJoiner` and `Joiner` to match the Polars and Pandas API

Small number of cosmetic changes to docstrings


skrub/_interpolation_joiner.py
```diff
@@ -177,9 +177,9 @@ def __init__(
         self,
         aux_table,
         *,
+        key=None,
```
glemaitre, line 180: This is less of a trivial change here. I let @jeromedockes judging this one.

skrub/_join_utils.py
```diff
@@ -40,6 +40,7 @@ def check_key(
         How to refer to `aux_key` in error messages.
     key_name : str
```
glemaitre, line 41: Can you add the default also for `main_key_name` and `aux_key_name`

```suggestion
    key_name : str, default="key"
```

skrub/_joiner.py
```diff
@@ -220,19 +220,19 @@ def __init__(
         self,
         aux_table,
         *,
+        key=None,
```
glemaitre, line 223: Also referring to @jeromedockes

glemaitre (approved):
Only some nitpicks

glemaitre (approved):
Only some nitpicks

skrub/_interpolation_joiner.py
jeromedockes: the order proposed in this pr is the same as polars and pandas so  I think it is better

jeromedockes:
thanks a lot @TheooJ !

jeromedockes (approved):

PR #882, opened 2024-01-23
MAINT avoid tie breaking in joiner examples

glemaitre:
Locally, these tests and docstrings were flaky because the distances in the k-NN search where equals.

Find a new example without any tie.


jeromedockes:
thanks a lot @glemaitre !

PR #883, opened 2024-01-23
MAINT compatibility for scikit-learn 1.4

glemaitre:
Fixes some compatibilities issues with scikit-learn 1.4

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -21,7 +21,7 @@ from sklearn.compose import ColumnTransformer
 from sklearn.compose._column_transformer import _get_transformer_list
 from sklearn.preprocessing import OneHotEncoder
 from sklearn.utils import Bunch
-from sklearn.utils._set_output import _safe_set_output
+from sklearn.utils._set_output import _get_output_config, _safe_set_output
 from sklearn.utils.validation import check_is_fitted
 
 from skrub import DatetimeEncoder, GapEncoder, to_datetime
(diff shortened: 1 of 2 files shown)


skrub/_table_vectorizer.py
```diff
@@ -567,7 +567,12 @@ def _check_X(self, X):
                     "your data has a single feature or array.reshape(1, -1) "
                     "if it contains a single sample."
                 )
-            feature_names = getattr(self, "feature_names_in_", None)
+            feature_names = getattr(
+                self,
+                "feature_names_in_",
+                # by default, we will use "x0", "x1", ...
+                np.asarray([f"x{i}" for i in range(X_array.shape[1])], dtype=object),
```
glemaitre, line 574: Because we did not need the feature names then it was maybe fine. But we were having inconsistent names: https://github.com/scikit-learn/scikit-learn/pull/27801

glemaitre: Actually, we might have a scikit-learn bug here. I have to check if we should forward the `.set_output` to the underlying `FunctionTransformer` if this is set on the `ColumnTransformer`.

glemaitre: This is a bug.

jeromedockes:
thanks a lot @glemaitre !
there was 1 remaining "passthrough" vs FunctionTransformer issue in a docstring
the set_output one was fixed by making sure the dataframe has string column names, which is something we want to enforce in skrub anyway

glemaitre:
> the set_output one was fixed by making sure the dataframe has string column names, which is something we want to enforce in skrub anyway

OK good enough workaround for `skrub` then. I'll try to fix it upstream for when we don't have only string name but this is another story (this is good to work on different package)

PR #884, opened 2024-01-24
[MAINT] ignore warning raised by polars.to_pandas()

jeromedockes:


jeromedockes:
not sure if really needed, I was having it locally but not anymore after updating pyarrow

glemaitre:
Let's skip it for the moment. Until we don't have a way to have the dependencies kind of build in the CI, we are going to run behind those.

Since it disappear, let's close.

PR #885, opened 2024-01-25
add polars support to the joiner

jeromedockes:
probably still a bit too early; we would need to install dataframe-api-compat from git because otherwise accessing `dtype` on a dataframe api column which represents a pandas datetime column raises an exception


jeromedockes:
this was trying out the dataframe api but we went in another direction

PR #887, opened 2024-02-13
MAINT ignore deprecation warning arising from pandas after numpy 1.25

jeromedockes:


pyproject.toml
```diff
@@ -47,6 +47,9 @@ filterwarnings = [
     'ignore:.*Attribute n is deprecated.*:DeprecationWarning',
     # FIXME we need to decided what to do with pyarrow that is required by pandas >= 3.0
     'ignore:(?s).*Pyarrow will become a required dependency of pandas.*:DeprecationWarning',
+    # accessing .values on a pandas dataframe raises this warning after numpy 1.25;
```
glemaitre, line 50: I prefer to call `.to_numpy()` instead of `.values`. I don't know if the warning would still be raised.

glemaitre (approved):

Issue #870, opened 2023-12-19
Consider casting to float32 by default in TableVectorizer


GaelVaroquaux:
### Problem Description

Using float64 instead of float32 typically incurs compute and memory loads, and users do not have this in mind.

### Feature Description

We should add an option to the TableVectorizer to output float32. We should consider whether this is the default.

### Alternative Solutions

N/A

### Additional Context

N/A

jeromedockes:
that also applies (maybe even more) to encoders, for example MinHash outputs float64

GaelVaroquaux:
> that also applies (maybe even more) to encoders, for example MinHash outputs float64

Absolutely! Thanks for raising this. Maybe we should start there

PR #888, opened 2024-02-14
Add dataframe dispatch

jeromedockes:
The goal of this PR is to make it easy to support both polars and pandas in skrub.

To make a function compatible with both polars and pandas, we use the `dispatch`
decorator. The function then has an attribute `specialize` which we can use to
register implementations for polars or for pandas (or for other backends we may
add in the future).

Compared to the current approach of having a `_pandas` and a `_polars` module and a function `_get_df_namespace` which returns the module
corresponding to a dataframe, this has several advantages:

- the callers don't need to know whether a function is explicitly dispatched or
  not in order to know where to import it from and how to call it. This allows
  to change how a function is implemented (eg introduce the dataframe API)
  without changing all call sites.
- specializations may be defined in any skrub module, so all skrub modules are
  not forced to put their helpers that need to be dispatched in the `_pandas`
  and `_polars` modules. Functions can be grouped by functionality (as usual)
  rather than by backend.
- a function and its specializations can be found next to each other.

```
>>> import pandas as pd

>>> from skrub._dispatch import dispatch
>>> @dispatch
... def drop_nulls(column):
...     raise NotImplementedError()

We can now register specializations for pandas and polars

>>> @drop_nulls.specialize("pandas")
... def _drop_nulls_pandas(column):
...     return column.dropna()


>>> @drop_nulls.specialize("polars")
... def _drop_nulls_polars(column):
...     return column.drop_nulls()


>>> df = pd.DataFrame(dict(A=[0, 1, None, 3]))
>>> df
     A
0  0.0
1  1.0
2  NaN
3  3.0
>>> drop_nulls(df)
     A
0  0.0
1  1.0
3  3.0
>>> import polars as pl
>>> polars_df = pl.from_pandas(df)
>>> polars_df
shape: (4, 1)
┌──────┐
│ A    │
│ ---  │
│ f64  │
╞══════╡
│ 0.0  │
│ 1.0  │
│ null │
│ 3.0  │
└──────┘
>>> drop_nulls(polars_df)
shape: (3, 1)
┌─────┐
│ A   │
│ --- │
│ f64 │
╞═════╡
│ 0.0 │
│ 1.0 │
│ 3.0 │
└─────┘

It is also possible to define a specialization specifically for dataframes or
for series.

>>> @dispatch
... def f(obj):
...     pass

>>> @f.specialize("pandas", "DataFrame")
... def _(df):
...     print("DataFrame")

>>> @f.specialize("pandas", "Column")
... def _(df):
...     print("Column")

>>> f(pd.DataFrame())
DataFrame
>>> f(pd.Series([0]))
Column

```

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -242,26 +242,30 @@ def _make_column_like_polars(column, values, name):
 
 
 @dispatch
-def all_null_like(column, length=None, dtype=None):
+def all_null_like(column, length=None, dtype=None, name=None):
     raise NotImplementedError()
 
 
 @all_null_like.specialize("pandas")
-def _all_null_like_pandas(column, length=None, dtype=None):
+def _all_null_like_pandas(column, length=None, dtype=None, name=None):
     if length is None:
         length = len(column)
     if dtype is None:
         dtype = column.dtype
-    return pd.Series(dtype=dtype, index=column.index)
+    if name is None:
+        name = column.name
+    return pd.Series(dtype=dtype, index=column.index, name=name)
 
 
 @all_null_like.specialize("polars")
-def _all_null_like_polars(column, length=None, dtype=None):
+def _all_null_like_polars(column, length=None, dtype=None, name=None):
     if length is None:
         length = len(column)
     if dtype is None:
         dtype = column.dtype
-    return pl.Series([None] * length, dtype=dtype)
+    if name is None:
+        name = column.name
+    return pl.Series(name, [None] * length, dtype=dtype)
 
 
 @dispatch
@@ -669,7 +673,7 @@ def is_null(column):
 
 @is_null.specialize("pandas")
 def _is_null_pandas(column):
-    return column.isna()
+    return rename(column.isna(), name(column))
 
 
 @is_null.specialize("polars")
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -560,8 +560,6 @@ def is_object(column):
 
 @is_object.specialize("pandas")
 def _is_object_pandas(column):
-    if pandas.api.types.is_string_dtype(column):
-        return False
     return pandas.api.types.is_object_dtype(column)
 
 
@@ -605,7 +603,10 @@ def _to_datetime_pandas(column, format, strict=True):
 def _to_datetime_polars(column, format, strict=True):
     if _is_anydate_polars(column):
         return column
-    return column.str.to_datetime(format=format, strict=strict)
+    try:
+        return column.str.to_datetime(format=format, strict=strict)
+    except pl.ComputeError as e:
+        raise ValueError("Failed to convert to datetime") from e
 
 
 @dispatch
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -1,8 +1,6 @@
-try:
-    import pandas as pd
-    import pandas.api.types
-except ImportError:
-    pass
+import pandas as pd
+import pandas.api.types
+
 try:
     import polars as pl
 except ImportError:
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -464,15 +464,15 @@ def is_numeric(column):
 
 @is_numeric.specialize("pandas")
 def _is_numeric_pandas(column):
-    # polars and pandas disagree about whether Booleans are numbers
-    return pandas.api.types.is_numeric_dtype(
+    return pandas.api.types.is_numeric_dtype(column) or pandas.api.types.is_bool_dtype(
         column
-    ) and not pandas.api.types.is_bool_dtype(column)
+    )
 
 
 @is_numeric.specialize("polars")
 def _is_numeric_polars(column):
-    return column.dtype.is_numeric()
+    # polars and pandas disagree about whether Booleans are numbers
+    return column.dtype.is_numeric() or column.dtype == pl.Boolean
 
 
 @dispatch
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -1,0 +1,737 @@
+try:
+    import pandas as pd
+    import pandas.api.types
+except ImportError:
+    pass
+try:
+    import polars as pl
+except ImportError:
+    pass
+
+from .._dispatch import dispatch
+
+__all__ = [
+    "skrub_namespace",
+    "dataframe_module_name",
+    "is_pandas",
+    "to_pandas",
+    "is_polars",
+    "shape",
+    "is_dataframe",
+    "is_lazyframe",
+    "is_column",
+    "col",
+    "to_array",
+    "pandas_convert_dtypes",
+    "column_names",
+    "name",
+    "rename",
+    "column_like",
+    "all_null_like",
+    "dataframe_from_columns",
+    "dataframe_from_dict",
+    "dataframe_like",
+    "concat_horizontal",
+    "to_column_list",
+    "is_in",
+    "is_null",
+    "drop_nulls",
+    "n_unique",
+    "is_bool",
+    "is_numeric",
+    "to_numeric",
+    "to_float32",
+    "is_string",
+    "is_object",
+    "is_anydate",
+    "set_column_names",
+    "collect",
+    "is_categorical",
+    "to_categorical",
+    "to_datetime",
+    "unique",
+    "dtype",
+    "cast",
+    "where",
+    "sample",
+    "replace",
+    "replace_regex",
+]
+
+
+# TODO: skrub_namespace is temporary; all code in those modules should be moved
+# here or in the corresponding skrub modules and use the dispatch mechanism.
+@dispatch
+def skrub_namespace(obj):
+    raise NotImplementedError()
+
+
+@skrub_namespace.specialize("pandas")
+def _skrub_namespace_pandas(obj):
+    from . import _pandas
+
+    return _pandas
+
+
+@skrub_namespace.specialize("polars")
+def _skrub_namespace_polars(obj):
+    from . import _polars
+
+    return _polars
+
+
+@dispatch
+def dataframe_module_name(obj):
+    return None
+
+
+@dataframe_module_name.specialize("pandas")
+def _dataframe_module_name_pandas(obj):
+    return "pandas"
+
+
+@dataframe_module_name.specialize("polars")
+def _dataframe_module_name_polars(obj):
+    return "polars"
+
+
+def is_pandas(obj):
+    return dataframe_module_name(obj) == "pandas"
+
+
+@dispatch
+def to_pandas(obj):
+    raise NotImplementedError()
+
+
+@to_pandas.specialize("pandas")
+def _to_pandas_pandas(obj):
+    return obj
+
+
+@to_pandas.specialize("polars")
+def _to_pandas_polars(obj):
+    return obj.to_pandas()
+
+
+def is_polars(obj):
+    return dataframe_module_name(obj) == "polars"
+
+
+@dispatch
+def is_dataframe(obj):
+    return False
+
+
+@is_dataframe.specialize("pandas", "DataFrame")
+def _is_dataframe_pandas(obj):
+    return True
+
+
+@is_dataframe.specialize("polars", ["DataFrame", "LazyFrame"])
+def _is_dataframe_polars(obj):
+    return True
+
+
+@dispatch
+def is_column(obj):
+    return False
+
+
+@is_column.specialize("pandas", "Column")
+def _is_column_pandas(obj):
+    return True
+
+
+@is_column.specialize("polars", "Column")
+def _is_column_polars(obj):
+    return True
+
+
+@dispatch
+def to_array(obj):
+    raise NotImplementedError()
+
+
+@to_array.specialize("pandas")
+def _to_array_pandas(obj):
+    return obj.to_numpy()
+
+
+@to_array.specialize("polars")
+def _to_array_polars(obj):
+    return obj.to_numpy()
+
+
+@dispatch
+def pandas_convert_dtypes(obj):
+    return obj
+
+
+@pandas_convert_dtypes.specialize("pandas")
+def _pandas_convert_dtypes_pandas(obj):
+    return obj.convert_dtypes()
+
+
+@dispatch
+def col(df, col_name):
+    raise NotImplementedError()
+
+
+@col.specialize("pandas")
+def _col_pandas(df, col_name):
+    return df[col_name]
+
+
+@col.specialize("polars")
+def _col_polars(df, col_name):
+    return df[col_name]
+
+
+@dispatch
+def column_names(df):
+    raise NotImplementedError()
+
+
+@column_names.specialize("pandas")
+def _column_names_pandas(df):
+    return list(df.columns.values)
+
+
+@column_names.specialize("polars")
+def _column_names_polars(df):
+    return df.columns
+
+
+@dispatch
+def name(col):
+    raise NotImplementedError()
+
+
+@name.specialize("pandas")
+def _name_pandas(col):
+    return col.name
+
+
+@name.specialize("polars")
+def _name_polars(col):
+    return col.name
+
+
+@dispatch
+def rename(col, new_name):
+    raise NotImplementedError()
+
+
+@rename.specialize("pandas")
+def _rename_pandas(col, new_name):
+    return col.rename(new_name)
+
+
+@rename.specialize("polars")
+def _rename_polars(col, new_name):
+    return col.rename(new_name)
+
+
+@dispatch
+def column_like(column, values, name):
+    return NotImplementedError()
+
+
+@column_like.specialize("pandas")
+def _column_like_pandas(column, values, name):
+    return pd.Series(values, name=name)
+
+
+@column_like.specialize("polars")
+def _column_like_polars(column, values, name):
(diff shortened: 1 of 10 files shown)


MarcoGorelli:
hey - so, in https://github.com/skrub-data/skrub/pull/786#issuecomment-1944232537 I was asked for my thoughts on this

first reaction to the dispatch mechanism - well done! this looks broadly useful, beyond skrub, perhaps it could be its own separate package? a lot of libraries are now trying to support both pandas and polars and so a community of people interested in being able to reuse code may well arise?

jeromedockes:
> first reaction to the dispatch mechanism - well done! this looks broadly useful, beyond skrub, perhaps it could be its own separate package? a lot of libraries are now trying to support both pandas and polars and so a community of people interested in being able to reuse code may well arise?

Thanks a lot for having a look! Your suggestion makes a lot of sense, but then I wonder if the scope would overlap with the dataframe-api-compat package... I guess we could start using the dispatch internally in skrub and see how it plays out in practice. And then, if we find it convenient to work with, consider moving it out?

MarcoGorelli:
I'm tempted to repurpose what I'd worked on as a `polars-api-compat`, so you can write pandas/polars/cudf/modin-agnostic code with a subset of the Polars API. Maybe this would be useful to you? If we don't make any claims about trying to be make a Standard or anything like that, but _just_ compatibility with the Polars API, then I don't think we wouldn't step on each others feet

So then, you could write something like
```python
def my_agnostic_function(df):
    dfx, plx = polars_api_compat.convert(df, api_version="0.20")

    # use some stable subset of Polars API, with `plx` as namespace
    result = (
        dfx.filter(plx.col("l_shipdate") <= var_1)
        .group_by("l_returnflag", "l_linestatus")
        .agg(
            plx.sum("l_quantity").alias("sum_qty"),
            plx.sum("l_extendedprice").alias("sum_base_price"),
            (plx.col("l_extendedprice") * (1 - plx.col("l_discount")))
            .sum()
            .alias("sum_disc_price"),
            (
                plx.col("l_extendedprice")
                * (1.0 - plx.col("l_discount"))
                * (1.0 + plx.col("l_tax"))
            )
            .sum()
            .alias("sum_charge"),
            plx.mean("l_quantity").alias("avg_qty"),
            plx.mean("l_extendedprice").alias("avg_price"),
            plx.mean("l_discount").alias("avg_disc"),
            plx.len().alias("count_order"),
        )
        .sort("l_returnflag", "l_linestatus")
    ).collect()

    # return result in original dataframe class
    return result.dataframe


my_agnostic_function(pandas.read_parquet("lineitem.parquet"))  # works, returns pandas dataframe
my_agnostic_function(polars.scan_parquet("lineitem.parquet"))  # works, returns polars dataframe
```

>  I guess we could start using the dispatch internally in skrub and see how it plays out in practice. And then, if we find it convenient to work with, consider moving it out?

agree - start with getting something working, add tests, and then consider generalising 👍

jeromedockes:
I think that would be very useful, yes. Indeed a compatibility layer for a couple or a few packages might be achievable faster than a more general standard

machow:
Hey! @MarcoGorelli pointed me here. I've been working on solutions this problem, and documented in a tool called [databackend](https://github.com/machow/databackend). It supports python singledispatch, without requiring the underlying library being dispatched on.

I've been working on how to get good type hints, and think there's a pretty good solution (see [this comment in this plum issue](https://github.com/beartype/plum/issues/89#issuecomment-1659458725)).

Happy to work more on this, but wanted to drop what I've got on this type of problem!

jeromedockes:
Thanks for sharing this, @machow! I guess if we look at it from a distance we landed on somewhat similar solutions, with a decorator for defining generic functions and registering single dispatch implementations without needing to import the actual types, + a private module of generic helper functions defined with this dispatch mechanism. The `ABC` approach you use in `databackend` has the advantage of allowing explicit `isinstance` checks as an alternative to calling generic functions, and allows registering implementations by providing type hints rather than an argument to the decorator.

jeromedockes:
In the short term for skrub, I suggest we move forward with the `dispatch` decorator and the few helper functions introduced in this PR and see how far that takes us. As this is an internal implementation detail, it should be easy to swap (or complement) this approach with a more complete external library of backend-agnostic dataframe functions in the future, whether it relies on generic functions / dispatch or on the dataframe_api_compat / polars_api_compat way of getting to the concrete function definitions

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+attribute that can be used to register implementations.
+
+>>> from skrub._dispatch import dispatch
+
+>>> @dispatch
+... def my_function(df, other_arg=1.0):
+...     # here we can either provide a default implementation or:
+...     raise NotImplementedError()
+
+my_function now has a ``specialize`` attribute we can use to register
```
glemaitre, line 30:
```suggestion
``my_function`` now has a ``specialize`` attribute we can use to register
```

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+implementations:
+
+>>> @my_function.specialize("pandas")
+... def _my_function_pandas(df, other_arg=1.0):
+...     print("calling pandas implementation")
+
+The name ``_my_function_pandas`` is arbitrary, we could use anything such as
+``_`` (unless we want to be able to call the specialized implementation
+directly, eg for testing).
```
glemaitre, line 39:
```suggestion
directly, e.g. for testing).
```

MarcoGorelli:
Totally agree! If you start with this and get it working, then it's going to be a lot easier for me to take a step back and ask "what's skrub doing, and what can we abstract into an easily vendorable reusable solution?"

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+...     import skrub._dataframe as ns
+...     ns.shape(df)
+
+
+We can inspect which implementations were registered in ``my_function.registry``.
+
+It is also possible to register a specialization specifically for dataframes or
+for columns:
+
+>>> @my_function.specialize("pandas", "Column")
```
glemaitre, line 96: At a first glance, I would prefer to enforce keyword, if you further specialization:

```python
@my_function.specialize(lib="pandas", container="dataframes")
```

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+>>> my_function(pd.DataFrame())
+Any pandas (Series or DataFrame)
+
+Importantly, single dispatch is done based on the type of the
+**first argument**. This means our generic function should be written in such a way
+that the first argument is the dataframe or column that will determine which
+implementation to use. For example the following would not work:
+
+>>> @dispatch
+... def sample(n_rows, df):
```
glemaitre, line 125: I assume that we could raise if the first argument does not live in the namespace of the library or something like that.

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+        return {
+            "module": polars,
+            "types": {
+                "DataFrame": [polars.DataFrame, polars.LazyFrame],
+                "LazyFrame": [polars.LazyFrame],
+                "EagerFrame": [polars.DataFrame],
+                "Column": [polars.Series],
+            },
+        }
+    raise KeyError(f"Unknown module: {name}")
```
glemaitre, line 162: We should probably raise an error with the available backend.

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+... def sample(df, n_rows):
+...     ...
+"""
+
+from functools import singledispatch
+
+
+def _load_dataframe_module_info(name):
+    # if the module is not installed, import errors are propagated
+    if name == "pandas":
```
glemaitre, line 140: I'm wondering if this is an overkill for now, but I would register the supported version into a `dict`-like class. This might be one step closer to register another backend a just write the specialize version.

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+
+def dispatch(function):
+    dispatched = singledispatch(function)
+
+    def specialize(module_name, generic_type_names=None):
+        try:
+            module_info = _load_dataframe_module_info(module_name)
+        except (ImportError, KeyError):
+
+            def decorator(specialized_impl):
```
glemaitre, line 173: I think this worth a comment here. If I'm not wrong this is falling back the generic implementation but we call this `specialized_impl` even in this case.

skrub/_dispatch.py
glemaitre: (Maybe going towards over-engineering): We could something similar to scikit-learn with the `set_output` (https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_set_output.py) by creating a manager: https://github.com/scikit-learn/scikit-learn/blob/main/sklearn/utils/_set_output.py#L183-L197

How it looks like, we return a dictionary for the moment, so it means that we don't really need a protocol, but we could use a dataclass instead.

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+
+def _load_dataframe_module_info(name):
+    # if the module is not installed, import errors are propagated
+    if name == "pandas":
+        import pandas
+
+        return {
+            "module": pandas,
+            "types": {
+                "DataFrame": [pandas.DataFrame],
```
glemaitre, line 146: I would tend to have tuples instead of something mutable

skrub/_dispatch.py
```diff
@@ -0,0 +1,194 @@
...
+
+            def decorator(specialized_impl):
+                return specialized_impl
+
+            return decorator
+
+        if generic_type_names is None:
+            generic_type_names = list(module_info["types"].keys())
+        if isinstance(generic_type_names, str):
+            generic_type_names = [generic_type_names]
```
glemaitre, line 181: a tuple as well

skrub/tests/_test_dispatch.py
```diff
@@ -0,0 +1,40 @@
+import pytest
```
glemaitre, line 1: is there a reason to start the name of the file by `_`?

glemaitre: Is the file going to be discovered automatically by `pytest`?

glemaitre (commented):
A couple of thoughts regarding the dispatcher.

GaelVaroquaux:
> Totally agree! If you start with this and get it working, then it's going to be a lot easier for me to take a step back and ask "what's skrub doing, and what can we abstract into an easily vendorable reusable solution?"

That's exactly what I hope will happen: we demonstrate something that works for us on a couple of dataframe models, and we call the community to see how it can abstract away and bring in new dataframe models.

By the way, London is not very far from Paris, if/when we organize a sprint, we'll ping you :)

skrub/_dataframe/tests/test_common.py
```diff
@@ -0,0 +1,147 @@
...
+
+from skrub._dataframe import _common as ns
+
+#
+# Inspecting containers' type and module
+# ======================================
+#
+
+
+def test_skrub_namespace(df_module):
```
glemaitre, line 13: I think that it would be worth to provide the location in a module docstring of this feature.

skrub/_dispatch.py
jeromedockes: the only function that access this dict is the `dispatch` function itself so it's relatively easy to make sure it doesn't modify the dict

skrub/_dispatch.py
jeromedockes: Here `specialized_impl` is the specialized implementation , eg `_drop_nulls_polars`. If polars isn't installed, we just return the function without registering it: we have no way of registering it and it would never get called anyway. so `_drop_nulls_polars` will exist in the module, but it will not appear in the generic function's registry, and it will never be called. If polars _is_ installed, we register it first but we also return the unmodified function afterwards

skrub/tests/_test_dispatch.py
jeromedockes: no reason, that's a typo :)

skrub/_dispatch.py
jeromedockes: also, the first one (the lib) is not optional but required


btw don't hesitate to suggest a different syntax if you have ideas eg we could have something like

```
@my_function.polars
def _(df):
   ...

@my_other_function.polars("DataFrame")
def _(df):
    ...

# or

@my_other_function.polars_dataframe
def _(df):
    ...
```

to type even less

glemaitre: > What would you think of making the first argument kwarg or positional and the second one kwarg-only?

Actually it was exactly what I thought.

skrub/_dispatch.py
glemaitre: > do you mean to allow dynamically registering backends?

Yes it was what I meant. But as I said, probably some over-engineering for the moment. Maybe using a `dataclass` instead of dict could be better to express what we those should return.

skrub/_dispatch.py
glemaitre: > dict that contains it is mutable too

It is where I was thinking of a dataclass indeed.

glemaitre: > the only function that access this dict is the dispatch function itself so it's relatively easy to make sure it doesn't modify the dict

Yep I'm not to worry about being modified. But having an immutable show to the reader that this is not supposed to be modified.

skrub/_dataframe/tests/test_common.py
glemaitre: > you mean state in test_common.__doc__ that df_module is a pytest fixture defined in conftest.py?

Just on the top of the file, adding a multiline docstring with the info that you mentioned. "module docstring" was a bit vague.

skrub/_dispatch.py
glemaitre: yes, it was what I had in mind.

skrub/_dataframe/_common.py
```diff
@@ -0,0 +1,780 @@
...
+
+
+@dispatch
+def is_numeric(column):
+    raise NotImplementedError()
+
+
+@is_numeric.specialize("pandas")
+def _is_numeric_pandas(column):
+    # polars and pandas disagree about whether Booleans are numbers
```
jeromedockes, line 467: here which should we pick for `is_numeric(Boolean)` ? pandas says yes, polars says no.
for example in the TableVectorizer do we want to apply the `numeric_transformer` to boolean columns?

for example, in a situation where the numeric transformer is simple cast to float32 and then we pass the output of the TableVectorizer to a scikit-learn estimators I think it would make sense to do the same thing for boolean columns

GaelVaroquaux:
> here which should we pick for is_numeric(Boolean) ? pandas says yes, polars says no.

Good catch!

> for example in the TableVectorizer do we want to apply the numeric_transformer to boolean columns?

I think for us boolean can be considered as numerical.

GaelVaroquaux:
> Would it not be more judicious to consider them as categorical data and encode with 1/0. If is_numeric means that we are going to apply the numerical transformer, I can foresee some use case that we saw in scikit-learn where you scale is really tiny and the division lead to numerical instability for the later solver of the ML model.

OK, point taken. I agree

jeromedockes:
> Would it not be more judicious to consider them as categorical data

so does this mean `is_categorical` should return `True` for bool columns?

and in the tablevectorizer we could always special-case bool columns and just cast them to float or something like that

GaelVaroquaux:
> and in the tablevectorizer we could always special-case bool columns and just cast them to float or something like that

I think that this would be nice

jeromedockes:
as this is an internal addition (and is not used yet to add polars support in any of the skrub estimators) I don't think it needs a whatsnew entry

apart from that @glemaitre you can have another look

glemaitre:
> should to_numeric output float32 by default?

You would tend to say no but there is maybe a consideration to have here. Are pandas and polars behaving the same?

jeromedockes:
> You would tend to say no but there is maybe a consideration to have here. Are pandas and polars behaving the same?

it depends on the input; when converting strings they output int64 or float64

glemaitre:
I just found this issue: https://github.com/skrub-data/skrub/issues/870

If we want to go this way then, we can make `np.float32` the primary data type.

glemaitre (approved):
LGTM.

glemaitre:
@GaelVaroquaux do you want to have a final look or we are good to merge as-is?

GaelVaroquaux:
> Merged [1]#888 into main.

Whoohoo! Congratulations!

MarcoGorelli:
Hey all - just wanted to raise that I'll likely be getting 3 months' funding to work on [Narwhals](https://github.com/MarcoGorelli/narwhals) with 2 interns 🥳 

If this would be useful to you and there's anything you'd like to see in it, please do give me a shout - anything that'd be useful to open source projects is considered in-scope (so long as it's already in Polars and easy-enough to do in pandas). The current API reference is here: https://marcogorelli.github.io/narwhals/api-reference/. I think it covers most of what you have here?

I hope I'm not coming across as trying to "force" Narhwals onto you, I'm just raising this issue in case it may help you reduce your cross-dataframe maintenance and focus on skrub's main mission

---

> By the way, London is not very far from Paris, if/when we organize a sprint, we'll ping you :)

I hope to be able to make it to PyData Paris, hopefully we can meet there!

jeromedockes:
Thanks for letting us know @MarcoGorelli ! that's great news. Indeed it seems to cover most of the subset of polars API we might need and we should definitely consider relying on it if/when the small set of private skrub functions we've added is not enough, maintaining it becomes a burden, using module-level functions becomes annoying and we want methods of generic dataframe objects as offered by narwhals, or we want to use something that exactly matches the polars API

PR #889, opened 2024-02-17
Add fairness paper

cmougan:
Hi,

I propose to add the following. (self-bias) about fairness implication of encoding prott. att.

This other one is a bit off topic https://link.springer.com/chapter/10.1007/978-3-030-85529-1_14


GaelVaroquaux:
I'm sorry, this reference is not about what's implemented in skrub.

PR #892, opened 2024-03-07
MNT import parse_version from sklearn.utils.fixes

jeromedockes:
in the scikit-learn main branch `parse_version` is not in `utils` anymore but can still be imported from `utils.fixes`


GaelVaroquaux (approved):
LGTM. Thanks!

PR #893, opened 2024-03-13
parametrize with keys so we get more informative test names

jeromedockes:
when using pytest -v with this change the tests that use this fixture will look like

```
skrub/_dataframe/tests/test_common.py::test_unique[pandas] PASSED                                                                                                                                  
skrub/_dataframe/tests/test_common.py::test_unique[polars] PASSED 
```

instead of 

```
skrub/_dataframe/tests/test_common.py::test_unique[df_module0] PASSED                                                                                                                                  
skrub/_dataframe/tests/test_common.py::test_unique[df_module1] PASSED 
```

### DIFF
--- a/skrub/conftest.py
+++ b/skrub/conftest.py
@@ -98,7 +98,7 @@ def all_dataframe_modules():
     return _DATAFAME_MODULES_INFO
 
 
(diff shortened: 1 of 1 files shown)


jeromedockes:
@TheooJ

PR #895, opened 2024-03-20
Add selectors

jeromedockes:
This adds the (private) `skrub._selectors` module.
It is kind of a weird PR because we decided to keep it private for now, and it is not yet used anywhere in skrub :sweat_smile: 

However it is a step towards column-wise transformations (#877), and in the longer term towards the skrub Pipe/Recipe/..., so selectors _will_ have a reason to exist at some point.

Moreover, I think in the future we will probably want to expose them to advanced users and accept them in skrub interfaces that currently accept a list of column names such as SelectCols (in addition to the list of columns, not instead).

The selectors achieve 2 things:

1: Delayed selection: passing a selection rule that describes what columns should be selected, when the dataframe on which the rule is evaluated and expanded into a concrete list of column names is not available yet. For example, instantiating a `SelectCols` that selects "all numeric columns except the column named 'ID'", even though the actual dataframe (and concrete list of column names) is only known at `fit` time, not instantiation time

A consequence is 1b: have a good value to represent "all columns" as a default argument for interfaces that expect a column list (`None` to mean "all" is not great, `_selectors.all()` is better).

2: Easily expressing complex selection rules because (i) the `_selectors` module offers a range of useful pre-defined rules such as `numeric()`, `cardinality_below(30)` or `glob('User_*')`, and (ii) because selectors can be combined with the operators `~`, `|`, `&`, `-`, `^`.

The docstring of `skrub/_selectors/__init__.py` provides more details but here is a summary:

- Selectors have a `.expand(dataframe)` method that returns the list of column names that should be selected.
- `make_selector` turns a column name, list of column names, or selector into a selector.
- `select` returns the part of a dataframe matched by a selector (ie indexes the dataframe with the result of `expand`).

```python
>>> from skrub import _selectors as s

>>> import pandas as pd
>>> df = pd.DataFrame(
...     {
...         "height_mm": [297.0, 420.0],
...         "width_mm": [210.0, 297.0],
...         "kind": ["A4", "A3"],
...         "ID": [4, 3],
...     }
... )

>>> s.all().expand(df)
['height_mm', 'width_mm', 'kind', 'ID']
>>> (s.all() - ['ID']).expand(df)
['height_mm', 'width_mm', 'kind']
>>> s.numeric().expand(df)
['height_mm', 'width_mm', 'ID']
>>> (s.glob('*_mm') | s.string()).expand(df)
['height_mm', 'width_mm', 'kind']
>>> s.cols('kind', 'ID').expand(df)
['kind', 'ID']
>>> s.make_selector(['kind', 'ID'])
cols('kind', 'ID')
>>> s.select(df, s.glob('*_mm'))
   height_mm  width_mm
0      297.0     210.0
1      420.0     297.0
```


jeromedockes:
I'm not sure why codecov reports some lines as missing. They are covered by the tests, for example the reported line [here](https://app.codecov.io/gh/skrub-data/skrub/pull/895?dropdown=coverage&src=pr&el=h1&utm_medium=referral&utm_source=github&utm_content=checks&utm_campaign=pr+comments&utm_term=skrub-data#a3ff24c982381190f626e24d4ac5dd9f-R24) is covered by [this test](https://github.com/skrub-data/skrub/pull/895/files#diff-20cf76c1b97e5b6702f1cd1d3a7b17d24f3094044019be4d64b46ea1a9848f0aR29); when running the tests locally I see 100% coverage for the `_selectors` module

skrub/_selectors/_selectors.py
```diff
@@ -0,0 +1,494 @@
...
+    str_      object
+    dtype: object
+
+    >>> s.select(df, s.integer())
+       i64  I64  i8
+    0    2    2   3
+
+    Use s.boolean() to also select Boolean columns:
+
+    >>> s.select(df, s.numeric() | s.boolean())
```
TheooJ, line 236:
```suggestion
    >>> s.select(df, s.integer() | s.boolean())
```

skrub/_selectors/_selectors.py
```diff
@@ -0,0 +1,494 @@
...
+    1        B
+
+    """
+    return Filter(sbd.is_categorical, name="categorical")
+
+
+def string():
+    """
+    Select columns that have a String data type.
+    In pandas, object columns containing (only) strings are also selected.
```
TheooJ, line 364:
```suggestion

    In pandas, object columns containing (only) strings are also selected.
```

skrub/_selectors/_selectors.py
```diff
@@ -0,0 +1,494 @@
...
+#
+# Selectors based on column values, computed statistics
+#
+
+
+def _cardinality_below(column, threshold):
+    try:
+        return sbd.n_unique(column) < threshold
+    except Exception:
+        # n_unique can fail for example for polars columns with dtype Object
```
TheooJ, line 449:
```suggestion
        # `n_unique` can fail for example for polars columns with dtype Object
```

skrub/_selectors/tests/test_selectors.py
```diff
@@ -0,0 +1,113 @@
...
+    assert (s.numeric() | s.boolean()).expand(df) == [
+        "int-col",
+        "float-col",
+        "bool-col",
+    ]
+    assert s.integer().expand(df) == ["int-col"]
+    assert s.float().expand(df) == ["float-col"]
+    assert s.string().expand(df) == ["str-col"]
+    assert s.categorical().expand(df) == ["cat-col"]
+
```
TheooJ, line 60:
```suggestion
assert s.any_date().expand(df) == ["datetime-col", "date-col"]

```

TheooJ (changes requested):
Thanks @jeromedockes !
I mostly have nitpicks about markdown

skrub/_selectors/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+-------------------------------------------
+
+filter and filter_names allow selecting columns based on arbitrary user-defined
+criteria. These are also used to implement many of the other selectors provided
+in this module.
+
+filter accepts a predicate that is called with a column (pandas or polars
+Series) and returns True if it should be selected.
+
+>>> s.filter(lambda col: "A4" in col.tolist()).expand(df)
```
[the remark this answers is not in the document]
jeromedockes: here it is an example and the dataframe is explicitly a pandas dataframe so I would keep things simple and use the pandas method

skrub/_selectors/_base.py
```diff
@@ -0,0 +1,505 @@
...
+    ...         "kind": ["A4", "A3"],
+    ...         "ID": [4, 3],
+    ...     }
+    ... )
+    >>> df
+       height_mm  width_mm kind  ID
+    0      297.0     210.0   A4   4
+    1      420.0     297.0   A3   3
+
+    >>> selector = s.filter(lambda col: 'A4' in col.values)
```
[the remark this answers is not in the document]
jeromedockes: here also, it is an example with a pandas dataframe so I think it's fine to use pandas code

skrub/_selectors/tests/test_selectors.py
jeromedockes: pandas doesn't have a notion of "date", only datetime AFAICT

skrub/_selectors/__init__.py
```diff
@@ -0,0 +1,269 @@
+"""
+Helpers for selecting columns in a dataframe.
+=============================================
+
+A selector represents a column selection rule, such as "all columns that have
+numerical data types, except the column 'User ID'. When applied to a dataframe,
+it expands into a list of column names that match the selection rule.
+This is implemented by the `Selector` class and its `expand` method.
```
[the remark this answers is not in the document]
jeromedockes: I applied all the other suggestions but I'm not sure about adding :class: -- this private module docstring doesn't appear in the documentation, so the source will be read not the rendered html. I think backticks still help separate symbols from the main text, but the sphinx `:class:` make the source a bit harder to read IMO, with little benefit if it doesn't actually get transformed to html

jeromedockes (commented):
thanks a lot for the thorough review @TheooJ !

skrub/_dataframe/_common.py
```diff
@@ -509,6 +543,36 @@ def _to_numeric_polars(column, dtype=None, strict=True):
     raise ValueError("Could not convert column to numeric dtype") from error
 
 
+@dispatch
+def is_integer(column):
+    raise NotImplementedError()
+
+
+@is_integer.specialize("pandas")
+def _is_integer_pandas(column):
```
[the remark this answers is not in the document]
jeromedockes: I 100% agree, however looking at the module there seems to be an almost balanced mix of `col` and `columns` (my bad). so let's harmonize it in another PR as this one isn't really related to the `dataframe` module so doing it here would obfuscate the diff with a ton of unrelated changes

skrub/_selectors/tests/test_selectors.py
jeromedockes: I updated the test to include `any_date`, distinguishing the pandas & polars cases

TheooJ (approved):
LGTM :) thanks

PR #896, opened 2024-03-25
Use dispatch and module fixture in tests

TheooJ:
The goal of this PR is to uniformize how we write tests: 
- to take advantage of `dispatch` for comparing dataframes with a custom implementation of `assert_frame_equal`,
- to avoid the boilerplate 
    ```
    import pandas as pd

    MODULES = [pd]
    ASSERT_TUPLES = [(pd, assert_frame_equal)]
    
    if POLARS_SETUP:
        import polars as pl
        from polars.testing import assert_frame_equal as assert_frame_equal_pl
    
        MODULES.append(pl)
        ASSERT_TUPLES.append((pl, assert_frame_equal_pl))
    ```
    in each test module, by using the `df_module` fixture in `conftest` instead.

I propose to use `df_module` to parametrize tests and create the dataframes (when we wouldn't have access to a dataframe to feed as a first parameter of a dispatched function), and to use functions in `_dataframe._common` and `_test_utils` for all other backend-specific functions.

Here's an example of this implementation on `test_joiner`, let me know what you think. I can apply the changes to all other test modules once we settle on the method.

cc @jeromedockes

### DIFF
--- a/skrub/_dataframe/_test_utils.py
+++ b/skrub/_dataframe/_test_utils.py
@@ -1,6 +1,33 @@
+import pandas as pd
+import pandas.testing
+
+try:
+    import polars as pl
+    import polars.testing
+except ImportError:
+    pass
+
+from .._dispatch import dispatch
+
+
 def is_module_pandas(px):
     return px.__name__ == "pandas"
 
 
 def is_module_polars(px):
     return px.__name__ == "polars"
+
+
+@dispatch
+def assert_frame_equal(left, right):
+    raise NotImplementedError()
+
+
+@assert_frame_equal.specialize("pandas")
+def _assert_frame_equal_pandas(left, right):
+    return pd.testing.assert_frame_equal(left, right)
+
+
+@assert_frame_equal.specialize("polars")
+def assert_frame_equal_polars(left, right):
+    return pl.testing.assert_frame_equal(left, right)
--- a/skrub/tests/test_joiner.py
+++ b/skrub/tests/test_joiner.py
@@ -2,25 +2,15 @@ import numpy as np
 import pandas as pd
 import pytest
 from numpy.testing import assert_array_equal
-from pandas.testing import assert_frame_equal
 
 from skrub import Joiner
-from skrub._dataframe._polars import POLARS_SETUP
-
-MODULES = [pd]
-ASSERT_TUPLES = [(pd, assert_frame_equal)]
-
-if POLARS_SETUP:
-    import polars as pl
-    from polars.testing import assert_frame_equal as assert_frame_equal_pl
-
-    MODULES.append(pl)
-    ASSERT_TUPLES.append((pl, assert_frame_equal_pl))
+from skrub._dataframe import _common as ns
+from skrub._dataframe._test_utils import assert_frame_equal
 
 
-@pytest.mark.parametrize("px", MODULES)
-def test_joiner(px):
-    main_table = px.DataFrame(
+# TODO: change this name
+def test_joiner(df_module):
+    main_table = df_module.make_dataframe(
         {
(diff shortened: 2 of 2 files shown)


jeromedockes:
thanks @TheooJ ! we seem to have some test failures with the ken embeddings that are unrelated to this PR, @jovan-stojanovic I think this could be due to the recent upload to figshare?

skrub/tests/test_joiner.py
```diff
@@ -30,7 +20,7 @@ def test_joiner(px):
         }
     )
 
-    aux_table = px.DataFrame(
+    aux_table = df_module.make_dataframe(
```
jeromedockes, line 23: in theory we should strive to stick to generic functions only so below `ns.shape(big_table)`, `ns.to_numpy(ns.col(big_table, "Population"))` instead of the direct calls to the libraries (which only happen to work because they use the same names for this).

skrub/tests/test_joiner.py
```diff
@@ -2,25 +2,15 @@
...
-    from polars.testing import assert_frame_equal as assert_frame_equal_pl
-
-    MODULES.append(pl)
-    ASSERT_TUPLES.append((pl, assert_frame_equal_pl))
-
-
-@pytest.mark.parametrize("px", MODULES)
-def test_joiner(px):
-    main_table = px.DataFrame(
+# TODO: change this name
```
jeromedockes, line 11: could you either rename it in this PR or remove the comment or make it more explicit to explain why it should be changed and why it's not done yet?

jeromedockes (commented):
Thanks @TheooJ I think that's the right approach! it should simplify the tests quite a bit. for completenesse could we remove all pandas-specific code from the `test_joiner`, eg `df.to_numpy()`

I guess now we have the tests the next step will be to make the joiner actually work with polars instead of converting to pandas under the hood...

skrub/_dataframe/_test_utils.py
```diff
@@ -1,6 +1,33 @@
...
+
+try:
+    import polars as pl
+    import polars.testing
+except ImportError:
+    pass
+
+from .._dispatch import dispatch
+
+
```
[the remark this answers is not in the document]
TheooJ: `is_module_pandas` is actually used nowhere
`is_module_polars` is used to raise `xfails` in some tests, it was just a convenience

skrub/tests/test_joiner.py
TheooJ: Yes I wanted to change it because
- having a test named `test_joiner` in the `test_joiner.py` file isn't very informative
- I wasn't able to run this test only using `pytest -k 'test_joiner'` because it shares name with the file

It's possible we want to split it into something like `test_fit_transform`, `test_wrong_main_key`, `test_wrong_aux_key`

skrub/tests/test_joiner.py
TheooJ: You're right, I've changed them
I also need a `to_list` for but you're already implementing it in https://github.com/skrub-data/skrub/pull/895

skrub/_dataframe/_test_utils.py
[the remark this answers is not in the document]
jeromedockes: ok I guess once we have updated most tests we won't need most of those xfails anymore, and with the new fixture it will just be `df_module.name == 'polars'`, and at that point we can remove these functions

skrub/tests/test_joiner.py
jeromedockes, line 11:
```suggestion
# TODO: change this name
# see https://github.com/skrub-data/skrub/pull/896#discussion_r1538167538
```

Issue #900, opened 2024-04-25
Failing fetch ken embeddings test


jeromedockes:
[This](https://github.com/skrub-data/skrub/actions/runs/8816192419/job/24199790449?pr=895#step:5:43) test has started failing, saying it does not find the footer of the downloaded parquet file.

I cannot reproduce locally; I am wondering if it could be that the file is only partly downloaded but it is weird that it only seems to happen in the python 3.11 environment.

In any case we should avoid tests that rely on network resources and mock the downloads, especially tests that download large files. So we need a quick fix for this test but in the medium term we should mock downloads in the main test suite

jeromedockes:
the ken embedding tests download 864M of data and takes over a minute to run locally so we probably want to start with those.
A good part of the data seems to be the embedding type ID (file 39266300) which is 652M; @jovan-stojanovic  do you know if this is one of the files that were recently updated on figshare?

jeromedockes:
I am also wondering about the way files are downloaded: the remote file is opened with pyarrow, then chunks of it are loaded into pandas dataframes and written in separate files with pandas.to_parquet.
However, in all cases the whole file is downloaded, and when using the skrub fetcher, the whole file is loaded into memory. So why do we use this approach rather than streaming the download into a single local file?

PR #901, opened 2024-04-25
temporarily skip ken embedding download tests

jeromedockes:
A quick fix for #900 (which should remain open until the downloads and test are actually improved and the skip is removed)


jeromedockes:
after discussion during the skrub meeting today, we decided to merge this as a short-term fix

PR #911, opened 2024-05-28
fix the import of parse_version broken after sklearn 1.5 release

jeromedockes:


jeromedockes:
actually there are also fixes for pandas and other scikit-learn breaking changes we should include

PR #913, opened 2024-05-28
Revert "Add column-wise transforms & refactor TableVectorizer"

glemaitre:
Reverts skrub-data/skrub#902


GaelVaroquaux:
We shouldn't do this revert. It's good to have this in main

PR #920, opened 2024-05-31
Make GapEncoder a single-column transformer

jeromedockes:
As discussed in skrub meetings.
Currently the actual functionality of the GapEncoder is implemented by the GapEncoderColumn. The GapEncoder is a wrapper that takes care of fitting a separate GapEncoderColumn to each column and collecting the results in a 2D array.
That part is now handled by the TableVectorizer, so we only need the GapEncoderColumn.

This PR essentially removes the GapEncoder, then renames GapEncoderColumn to GapEncoder

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -29,6 +29,7 @@ __all__ = [
     "to_list",
     "to_numpy",
     "to_pandas",
+    "reset_index",
     "make_dataframe_like",
     "make_column_like",
     "null_value_for",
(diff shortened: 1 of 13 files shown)


skrub/_dataframe/_common.py
```diff
@@ -29,6 +29,7 @@
     "to_list",
     "to_numpy",
     "to_pandas",
+    "reset_index",
```
jeromedockes, line 32: changes in this file are take from #919 and should be reviewed there

skrub/_gap_encoder.py
```diff
@@ -23,18 +22,164 @@
...
 
-from ._utils import check_input
+from . import _dataframe as sbd
+from ._on_each_column import SingleColumnTransformer
 
 
-class GapEncoderColumn(BaseEstimator, TransformerMixin):
-    """GapEncoder for encoding a single column.
+class GapEncoder(SingleColumnTransformer, TransformerMixin):
+    """Constructs latent topics with continuous encoding.
```
jeromedockes, line 30: the docstring has been moved essentially unchanged from the previous GapEncoder with 2 minor changes:

- the docstring still mentioned 'empty_impute', but apparently the parameter has been renamed 'zero_impute'
- the docstring mentioned 'inverse_transform' but it doesn't seem to be defined

glemaitre:
@jeromedockes could you solve the conflicts

skrub/_gap_encoder.py
glemaitre: I assume that this was resolved after what we discuss in the meeting.

glemaitre (approved):

PR #923, opened 2024-06-03
make minhash encoder a single-column transformer

jeromedockes:


jeromedockes:
ok @glemaitre  this one has been updated and it should be OK to review/merge !

skrub/_minhash_encoder.py
```diff
@@ -9,19 +9,21 @@
...
+from ._on_each_column import RejectColumn, SingleColumnTransformer
 from ._string_distances import get_unique_ngrams
-from ._utils import LRUDict, check_input
+from ._utils import LRUDict
 
 NoneType = type(None)
 
 
-class MinHashEncoder(TransformerMixin, BaseEstimator):
+class MinHashEncoder(SingleColumnTransformer, TransformerMixin):
```
glemaitre, line 26: For the inheritance, the Mixin needs to be on the right of the `BaseEstimator`. Otherwise the tag do not work properly via the MRO.

skrub/_minhash_encoder.py
```diff
@@ -93,26 +95,24 @@ class MinHashEncoder(TransformerMixin, BaseEstimator):
...
-            -1.48574446e+09, -1.68729890e+09],
-           [-1.97582893e+09, -2.09500033e+09, -1.59652117e+09,
-            -1.81759383e+09, -2.09569333e+09],
-           [-1.97582893e+09, -2.09500033e+09, -1.53072052e+09,
-            -1.45918266e+09, -1.58098831e+09]])
+             city_0        city_1        city_2        city_3        city_4
+    0 -1.783375e+09 -1.588270e+09 -1.663592e+09 -1.819887e+09 -1.962594e+09
+    1 -8.480470e+08 -1.766579e+09 -1.558912e+09 -1.485745e+09 -1.687299e+09
+    2 -1.975829e+09 -2.095000e+09 -1.596521e+09 -1.817594e+09 -2.095693e+09
+    3 -1.975829e+09 -2.095000e+09 -1.530721e+09 -1.459183e+09 -1.580988e+09
```
glemaitre, line 115: We should certainly switch to scipy-doctest at some points: https://github.com/scipy/scipy_doctest

skrub/_minhash_encoder.py
```diff
@@ -256,10 +254,20 @@ def fit(self, X: ArrayLike, y=None) -> "MinHashEncoder":
...
                 "any of {'error', 'zero_impute'}. "
             )
+        if self.minmax_hash and self.n_components % 2 != 0:
+            raise ValueError(
+                "n_components should be even when using"
+                f"minmax_hash encoding, got {self.n_components}"
+            )
+        if self.hashing == "murmur" and self.minmax_hash:
+            raise ValueError(
+                "minmax_hash encoding is not supportedwith the murmur hashing function"
```
glemaitre, line 264:
```suggestion
                "minmax_hash encoding is not supported with the murmur hashing function"
```

skrub/_minhash_encoder.py
glemaitre: I'll open a PR to solve that because I probably merge some other PR with this issue.

glemaitre (approved):

glemaitre (approved):
Apart from those little things. LGTM.

skrub/_minhash_encoder.py
jeromedockes: yep, looks super useful!

glemaitre:
I open https://github.com/skrub-data/skrub/pull/931 to solve the issue for the GapEncoder (and linter).

skrub/_minhash_encoder.py
jeromedockes: I'll change it here and in the gap encoder. Later we can consider removing the transformermixin because those transformers always output the correct dataframe type so have no need for the set_output stuff, and they are not used directly but always inside a tablevectorizer or similar

PR #928, opened 2024-06-06
MAINT use pixi to manage the CI dependencies

### LINKED ISSUE #860: development status in setup.cfg
in the setup.cf we have as development status "5 - Production/Stable". We should maybe change that to alpha or beta? here are the options: https://pypi.org/classifiers/

### LINKED ISSUE #899: incorrect github action matrix description
some errors in the 'include' value [here](https://github.com/skrub-data/skrub/blob/f5dfdc1ef8488b7cdeae75fd86903dbb32c61f0d/.github/workflows/testing.yml#L45):
- the dependencies-version min-py310 only has effect on the python 3.10 environment, so other jobs listed as "with minimal dependencies" actually use the latest versions of the dependencies. that may be relevant @jovan-stojanovic  when you look into the fetch ken embeddings failure on macos with python 3.11 and "minimal" dependencies -- check the actual versions used in the log they are the most recent
- the "dev, polars" for py 3.11 overwrites dev, pyarrow so the latter is never used

glemaitre:
closes #708 , closes #899, closes #860

Let's see if we can use `pixi` to have kind of lock files system to not have failing CI when contributing.

Here, I wanted to check that everything works before to go further.

It should be noted that the `pixi.toml` file can be merged with `pyproject.toml` and I don't think that we need anymore the `setup.cfg` as well.


pixi.lock
jeromedockes, line 1: I'm a bit surprised by the size of this file :D

glemaitre: Indeed. Basically, this the lock file for all the different environment and all different platforms. So here this is consolidated in a single file compare to scikit-learn. In addition, they store metadata for the package (I assume for whatever caching mechanism).

.github/workflows/pixi-testing.yml
```diff
@@ -0,0 +1,29 @@
+name: Unit Tests
```
jeromedockes, line 1: I assume in the future this would replace `testing.yml`?

glemaitre: Yep it was to have a first push to not remove the previous file :)

jeromedockes:
while you're tinkering with the CI you may want to see #899 as well

glemaitre:
I think the CI is failing because the "Required" CI are failing because I removed the script.

build_tools/circle/build_doc.sh
```diff
@@ -126,7 +126,7 @@ pip install .[doc]
 pip install scikit-learn
 
 # Build and install the project in dev mode
-python setup.py develop
+pip install .
```
glemaitre, line 129:
```suggestion
pip install -e .
```

skrub/_dataframe/tests/test_polars.py
```diff
@@ -29,7 +31,11 @@
 
 def test_join():
     joined = join(left=main, right=main, left_on="movieId", right_on="movieId")
-    expected = main.join(main, on="movieId", how="left", coalesce=True)
+    if "coalesce" in inspect.signature(main.join).parameters:
```
glemaitre, line 34: @jeromedockes I think that we need to make an extra change here as well.

skrub/tests/test_table_vectorizer.py
```diff
@@ -682,6 +682,10 @@ def test_accept_pipeline():
     tv.fit(df)
 
 
+@pytest.mark.skipif(
```
glemaitre, line 685: By checking properly the "polars" + sklearn, I found this issue as well.

.github/workflows/testing.yml
```diff
@@ -9,113 +9,69 @@ on:
...
+        run: pixi run -e ${{ matrix.environment }} test -n 3
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4.4.1
         with:
-          directory: ${{ runner.temp }}
-          verbose: true
-        if: success()
-        name: 'Upload coverage to CodeCov'
+          token: ${{ secrets.CODECOV_TOKEN }}
```
glemaitre, line 41: We might need to setup the secret in GitHub if it does not exist.

skrub/_dataframe/tests/test_polars.py
jeromedockes: which one?

skrub/tests/test_table_vectorizer.py
jeromedockes: that's a good point -- it doesn't affect only this test but all the scikit-learn transformers. I'll try to see what would be involved to have polars support on scikit-learn versions that don't support it. If it's complicated I think at first we could say that  skrub only supports polars dataframe if you have a recent enough scikit-learn version installed

pyproject.toml
```diff
@@ -1,11 +1,159 @@
...
 
+[project]
+name = "skrub"
+version = "0.2.0.dev0"
+description = "Prepping tables for machine learning"
+authors = [
+    { name="Patricio Cerda", email="patricio.cerda@inria.fr"},
+]
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
```
jeromedockes, line 13:
```suggestion
    "Development Status :: 4 - Beta",
```

while you're at it can you change this so we can close #860

skrub/tests/test_table_vectorizer.py
jeromedockes: this has been fixed by #941

skrub/_dataframe/tests/test_polars.py
glemaitre: I mean we did not to this change in `main` when merging the previous PR.

jeromedockes: ah right -- we did it in the skrub code but not in the test

glemaitre: Since we are going to merge this PR, let's doing it here. This is a minimal change.

.github/workflows/testing.yml
```diff
@@ -9,116 +9,69 @@ on:
...
+          # latest version time to time.
+          frozen: true
+
+      - name: Run tests
+        run: pixi run -e ${{ matrix.environment }} test -n 3
+
+      - name: Upload coverage reports to Codecov
+        uses: codecov/codecov-action@v4.4.1
         with:
-          directory: ${{ runner.temp }}
```
jeromedockes, line 88: does this mean we are now running the tests from the source directory? I think that's not ideal with the current layout of the package because it could end up being imported from the pwd? or maybe I am missing something

.github/workflows/testing.yml
```diff
@@ -9,116 +9,69 @@ on:
...
         with:
             python-version: "3.11"
         name: 'Setup python'
       - shell: bash {0}
-        run: ./build_tools/github/install.sh
-        name: 'Install skrub'
-        env:
-          INSTALL_NIGHTLY: "true"
-          DEPS_VERSION: "dev"
+        run: |
```
jeromedockes, line 68: I guess there's no way to specify this environment with pixi?

glemaitre: Here, it does not make sense to use `pixi` because you don't want to pin the nightly uild.

In this manner, we also test that our `pip` command of `pyproject.toml` work

jeromedockes (approved):
thanks @glemaitre ! this simplifies the CI configuration while fixing several issues

PR #929, opened 2024-06-06
MAINT update pre-commig-config

glemaitre:
Update the `pre-commit` configuration.

The idea is to:

- use the same as `scikit-learn`
  - update the version
  - use `ruff` and `black` instead of `isort` and `flake8`
- centralize a bit more the config into the `pyproject.toml`
- remove `mypy` since it is not used.

I kept the traceback over 88 characters in this PR but I needed to switch from `# noqa` to `# noqa: E501` at a different location.


glemaitre:
@jeromedockes there was not real changes in the file that should affect the code. This is only cosmetic due to updating the tools.

jeromedockes (approved):
thanks for doing this maintenance @glemaitre

glemaitre:
I see this is going towards the direction of https://github.com/skrub-data/skrub/pull/708

jeromedockes:
merging. here come the git conflicts :D

PR #931, opened 2024-06-07
FIX make sure to have Mixin on the left of BaseEstimator

glemaitre:
From the top of the head, we should always have the mixins on the left of `BaseEstimator` for the MRO to work properly (e.g. for tags inheritance).

Quick fix to solve issues that I let pass in a previous PR.


jeromedockes (approved):
thanks @glemaitre

PR #935, opened 2024-06-10
use _selectors.select in SelectCols and DropCols

jeromedockes:
small simplificaton of the implementation of `SelectCols` and `DropCols` that removes some duplicated code


TheooJ (approved):
LGTM, thanks @jeromedockes

PR #936, opened 2024-06-10
Remove pandas and polars specific code

TheooJ:
- Removed `make_dataframe`, `make_series` as we’re using dispatched functions now
- Removed `skrub._dataframe._polars import POLARS_SETUP` → use `conftest`’s `_POLARS_INSTALLED` instead
- Removed `DATAFRAME_MODULE_NAME` in `_pandas` and `_polars`, only used in `ns.skrub_namespace`

I plan on removing more stuff from `_dataframe._pandas/_polars/_namespace` in my next PRs

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -16,7 +16,6 @@ __all__ = [
     #
     # Inspecting containers' type and module
     #
-    "skrub_namespace",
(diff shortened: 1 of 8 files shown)


jeromedockes (approved):
thanks a lot @TheooJ !

PR #939, opened 2024-06-11
MAINT  add support for Python 3.9

### LINKED ISSUE #815: support python 3.8 & 3.9

jeromedockes:
closes #868, closes #815

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -35,7 +35,7 @@ jobs:
         fail-fast: false
         matrix:
             os: ["ubuntu-latest", "macos-latest", "windows-latest"]
-            python-version: ["3.9", "3.11", "3.12"]
+            python-version: ["3.9", "3.10", "3.11", "3.12"]
             # Install the min or latest dependencies for skrub
             # as defined in setup.cfg at [options.extra_require].
             #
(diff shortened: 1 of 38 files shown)


jeromedockes:
the macos + python 3.8 fails to install pre-commit; not really related to skrub. As  #928 will redo the CI configuration I won't edit it here

jeromedockes:
the latest scikit-learn release only supports 3.9 so I wonder if we should set that as our minimum supported version? OTOH we support older scikit-learn versions that work with python 3.8

glemaitre (approved):
LTGM. Not really sure why we have those "required" GHA.

jeromedockes:
> LTGM. Not really sure why we have those "required" GHA.

ah I see what happened -- we require all tests to pass in the CI before allowing to merge a PR. That  includes the py3.10 tests, but they don't exist anymore because i replaced 3.10 with 3.9 (the matrix of os/python/dependencies combinations is already huge) I guess I'll just add 3.10 back; anyway this is temporary because your pixi PR will redo the configuration

PR #940, opened 2024-06-12
enforce that sklearn >= 1.4 to use polars

jeromedockes:
the "polars" option to `set_output` was added in 1.4

this addresses this [comment](https://github.com/skrub-data/skrub/pull/928#discussion_r1632086377)

If we think asking for such a recent version to support polars is not reasonable, we can check the scikit-learn version in the code and do the conversion to polars ourselves if necessary -- `set_output` doesn't do anything complicated. WDYT?

### DIFF
--- a/skrub/tests/test_check_input.py
+++ b/skrub/tests/test_check_input.py
@@ -1,6 +1,8 @@
 import numpy as np
 import pandas as pd
 import pytest
+import sklearn
+from sklearn.utils.fixes import parse_version
 
 from skrub import _dataframe as sbd
 from skrub._check_input import CheckInputDataFrame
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
> If we think asking for such a recent version to support polars is not reasonable,

I think that it is OK

skrub/conftest.py
```diff
@@ -94,7 +96,7 @@ def _example_data_dict():
 except ImportError:
     _POLARS_INSTALLED = False
 
-if _POLARS_INSTALLED:
+if _POLARS_INSTALLED and parse_version("1.4") <= parse_version(sklearn.__version__):
```
TheooJ, line 99: nit : 
```suggestion
if _POLARS_INSTALLED and parse_version(sklearn.__version__) >= parse_version("1.4"):
```

skrub/tests/test_check_input.py
```diff
@@ -38,6 +40,15 @@ def test_input_is_an_array():
         check.fit_transform(np.ones((2,)))
 
 
+def test_sklearn_version_check():
+    if parse_version("1.4") <= parse_version(sklearn.__version__):
```
TheooJ, line 44:
```suggestion
    if parse_version(sklearn.__version__) >= parse_version("1.4"):
```

TheooJ (approved):

jeromedockes:
thinking about it a bit more I don't really like having a constraint on dependencies that is not expressed in the setup.cfg / checked during installation, but results in an error at runtime.
Also, scikit-learn 1.4 is very recent (last January), and older versions can cope with polars inputs, they just don't support outputing polars.
so #941 is probably a better solution

jeromedockes:
superseded by #941

PR #941, opened 2024-06-12
FIX handle polars transformer output in scikit-learn < 1.4

jeromedockes:
I think this may be a better option than #940 

when the input is a polars dataframe and the transformer given to OnEachColumn or OnSubFrame is a scikit-learn transformer which exposes the `set_output` API, we would like to call `set_output(transform='polars')` so that it returns us a polars dataframe.
However, in scikit-learn < 1.3, the 'polars' option is not available, only 'pandas'.
so we have a few options:
- force scikit-learn >= 1.4 but that is quite constraining for users because it is very recent
- basically re-implement set_output ourselves, which is not good because it duplicates code that is already in scikit-learn 
- use `set_output(transform='pandas')` and convert the result to `polars` with `polars.from_pandas`. This lets scikit-learn deal with the conversion from numpy, getting the column names from `get_feature_names_out` etc and produce a pandas dataframe which is easy to convert to polars. This is what this PR does. (for scikit-learn >= 1.4 it just uses `set_output(transform='polars')`

it does not complicate the code much, and avoids complicating the combinations of versions that we allow. it may also result in better error messages for custom estimators that don't support the set_output api.

In addition to doing the logic around scikit-learn 1.4 and pandas-> polars conversion this PR improves the checks of the transformer's output type

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -1,4 +1,4 @@
-from collections.abc import Mapping
+from collections.abc import Mapping, Sequence
 
 import numpy as np
 import pandas as pd
@@ -33,6 +33,7 @@ __all__ = [
     "null_value_for",
     "all_null_like",
     "concat_horizontal",
+    "is_column_list",
     "to_column_list",
     "col",
     "collect",
@@ -314,12 +315,22 @@ def _concat_horizontal_polars(*dataframes):
     return pl.concat(dataframes, how="horizontal")
 
 
+def is_column_list(obj):
+    if not isinstance(obj, Sequence):
+        return False
+    if not len(obj):
+        return True
+    if is_column(obj[0]):
+        return True
+    return False
+
+
 def to_column_list(obj):
     if is_column(obj):
         return [obj]
     if is_dataframe(obj):
         return [col(obj, c) for c in column_names(obj)]
-    if not hasattr(obj, "__iter__") or (len(obj) and not is_column(next(iter(obj)))):
+    if not is_column_list(obj):
         raise TypeError("obj should be a DataFrame, a Column or a list of Columns.")
     return obj
 
--- a/skrub/_utils.py
+++ b/skrub/_utils.py
@@ -1,7 +1,6 @@
 import collections
 import importlib
 import secrets
-import warnings
 from collections.abc import Hashable
 from typing import Any, Iterable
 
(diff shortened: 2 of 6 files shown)


skrub/_utils.py
```diff
@@ -156,22 +159,85 @@ def repr_args(args, kwargs, defaults={}):
...
-    module = sbd.dataframe_module_name(transform_input)
+def set_output(transformer, X):
+    if not hasattr(transformer, "set_output"):
+        return
+    module_name = sbd.dataframe_module_name(X)
+    if module_name == "polars" and parse_version(sklearn.__version__) < parse_version(
+        "1.4"
+    ):
+        # TODO: remove when scikit-learn 1.3 support is dropped.
+        warnings.warn(
```
jeromedockes, line 170: I'm not sure if we should have this warning. Hopefully in a growing number of cases polars can do the conversion from pandas without a copy

skrub/_dataframe/_common.py
```diff
@@ -314,12 +315,22 @@ def _concat_horizontal_polars(*dataframes):
     return pl.concat(dataframes, how="horizontal")
 
 
+def is_column_list(obj):
```
glemaitre, line 318: This will also accept `tuple` but I guess that we can silently support it.

skrub/_utils.py
glemaitre: I would skip the warning. At least with a have a better support in new version but warning could be quite annoying.

skrub/_utils.py
```diff
@@ -156,22 +159,85 @@ def repr_args(args, kwargs, defaults={}):
...
+            "To use polars, we recommend using scikit-learn version 1.4 or newer. The"
+            f" current version is {sklearn.__version__!r}. This will produce correct"
+            " results but may cause an extra copy of the transformer's output."
+        )
+        transformer.set_output(transform="pandas")
+        return
```
glemaitre, line 176: This is not cover but I assume that it will be in my PR.

skrub/tests/test_table_vectorizer.py
```diff
@@ -684,6 +684,8 @@ def test_accept_pipeline():
 
 def test_clean_null_downcast_warning():
     # non-regression test for https://github.com/skrub-data/skrub/issues/894
+    if parse_version(sklearn.__version__) < parse_version("1.4"):
```
glemaitre, line 687: It would be best to decorate the test:

```python
pytest.mark.skipif(
    parse_version(sklearn.__version__) < parse_version("1.4"),
    reason="xxx"
)
def test_...
```

skrub/_dataframe/_common.py
jeromedockes: yes any sequence will do but `is_column_list` is a bit simpler name than `is_column_sequence` and resembles `to_column_list`

skrub/_utils.py
jeromedockes: yes exactly -- that's how we found the polars and scikit-learn 1.3 issue in the first place. I thought it's not worth adding a dedicated workflow in the CI here because it will be superseded by your PR anyway

skrub/tests/test_table_vectorizer.py
jeromedockes: thanks. I'll keep that in mind for next time. here as we've just decided to remove the warning we don't need to skip the test anymore

glemaitre (approved):
LGTM.

PR #944, opened 2024-06-13
[MAINT] Harmonize parameter names in private helpers: col, df & specify argument_type in dispatched .specialize

TheooJ:
- Uniformize `df`, `col`, `obj` arguments in dispatched functions
- Specify `argument_type` when applicable


jeromedockes (approved):
thanks @TheooJ

PR #945, opened 2024-06-13
ENH Polars support in Joiner

TheooJ:
Refactoring the `Joiner`:
- Support for polars dataframes, use selectors + the dispatch mechanism in `Joiner` & `fuzzy_join()`
- Better handling of duplicate column names, and input dataframe checks
- Dispatch `left_join()` to have a single join utils
  - Cleanup in `agg_joiner` to remove `_pandas/_polars.join()`
- Dispatch `with_columns`


skrub/_dataframe/_common.py
```diff
@@ -955,3 +956,9 @@ def _replace_pandas(col, old, new):
 @replace.specialize("polars", argument_type="Column")
 def _replace_polars(col, old, new):
     return col.replace(old, new)
+
+
+def with_columns(df, **new_columns):
+    columns = {col_name: col(df, col_name) for col_name in column_names(df)}
+    columns.update({n: make_column_like(df, c, n) for n, c in new_columns.items()})
```
jeromedockes: as we discussed IRL, we will want to add a separate test to check that `make_column_like` does what we want when the input is already a column

jeromedockes: checking also that the result name is n even if the input column has a different name (that is already the case in the main branch but there is no test to check that aspect)

skrub/_join_utils.py
```diff
@@ -218,3 +228,107 @@ def _get_new_name(suggested_name, forbidden_names):
...
+        if "coalesce" in inspect.signature(left.join).parameters:
+            kw = {"coalesce": True}
+        else:
+            kw = {}
+        return left.join(right, how="left", left_on=left_on, right_on=right_on, **kw)
+    else:
+        raise TypeError(
+            "'left' and 'right' must be polars dataframes or lazyframes, "
+            f"got {type(left)!r} and {type(right)!r}."
+        )
```
jeromedockes, line 334:
```suggestion
def left_join(left, right, left_on, right_on, rename_right_cols="{}"):
    # TODO -- replace AssertionErrors with exceptions with appropriate type &
    # message
    assert sbd.is_dataframe(left)
    assert sbd.is_dataframe(right)
    assert sbd.dataframe_module_name(left) == sbd.dataframe_module_name(right)

    left_cols = sbd.column_names(left)
    original_right_cols = sbd.column_names(right)
    right_cols = map(_utils.renaming_func(rename_right_cols), original_right_cols)
    right_cols = pick_column_names(right_cols , forbidden_names=left_cols)
    renaming = dict(zip(original_right_cols, right_cols))
    right = sbd.set_column_names(right, right_cols)
    if isinstance(right_on, str):
        right_on = renaming[right_on]
        right_on_selector = s.cols(right_on)
    else:
        right_on = tuple(renaming[c] for c in right_on)
        right_on_selector = s.cols(*right_on)
    joined = _do_left_join(left, right, left_on, right_on)
    joined = s.select(joined, ~right_on_selector)
    return joined


@dispatch
def _do_left_join(left, right, left_on, right_on):
    raise NotImplementedError()


@_do_left_join.specialize("pandas", argument_type="DataFrame")
def _do_left_join_pandas(left, right, left_on, right_on):
    return left.merge(right, left_on=left_on, right_on=right_on, how="left", suffixes=("", ""))


@_do_left_join.specialize("polars", argument_type="DataFrame")
def _do_left_join_polars(left, right, left_on, right_on):
    if "coalesce" in inspect.signature(left.join).parameters:
        kw = {"coalesce": True}
    else:
        kw = {}
    return left.join(
        right, left_on=left_on, right_on=right_on, how="left", suffix="", **kw
    )
```

jeromedockes: This is what we wrote while pair-programming today. You will need these imports:
```python
import inspect
from skrub import _utils
from skrub import _selectors as s
from skrub._dispatch import dispatch
```

Here is an example:

```
>>> df
shape: (2, 4)
┌─────┬─────┬─────┬─────┐
│ a   ┆ b   ┆ c   ┆ d   │
│ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╡
│ 1   ┆ 3   ┆ 5   ┆ 7   │
│ 2   ┆ 4   ┆ 6   ┆ 8   │
└─────┴─────┴─────┴─────┘
>>> df2
shape: (2, 4)
┌─────┬─────┬─────┬─────┐
│ a   ┆ bb  ┆ cc  ┆ d   │
│ --- ┆ --- ┆ --- ┆ --- │
│ i64 ┆ i64 ┆ i64 ┆ i64 │
╞═════╪═════╪═════╪═════╡
│ 2   ┆ 4   ┆ 5   ┆ 7   │
│ 1   ┆ 3   ┆ 6   ┆ 8   │
└─────┴─────┴─────┴─────┘
>>> left_join(df, df2, left_on=("a", "b"), right_on=("a", "bb"))
shape: (2, 6)
┌─────┬─────┬─────┬─────┬─────┬─────────────────────┐
│ a   ┆ b   ┆ c   ┆ d   ┆ cc  ┆ d__skrub_add41ba6__ │
│ --- ┆ --- ┆ --- ┆ --- ┆ --- ┆ ---                 │
│ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64                 │
╞═════╪═════╪═════╪═════╪═════╪═════════════════════╡
│ 1   ┆ 3   ┆ 5   ┆ 7   ┆ 6   ┆ 8                   │
│ 2   ┆ 4   ┆ 6   ┆ 8   ┆ 5   ┆ 7                   │
└─────┴─────┴─────┴─────┴─────┴─────────────────────┘

>>> left_join(
...     df, df2, left_on=("a", "b"), right_on=("a", "bb"), rename_right_cols="right.{}"
... )
shape: (2, 6)
┌─────┬─────┬─────┬─────┬──────────┬─────────┐
│ a   ┆ b   ┆ c   ┆ d   ┆ right.cc ┆ right.d │
│ --- ┆ --- ┆ --- ┆ --- ┆ ---      ┆ ---     │
│ i64 ┆ i64 ┆ i64 ┆ i64 ┆ i64      ┆ i64     │
╞═════╪═════╪═════╪═════╪══════════╪═════════╡
│ 1   ┆ 3   ┆ 5   ┆ 7   ┆ 6        ┆ 8       │
│ 2   ┆ 4   ┆ 6   ┆ 8   ┆ 5        ┆ 7       │
└─────┴─────┴─────┴─────┴──────────┴─────────┘

```

skrub/_join_utils.py
```diff
@@ -218,3 +222,95 @@ def _get_new_name(suggested_name, forbidden_names):
...
+
+def left_join(left, right, left_on, right_on, rename_right_cols="{}"):
+    """Left join two dataframes of the same type.
+
+    The input dataframes type must agree: either both `left` and `right` need to be
+    :obj:`pandas.DataFrame`, :obj:`polars.DataFrame` or :obj:`polars.LazyFrame`.
+    Mixing types will raise an error.
+
+    `rename_right_cols` can be used to format the right dataframe columns, e.g. use
+    "right_.{}" to rename all right cols with a leading "right_".
```
jeromedockes, line 235:
```suggestion
    "right_.{}" to rename all right cols with a leading "right_.".
```

skrub/_join_utils.py
```diff
@@ -218,3 +222,95 @@ def _get_new_name(suggested_name, forbidden_names):
...
+    a __skrub_<random string>__ is added at the end of columns that would otherwise
+    be duplicates.
+
+    Parameters
+    ----------
+    left : pd.DataFrame or pl.DataFrame or pl.LazyFrame
+        The left dataframe of the left-join.
+    right : pd.DataFrame or pl.DataFrame or pl.LazyFrame
+        The right dataframe of the left-join.
+    left_on : str or Iterable[str]
```
jeromedockes, line 247: For readability I would prefer "str or list of str", or if you really want to insist it can be any sequence "str or sequence of str" rather than "Iterable[str]"

skrub/_join_utils.py
```diff
@@ -218,3 +222,95 @@ def _get_new_name(suggested_name, forbidden_names):
...
+        Left keys to merge on.
+    right_on : str or Iterable[str]
+        Right keys to merge on.
+    rename_right_cols : str, optional
```
jeromedockes, line 251: it can also be a callable that gets called with the original name as the sole argument

skrub/_join_utils.py
```diff
@@ -218,3 +222,95 @@ def _get_new_name(suggested_name, forbidden_names):
...
+        Formatting used to rename right cols, by default "{}".
+
+    Returns
+    -------
+    pd.DataFrame or pl.DataFrame or pl.LazyFrame
```
jeromedockes, line 256: maybe just "dataframe"? Let's try to limit the number of places where we explicitly list the possible dataframe libraries so we have less to edit if we ever add Ibis

skrub/_join_utils.py
```diff
@@ -218,3 +222,95 @@ def _get_new_name(suggested_name, forbidden_names):
         return suggested_name
     token = _utils.random_string()
     return f"{untagged_name}__skrub_{token}__"
+
+
+def left_join(left, right, left_on, right_on, rename_right_cols="{}"):
+    """Left join two dataframes of the same type.
+
+    The input dataframes type must agree: either both `left` and `right` need to be
+    :obj:`pandas.DataFrame`, :obj:`polars.DataFrame` or :obj:`polars.LazyFrame`.
```
jeromedockes, line 231: let's not mention LazyFrame for now; we don't need it in this private utility because we don't have support for lazy frames in the user-facing estimators

skrub/_join_utils.py
```diff
@@ -218,3 +222,95 @@ def _get_new_name(suggested_name, forbidden_names):
...
+        )
+    if not sbd.is_dataframe(right):
+        raise TypeError(
+            f"`right` must be a pandas or polars dataframe, got {type(right)}."
+        )
+    if not sbd.dataframe_module_name(left) == sbd.dataframe_module_name(right):
+        raise TypeError(
+            "`left` and `right` must be of the same dataframe type, got"
+            f"{type(left)} and {type(right)}."
+        )
```
jeromedockes, line 277: here we haven't checked the case of lazy frames but I don't think we need to; ATM the callers of this function shouldn't create laz frames

skrub/_joiner.py
```diff
@@ -120,7 +121,7 @@ class Joiner(TransformerMixin, BaseEstimator):
 
     Parameters
     ----------
-    aux_table : :obj:`~pandas.DataFrame`
+    aux_table : :obj:`~pandas.DataFrame` or :obj:`~polars.DataFrame`
```
jeromedockes, line 124: maybe just "dataframe"?

skrub/_joiner.py
```diff
@@ -271,7 +261,7 @@ def _check_max_dist(self):
     def _check_ref_dist(self):
         if self.ref_dist not in _MATCHERS:
             raise ValueError(
-                f"ref_dist should be one of {list(_MATCHERS.keys())}, got"
+                f"`ref_dist` should be one of {list(_MATCHERS.keys())}, got"
```
jeromedockes, line 264: this doesn't get rendered so we don't really need markub maybe we could use regular quotes instead

skrub/_joiner.py
```diff
@@ -305,11 +296,11 @@ def fit(self, X, y=None):
             X, self._aux_table, self.suffix, main_table_name="X"
         )
         self.vectorizer_ = _make_vectorizer(
-            self._aux_table[self._aux_key],
+            ns.col(self._aux_table, self._aux_key),
```
jeromedockes, line 299: here I think aux_key can be a tuple?

skrub/_join_utils.py
TheooJ: In the `AggJoiner` we went for `str or iterable of str`, WDYT ?

skrub/_join_utils.py
TheooJ: Keeping all the types only once in the docstring so that it's clear, replacing other occurrences with "dataframe"

skrub/_joiner.py
TheooJ: Replacing in `fit` and `transform` too

skrub/_join_utils.py
TheooJ: Will do another pass on this one, thanks

TheooJ:
Should we rename `self._matching` into `self._matcher`, as well as the associated files `_matching` into `_matcher` ? Or do we keep this for an upcoming PR ?

jeromedockes:
> Should we rename `self._matching` into `self._matcher`, as well as the associated files `_matching` into `_matcher` ? Or do we keep this for an upcoming PR ?

I would prefer to leave it for another PR

skrub/_join_utils.py
TheooJ: Taking `str or list of str` as discussed in https://github.com/skrub-data/skrub/pull/958

skrub/_joiner.py
TheooJ: Used selectors instead :) thanks

skrub/_agg_joiner.py
```diff
@@ -289,9 +289,9 @@ def transform(self, X):
         X, _ = self._check_dataframes(X, self.aux_table_)
         _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
 
-        skrub_px, _ = get_df_namespace(self.aux_table_)
-        X = skrub_px.join(
-            left=X,
+        # skrub_px, _ = get_df_namespace(self.aux_table_)
```
jeromedockes, line 292:
```suggestion
```

skrub/_joiner.py
```diff
@@ -12,20 +11,23 @@
 from sklearn.utils.validation import check_is_fitted
 
 from skrub import _join_utils, _matching, _utils
-from skrub._dataframe._namespace import is_pandas, is_polars
+from skrub import _selectors as s
+from skrub._check_input import CheckInputDataFrame
+from skrub._dataframe import _common as ns
```
jeromedockes, line 16:
```suggestion
from . import _dataframe as ns
```

also can you replace the other imports with relative ones, and I'm not too sure about the alias "ns"

skrub/_joiner.py
```diff
@@ -12,20 +11,23 @@
...
 
 
-def _as_str(column):
-    return column.fillna("").astype(str)
+def _as_str(col):
+    return ToStr().fit_transform(col)
 
 
 DEFAULT_STRING_ENCODER = make_pipeline(
     FunctionTransformer(_as_str),
```
jeromedockes, line 28:
```suggestion
    ToStr(),
```

and you can get rid of the function `_as_str`

skrub/_joiner.py
```diff
@@ -12,20 +11,23 @@
...
     HashingVectorizer(analyzer="char_wb", ngram_range=(2, 4)),
```
jeromedockes, line 29: here to deal with missing values you will need to pass an appropriate preprocessor to the HashingVectorizer, as is done [here](https://github.com/skrub-data/skrub/blob/da849a5c5e7846715d6ce0c82e1f6a5fcf7ed0c5/skrub/_gap_encoder.py#L228) for the GapEncoder. you can move the private `_preprocess_text` from `_gap_encoder` to `_utils` and rename it `preprocess_text`

skrub/_joiner.py
```diff
@@ -247,31 +248,25 @@ def __init__(
...
         if (
             self.max_dist is None
             or isinstance(self.max_dist, str)
             and self.max_dist == "inf"
         ):
             self.max_dist_ = np.inf
-        else:
+        elif isinstance(self.max_dist, int) or isinstance(self.max_dist, float):
             self.max_dist_ = self.max_dist
+        else:
```
jeromedockes, line 260: I don't think we need this. what do you think someone could mistakenly pass instead of a float and what does the exception look like in that case if we remove the check?

skrub/_joiner.py
```diff
@@ -318,52 +316,47 @@ def transform(self, X, y=None):
...
         _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
         _join_utils.check_column_name_duplicates(
             X, self._aux_table, self.suffix, main_table_name="X"
         )
         main = self.vectorizer_.transform(
-            X[self._main_key].set_axis(self._aux_key, axis="columns")
+            ns.set_column_names(s.select(X, s.cols(*self._main_key)), self._aux_key)
+        )
+        aux_table = ns.reset_index(
+            _join_utils.add_column_name_suffix(self._aux_table, self.suffix)
```
jeromedockes, line 340: maybe we could use the `left_join` `rename_right_cols` parameter and not need `add_column_name_suffix` anymore?

skrub/_joiner.py
```diff
@@ -318,52 +316,47 @@ def transform(self, X, y=None):
...
         )
-        match_result = self._matching.match(main, self.max_dist_)
-        aux_table = _join_utils.add_column_name_suffix(
-            self._aux_table, self.suffix
-        ).reset_index(drop=True)
-        matching_col = match_result["index"].copy()
-        matching_col[~match_result["match_accepted"]] = -1
+        _match_result = self._matching.match(main, self.max_dist_)
```
jeromedockes, line 342: why do you need to rename `_match_result`?

skrub/_joiner.py
```diff
@@ -318,52 +316,47 @@ def transform(self, X, y=None):
...
-        join = join.drop([left_key_name, right_key_name], axis=1)
+        join = s.select(join, ~s.cols(left_key_name))
         if self.add_match_info:
             for info_key, info_col_name in self._match_info_key_renaming.items():
-                join[info_col_name] = match_result[info_key]
-        if input_is_polars:
-            import polars as pl
-
-            join = pl.from_pandas(join)
+                join = ns.with_columns(join, **{info_col_name: _match_result[info_key]})
```
jeromedockes, line 361: I think we should build the dictionary in the loop and call `with_columns` only once it will be faster and a bit clearer

skrub/_joiner.py
TheooJ: As of now we've used :
- `from ._dataframe import common as ns`
- `from . import _dataframe as sbd`

so I can put `from . import _dataframe as sbd` here

skrub/_joiner.py
jeromedockes: sorry actually that won't be enough because the decoding will fail before we reach the preprocessor. so instead we can have `make_pipeline(FunctionTransformer(partial(sbd.fill_null, value="")), ToStr(), HashingVectorizer())`

skrub/_joiner.py
TheooJ: I've used a false parameter as if the user copy / pasted the previous argument and it raises `UFuncTypeError: ufunc 'less_equal' did not contain a loop with signature matching types (, ) -> None` on the line `"match_accepted": rescaled_distances <= max_dist`

Removing it as it might be too narrow

skrub/_joiner.py
TheooJ: In a previous commit I was converting it to a dataframe, I don't need to rename :)

skrub/_joiner.py
```diff
@@ -3,29 +3,25 @@
...
 from ._wrap_transformer import wrap_transformer
 
-
-def _as_str(column):
-    return column.fillna("").astype(str)
-
-
 DEFAULT_STRING_ENCODER = make_pipeline(
-    FunctionTransformer(_as_str),
+    ToStr(),
```
jeromedockes, line 22:
```suggestion
    FunctionTransformer(partial(sbd.fill_nulls, value="")),
    ToStr(),
```

you will need `from functools import partial` and to restore the `FunctionTransformer` import

skrub/_joiner.py
TheooJ: Thanks !

jeromedockes:
the CI failure is due to the fact polars support in the column transformer was only added in scikit-learn 1.4, when polars was added to the set_output API: https://github.com/scikit-learn/scikit-learn/pull/27315/files#diff-66316a90c42bbe375eef74f1b9dcbd6a1b5b99f60feb692f676067a633d30f60R226

jeromedockes:
I guess a workaround would be to provide the columns directly as integer indices to the columntransformer instead of strings

jeromedockes:
> I guess a workaround would be to provide the columns directly as integer indices to the columntransformer instead of strings

if we do this we get another error when scikit-learn tries to access `.ndims`. but actually here we are just vectorizing the key columns and we want the output as numpy or a scipy sparse matrix anyway, I think we can just convert to pandas before vectorizing.

```diff
diff --git a/skrub/_joiner.py b/skrub/_joiner.py
index 0c325a09..795f71c2 100644
--- a/skrub/_joiner.py
+++ b/skrub/_joiner.py
@@ -5,11 +5,13 @@ The Joiner provides fuzzy joining as a scikit-learn transformer.
 from functools import partial
 
 import numpy as np
+import sklearn
 from sklearn.base import BaseEstimator, TransformerMixin, clone
 from sklearn.compose import make_column_transformer
 from sklearn.feature_extraction.text import HashingVectorizer, TfidfTransformer
 from sklearn.pipeline import make_pipeline
 from sklearn.preprocessing import FunctionTransformer, StandardScaler
+from sklearn.utils.fixes import parse_version
 from sklearn.utils.validation import check_is_fitted
 
 from . import _dataframe as sbd
@@ -39,6 +41,12 @@ _MATCHERS = {
 DEFAULT_REF_DIST = "random_pairs"
 
 
+def _compat_df(df):
+    if parse_version(sklearn.__version__) < parse_version("1.4"):
+        return sbd.to_pandas(df)
+    return df
+
+
 def _make_vectorizer(table, string_encoder, rescale):
     """Construct the transformer used to vectorize joining columns.
 
@@ -299,7 +307,7 @@ class Joiner(TransformerMixin, BaseEstimator):
             rescale=self.ref_dist != "no_rescaling",
         )
         aux = self.vectorizer_.fit_transform(
-            s.select(self._aux_table, s.cols(*self._aux_key))
+            _compat_df(s.select(self._aux_table, s.cols(*self._aux_key)))
         )
         self._matching.fit(aux)
         return self
@@ -327,7 +335,11 @@ class Joiner(TransformerMixin, BaseEstimator):
             X, self._aux_table, self.suffix, main_table_name="X"
         )
         main = self.vectorizer_.transform(
-            sbd.set_column_names(s.select(X, s.cols(*self._main_key)), self._aux_key)
+            _compat_df(
+                sbd.set_column_names(
+                    s.select(X, s.cols(*self._main_key)), self._aux_key
+                )
+            )
         )
         match_result = self._matching.match(main, self.max_dist_)
         matching_col = match_result["index"].copy()
```

this fixes the issue

jeromedockes:
@TheooJ I opened a pr on your branch to fix the issue for old scikit-learn versions: https://github.com/TheooJ/skrub/pull/2

skrub/_joiner.py
```diff
@@ -292,8 +287,9 @@ def fit(self, X, y=None):
...
         """
         del y
-        X = self._check_dataframe(X)
-        self._aux_table = self._check_dataframe(self.aux_table)
+        self._aux_table = CheckInputDataFrame().fit_transform(self.aux_table)
+        self._main_check_input = CheckInputDataFrame()
+        X = self._main_check_input.fit_transform(X)
         self._check_ref_dist()
         self._check_max_dist()
         self._main_key, self._aux_key = _join_utils.check_key(
```
jeromedockes, line 295: we may not need all of check_key anymore, in particular left_join can now handle duplicate names in the result, but let's clean up the checks in a follow-up pr

skrub/tests/test_joiner.py
```diff
@@ -76,25 +71,25 @@ def test_multiple_keys(df_module):
...
     )
     result = joiner_list.fit_transform(df)
-    assert_frame_equal(result, expected)
+    df_module.assert_frame_equal(result, expected)
 
 
-def test_pandas_aux_table_index():
-    main_table = pd.DataFrame({"Country": ["France", "Italia", "Georgia"]})
-    aux_table = pd.DataFrame(
+def test_pandas_aux_table_index(df_module):
```
jeromedockes, line 90: this test is about handling of pandas indexes, so it is really specific to pandas -- there is no need to run it on polars so I would not use df_module here

skrub/tests/test_joiner.py
```diff
@@ -115,15 +110,47 @@ def test_pandas_aux_table_index():
...
-    table = pd.DataFrame({"A": [1, 2]})
-    joiner = Joiner(table, key="A", ref_dist="bad")
-    with pytest.raises(ValueError, match="got 'bad'"):
-        joiner.fit(table)
+def test_wrong_ref_dist(main_table, aux_table):
+    joiner = Joiner(
+        aux_table, main_key="Country", aux_key="country", ref_dist="wrong_ref_dist"
+    )
+    with pytest.raises(
+        ValueError, match=r"('ref_dist' should be one of)*(Got 'wrong_ref_dist')"
```
jeromedockes, line 118:
```suggestion
        ValueError, match=r"'ref_dist' should be one of.* Got 'wrong_ref_dist'"
```

jeromedockes (commented):
thanks @TheooJ ! a few more comments :)

skrub/_joiner.py
```diff
@@ -304,12 +300,15 @@ def fit(self, X, y=None):
         _join_utils.check_column_name_duplicates(
             X, self._aux_table, self.suffix, main_table_name="X"
         )
+        self._suffix = f"{{}}{self.suffix}".format
```
jeromedockes, line 303: maybe `self._right_cols_renaming`

skrub/tests/test_fuzzy_join.py
```diff
@@ -137,41 +123,41 @@ def test_drop_unmatched(df_module):
...
-    result = pd.merge(left, right, left_on="key", right_on="key_")
+    result = _join_utils.left_join(left, right, left_on="key", right_on="key_")
     result_fj = fuzzy_join(
         left, right, left_on="key", right_on="key_", add_match_info=False
     )
+    # `fuzzy_join`` keeps the vectorized col, so we must drop it
+    result_fj = s.select(result_fj, ~s.cols("key_"))
 
-    assert_frame_equal(result, result_fj)
+    df_module.assert_frame_equal(result, result_fj)
```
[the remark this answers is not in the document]
TheooJ: Thanks !

skrub/tests/test_joiner.py
TheooJ: Thanks

Removing the dispatching too to make it clearer

jeromedockes (approved):
great! thanks a ton @TheooJ  this is a big one!!

Issue #671, opened 2023-07-20
API: shorter names / argument names


GaelVaroquaux:
Long names lead to long lines of code, and beginners do not know how to format them right, and as a consequence their code is hard to read and they struggle. I am starting to think that we need to try to avoid long names when possible.

1. [ ] Rename all arguments in TableVectorizer that end with "_transformer" to not end with "_transformer", eg: "high_card_cat_transformer" -> "high_card_cat"
2. [ ] Use short hands when they are obvious: "col" instead of "columns" (no clear list in my head of where this should be done, maybe we should audit our API to find out)
3. [ ] Use verbs instead of nouns? (this is more open IMHO)

Vincent-Maladiere:
Regarding the TableVectorizer, do you think "high_cardinality"/"low_cardinality" would be more suited than "high_card_cat"/"low_card_cat"? It's more explicit but still short.

jeromedockes:
let's make a decision on the TableVectorizer parameters before the next release

rcap107:
We may want to go over the library again to see if there are more parameters that may need to be shortened

PR #947, opened 2024-06-17
Rename TableVectorizer parameters

jeromedockes:
rename the TableVectorizer parameters and corresponding attributes according to #671 and this morning's skrub meeting discussion:

```
cardinality_threshold → n_unique_threshold
high_cardinality_transformer → many_unique
low_cardinality_transformer → few_unique
datetime_transformer → datetime
numeric_transformer → numeric
```

I'm not 100% sure the new names are better esp. the "cardinality" replacement  :thinking:

### DIFF
--- a/benchmarks/bench_gap_divergence.py
+++ b/benchmarks/bench_gap_divergence.py
@@ -206,7 +206,7 @@ def benchmark(max_iter_e_step, dataset_name):
                 (
                     "encoding",
                     TableVectorizer(
-                        high_cardinality_transformer=ModifiedGapEncoder(
+                        many_unique=ModifiedGapEncoder(
                             min_iter=5,
                             max_iter=5,
                             max_iter_e_step=max_iter_e_step,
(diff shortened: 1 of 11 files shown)


TheooJ:
I like the removal of `transformer` everywhere but I'm not sold on `many_unique` and `few_unique` yet

GaelVaroquaux:
I started a twitter poll on this: https://x.com/skrub_data/status/1803508324941402198

jeromedockes:
it's a detail but I put `many_unique` without an S instead of `many_uniques` because the polars and pandas functions are called `n_unique` (no s) and I don't think 'unique' as a noun exists in english

GaelVaroquaux:
> it's a detail but I put many_unique without an S instead of many_uniques because the polars and pandas functions are called n_unique (no s)

Good thinking

TheooJ:
I prefer this option too

jeromedockes:
ok then I'll update the PR to just drop the `transformer` part and keep 'cardinality'

PR #950, opened 2024-06-17
MAINT require optional dependency to build the documentation

glemaitre:
This should fix `main` branch CIs build

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -147,7 +147,7 @@ test = { cmd = "pytest -vsl --cov=skrub --cov-report=xml skrub" }
 
 [tool.pixi.environments]
 lint = ["lint"]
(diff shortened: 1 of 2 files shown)


pyproject.toml
```diff
@@ -147,7 +147,7 @@ test = { cmd = "pytest -vsl --cov=skrub --cov-report=xml skrub" }
 
 [tool.pixi.environments]
 lint = ["lint"]
-doc = ["doc"]
+doc = ["optional", "doc"]
```
glemaitre, line 150: @jeromedockes This should be the culprit since the doc CI only run on changed example.

I trigger a full build here to make sure.

jeromedockes (approved):
thanks!

PR #956, opened 2024-06-17
FIX always take into account suffix parameter in fetch_ken_embeddings

glemaitre:
When `pca_components=None`, the parameter `suffix` was disregarded in `fetch_ken_embeddings`.

This PR fixes this bug and add a non-regression test.


jeromedockes (approved):
LGTM. thanks a lot @glemaitre !

PR #959, opened 2024-06-18
FIX remove the unwanted suffix in example 6

jeromedockes:
the example needs this small change after the bugfix in #956


jeromedockes:
most likely circleCI will fail here because the example uses too much memory but that's a separate issue

PR #962, opened 2024-06-18
DOC nitpicks on the documentation of `tabular_learner`

glemaitre:
Those are comments that I wanted to raise regarding the documentation. There is nothing radical here. I just wanted to expand a bit the user guide by providing some information regarding the parameter of the `TableVectorizer`

### DIFF
--- a/doc/encoding.rst
+++ b/doc/encoding.rst
@@ -10,6 +10,8 @@ Encoding or vectorizing creates numerical features from the data,
 converting dataframes, strings, dates... Different encoders are suited
 for different types of data.
 
+.. _table_vectorizer:
+
 Turning a dataframe into a numerical feature matrix
 ---------------------------------------------------
 
--- a/doc/encoding.rst
+++ b/doc/encoding.rst
@@ -59,6 +59,17 @@ type of the final estimator.
      - Native support
      - :class:`~sklearn.impute.SimpleImputer`
 
+With tree-based models, the :obj:`MinHashEncoder` is used for high-cardinality
+categorical features. It does not provide interpretable features as the
+default :obj:`GapEncoder` but it is much faster. For low-cardinality, these models
+relies on either the native support of the model or the :obj:`OrdinalEncoder`.
+
+With linear models or unknown models, the default values of the different parameters
+are used. Therefore, the :obj:`GapEncoder` is used for high-cardinality categorical
+features and the :obj:`OneHotEncoder` for low-cardinality ones. If the final estimator
+does not support missing values, a :obj:`SimpleImputer` is added before the final
+estimator. Those choices are not optimal but they are methodologically safe.
+
 .. _dirty_categories:
 
 Encoding open-ended entries and dirty categories
--- a/doc/encoding.rst
+++ b/doc/encoding.rst
@@ -11,26 +11,56 @@ converting dataframes, strings, dates... Different encoders are suited
 for different types of data.
 
 Turning a dataframe into a numerical feature matrix
-----------------------------------------------------
+---------------------------------------------------
 
 A dataframe can comprise columns of all kind of types. A good numerical
 representation of these columns help analytics and statistical learning.
 
 The :class:`TableVectorizer` gives a turn-key solution by applying
-different data-specific encoder to the different columns. It makes
-heuristic choices that are not necessarily optimal but is typically a
-very good baseline.
+different data-specific encoder to the different columns. It makes reasonable
+heuristic choices that are not necessarily optimal since it is not aware of the learner
+used for the machine learning task). However, it already provides a typically very good
+baseline.
 
-The function :func:`tabular_learner` is an easy way to create a machine-learning
-model that works well on tabular data. It is given a scikit-learn estimator and
-creates a :class:`~sklearn.pipeline.Pipeline` combining a
-:class:`TableVectorizer`, optional missing value imputation and the provided
-estimator.
+The function :func:`tabular_learner` goes the extra mile by creating a machine-learning
+model that works well on tabular data. This model combines a :class:`TableVectorizer`
+with a provided scikit-learn estimator. Depending whether or not the final estimator
+natively support missing values, a missing value imputer step is added before the
+final estimator. The parameters of the :class:`TableVectorizer` are chosen based on the
+type of the final estimator.
+
+.. list-table:: Parameter values choice of :class:`TableVectorizer` when using :func:`tabular_learner` function
+   :header-rows: 1
+
+   * -
+     - ``RandomForest`` models
+     - ``HistGradientBoosting`` models
+     - Linear models and others
+   * - Low-cardinality encoder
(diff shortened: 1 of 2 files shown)


skrub/_tabular_learner.py
```diff
@@ -104,7 +114,7 @@ def tabular_learner(estimator, n_jobs=None):
     ...         "fasting_glucose": [35, 140, 44, 137],
     ...     }
     ... )
-    >>> y = [False, True, False, True]
+    >>> y = [0, 1, 0, 1]
```
glemaitre, line 117: I would not encourage to not use integral classification values when this is possible

skrub/_tabular_learner.py
```diff
@@ -129,34 +136,34 @@ def tabular_learner(estimator, n_jobs=None):
...
 
-    By applying only the first pipeline step we can see the transformed data
-    that is sent to the supervised estimator (see the :obj:`TableVectorizer`
-    documentation for details):
+    By applying only the first pipeline step we can see the transformed data that is
+    sent to the supervised estimator (see the :obj:`TableVectorizer` documentation for
+    details):
 
-    >>> model.named_steps['tablevectorizer'].transform(X)            # doctest: +SKIP
+    >>> model.named_steps['tablevectorizer'].transform(X)
```
glemaitre, line 143: I would avoid not testing the docstring. I think this is useful to find out that this is outdated.

skrub/_tabular_learner.py
```diff
@@ -129,34 +136,34 @@ def tabular_learner(estimator, n_jobs=None):
...
     Pipeline(steps=[('tablevectorizer', TableVectorizer()),
                     ('simpleimputer', SimpleImputer(add_indicator=True)),
                     ('logisticregression', LogisticRegression())])
 
-    We see that for the :obj:`~sklearn.linear_model.LogisticRegression` we get
-    the default configuration of the ``TableVectorizer`` which is intended to
-    work well for a wide variety of downstream estimators. Moreover, as the
-    ``LogisticRegression`` cannot handle missing values, an imputation step is
-    added.
+    We see that for the :obj:`~sklearn.linear_model.LogisticRegression` we get the
```
glemaitre, line 157: I wondering if the discussion here and below should not go directly to the user guide.

My own habit is to use this section to quickly copy pasting code to get started. If I want to get inside I would naturally go read the information in the user guide.

I was wondering if we want a similar pattern or this is just me?

skrub/_tabular_learner.py
```diff
@@ -79,20 +83,26 @@ def tabular_learner(estimator, n_jobs=None):
...
+    :obj:`~sklearn.ensemble.HistGradientBoostingRegressor`.
+
+    >>> tabular_learner('classifier')
     Pipeline(steps=[('tablevectorizer',
                      TableVectorizer(high_cardinality_transformer=MinHashEncoder(),
                                      low_cardinality_transformer=ToCategorical())),
                     ('histgradientboostingclassifier',
                      HistGradientBoostingClassifier(categorical_features='from_dtype'))])
 
+    When requesting a ``'classifier'``, the last of the pipeline is set to a
```
TheooJ, line 105:
```suggestion
    When requesting a ``'classifier'``, the last step of the pipeline is set to a
```

skrub/_tabular_learner.py
```diff
@@ -79,20 +83,26 @@ def tabular_learner(estimator, n_jobs=None):
...
-    >>> tabular_learner('regressor')                             # doctest: +SKIP
+    >>> tabular_learner('regressor')
     Pipeline(steps=[('tablevectorizer',
                      TableVectorizer(high_cardinality_transformer=MinHashEncoder(),
                                      low_cardinality_transformer=ToCategorical())),
                     ('histgradientboostingregressor',
                      HistGradientBoostingRegressor(categorical_features='from_dtype'))])
 
-    >>> tabular_learner('classifier')                            # doctest: +SKIP
+    When requesting a ``'regressor'``, the last of the pipeline is set to a
```
TheooJ, line 95:
```suggestion
    When requesting a ``'regressor'``, the last step of the pipeline is set to a
```

doc/encoding.rst
```diff
@@ -10,27 +10,59 @@ Encoding or vectorizing creates numerical features from the data,
...
+baseline.
+
+The function :func:`tabular_learner` goes the extra mile by creating a machine-learning
+model that works well on tabular data. This model combines a :class:`TableVectorizer`
+with a provided scikit-learn estimator. Depending whether or not the final estimator
+natively support missing values, a missing value imputer step is added before the
+final estimator. The parameters of the :class:`TableVectorizer` are chosen based on the
+type of the final estimator.
+
+.. list-table:: Parameter values choice of :class:`TableVectorizer` when using :func:`tabular_learner` function
```
glemaitre, line 34: This table should look like:

image

doc/encoding.rst
```diff
@@ -11,26 +11,56 @@ converting dataframes, strings, dates... Different encoders are suited
...
+   :header-rows: 1
+
+   * -
+     - ``RandomForest`` models
```
TheooJ, line 38: +1 for this, really useful

TheooJ (approved):

skrub/_tabular_learner.py
glemaitre: OK now I understand why we were skipping :)

glemaitre:
Now I understand why we were skipping the test.

I will have to change a bit the narrative. I think in the user guide, we have space to have an (*) to explain that the native support was added at some point. If this is not the case then we fallback to the same strategy than the forest.

This could avoid some user question. I'll do that tomorrow.

doc/encoding.rst
jeromedockes: that's a good idea, super useful! the non-hist gradient boosting would go in the same column as randomforest but just "RandomForest" is good it's better to keep the headings short

jeromedockes:
> Now I understand why we were skipping the test.

yes the choice depend on the scikit-learn version because in 1.2 the HGBT did not have the "from_dtype" option, and randomforest did not have missing value support so the output of the doctests would be a bit different. an alternative to skipping the test altogether could be adding ellipsis in the output

skrub/_tabular_learner.py
```diff
@@ -48,19 +48,25 @@ def tabular_learner(estimator, n_jobs=None):
       missing values itself.
     - the last step is the provided ``estimator``.
 
-    **Note:** ``tabular_learner`` is a recent addition and the heuristics used
-    to define an appropriate preprocessing based on the ``estimator`` may change
-    in future releases.
+    Read more in the :ref:`User Guide <table_vectorizer>`.
+
+    .. note::
```
jeromedockes, line 53: :) I had done this at first but I thought the way notes are rendered with our template gives them too much visual importance for this comment

jeromedockes (approved):
thanks a lot @glemaitre !

jeromedockes:
>  I'll do that tomorrow.

do you want to do that in this PR or should I merge it?

PR #963, opened 2024-06-18
FIX use keyword-only argument

glemaitre:
Small nitpicks where I would like to require all arguments apart from `estimator` to be keyword only. It would allow us to introduce any new parameter in the future without caring that our user was passing by position or not.

It is removing the future hassle questioning if we should put a new parameter or not of the signature.


jeromedockes (approved):
good idea. we might want to do that in a few other places

PR #964, opened 2024-06-19
remove _preprocess_text which is not necessary

jeromedockes:
handling missing values is actually handled in the computation of unique values now, with the addition of `unique_strings` so passing a custom preprocessor is not necessary and anyway it works for None and pd.NA but not for numpy nan


TheooJ (approved):
Thanks !

PR #966, opened 2024-06-19
add standard scaling to tabular_learner

jeromedockes:
WIP: I'll wait for #962  to edit the documentation

### DIFF
--- a/doc/encoding.rst
+++ b/doc/encoding.rst
@@ -49,7 +49,7 @@ type of the final estimator.
    * - Numerical preprocessor
      - No processing
      - No processing
-     - No processing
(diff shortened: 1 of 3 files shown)


TheooJ:
We need a - StandardScaler bulletpoint in the doc too

GaelVaroquaux (approved):
Looks good to me.

I note that we are not documenting this behavior. I think that this is fine, as documenting the details of tabular_learner would lead to an unreadable documentation hard to maintain.

jeromedockes:
I wanted to merge #962 before updating the doc to avoid dealing with conflicts, I just did both so the standard scaling is documented and added to the table.

about the table I wonder if the rows should be ordered eg according to their position in the pipeline

PR #967, opened 2024-06-19
Simple start of the example

GaelVaroquaux:
My attention span is tiny. I need to fit everything on a screen to understand.

I removed a few lines / compressed the text to be able to see the important points quicker.


jeromedockes (approved):
Thanks @GaelVaroquaux ! I checked the exemple renders well in the online doc

GaelVaroquaux:
Thanks, that was fast!!

PR #968, opened 2024-06-19
Fixes

GaelVaroquaux:
Sorry, something on top of https://github.com/skrub-data/skrub/pull/967 : a few typos, a bit of layout (the "|" adds vertical space), and an additional line which I believe helps.

### DIFF
--- a/examples/01_encodings.py
+++ b/examples/01_encodings.py
@@ -62,11 +62,11 @@ employees
 # set of columns: some are numerical (``'year_first_hired'``), some dates
 # (``'date_first_hired'``), some have a few categorical entries
 # (``'gender'``), some many (``'employee_position_title'``). Therefore
-# our table needs to be "vectorized": processed to extracted numeric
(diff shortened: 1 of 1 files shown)


examples/01_encodings.py
```diff
@@ -62,11 +62,11 @@
 # set of columns: some are numerical (``'year_first_hired'``), some dates
```
TheooJ, line 62: some are dates ?

GaelVaroquaux: Actually, I think that this is proper English

TheooJ:
Should it be `skrub`'s ? or do we keep skrub's ?

GaelVaroquaux:
> Should it be `skrub`'s ? or do we keep skrub's ?

It doesn't matter much to me

TheooJ (approved):
Merging then !

GaelVaroquaux:
Thanks!!

PR #971, opened 2024-06-24
MAINT xfail the fetch_movielens test

jeromedockes:
this test if causing the CI to fail due to a download error. here also, (as for the ken embeddings) the real fix is to mock doownloads in tests but in the meanwhile this allows the ci to run @glemaitre


jeromedockes:
also fix the unbound local variable error

glemaitre (approved):

PR #977, opened 2024-07-01
FIX deduplication of gapencoder output columns

jeromedockes:
fixes a regression introduced when the GapEncoderColumn was made the GapEncoder: appending column numbers to topic names was not done correctly anymore

### DIFF
--- a/skrub/_gap_encoder.py
+++ b/skrub/_gap_encoder.py
@@ -551,7 +551,7 @@ class GapEncoder(TransformerMixin, SingleColumnTransformer):
             label = ", ".join(labels)
             if self._input_name:
                 label = f"{self._input_name}: {label}"
-            # Avoid having twice the same name for the different features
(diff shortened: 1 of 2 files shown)


skrub/_gap_encoder.py
```diff
@@ -555,12 +549,11 @@ def get_feature_names_out(
             x = encoding[:, i]
             labels = vocabulary[np.argsort(-x)[:n_labels]]
             label = ", ".join(labels)
-            label = prefix + label
-            # Avoid having twice the same name for the different features
-            if label in topic_labels:
-                label += " ({:})".format(i)
             if self._input_name:
                 label = f"{self._input_name}: {label}"
+            # Avoid having twice the same name for the different features
```
TheooJ, line 554:
```suggestion
            # Avoid having the same name twice for the different features
```

TheooJ (approved):
Thanks @jeromedockes !

PR #992, opened 2024-07-12
DOC reorganize a bit user guide structure

glemaitre:
This PR will:

- move the development page to its own section to have the user guide only about the library tools
- add the changelog out of the development page
- remove the left sidebar when there is no item to click on
- add a title to the index page
- remove the footer on the page. We should probably make it more visible to get access to the changelog and contributing guideline.

### DIFF
--- a/doc/index.rst
+++ b/doc/index.rst
@@ -1,9 +1,9 @@
+.. title:: Index
 
 .. toctree::
    :maxdepth: 2
 
 .. currentmodule:: skrub
-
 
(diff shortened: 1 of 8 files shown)


doc/index.rst
```diff
@@ -1,14 +1,16 @@
+.. title:: Index
```
jeromedockes, line 1:
```suggestion
.. title:: Home
```

jeromedockes (approved):

PR #995, opened 2024-07-12
Revamp README

TheooJ:
- Add example snippet
- Fix broken installation link
- Remove refs & additional resources


README.rst
```diff
@@ -1,4 +1,4 @@
-`skrub`
+skrub
 =======
```
glemaitre, line 2:
```suggestion
skrub
=====
```

glemaitre (approved):
LGTM

README.rst
```diff
@@ -17,23 +17,47 @@
...
-
-`skrub` provides data assembling tools (``TableVectorizer``, ``fuzzy_join``...) and
-encoders (``GapEncoder``, ``MinHashEncoder``...) for **morphological similarities**,
-for which we usually identify three common cases: **similarities, typos and variations**
+Website: https://skrub-data.org/
+
+What can skrub do?
+------------------
+
+The goal of skrub is to facilitate building and deploying machine-learning
```
jeromedockes, line 31: I don't think it helps with "deploying"

README.rst
```diff
@@ -17,23 +17,47 @@
...
+>>> from skrub import tabular_learner
+>>> cross_val_score(tabular_learner('regressor'), df, y)
+array([0.89370447, 0.89279068, 0.92282557, 0.92319094, 0.92162666])
 
 See our `examples <https://skrub-data.org/stable/auto_examples>`_.
 
-What `skrub` cannot do
-~~~~~~~~~~~~~~~~~~~~~~
+What skrub cannot do
+~~~~~~~~~~~~~~~~~~~~
```
jeromedockes, line 60: that section seems more specific to stuff that used to be the focus of dirty_cat, in the context of skrub it seems a bit weird -- I wonder if we should drop the what skrub cannot do section

jeromedockes (approved):

README.rst
TheooJ: Ah you're right, I took this from the VISION.rst

I'll just keep the building part

README.rst
TheooJ: Another argument in favor of removing is that it's rare to see a readme explicitly mention what a package *cannot* do

@glemaitre wdyt ?

PR #1000, opened 2024-07-12
DOC reorder properly the numpydoc docstring

glemaitre:
This should solve the issue having the attributes appearing after the examples.


jeromedockes (approved):
PR #1000 :tada: !!

PR #1003, opened 2024-07-16
Set sphinx gallery default thumbnail?

jeromedockes:
the default sphinx gallery thumbnail for examples without a plot is faded out giving the impression that something went wrong or something is missing. But it's valid for an example not to create any matplotlib plots, for example example 0 getting started.

shall we consider replacing the default thumbnail with something else, like the skrub logo?


GaelVaroquaux:
Yes! I was thinking exactly this looking at the example gallery with the latest merge. Thanks heaps, merging!

GaelVaroquaux (approved):

PR #1004, opened 2024-07-16
DOC: Add version switcher

GaelVaroquaux:
Add a version switcher to easily go from dev to stable.
image

For now, the version.json file points to this PR, but we will need to change it to point to main branch on skrub


GaelVaroquaux:
This is ready to merge. If people are :+1:

I suggest that I do a second PR to have the json file point to its final version, as if we do it in one step, I fear that we'll have a build failure on main, because of a race condition

TheooJ:
Great idea ! I wanted people to see the new doc without needing to release yet

TheooJ (approved):
Thanks @GaelVaroquaux !

PR #1008, opened 2024-07-22
Remove pypi button

TheooJ:
Removing the pypi button with no logo

before:
Capture d’écran 2024-07-22 à 14 30 51

after:
Capture d’écran 2024-07-22 à 16 17 45


Vincent-Maladiere (approved):
Nice one

PR #1009, opened 2024-07-24
Discussion: apply column filter to similarity table in TableReport

jeromedockes:
As @lesteve  rightly pointed out, the fact that the column filter doesn't apply to the table of column similarities is surprising. This PR filters the similarity table: the row containing the similarity of column A and column B only appears if either A or B matches the column filter. 
There is also a special case for the "Columns with high similarity" filter: only rows corresponding to the high similarities appear

I also noted in the code there is a mix of "interactions", "associations" and "similarities" as our preferred term for this has evolved, I'll address that in a separate PR

### DIFF
--- a/skrub/_reporting/_data/templates/dataframe-columns.css
+++ b/skrub/_reporting/_data/templates/dataframe-columns.css
@@ -49,7 +49,7 @@
 
 /* Hide cards of columns that do not match the filter (eg "all()", "numeric()", ...) */
 
-.filterable-column[data-is-excluded-by-filter] {
+:is(.filterable-column, .filterable-column-pair)[data-is-excluded-by-filter] {
     display: none;
 }
 
--- a/skrub/_reporting/_data/templates/dataframe-interactions.html
+++ b/skrub/_reporting/_data/templates/dataframe-interactions.html
@@ -1,41 +1,58 @@
-<article class="flex-reverse wrapper wrapper-vert-l gap-l">
+<article>
     {% if summary["top_associations"] %}
+    <div class="if-else interactions-toggle"
+         data-predicate="true"
+         id="interactions-content"
+         data-similarity-threshold="{{ column_similarity_threshold }}" >
+        <div class="flex-reverse wrapper gap-l wrapper-vert-l">
+            <div class="horizontal-scroll">
+                <table class="pure-table">
+                    <thead>
+                        <tr>
+                            <th>Column 1</th>
+                            <th>Column 2</th>
+                            <th><a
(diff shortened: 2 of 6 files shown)


GaelVaroquaux:
I think that I would prefer graying out the selector on this tab than applying it. The reason is that I can see someone moving to the tab and not seeing columns that are highly similar

jeromedockes:
@GaelVaroquaux  is [this](https://output.circle-artifacts.com/output/job/e0bf48ba-4054-48a9-b929-b4bbb85eeced/artifacts/0/doc/auto_examples/00_getting_started.html#sphx-glr-auto-examples-00-getting-started-py) what you have in mind?

I'm not sure which I like better, from the user's point of view it's not obvious why the filter they chose would be deactivated when they visit this tab.

OTOH applying the filter isn't super useful, I don't really see a use case for "all similarities involving a numeric column" for example

>  not seeing columns that are highly similar

I see what you mean but arguably that's precisely the goal of the filter: not to see columns that don't match it. so it might make sense to have it behave consistently across all 3 tabs. Note the default is to show all columns and this only changes if the user manually goes and selects something else.


maybe yet another option could be to gray out the columns that don't match the filter. I'm not sure I like it either: it might not be immediately obvious why some cells are grayed out, and we lose the benefit of saving space by not showing them 

:thinking:

jeromedockes:
In the future we might add more tabs for which filtering columns doesn't make sense; that would be an argument for the "gray-out rather than apply" option.

jeromedockes:
also, instead of graying it out we could hide it altogether, that might hint better that it doesn't apply (rather than is disabled), and also if users don't see the dropdown they are less likely to notice it and wonder why it's disabled

jeromedockes:
> instead of graying it out we could hide it altogether

you can check #1011 to see what that option would look like

TheooJ:
I like the option of having the filter, it could be useful for (i) users that want the same columns they were seeing in the "Table preview" and "Column summaries" tabs when they go to "Column similarities" and (ii) users that have dataframes with many columns. If we're concerned about a user missing very similar columns, we could add a "beware: some very similar columns are not shown here" button

If we go with the other option of disabling this though, I agree with @lesteve that I'd rather have nothing than something greyed out

jeromedockes:
thanks for following up! ok let's merge the one that completely hides the dialog for now, and I'll leave this one open in case we want to discuss more / revisit when we add some more useful filters

jeromedockes:
it seems we're fine with the current status (hiding the select when in this tab) so i'll close this issue

PR #1010, opened 2024-07-24
disable column filter when column associations tab is selected

jeromedockes:
alternative to #1009 as suggested in [comment](https://github.com/skrub-data/skrub/pull/1009#issuecomment-2247015741)  (at least IIUC): disable the column filter menu when in the similarities tab


jeromedockes:
closing in favor of #1011

PR #1011, opened 2024-07-24
hide column filter selector when in associations tab

jeromedockes:
variant of #1010 : hide the select altogether instead of disabling it


TheooJ (approved):
Thanks @jeromedockes !

PR #1012, opened 2024-07-24
use browser default style for focus-visible

jeromedockes:
in the report we include the copy of normalize.css distributed with pure which provides some baseline default styling and makes stuff more similar across browsers.

one problem with it is it hides the browser's default highlighting of the element that has keyboard focus, this PR removes that rule


TheooJ (approved):
LGTM !

PR #1016, opened 2024-07-25
ENH polars support for InterpolationJoiner

### LINKED ISSUE #897: InterpolationJoiner - polars
### Describe the bug

Tried a simple join as follows:

joiner = InterpolationJoiner(
    data_store["depth_0"][0],
    key=["case_id"],
    suffix="_depth_0",
).fit(data_store["df_base"])
join = joiner.transform(data_store["df_base"])
join.head()

--

data_store["depth_0"][0] - polars Dataframe
data_store["df_base"] - polars Dataframe

--

### Steps/Code to Reproduce

joiner = InterpolationJoiner(
    data_store["depth_0"][0],
    key=["case_id"],
    suffix="_depth_0",
).fit(data_store["df_base"])
join = joiner.transform(data_store["df_base"])
join.head()

### Expected Results

Wanted to see the join, as in: https://skrub-data.org/stable/auto_examples/09_interpolation_join.html

### Actual Results

KeyError                                  Traceback (most recent call last)
File /opt/conda/lib/python3.10/site-packages/polars/_utils/deprecation.py:95, in deprecate_parameter_as_positional.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
     94 try:
---> 95     param_args = kwargs.pop(old_name)
     96 except KeyError:

KeyError: 'columns'

During handling of the above exception, another exception occurred:

Cell In[8], line 5
      1 joiner = InterpolationJoiner(
      2     data_store["depth_0"][0],
      3     key=["case_id"],
      4     suffix="_depth_0",
----> 5 ).fit(data_store["df_base"])
      6 join = joiner.transform(data_store["df_base"])
      7 join.head()

File /opt/conda/lib/python3.10/site-packages/skrub/_interpolation_joiner.py:225, in InterpolationJoiner.fit(***failed resolving arguments***)
    223     _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
    224 key_values = self.vectorizer_.fit_transform(self.aux_table[self._aux_key])
--> 225 estimators = self._get_estimator_assignments()
    226 fit_results = joblib.Parallel(self.n_jobs)(
    227     joblib.delayed(_fit)(
    228         key_values,
   (...)
    233     for assignment in estimators
    234 )
    235 fit_results = self._check_fit_results(fit_results)

File /opt/conda/lib/python3.10/site-packages/skrub/_interpolation_joiner.py:356, in InterpolationJoiner._get_estimator_assignments(self)
    339 def _get_estimator_assignments(self):
    340     """Identify column groups to be predicted together and assign them an estimator.
    341 
    342     In many cases, a single estimator cannot handle all the target columns.
   (...)
    354     separately to each column.
    355     """
--> 356     aux_table = self.aux_table.drop(self._aux_key, axis=1)
    357     assignments = []
    358     regression_table = aux_table.select_dtypes("number")

File /opt/conda/lib/python3.10/site-packages/polars/_utils/deprecation.py:97, in deprecate_parameter_as_positional.<locals>.decorate.<locals>.wrapper(*args, **kwargs)
     95     param_args = kwargs.pop(old_name)
     96 except KeyError:
---> 97     return function(*args, **kwargs)
     99 issue_deprecation_warning(
    100     f"named `{old_name}` param is deprecated; use positional `*args` instead.",
    101     version=version,
    102 )
    103 if not isinstance(param_args, Sequence) or isinstance(param_args, str):

TypeError: DataFrame.drop() got an unexpected keyword argument 'axis'

### Versions

```shell
'0.1.0'
```

TheooJ:
Closes #897 

```
>>> import polars as pl
>>> from sklearn.neighbors import KNeighborsRegressor
>>> from skrub import InterpolationJoiner

>>> buildings = pl.DataFrame(
...     {"latitude": [1.0, 2.0], "longitude": [1.0, 2.0], "n_stories": [3, 7]}
... )
>>> annual_avg_temp = pl.DataFrame(
...     {
...         "latitude": [1.2, 0.9, 1.4, 1.7, 5.0],
...         "longitude": [0.8, 1.1, 1.8, 1.8, 5.0],
...         "avg_temp": [10.0, 11.0, 15.0, 16.0, 20.0],
...     }
... )

>>> joiner = InterpolationJoiner(
...     annual_avg_temp,
...     key=["latitude", "longitude"],
...     regressor=KNeighborsRegressor(2),
... )
>>> joiner.fit_transform(buildings)

┌──────────┬───────────┬───────────┬──────────┐
│ latitude ┆ longitude ┆ n_stories ┆ avg_temp │
│ ---      ┆ ---       ┆ ---       ┆ ---      │
│ f64      ┆ f64       ┆ i64       ┆ f64      │
╞══════════╪═══════════╪═══════════╪══════════╡
│ 1.0      ┆ 1.0       ┆ 3         ┆ 10.5     │
│ 2.0      ┆ 2.0       ┆ 7         ┆ 15.5     │
└──────────┴───────────┴───────────┴──────────┘
```

As discussed with @jeromedockes, if the joined dataframe could contain duplicate column names, we add random suffixes and don't raise -- to be consistent with the behavior of the other joiners

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -301,7 +301,7 @@ def _all_null_like_pandas(col, length=None, dtype=None, name=None):
         dtype = col.dtype
     if name is None:
         name = col.name
-    return pd.Series(dtype=dtype, index=col.index, name=name).head(length)
+    return pd.Series(dtype=dtype, index=np.arange(length), name=name)
 
 
 @all_null_like.specialize("polars", argument_type="Column")
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -7,6 +7,8 @@ import pandas as pd
 import pandas.api.types
 from sklearn.utils.fixes import parse_version
 
+from skrub import _join_utils
+
 try:
     import polars as pl
 except ImportError:
@@ -321,11 +323,13 @@ def concat_horizontal(*dataframes):
 @concat_horizontal.specialize("pandas")
 def _concat_horizontal_pandas(*dataframes):
     dataframes = [df.reset_index(drop=True) for df in dataframes]
+    dataframes = _join_utils.make_column_names_unique(dataframes)
     return pd.concat(dataframes, axis=1, copy=False)
 
 
 @concat_horizontal.specialize("polars")
 def _concat_horizontal_polars(*dataframes):
+    dataframes = _join_utils.make_column_names_unique(dataframes)
     return pl.concat(dataframes, how="horizontal")
 
 
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -320,12 +320,36 @@ def concat_horizontal(*dataframes):
 
 @concat_horizontal.specialize("pandas")
 def _concat_horizontal_pandas(*dataframes):
+    parsed_names = set()
+    overlap = set()
+    for df in dataframes:
+        col_names = set(column_names(df))
+        overlap.update(parsed_names.intersection(col_names))
+        parsed_names.update(col_names)
+    if overlap:
+        raise ValueError(
+            "Can't concatenate dataframes containing duplicate column names."
+            f" The following column names are found in multiple tables: {overlap}."
+            " Please make sure column names do not overlap before concatenating them."
+        )
     dataframes = [df.reset_index(drop=True) for df in dataframes]
     return pd.concat(dataframes, axis=1, copy=False)
 
 
 @concat_horizontal.specialize("polars")
 def _concat_horizontal_polars(*dataframes):
+    parsed_names = set()
+    overlap = set()
+    for df in dataframes:
+        col_names = set(column_names(df))
+        overlap.update(parsed_names.intersection(col_names))
+        parsed_names.update(col_names)
+    if overlap:
+        raise ValueError(
+            "Can't concatenate dataframes containing duplicate column names."
+            f" The following column names are found in multiple tables: {overlap}."
+            " Please make sure column names do not overlap before concatenating them."
+        )
     return pl.concat(dataframes, how="horizontal")
 
 
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -323,13 +323,13 @@ def concat_horizontal(*dataframes):
 @concat_horizontal.specialize("pandas")
 def _concat_horizontal_pandas(*dataframes):
     dataframes = [df.reset_index(drop=True) for df in dataframes]
-    dataframes = _join_utils.make_column_names_unique(dataframes)
+    dataframes = _join_utils.make_column_names_unique(*dataframes)
     return pd.concat(dataframes, axis=1, copy=False)
 
 
 @concat_horizontal.specialize("polars")
 def _concat_horizontal_polars(*dataframes):
-    dataframes = _join_utils.make_column_names_unique(dataframes)
+    dataframes = _join_utils.make_column_names_unique(*dataframes)
     return pl.concat(dataframes, how="horizontal")
 
 
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -320,36 +320,12 @@ def concat_horizontal(*dataframes):
 
 @concat_horizontal.specialize("pandas")
 def _concat_horizontal_pandas(*dataframes):
-    parsed_names = set()
-    overlap = set()
-    for df in dataframes:
-        col_names = set(column_names(df))
-        overlap.update(parsed_names.intersection(col_names))
-        parsed_names.update(col_names)
-    if overlap:
-        raise ValueError(
-            "Can't concatenate dataframes containing duplicate column names."
-            f" The following column names are found in multiple tables: {overlap}."
-            " Please make sure column names do not overlap before concatenating them."
-        )
     dataframes = [df.reset_index(drop=True) for df in dataframes]
     return pd.concat(dataframes, axis=1, copy=False)
 
 
 @concat_horizontal.specialize("polars")
 def _concat_horizontal_polars(*dataframes):
-    parsed_names = set()
-    overlap = set()
-    for df in dataframes:
-        col_names = set(column_names(df))
-        overlap.update(parsed_names.intersection(col_names))
-        parsed_names.update(col_names)
-    if overlap:
-        raise ValueError(
-            "Can't concatenate dataframes containing duplicate column names."
-            f" The following column names are found in multiple tables: {overlap}."
-            " Please make sure column names do not overlap before concatenating them."
-        )
     return pl.concat(dataframes, how="horizontal")
 
 
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -322,9 +322,12 @@ def concat_horizontal(*dataframes):
 
 @concat_horizontal.specialize("pandas")
 def _concat_horizontal_pandas(*dataframes):
+    init_index = dataframes[0].index
     dataframes = [df.reset_index(drop=True) for df in dataframes]
     dataframes = _join_utils.make_column_names_unique(*dataframes)
-    return pd.concat(dataframes, axis=1, copy=False)
+    result = pd.concat(dataframes, axis=1, copy=False)
+    result.index = init_index
+    return result
 
 
 @concat_horizontal.specialize("polars")
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -301,7 +301,7 @@ def _all_null_like_pandas(col, length=None, dtype=None, name=None):
         dtype = col.dtype
     if name is None:
         name = col.name
-    return pd.Series(dtype=dtype, index=col.index, name=name)
+    return pd.Series(dtype=dtype, index=col.index, name=name).head(length)
 
 
 @all_null_like.specialize("polars", argument_type="Column")
--- a/skrub/_dataframe/tests/test_common.py
+++ b/skrub/_dataframe/tests/test_common.py
@@ -167,10 +167,20 @@ def test_all_null_like(df_module):
 
 
 def test_concat_horizontal(df_module, example_data_dict):
-    df1 = df_module.make_dataframe(example_data_dict)
+    df1 = df_module.example_dataframe
     df2 = ns.set_column_names(df1, list(map("{}1".format, ns.column_names(df1))))
     df = ns.concat_horizontal(df1, df2)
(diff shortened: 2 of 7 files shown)


skrub/_dataframe/tests/test_common.py
```diff
@@ -165,6 +165,9 @@ def test_all_null_like(df_module):
         expected = expected.astype("bool")
     df_module.assert_column_equal(ns.is_null(col), expected)
 
+    col = ns.all_null_like(df_module.example_column, length=2)
```
jeromedockes, line 168: could we also have a test where the required length is bigger than the input column?

skrub/_dataframe/_common.py
```diff
@@ -321,11 +323,13 @@ def concat_horizontal(*dataframes):
 @concat_horizontal.specialize("pandas")
 def _concat_horizontal_pandas(*dataframes):
     dataframes = [df.reset_index(drop=True) for df in dataframes]
+    dataframes = _join_utils.make_column_names_unique(*dataframes)
```
jeromedockes, line 326: can we add tests for the case of concatenating dataframes with identical names?

skrub/tests/test_interpolation_join.py
```diff
@@ -28,35 +30,34 @@ def weather():
...
-def test_interpolation_join(df_module, buildings, weather, key, with_nulls):
-    if df_module.name == "polars":
-        pytest.xfail(
-            reason=(
-                "In polars, DataFrame.drop() got an unexpected keyword argument 'axis'."
-            )
-        )
-    weather = weather.fillna(0.0)
+def test_simple_join(df_module, buildings, weather):
+    weather = df_module.DataFrame(weather)
```
jeromedockes, line 34: I think we can parametrize the fixtures themselves with `df_module` but i think you planned that for a future PR

skrub/tests/test_interpolation_join.py
```diff
@@ -76,117 +91,63 @@ def transform(self, X):
     assert_array_equal(ns.col(join, "B"), [0, 1])
 
 
-def test_no_multioutput(df_module, buildings, weather):
```
jeromedockes, line 79: why do we remove this test?

skrub/tests/test_interpolation_join.py
```diff
@@ -76,117 +91,63 @@ def transform(self, X):
...
+def test_duplicate_column(df_module):
     main = df_module.make_dataframe({"A": [0, 1, 2]})
-    aux = df_module.make_dataframe({"A": [0, 1, 2], "rB": [2, 0, 1], "C": [10, 11, 12]})
+    aux = df_module.make_dataframe({"A": [0, 1, 2], "C": [10, 11, 12]})
     join = InterpolationJoiner(
         aux, key="A", regressor=KNeighborsRegressor(1)
     ).fit_transform(main)
     assert_array_equal(ns.to_list(ns.col(join, "C")), [10, 11, 12])
 
-    join = InterpolationJoiner(
```
jeromedockes, line 110: here why is this check not necessary anymore?

skrub/tests/test_interpolation_join.py
```diff
@@ -76,117 +91,63 @@ def transform(self, X):
...
-            )
-        )
     df = df_module.make_dataframe({"A": [0, 1], "B": [0, 1]})
     join = InterpolationJoiner(
         df, key="A", suffix="_aux", regressor=KNeighborsRegressor(1)
     ).fit_transform(df)
     assert_array_equal(ns.column_names(join), ["A", "B", "B_aux"])
 
 
-def test_mismatched_indexes(df_module):
```
jeromedockes, line 145: why do we remove this test?

jeromedockes (commented):
a first pass

skrub/tests/test_interpolation_join.py
TheooJ: Yes I will do multiple follow-ups after merging this

skrub/tests/test_interpolation_join.py
TheooJ: The case where we join [0, 1] on [1, 0] is tested in `test_fit_on_none`. I can separate into two tests if necessary

skrub/_dataframe/tests/test_common.py
TheooJ: Certainly ! 

I've added the option to create null columns longer than the input column, by removing `index=col.index` for `index=np.arange(length)`. We don't keep the index -- bu we weren't using it anywhere -- and now it's possible to create longer columns. I think it makes sense, but let me know if you'd rather have a max length set to `len(col)`

skrub/tests/test_interpolation_join.py
TheooJ: The test wasn't checking for no multioutput, since we use HGBT by default and they are multioutput

I will add a specific test

skrub/tests/test_interpolation_join.py
TheooJ: I readded a test (for pandas only). Note that because of `concat_horizontal`, the initial index is reset. We could store it and reapply it in the case of pandas

It's possible to store the index of the first arg of `concat_horizontal` and set the output to have this index, WDYT ?

skrub/tests/test_interpolation_join.py
TheooJ: We were just fitting on two different column names ("A" and "rB"). Now the test checks that the double fitting returns unique column names. Other tests check fitting on left + right keys

jeromedockes:
@TheooJ  could you do `git commit --allow-empty -m '[doc build]'` and push? so that we can see the rendered example

skrub/_dataframe/_common.py
```diff
@@ -299,7 +301,7 @@ def _all_null_like_pandas(col, length=None, dtype=None, name=None):
         dtype = col.dtype
     if name is None:
         name = col.name
-    return pd.Series(dtype=dtype, index=col.index, name=name)
+    return pd.Series(dtype=dtype, index=np.arange(length), name=name)
```
jeromedockes, line 304: I think we could keep the original index when length is not specified. not sure if we care

skrub/_dataframe/tests/test_common.py
```diff
@@ -165,13 +166,28 @@ def test_all_null_like(df_module):
...
     df = ns.concat_horizontal(df1, df2)
     assert ns.column_names(df) == ns.column_names(df1) + ns.column_names(df2)
 
+    # Test concatenating dataframes with the same column names
+    df2 = df1
+    df = ns.concat_horizontal(df1, df2)
+    assert ns.shape(df) == (4, 16)
+    for n in ns.column_names(df)[8:]:
+        assert re.match(r".*__skrub_[0-9a-f]+__", n)
+
```
jeromedockes, line 190: I realize we don't have a test for concat_horizontal for pandas dataframes that have different indexes (of the same length). it's important that we reset the indexes in concat_horizontal because by default pandas will align them. In this or a later PR it would be nice to add a test

skrub/_interpolation_joiner.py
```diff
@@ -287,43 +291,46 @@ def transform(self, X):
...
+            main_table, prediction_results
+        )
         predictions = [res["predictions"] for res in prediction_results]
         predictions = [
             _join_utils.add_column_name_suffix(df, self.suffix) for df in predictions
         ]
-        for part in predictions:
-            part.index = main_table.index
-        return pd.concat([main_table] + predictions, axis=1)
+        return sbd.concat_horizontal(main_table, *predictions)
```
jeromedockes, line 315: now we have a different index than in the input for pandas dataframes. either here or in the `concat_horizontal` it might be good to preserve the index of the first dataframe, WDYT?

skrub/_interpolation_joiner.py
```diff
@@ -287,43 +291,46 @@ def transform(self, X):
...
-                    dtype=dtype,
-                )
+                _pred = [
+                    sbd.all_null_like(
+                        sbd.col(self.aux_table, col),
+                        length=res["shape"][0],
+                    )
+                    for col in res["columns"]
+                ]
+                _pred = [
```
jeromedockes, line 330: do we need to convert to dataframe and then concatenate? couldn't we just use `make_dataframe` and pass to it the list of columns in `_pred`?

also it is not usual to have local variable names starting with an underscore (except the special case of `_`)

skrub/_interpolation_joiner.py
```diff
@@ -396,13 +408,17 @@ def _handles_multioutput(estimator):
...
+    key_values = sbd.filter(key_values, ~discarded_rows)
+    Y = np.concatenate(
+        [sbd.to_numpy(col).reshape(-1, 1) for col in sbd.to_column_list(target_table)],
+        axis=1,
+    )[~discarded_rows]
 
     # Estimators that expect a single output issue a DataConversionWarning if
     # passing a column vector rather than a 1-D array
-    if len(target_table.columns) == 1:
+    if len(sbd.column_names(target_table)) == 1:
```
jeromedockes, line 421: maybe we could do sbd.shape(target_table)[1] instead?

skrub/_interpolation_joiner.py
```diff
@@ -426,10 +446,17 @@ def _predict(key_values, columns, estimator, propagate_exceptions):
     if failed:
         predictions = None
     else:
-        predictions = pd.DataFrame(data=Y_values, columns=columns)
+        # If more than one prediction, need to set predictions as columns
+        # in order to create the dataframe
+        if len(columns) > 1:
+            Y_values = np.swapaxes(Y_values, 0, 1)
```
jeromedockes, line 452: maybe just Y_values.T ? or am I missing something

skrub/_interpolation_joiner.py
```diff
@@ -426,10 +446,17 @@ def _predict(key_values, columns, estimator, propagate_exceptions):
...
+            data = dict(zip(columns, Y_values))
+        else:
+            data = {columns[0]: Y_values}
```
jeromedockes, line 455: maybe Y_values.ravel() here just to be sure

jeromedockes (commented):
thanks a lot @TheooJ I think it's almost ready

skrub/_interpolation_joiner.py
TheooJ: Yes good catch

skrub/_interpolation_joiner.py
TheooJ: I agree, let's do this in the `concat_horizontal` so that we have a consistent behavior if we use it elsewhere

jeromedockes (approved):
very cool, thanks @TheooJ !

skrub/_dataframe/_common.py
TheooJ: You're right, we could address this in another PR if it comes up

PR #1017, opened 2024-07-25
MAINT update pixi and lock files

glemaitre:
Update pixi from 0.24.0 to 0.26.1
At the same time, I'm updating the latest dependencies. I did not observe a segfault but we have some deprecation warnings.

### DIFF
--- a/.github/workflows/run-code-format-checks.yaml
+++ b/.github/workflows/run-code-format-checks.yaml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: prefix-dev/setup-pixi@v0.8.1
         with:
-          pixi-version: v0.24.0
+          pixi-version: v0.26.1
           frozen: true
 
       - name: Run tests
--- a/.github/workflows/test-javascript.yml
+++ b/.github/workflows/test-javascript.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: prefix-dev/setup-pixi@v0.8.1
         with:
-          pixi-version: v0.24.0
+          pixi-version: v0.26.1
           environments: ci-py312-latest-optional-deps
           # we can freeze the environment and manually bump the dependencies to the
           # latest version time to time.
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: prefix-dev/setup-pixi@v0.8.1
         with:
-          pixi-version: v0.24.0
+          pixi-version: v0.26.1
           environments: ${{ matrix.environment }}
           # we can freeze the environment and manually bump the dependencies to the
           # latest version time to time.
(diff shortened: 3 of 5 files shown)


jeromedockes:
```diff
diff --git a/skrub/_dataframe/_common.py b/skrub/_dataframe/_common.py
index f966fff4..81058c54 100644
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -774,6 +774,7 @@ def _to_datetime_pandas(col, format, strict=True):
 
 @to_datetime.specialize("polars", argument_type="Column")
 def _to_datetime_polars(col, format, strict=True):
+    import polars.exceptions
     if _is_any_date_polars(col):
         return col
     try:
@@ -781,7 +782,7 @@ def _to_datetime_polars(col, format, strict=True):
         # differently.
         format = format.replace(".%f", "%.f")
         return col.str.to_datetime(format=format, strict=strict)
-    except pl.ComputeError as e:
+    except (pl.exceptions.ComputeError, pl.exceptions.InvalidOperationError) as e:
         raise ValueError("Failed to convert to datetime") from e
 
 
diff --git a/skrub/_dataframe/tests/test_common.py b/skrub/_dataframe/tests/test_common.py
index 651bad73..e971a1d2 100644
--- a/skrub/_dataframe/tests/test_common.py
+++ b/skrub/_dataframe/tests/test_common.py
@@ -648,7 +648,7 @@ def test_drop_nulls(df_module):
 
 def test_fill_nulls(df_module):
     # Test on dataframe
-    df = df_module.make_dataframe({"col_1": [0, np.nan, 2], "col_2": [0, None, 2.0]})
+    df = df_module.make_dataframe({"col_1": [0.0, np.nan, 2.0], "col_2": [0.0, None, 2.0]})
     df_module.assert_frame_equal(
         ns.fill_nulls(df, -1),
         df_module.make_dataframe({"col_1": [0.0, -1, 2.0], "col_2": [0.0, -1.0, 2.0]}),
```

this fixes the tests

skrub/_dataframe/_common.py
```diff
@@ -781,7 +781,7 @@ def _to_datetime_polars(col, format, strict=True):
         # differently.
         format = format.replace(".%f", "%.f")
         return col.str.to_datetime(format=format, strict=strict)
-    except pl.ComputeError as e:
+    except (pl.exceptions.ComputeError, pl.exceptions.InvalidOperationError) as e:
```
jeromedockes, line 784: here I wonder if we should just catch Exception... there may be more and it is bound to change again in other polars releases

glemaitre: Exception is a bit large ;). But maybe this is better.

skrub/conftest.py
```diff
@@ -102,7 +102,7 @@ def _example_data_dict():
             "module": pl,
             "DataFrame": pl.DataFrame,
             "Column": pl.Series,
-            "make_dataframe": pl.from_dict,
+            "make_dataframe": lambda data: pl.from_dict(data, strict=False),
```
jeromedockes, line 105:
```suggestion
            "make_dataframe": pl.from_dict,
```

glemaitre: you need the same for `make_column`

jeromedockes (approved):
:sweat_smile:

PR #1023, opened 2024-07-31
Cleanup & misc improvements of TableReport

jeromedockes:
- add more tests
- rename column interactions -> associations everywhere
- rename some files, functions & classes to make it easier to find where each part of the report is implemented
- remove the checkbox to regroup constant columns (as it was not really necessary, I had added it to save space for datasets with many constant columns but that's unusual and now we have the filters and the tabbed interface so it's less than an issue. we could consider adding constant/not constant filters)
- add icons for the buttons & improve the copy buttons
- improve the tabbed interface (keyboard arrows navigation + better display of the tabs)
- allow wrapping of long column names in column summaries & ellipsis in tables
- some refactoring of the reports' (javascript) code
- print progress messages while generating the report


TheooJ (approved):
Thanks a lot @jeromedockes  !

PR #1024, opened 2024-07-31
Follow-up on DatatimeEncoder example

TheooJ:
Addresses minor comments left on #997


jeromedockes:
very nice! I think we just have a problem with the x labels in the plot

screenshot_2024-07-31T18:15:01+02:00

PR #1025, opened 2024-08-01
flush stdout after computing all column summaries

jeromedockes:
tiny detail: make sure the stream is flushed otherwise the last line may take longer to appear eg in ci logs


Vincent-Maladiere (approved):
Nice trick!

PR #1028, opened 2024-08-02
run linter + misc details

jeromedockes:


skrub/_reporting/_html.py
```diff
@@ -142,4 +143,6 @@ def to_html(summary, standalone=True, column_filters=None):
 
 
 def _b64_encode(obj):
-    return base64.b64encode(json.dumps(obj).encode("utf-8")).decode("utf-8")
+    return base64.b64encode(json.dumps(obj, ensure_ascii=True).encode("utf-8")).decode(
```
jeromedockes, line 146: `ensure_ascii=True` is actually the default, this is just specified explicitly here to make it more obious that the base64 once decoded will be ascii text and thus does not need text decoding before being passed to the json parser

TheooJ (approved):

PR #1030, opened 2024-08-02
Format InterpolationJoiner

TheooJ:
Follow-up after https://github.com/skrub-data/skrub/pull/1016

- Docstrings improvements
- Parametrize fixtures with `df_module`
- Relative imports in source, absolute imports in tests
- Some renaming and reordering of assignments


jeromedockes (approved):
thanks a lot!

PR #1036, opened 2024-08-05
Remove inexistant attributes from TableReport docstring

jeromedockes:
These are actually provided by methods now, but were never removed from the `Attributes` section of the docstring


TheooJ (approved):
LGTM

Issue #1037, opened 2024-08-05
Font size in TableReport


GaelVaroquaux:
### Describe the bug

Fonts are too big in some environments (bigger than the fonts 
![image](https://github.com/user-attachments/assets/592cdf67-3128-4092-b61b-bb483b3678c1)

![image](https://github.com/user-attachments/assets/6a9616f5-e618-4d03-ba27-a7ce0d785e74)

### Steps/Code to Reproduce

Use it in vscode or check the render documentation

### Expected Results

The size of the font should be the same or smaller than the size of the code

### Actual Results

See figures above

### Versions

```shell
dev
```

jeromedockes:
The sphinx documentation theme sets a smaller font size for preformatted text and code elements, that's why the code cell text is smaller. 

The skrub report uses the font size of the root element, that is the default chosen by the user agent or by the user in their configuration. (That is also the size of the non-code text and non-preformatted outputs in the documentation)

I think it might be a good idea to use the font size of the parent element instead (in this case the cell output), to allow the page containing the report to control its font size easily. however that wouldn't change the font in the sphinx gallery, because the default size of cell outputs is the current one.

we can also force a smaller font size in all cases, but then some users might have difficulty reading it

jeromedockes:
I think it might be a question of the font itself rather than the font size because in the documentation the text in the report appears bigger than the main documentation text, even though it is set at the same size, because it is a different font

jeromedockes:
I'll open open a pr to inherit the font & font-size when embedding the report.

Setting systematically a small font size is more up to debate IMO, let's first see how it looks after the first changes are done

jeromedockes:
fixed in #1038

PR #1038, opened 2024-08-05
Fonts in inline report

jeromedockes:
partly addresses #1037 

- [x] don't set the font when the report is inserted in another page
- [x] reduce monspaced text size
- [x] reduce banner font size

can be discussed in another pr:

- [ ] scale report according to parent element's font size
- [ ] reduce the font size further

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,9 @@ Major changes
 
 Minor changes
 -------------
+* The reports created by :class:`TableReport`, when inserted in an html page (or
+  displayed in a notebook), now use the same font as the surrounding page.
+  :pr:`1038` by :user:`Jérôme Dockès <jeromedockes>`.
 
 
 Release 0.3.0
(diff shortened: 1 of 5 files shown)


jeromedockes:
here is a screenshot in vscode

screenshot_2024-08-05T19:51:19+02:00

GaelVaroquaux (approved):
That's already a great improvement.

I was about to merge, but maybe a changelog entry? I suspect that this is a change that people want to know about

jeromedockes:
thanks! I added the changelog entry

jeromedockes:
now in vscode and the sphinx documentation the reports use the same font at the same size as the rest of the page.

in a jupyter notebook, by default in most browsers the text will be 15px and the report will be 16px.

if a user changes the default font size in their browser settings (eg to accomodate their eyesight), the difference will be bigger because the report will scale accordingly and the notebook text won't and will stay fixed at 15px. But I would argue the correct behavior is the report's in this case

GaelVaroquaux:
This is a net improvement. Merging. Thanks!!

PR #1043, opened 2024-08-29
MAINT update version pixi and lock file

glemaitre:
It looks like the workflow failed in the past:

https://github.com/skrub-data/skrub/actions/runs/10553902445/job/29234929192


glemaitre:
I could not reproduce the issue locally. Let's try to update locally and update the lock files and check if the CIs work.

We can then try to debug the workflow.

jeromedockes (approved):
thanks @glemaitre

PR #1049, opened 2024-09-02
DOC fix backreferences of doc gallery in API references

glemaitre:
This PR should fix the issue in the API references pages where we don't see the link to the example gallery.

The following changes are done:

- make sure that examples are generated in `reference/generated`
- update the `conf.py` with the right link
- update `pyproject.toml` to make sure that we do a proper cleaning when calling `pixi run clean-doc` 

So now, we have for sure this section in the API reference:

image


jeromedockes (approved):
awesome, thanks a log @glemaitre !

PR #1051, opened 2024-09-02
FEA add link for HTML representation

glemaitre:
This PR add the necessary class and function to add a link to the documentation when showing the HTML representation of an estimator. Unfortunately, there is a bug in scikit-learn that will be corrected in 1.6 and we need to vendor the fix and inherit from the mixin.


jeromedockes:
if the bug is in scikit-learn and this is not critical, could we just wait for it to be fix in scikit-learn without vendoring the htmlmixin class? also once it is fixed in scikit-learn would we still need to inherit from a private scikit-learn class?

glemaitre:
> could we just wait for it to be fix in scikit-learn without vendoring the htmlmixin class?

Yes but we need to bump the minimum version

also once it is fixed in scikit-learn would we still need to inherit from a private scikit-learn class?

Actually, we don't need to. We only need to set the class attributes. `BaseEstimator` is inheriting from this private mixin.

Vincent-Maladiere:
> Yes but we need to bump the minimum version

This is somewhat unrelated to this PR, but we were thinking with @jeromedockes whether or not we should try to support as many Python and scikit-learn versions as possible, for many potential users are often stuck with an old Python version (e.g. APHP). We can talk about this idea somewhere else, though.

skrub/_repr.py
```diff
@@ -0,0 +1,103 @@
+import itertools
```
jeromedockes, line 1: could we name this module something more explicit like "_sklearn_html_repr_utils" or similar?

skrub/_repr.py
```diff
@@ -0,0 +1,103 @@
...
+import sklearn
+from sklearn.utils.fixes import parse_version
+
+sklearn_version = parse_version(sklearn.__version__)
+
+# TODO: remove when scikit-learn 1.6 is the minimum supported version
+# TODO: subsequently, we should remove the inheritance from _HTMLDocumentationLinkMixin
+# for each estimator then.
+if sklearn_version > parse_version("1.6"):
+    from sklearn.utils._estimator_html_repr import _HTMLDocumentationLinkMixin
```
jeromedockes, line 12: IIUC we will now have 2 of those in every skrub estimator's parents, one directly and one through the BaseEstimator?

glemaitre:
@Vincent-Maladiere I share the same feeling and this is the reason why to vendor for the moment as a fix.

jeromedockes (commented):
I find the fix a bit too intrusive for having early access to a scikit-learn feature that doesn't affect skrub itself, but OTOH I completely agree the functionality is useful :thinking:

jeromedockes:
@glemaitre  could you do a 
```
git commit --allow-empty -m '[doc build]' && git push
```

so we can see the new TableVectorizer display in the examples? thanks!

skrub/_repr.py
```diff
@@ -101,3 +101,9 @@ def doc_link_url_param_generator(estimator):
         "estimator_module": estimator_module,
         "estimator_name": estimator_name,
     }
+
+
+class _SkrubHTMLDocumentationLinkMixin(_HTMLDocumentationLinkMixin):
```
jeromedockes, line 106: even when the scikit-learn fix is released, we will most likely want this mixin to avoid repeating these 3 attributes everywhere. so in the end all we will have to remove is the `HTMLDocumentationLinkMixin` which is confined to this module.

so after all I am +1 for this PR

Vincent-Maladiere:
@glemaitre ok, I misread your first message. I'm +0 or +1 as long as we don't need to bump the minimal version.

rcap107:
What is the status of this PR?

Vincent-Maladiere:
@glemaitre I can take over this one if you prefer

rcap107:
Is this still relevant @glemaitre?

jeromedockes:
fixed in #2036

Issue #1046, opened 2024-08-30
Add example with the e-commerce fraud detection dataset


jeromedockes:
ping @Vincent-Maladiere

PR #1053, opened 2024-09-03
DOC Replace the MovieLens dataset with the fraud dataset

Vincent-Maladiere:
Addresses https://github.com/skrub-data/skrub/issues/1046

This PR proposes replacing the MovieLens dataset used in the AggJoiner example with the fraud dataset from Figshare. It highlights the performance gained when using AggJoiner for a modeling problem involving tables with one-to-many relationships. The previous problem of self-join was somewhat niche and not representative of the majority of the AggJoiner use cases.

It also discusses the current limitations of both Joiner and AggJoiner when it comes to deploying models.


examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+# To consider the problem from a business perspective, we define our utility function
+# by the cost matrix in the function ``credit_gain_score``. False positive and false
+# negative predictions incur a negative gain.
 #
-# Our first step is to extract features from the timestamp, using the
-# |TableVectorizer|. Natively, it uses the |DatetimeEncoder| on datetime
-# columns, and doesn't interact with numerical columns.
-from skrub import DatetimeEncoder, TableVectorizer
+# Ultimately, we want to maximize this metric. To do so, we can train our learner to
+# minimize a proper scoring rule like the log loss.
```
jeromedockes, line 108: actually if the goal of the example is to show how to use the aggjoiner I wonder if the business metric / proper scoring rule stuff does not distract from that? it's definitely interesting in its own right but it complicates a bit the example (we need to define more functions, use the metadata routing and what not) and does not yield a better understanding of the joiner AFAICT

jeromedockes: but it does make for a more realistic example, and we need some of those to check that skrub has enough flexibility for real-world scenarios and not just toy examples :thinking:

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
-###############################################################################
-# We also display the distribution of our target ``y``.
-rating_count = y.value_counts().sort_index()
 
-make_barplot(
-    x=rating_count.index,
-    y=rating_count.values,
-    title="Distribution of ratings given to movies",
-)
+sklearn.set_config(enable_metadata_routing=True)
```
jeromedockes, line 148: not necessarily important but I think this config will persist for subsequent examples

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+from sklearn.preprocessing import TargetEncoder
+
+from skrub import TableVectorizer
 
-pipeline = make_pipeline(
-    table_vectorizer,
-    agg_target_user,
-    agg_target_movie,
-    HistGradientBoostingRegressor(learning_rate=0.1, max_depth=4, max_iter=40),
+X_train_, X_val, y_train_, y_val = train_test_split(
```
jeromedockes, line 209: maybe a comment to highlight we are further splitting the training data here

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+            color=palette[name],
+            n_bins=15,
+        )
+    ax.set_xlim([-0.001, 0.13])
+    ax.set_ylim([-0.001, 0.13])
+    ax.set_title("Calibration curve")
+
+
+# %%
+# We see below that the low effort classifier significantly improves our gains compared
```
jeromedockes, line 324: I would maybe remind the reader that there aren't that many frauds and we show the total gain so even if we completely ignore frauds we still have a high positive gain. otherwise we may think that there is a difference but in proportion it is not huge

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+plot_calibration_curve(results)
+
+
+# %%
+# Agg-Joiner based estimator
+# --------------------------
 #
-# Note that you need the name of the pipeline elements to assign them
-# hyper-parameters search.
+# We first need to split the dataframe between a dataframe representing baskets and a
```
jeromedockes, line 343: maybe more of a transition here. it seemed like we were done with our analysis and now we jump with "we first need" and it's not clear why.

but to make the example shorter I think we could split the tables in the fetcher and start the example here

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+# %%
+# Let's now detail how to leverage |AggJoiner| here. We have just encoded each product
+# attributes, and now we need to somehow aggregate these product encodings into their
+# respective baskets.
+#
+# By aggregating instead of concatenating, we obtain an invariant number of columns,
+# and we remove the sparsity of the dataset.
+#
+# But which aggregation operation should we choose? Since the |MinHashEncoder| hashes
+# ngrams with different hashing functions and return their minimum, it makes sense to
```
jeromedockes, line 451: this part about the minhash is probably hard to follow for someone who isn't very familiar with the minhash and skrub jargon

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
-for train_idx, test_idx in TimeSeriesSplit(n_splits=10).split(X, y):
-    all_baseline_r2.append(baseline_r2(X, y, train_idx, test_idx))
+from skrub import AggJoiner
+from skrub import _selectors as s
+
+minhash_cols = "ID" | s.glob("item_*") | s.glob("model_*") | s.glob("make_*")
+single_cols = ["ID", "goods_code", "Nbr_of_prod_purchas", "cash_price"]
+
+pipe_agg_joiner = make_pipeline(
+    AggJoiner(
```
jeromedockes, line 479: I like this approach of having several aggjoiners, each doing a uniform aggregation but on a subset of columns. it is simpler than having a columntransformery approach where we specify several transformations in the aggjoiner, and it will become even simpler with the recipe. we might even think if this could allow simplifying the aggjoiner's logic with the numeric vs categorical columns

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+
+TableReport(basket_train_transformed)
+
+# %%
+# Now that we get a sense of how the |AggJoiner| can help us, we complete this pipeline
+# with a |HGBC| and evaluate our final model.
+
+tic = time()
+agg_join_estimator = make_pipeline(
+    pipe_agg_joiner,
```
jeromedockes, line 500: maybe we could append to the pipeline instead of wrapping in a new one? or we can keep it like this now and it will just become a `recipe.add()` later (where later is before the next release anyway)

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+plot_calibration_curve(results)
 
-###############################################################################
-# The naive estimator has a lower performance than our pipeline, which means
-# that our extracted features brought some predictive power.
+# %%
+# Conclusion
+# ----------
+#
+# Many problems involve tables where IDs have a one-to-many relationship. To simplify
```
jeromedockes, line 532: I think this could go towards the beginning to motivate the example

jeromedockes (commented):
thanks @Vincent-Maladiere , this is a really cool example! it showcases several important features while offering a more realistic scenario than the previous one.

the drawback of such a rich example is that it is very complex and thus can be a bit scary for users. Indeed it touches on several fine and somewhat orthogonal points:

- the representation of the input data, normalized vs sparse/concatenated
- using a business metric and the tuned threshold (which in turn triggers the introduction of the metadata routing)
- appropriate encoding of the different columns (some need target encoding, other minhashing)
- what minhash really does in order to understand why the appropriate aggregation function is min()
- how to use the aggjoiner

all of these are definitely interesting and at the center of skrub's focus, but it seems like a lot.

I wonder if we should have a simple version, which starts from the 2-table dataset (ie just the last section), and uses the AUC while stating why it is suboptimal, and a complete version in an examples "advanced" section. WDYT?

examples/08_join_aggregation.py
jeromedockes: although it is interesting to see that the "correct" approach using the aggjoiner is faster _and_ more effective

Vincent-Maladiere:
Thanks for the in-depth review @jeromedockes.

To answer your main concern, there is indeed a tradeoff between using an example that is not 100% trivial, which supports the case for a more realistic example gallery, and having a longer size or slightly higher complexity. In our IRL conversation, @glemaitre seemed to lean towards something more realistic. I agree that this could be simplified further, though. We need to decide where to put the cursor.

examples/08_join_aggregation.py
Vincent-Maladiere: Yes, demos like this are a good candidate for the recipe

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
 
-###############################################################################
-# We can now make a couple of plots and gain some insight on our dataset.
-import seaborn as sns
-from matplotlib import pyplot as plt
+def credit_gain_score(y_true, y_pred, amount):
+    """Define our utility function.
 
-sns.set_style("darkgrid")
+    These numbers are entirely made-up, don't try this at home!
```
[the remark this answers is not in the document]
Vincent-Maladiere: haha yeah, that's better

examples/08_join_aggregation.py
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, good idea. Alternatively, I guess scikit-learn has an entry for proper scoring rules somewhere in the user guide

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,548 @@
...
+- a category (marked by the column ``"item"``),
+- a model (``"model"``),
+- a brand (``"make"``),
+- a merchant code (``"goods_code"``),
+- a price per unit (``"cash_price"``),
+- a quantity selected in the basket (``"Nbr_of_prod_purchas"``)
+
+Since the number of products in each basket varies, the creators of this dataset have
+chosen to join all products and their attributes with their respective basket. They have
+arbitrarily decided to cut off the basket at the 24th product. However, since most
```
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, I was thinking about iterating on this. That would make more sense. If we want to keep the long version, we could still reproduce the naive joining method creating a lot of NaNs, before showcasing the benefits of the AggJoiner.

Vincent-Maladiere:
But I agree with you that we could add another, simpler example for the AggJoiner while using this one to build a more realistic use case.

What do you all think?

GaelVaroquaux:
I think that Jerome's suggestion to have two examples is a good one. It's useful to have an example that is not too scary

Vincent-Maladiere:
I simplified the example and put the previous long one in the FIXME folder. We can adapt it for an "advanced" example section later.

Vincent-Maladiere:
I need to add a few tests for the new dataset fetching function.

Also, the notebook takes 2min30 to run on the CI (the randomized search cv being the culprit). When I remove it, the model's performance gets worryingly close to the dummy model's. Should I hardcode the good hyper-parameters in the model instead, with a comment saying we found them using RandomizedSearchCV?

jeromedockes:
maybe we can keep the hyperparam search and shave some time off another example such as the "multiple key join" one :)

jeromedockes:
I saw this error before but I don't understand because even if pyarrow is an optional dependency it is supposed to be listed in the env of the doc build...

jeromedockes:
look for some keywords to nerd-snipe @glemaitre such as  #pixi #ci

jeromedockes:
ah right sorry I missed the fact this isn't for the example but the test. I think pyarrow as a mandatory dependency might scare off some potential users because it used to be a very heavy one with lots of dependencies of its own. I believe it has improved in that respect but still I would give it a little time

maybe we could have something like you must have either polars or pandas+pyarrow

but in the meanwhile I think your suggestion of storing the datasets as csv is a good one. some of them such as the airplanes may need to be subsampled

jeromedockes:
But maybe someone who was there when the first datasets were uploaded such as @GaelVaroquaux  or @jovan-stojanovic would have a better idea or more insights on this choice

Vincent-Maladiere:
Although I updated the pixi lock and added optional to test, I still have the missing pyarrow error 🤔
@glemaitre WDYT?

pyproject.toml
```diff
@@ -54,6 +54,7 @@ doc = [
     "sphinx-gallery",
     "sphinxext-opengraph",
     "statsmodels",
+    "pyarrow",
```
glemaitre, line 57: `project.optional-dependencies` represent the `pip` as in `requirements.txt`. Basically, you would need to add it also in the `tool.pixi.feature.doc.dependencies` to have it for `pixi` as well.

But at the end, the documentation `doc` environment uses default + doc + optional so it should include `pyarrow` already.

glemaitre:
> Although I updated the pixi lock and added optional to test, I still have the missing pyarrow error 🤔
@glemaitre WDYT?

I think that there is something wrong with the test CI: it also test the documentation. So when we have the min-dependencies without the optional dependencies because we run the example that requires more tests it fails.

So we should not be testing the documentation as part of the "test" builds but only for the ci dedicated to the documentation that relies on the `doc` environment containing as well the optional dependencies.

skrub/datasets/tests/test_fetching.py
```diff
@@ -219,3 +220,39 @@ def test_fetch_movielens():
             )
             mock_urlretrieve.assert_not_called()
             assert disk_loaded_info == returned_info
+
+
+def test_fetch_credit_fraud():
```
glemaitre, line 225: Actually, I'm wrong.

Here, you are testing something that request `pyarrow` but this is an optional one. So you should add:

```
pytest.importorskip("pyarrow")
```

to skip the test for when we don't have `pyarrow` installed.

Vincent-Maladiere: Thanks @glemaitre !

doc/conf.py
```diff
@@ -482,6 +482,7 @@ def notebook_modification_function(notebook_content, notebook_filename):
     "ColumnTransformer": "sklearn.compose.ColumnTransformer",
     "OneHotEncoder": "sklearn.preprocessing.OneHotEncoder",
     "Pipeline": "sklearn.pipeline.Pipeline",
+    "make_pipeline": "sklearn.pipeline.make_pipeline",
```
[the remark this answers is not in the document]
Vincent-Maladiere: Ok great, I can remove this line and open an issue to clean this in another PR? WDYT?

Vincent-Maladiere:
@jeromedockes Finally I kept the files as parquet, we can address that issue later.

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+of variable size. Each basket is linked to many products, e.g. basket 1 can have
+product A, B and C.
 
-The benefit of using  |AggJoiner| and |AggTarget| is that they readily
-provide a full pipeline, from the original tables to the prediction, that can
-be cross-validated or applied to new data to serve prediction. At the end of
-this example, we showcase hyper-parameter optimization on the whole pipeline.
+Our aim is to predict which baskets are fraudulent.
 
+The products dataframe can be joined on the basket dataframe using the ``basket_ID``
```
TheooJ, line 20:
```suggestion
The products dataframe can be joined on the baskets dataframe using the ``basket_ID``
```

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+# Let's explore a naive solution first.
+#
+# The first idea that comes to mind to merge these two tables is to aggregate the
+# products attributes into lists, using their basket IDs.
+products_grouped = products.groupby("basket_ID").agg(list)
+products_grouped
+
+# %%
+# Then, we can expand all lists into columns, as if we were "flattening" the dataframe.
+# We end up with a product dataframe ready to be joined on the basket dataframe, using
```
TheooJ, line 90:
```suggestion
# We end up with a products dataframe ready to be joined on the baskets dataframe, using
```

I would keep it plural when referencing the dataframes

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+# basket IDs as join keys.
```
TheooJ, line 91:
```suggestion
# ``"basket_ID"`` as the join key.
```

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
 
-In this notebook, we only deal with the main table "ratings".
-Our objective is **not to achieve state-of-the-art performance** on
-the explicit regression task, but rather to illustrate how to perform
-feature engineering in a simple way using |AggJoiner| and |AggTarget|.
-Note that our performance is higher than the baseline of using the mean
-rating per movies.
+The data consists of two distinct entities: e-commerce "baskets", and "products".
+Baskets can be tagged fraudulent (1) or not (0), and are essentially a list of products
+of variable size. Each basket is linked to many products, e.g. basket 1 can have
```
TheooJ, line 15:
```suggestion
of variable size. Each basket can be linked to many products, e.g. basket 1 can have
```

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+# created one column for each product attribute, for each position (up to 24) in
+# the dataframe.
 #
-# To solve this, the |AggTarget| transformer allows you to
-# aggregate the target ``y`` before joining it on the main table, without
-# risk of leaking. Note that to perform aggregation then joining on the features
-# ``X``, you need to use |AggJoiner| instead.
+# Moreover, if we wanted to replace text columns with encodings, we would create
+# :math:`d \times 24 \times 2` columns (encoding of dimensionality :math:`d`, for
+# 24 products, for the "item" and "make" columns), which would explode the memory usage.
```
TheooJ, line 114:
```suggestion
# 24 products, for the ``"item"`` and ``"make"`` columns), which would explode the memory usage.
```

I don't know if we should add this here and below. Pretty minor and fine with it either way

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+    ],
+)
+products_transformed = vectorizer.fit_transform(products)
+TableReport(products_transformed)
 
-###############################################################################
-# Hyper-parameters tuning and cross validation
-# --------------------------------------------
+# %%
+# Our objective is now to aggregate this dataframe by basket ID, then merging it
```
TheooJ, line 152:
```suggestion
# Our objective is now to aggregate this dataframe by ``"basket_ID"``, then merging it
```

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+# limitations of this design in the conclusion at the bottom of this notebook.
 #
-# Note that you need the name of the pipeline elements to assign them
-# hyper-parameters search.
+# The minimum is the most appropriate operation to aggregate encodings from
+# MinHashEncoder, for reasons that are out of the scope of this notebook.
 #
-# You can lookup the name of the pipeline elements by doing:
-list(pipeline.named_steps)
+# Notice that the ``basket_ID`` column is dropped from the joined table.
```
TheooJ, line 166:
```suggestion
# Notice that the ``"basket_ID"`` column is dropped from the joined table.
```

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+)
+products_transformed = vectorizer.fit_transform(products)
+TableReport(products_transformed)
 
-###############################################################################
-# Hyper-parameters tuning and cross validation
-# --------------------------------------------
+# %%
+# Our objective is now to aggregate this dataframe by basket ID, then merging it
+# on the basket dataframe, still on the basket ID.
```
TheooJ, line 153:
```suggestion
# on the baskets dataframe, still on the ``"basket_ID"``.
```

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+#
+# AggJoiner can help us achieve exactly this. We need to pass the product dataframe as
+# an auxiliary table argument to AggJoiner in ``__init__``. ``aux_key`` represent both
```
TheooJ, line 156: I'm not sure I'd mention ``__init__``, do we expect people to know this ? Completely fine with leaving it if we do

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+# the columns used to groupby on, and the columns used to join on.
 #
-# We can finally create our hyper-parameter search space, and use a
-# |GridSearchCV|. We select the cross validation splitter to be
-# the |TimeSeriesSplit| to prevent leakage, since our data are timestamped
-# logs.
+# The basket dataframe is our main table, and we indicate the columns to join with
```
TheooJ, line 159:
```suggestion
# The basket dataframe is our main table, and we indicate the columns to join on with
```

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+# The minimum is the most appropriate operation to aggregate encodings from
+# MinHashEncoder, for reasons that are out of the scope of this notebook.
 #
-# You can lookup the name of the pipeline elements by doing:
-list(pipeline.named_steps)
+# Notice that the ``basket_ID`` column is dropped from the joined table.
+from skrub import AggJoiner
+from skrub import _selectors as s
+
+minhash_cols_query = s.glob("item*") | s.glob("model*")
```
TheooJ, line 170: Maybe we should use a search of columns by hand if people are not familiar with selectors

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
-    movie_avg_rating = X_train.groupby("movieId")["y"].mean().to_frame().reset_index()
-
-    y_pred = X_test.merge(movie_avg_rating, on="movieId", how="left")["y"]
-    y_pred = y_pred.fillna(y_pred.mean())
-
-    return r2_score(y_true=y_test, y_pred=y_pred)
+# For the second AggJoiner, we use the mean, standard deviation, minimum and maximum
+# operations to extract a representative summary of each distribution.
+#
+# DropCols is another sklearn transformer which removes the "ID" column, which doesn't
```
TheooJ, line 193: I would link this + some of the other transformers above such as MinHashEncoder

examples/08_join_aggregation.py
```diff
@@ -1,302 +1,311 @@
...
+# The best hyper parameters are:
+
+search.best_params_
+
+# %%
+# To benchmark our performance, we plot the log loss of our model on the test set
+# against the log loss of a dummy model that always output the observed probability of
+# the two classes.
+#
+# As this dataset is extremely imbalanced, this dummy should be a good baseline.
```
TheooJ, line 255:
```suggestion
# As this dataset is extremely imbalanced, this dummy model should be a good baseline.
```

TheooJ (commented):
I have a couple of comments (mainly on typos and layout). I think the example is a bit complicated, and if I were a user trying to understand how it works "quickly" I could be a bit discouraged seeing this. I was wondering if we should use a full numeric dataset for the AggJoiner example (I'm on the lookout for such a dataset ATM), and use MinHash for the fraud, let me know what you think.

examples/08_join_aggregation.py
jeromedockes: that's true but on the other hand it will add a bunch of annoying / distracting boilerplate. maybe we can add a comment saying what it does. 

it will get nicer if the aggjoiner accepts a selector directly for the `cols=`; this will remove the need to do the s.select ourselves

examples/08_join_aggregation.py
jeromedockes, line 193:
```suggestion
# DropCols is another skrub transformer which removes the "ID" column, which doesn't
```

examples/08_join_aggregation.py
Vincent-Maladiere: "is linked to at least one" if we want to be thorough

examples/08_join_aggregation.py
Vincent-Maladiere: +1 for a clear comment

TheooJ:
I agree it's much nicer than the example on the main branch, let's merge it now that the smaller changes are addressed

TheooJ (approved):

Vincent-Maladiere:
Ok, TODO before merging:
- [ ] A comment on the selector
- [ ] Linking entities like DropCols
- [ ] Add a paragraph to skip the naive approach
- [ ] Update pixi.lock

Vincent-Maladiere:
I'm eager to try adding some diagrams to make the example more telling

Vincent-Maladiere:
WDYT?


## Data schema

Screenshot 2024-09-30 at 16 48 46

## AggJoiner

Screenshot 2024-09-30 at 16 50 52

jeromedockes:
I like the idea of having a schema! actually we may want to have it in the reference doc or user guide as well, and to have similar ones for some other joiners.

I have some small feedback about the diagram itself but it might be easier to do if we can wave and point at stuff, maybe we can talk about it on video call or IRL?

TheooJ:
I also like the idea ! Also +1 for having it somewhere in the user guide or on a specific page

For some inspiration : https://pandas.pydata.org/docs/user_guide/reshaping.html

Vincent-Maladiere:
Update

## Data schema

Screenshot 2024-10-01 at 10 11 24

## AggJoiner

Screenshot 2024-10-01 at 10 11 43

jeromedockes:
> Update

nice! LGTM

jeromedockes:
great, merging! very nice addition :)

PR #1054, opened 2024-09-03
Pixi dev env

glemaitre:
Adding the following libraries in the `dev` environment:

- `ipython`
- `ipykernel`
- `jupyterlab`

There are pretty handy stuff for interactive development. Ideally, `pixi` could one deal with a user-defined dev environment without relying on the project config.


jeromedockes (approved):
thanks a lot @glemaitre !!

jeromedockes:
> There are pretty handy stuff for interactive development. Ideally, pixi could one deal with a user-defined dev environment without relying on the project config.

I agree; IIUC that's not the case at the moment right?

glemaitre:
> I agree; IIUC that's not the case at the moment right?

Yep this is not.

PR #1056, opened 2024-09-04
Add summary stats panel in report

jeromedockes:


GaelVaroquaux:
Can you please do a trivial change (eg add an empty line) to the getting started example, so that our CI renders it, and we can check how the table report looks like via the rendered docs?

Thanks!

jeromedockes:
> Can you please do a trivial change (eg add an empty line) to the getting started example, so that our CI renders it, and we can check how the table report looks like via the rendered docs?
>
> Thanks!

yep, I did it on the previous one but forgot to redo it after the last
change.

FYI git commit has the `--allow-empty` option in cases where you want to
push a commit to trigger the CI but don't have any change to make

GaelVaroquaux:
> FYI git commit has the `--allow-empty` option in cases where you want to push a commit to trigger the CI but don't have any change to make

Indeed. But I think our CI detects that the example has not been changed and does not run, to decrease computation costs.

jeromedockes:
> Indeed. But I think our CI detects that the example has not been changed and does not run, to decrease computation costs.

ah yes that's a good point, [ci-build] runs all the examples

GaelVaroquaux:
> ah yes that's a good point, [ci-build] runs all the examples

But only once. And what we would like is that this specific example is rerun for each push, to facilitate looking at the TableReport.

Vincent-Maladiere (approved):
Very useful, LGTM!

GaelVaroquaux (approved):
Excellent!

This looks good to merge to me

GaelVaroquaux:
Two approvals, merging

PR #1057, opened 2024-09-05
Better distinguish focus & selection in sample table

jeromedockes:
ATM the selected table cell is outlined in black. that's ok in firefox but that is the color used by chrome to draw the focus ring, so there it makes it look like the selected cell has the keyboard focus even when it doesn't.

This PR makes the outline dotted when the focus is outside of the table and makes sure the cell has the default focus ring when it is in it

### DIFF
--- a/skrub/_reporting/_data/templates/dataframe-sample.css
+++ b/skrub/_reporting/_data/templates/dataframe-sample.css
@@ -22,12 +22,19 @@ tr:nth-child(2n -1) .table-cell[data-is-in-active-column] {
 
 
 .table-cell[data-is-active] {
-    outline: 2px solid black;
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
In stable we have the selected cell in yellow. We decided to drop that?

jeromedockes:
> In stable we have the selected cell in yellow. We decided to drop that?

yes @Vincent-Maladiere suggested using an outline instead to be more similar to spreadsheets (google, excel, libreoffice). I am fine with either the background color change or the outline, or even both. 

In any case I'll merge this PR because the yellow background has already been removed in main and this PR is more about showing keyboard focus, and I can do another small pr to restore the background if we decide we want it after all

PR #1058, opened 2024-09-05
Remove copy boxes selects

jeromedockes:
the select (dropdown) that changes the content of the bar above the table and of the frequent element representations are removed. now we always display the str of the value, and put the repr in the clipboard when the copy button is pressed. any code related to generating python snippets for filtering the dataframe is removed


Vincent-Maladiere (approved):
LGTM

PR #1060, opened 2024-09-05
always show top 20 associations regardless of their value

jeromedockes:
before we would show at most 20 but only those that have an association above 0.2
having 2 cutoffs was confusing: if usually there are 20 but on my data only 15 show up (even if I have many columns in the dataframe) it is not clear why

this pr always shows the top 20


Vincent-Maladiere (approved):
Nice!

GaelVaroquaux (approved):
LGTM

GaelVaroquaux:
Thanks!! Merging

PR #1061, opened 2024-09-05
use an icon for warning sign and replace clipboard with double squares

jeromedockes:


Vincent-Maladiere (approved):
I'm +1 for this copy-past icon

PR #1062, opened 2024-09-05
[DOC] Freshen up the homepage

Vincent-Maladiere:
This PR revamps the homepage to articulate what skrub does more clearly.
From top to bottom:
- Add a Discord link on the navbar
- Change the features on the gradient band
- Add a sentence describing what skrub aims for, in plain English
- Move the button from the left of the band to the center of the page
- Add three feature sections to replace the previous cards. These sections are feature-centric instead of module-centric.
- Add a contributors section (it's nice to put some faces, or at least icons, to the project)
- Add a footer


Vincent-Maladiere:
Just realized that the new footer wasn't playing nice with the current one in all other pages, I need to address that

Vincent-Maladiere:
Also, it would be good to put some large HTML chunks (like the TableReport) elsewhere to ease the readability of the index.html file

doc/_static/css/custom.css
```diff
@@ -233,17 +233,104 @@ div.sk-landing-footer {
...
+}
+
+
+a.sk-btn-orange:hover {
+    background-color: #fbe1ca;  /* --sk-orange-tint-3 */
 }
 
+
+.container {
+    max-width: 80%; /* Restrict container width for better readability */
```
jeromedockes, line 268: on narrow screens we don't want to waste 20% of the screen width. So I would suggest setting that in size rather than percentage eg max-width: 180ch;

doc/_static/css/custom.css
```diff
@@ -233,17 +233,104 @@ div.sk-landing-footer {
...
+}
+
+.feature-title {
+    color: #f99f44; /* Darker text color for the title */
```
jeromedockes, line 272: here also I think the contrast with the background is too low

doc/_static/css/custom.css
```diff
@@ -233,17 +233,104 @@ div.sk-landing-footer {
...
+    color: #555; /* Lighter grey for the description text */
+    font-size: 1rem; /* Regular font size for text */
+    margin-bottom: 0; /* Remove default bottom margin */
+}
+
+
+
+.hero-description {
+    text-align: center;
+    font-size: 20px;
```
jeromedockes, line 287:
```suggestion
    font-size: 1.25rem;
```

doc/_static/css/custom.css
```diff
@@ -233,17 +233,104 @@ div.sk-landing-footer {
...
+    width: 100%;
+    height: 0.5px; /* Adjust the thickness of the line */
+    background-color: #12134b; /* Change the color as needed */
+    border: none; /* Ensures no border is applied */
+    margin-top: 50px;
+    margin-bottom: 50px;
+}
+
+
+.contributors-list {
```
jeromedockes, line 302: this is a really nice addition; the only drawback is the github avatars make the page much bigger

jeromedockes (commented):
thanks a lot @Vincent-Maladiere  this is super cool! :tada: 

you may want to check how it looks on narrow screens or mobile phones:
- there is no padding in the top banner so the last of the 3 bullet points touches the bottom of the banner
- there is too much white space on each side of the screen due to the 80% max-width
- the gaps between the sections need to be tightened IMO


I'm not sure why there is no line between "effortless pipelines" and "powerful feature engineering" but there is one between the other sections?

I think we need to create the tablereport as part of the sphinx build process, rather than adding it to the repo. it will avoid having huge meaningless diffs in the repo and also this way when we improve the report the homepage will get updated without needing to regenerate it and add it manually. (we may also be able to do the highlighting of code snippets with pygments during the sphinx build to be able to write them in python and update them more easily). I'm ok with addressing this last point in another PR though

doc/_static/css/custom.css
Vincent-Maladiere: Ok, I can't see a difference for a single screen size, though

doc/_static/css/custom.css
Vincent-Maladiere: Why is that a drawback?

Vincent-Maladiere:
> I'm not sure why there is no line between "effortless pipelines" and "powerful feature engineering" but there is one between the other sections?

I don't see it, can you provide a screenshot? The horizontal lines split all feature sections.

> there is too much white space on each side of the screen due to the 80% max-width
> the gaps between the sections need to be tightened IMO

I'm afraid I have to disagree with both points. Having a small margin on a desktop is awkward. Spacing between sections also prevents it from being too crowded with code and demos. I agree with you on mobile, however. I reduced them for this format.

I also changed the orange color of the button and headers to bump the ratio to 3.33, which should be enough for large texts. I can't do more than that, because it wouldn't be orange any more.

TODO:
- [ ] generate the HTML table report from a Python script
- [ ] fix the 3 dots thingy on the banner (hopefully we will get rid of it someday)

doc/_static/css/custom.css
jeromedockes: yes but some users do change the font size in their browser settings

jeromedockes:
> I don't see it, can you provide a screenshot? The horizontal lines split all feature sections.

screenshot_2024-09-06T14:58:39+02:00


I think it is due to the 0.5px line width, the lines appear and disappear depending on the zoom level. on my laptop and external screen at the default zoom 100% on firefox it doesn't show up between those 2 sections. that's not a big deal either though

jeromedockes:
> Spacing between sections also prevents it from being too crowded with code and demos

sorry i wasn't clear, for the spacing between sections I meant on a phone screen only, it looks fine to me on a big screen. (I saw you changed it already for small screens)

doc/_static/css/custom.css
Vincent-Maladiere: Ok, to be kept in mind if things start to get slow. Also, since we're using github API, maybe we would run into rate limits at some points?

Vincent-Maladiere: Ok, to be kept in mind if things start to get slow. Also, since we're using github API, maybe we would run into rate limits at some point?

jeromedockes:
> I also changed the orange color of the button and headers to bump the contrast ratio to 3.33, which should be enough for large texts. I can't do more than that, because it wouldn't be orange any more.

thanks! If needed we could always make it not orange but I think it looks ok now. the hovered state still has a very low contrast but that's probably less important

doc/_static/css/custom.css
jeromedockes: I would be very surprised as it's just one request when loading the page. many projects seem to show contributors like this without problems, I wouldn't worry about it

doc/_static/css/custom.css
```diff
@@ -233,17 +233,102 @@ div.sk-landing-footer {
...
+    background-color: #fbe1ca; /* Fill with white on hover */
+    color: #DA7007; /* Text color changes to match the original background color */
+    border: 2px solid #DA7007; /* Border remains the same on hover */
+}
+
+.container {
+    max-width: 120ch;
+}
+
+@media (min-width: 768px) {
```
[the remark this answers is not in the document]
Vincent-Maladiere: I removed it entirely, do you think it works or should we add some padding?

GaelVaroquaux:
One "graphical" detail: the new homepage is made of horizontal blocks.

It would be very nice to have one out of two with a slightly greyed background. I do realize that this implies removing the "hr" elements and using the background color of the divs instead.

doc/_templates/index.html
```diff
@@ -11,230 +11,203 @@
...
+
+
+<div class="container mb-5 mt-5">
+  <div class="row justify-content-center"> <!-- Centers the row content -->
+    <div class="col-md-10">
+      <h2 class="text-center mt-0 mb-4">Our community</h2> <!-- Centers the header text -->
+      <p class="mb-5 text-center">
+        Here are some of the contributors to the Skrub project, whose efforts have helped shape
+        its development. Here we display a randomly selected group of
+        30 contributors.</p>
```
GaelVaroquaux, line 147:
```suggestion
        The Skrub project is powered by the efforts of a world-wide community of contributors. Here we display a randomly selected group of
        30 contributors.</p>
```

Vincent-Maladiere:
Feedback applied so far:
- [x] Rephrasing the headline (easing -> to ease)
- [x] Rephrasing the last bullet point for heteregenous types 
- [x] Adding / moving code on the feature section
- [x] Generate the table report at each build using a function called in conf.py
- [x] Showing the 30 randomly sampled contributors + a quick word about the contributor section

doc/_templates/index.html
Vincent-Maladiere: better indeed!

doc/_static/gap.png
GaelVaroquaux, line 1: Use an SVG here? Matplotlib saves very nice SVGs (I think that you need to use ", transparent=True" to savefig to have a transparent background, which comes in handy)

GaelVaroquaux:
Can we easily add a vignette saying "And many more" at the end of the contributor list?

Vincent-Maladiere:
> Can we easily add a vignette saying "And many more" at the end of the contributor list? When clicked it would go to the github contributor list.

yeswecan

doc/_static/css/custom.css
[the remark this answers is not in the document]
jeromedockes: as you prefer! to me it looks good now

doc/table_report.py
```diff
@@ -0,0 +1,14 @@
...
+
+
+def generate_demo():
+    X = fetch_employee_salaries().X
+    with open("_templates/demo_table_report_generated.html", "w") as f:
+        f.write(TableReport(X).html_snippet())
+
+
+# %%
+generate_demo()
```
jeromedockes, line 13: not a big deal but I think we end up doing it twice because it is done here upon import and then called again in conf.py. you may want to put this line inside a `if __name__ == '__main__'` block

doc/_templates/index.html
```diff
@@ -11,230 +11,205 @@
...
+        <a class="reference internal"
+        href="reference/generated/skrub.TableReport.html#tablereport"
+        title="skrub.TableReport"><code class="xref py py-class docutils
+        literal notranslate"><span class="pre">TableReport</span></code></a>
+        .</p>
+      <br>
+      {% include "demo_table_report_code.html" %}
+    </div>
+    <div class="col-md-7">
+      <p class="click-table-report-hint">Click anywhere on the table!</p>
```
jeromedockes, line 128: I don't really have a better suggestion but with "click on the table" people might look only at the table and miss the tabs and the column filter... maybe something like "the displayed report is interactive"?... I can't come up with a better wording and it's not a very big deal

jeromedockes (approved):
thanks a lot @Vincent-Maladiere !! I think it highlights the selling points of skrub really well.

I think the blue/orange banner at the top is too tall on narrow displays and ends up pushing the "getting started" button out of the first screen, but that is somewhat orthogonal to the changes done here so it could be addressed in another pr

doc/_static/gap.png
Vincent-Maladiere: The transparent makes it challenging in darkmode (see below), why do we want SVG? The image is not super big and is displayed in good quality even with a big screen

Screenshot 2024-09-09 at 18 06 46

doc/_templates/index.html
GaelVaroquaux: How about moving the text back to the left panel (for better usage of the space), and writing "Try it out: click anywhere ➔"

Vincent-Maladiere:
> One "graphical" detail: the new homepage is made of horizontal blocks.
> 
> It would be very nice to have one out of two with a slightly greyed background. I do realize that this implies removing the "hr" elements and using the background color of the divs instead.

The issue with adding a grey background to some sections is that their margins would remain white (notice how the line breaks don't fill the entire width on big screens).

That is because I can't get rid of the main container {% block docs_main %} defined somewhere in Jinja & Sphinx. This container has set some margins that I can't style.

I suggest adding an issue to try that after this PR because this is quite challenging. WDYT @GaelVaroquaux ?

doc/_static/css/custom.css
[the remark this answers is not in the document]
Vincent-Maladiere: I think it's ok

GaelVaroquaux:
@jeromedockes writes:

> I think the blue/orange banner at the top is too tall on narrow displays and ends up pushing the "getting started" button out of the first screen

Indeed, that's a good illustration of why we're always fighting for screen real estate :)

I think that the third bullet point can be made shorter: "Work on heterogenous types (numeric, categorical, dates, text, missing values...)" -> "heterogenous data (numbers, categories, dates, text, missing values...)"

doc/table_report.py
Vincent-Maladiere: typo, I forgot to remove the function call used for debugging

GaelVaroquaux:
> The issue with adding a grey background to some sections is that their margins would remain white (notice how the line breaks don't fill the entire width on big screens).

Here is the trick to do it: set the color on an enclosing div with max-width: none; width: 140vw; margin-left: -20vw; margin-left; -20vw; padding-left: 20vw; padding-right: 20vw;
In this enclosing div, use a div with width: 98vw; max-width: 120rem; margin-left: auto; margin-right:auto"

With this you can have the horizontal slabs

jeromedockes:
[here](https://github.com/pydata/pydata-sphinx-theme/blob/80221a4baa0ef1e3f13f7fdfb99ce4ced87ce7c9/src/pydata_sphinx_theme/theme/pydata_sphinx_theme/layout.html#L88) is the template from the pydata sphinx theme that you would need to override by having a file with the same name in the template path. You could also see on which element there is padding and force removing it with something like `main {padding: 0 !important;}`

however,
- there are a few gray elements on the page such as the striped tables and the code blocks, which may not play very well with a gray background
- there is the theme's dark mode thingy :roll_eyes: so you would probably need to choose colors for both modes or adjust the text color as well
- IMHO the current combination of headings, horizontal rules and whitespace works fine to separate the different sections

so I would lean towards not considering this a blocker for merging this PR

> But a cleaner way to do it would be by modifying a bit more the html structure of the page: closing the "main" and it's two enclosing divs, and opening new ones below.

I think I would prefer the "less clean" way you proposed (or adding a css rule, or overriding `layout.html`) because the page must have only one (non-hidden) `main` element and it should hold the main content

GaelVaroquaux:
Fair enough with the however, I'm happy giving up the grey / white alternance. But it brings me to the core problem: the current design creates too much vertical scrolling. I was trying to use grey / white alternance to decrease the empty space between elements.

We should also work on a bit more use of 2 panes on the same horizontal line: the getting started button on the right, the contributors on the left...

Vincent-Maladiere:
Ok, should we merge this version and try to reduce the scrolling in subsequent PRs?

GaelVaroquaux:
Can we do the horizontal splitting for the first and last rows.

Vincent-Maladiere:
Ok let's try that

GaelVaroquaux:
Also, in the interest of vertical scrolling, I think that the "Click anywhere on the table!" should be moved to the left panel, below the code and maybe with an arrow added: "Click anywhere on the table! ➡", possibly "Try it! Click anywhere on the table ➡"

Vincent-Maladiere:
> Also, in the interest of vertical scrolling, I think that the "Click anywhere on the table!" should be moved to the left panel, below the code and maybe with an arrow added: "Click anywhere on the table! ➡", possibly "Try it! Click anywhere on the table ➡"

What about this? https://github.com/skrub-data/skrub/pull/1062#discussion_r1750610079

GaelVaroquaux:
Good point about the fact that the arrow may not point in the right direction on small screen (I like attention to details!). How about using this arrow: "🡾" (or other on the page https://stackoverflow.com/questions/13973106/unicode-ascii-arrows-in-8-directions)?

Vincent-Maladiere:
Ok, I can also try that. I'm just a little afraid of special character rendering

Screenshot 2024-09-10 at 18 32 30

GaelVaroquaux:
To avoid using special characters, one option is to use fontawesome (which I believe is included in our theme). The only icon that I found that is pointing to the right direction is https://fontawesome.com/icons/arrow-trend-down?f=classic&s=solid

Vincent-Maladiere:
Or this one? https://fontawesome.com/icons/arrow-right?s=solid

jeromedockes:
> Or this one? https://fontawesome.com/icons/arrow-right?s=solid

I think Gael's trick was to have an arrow that points towards the south-east so that it's in the right direction whether it is above or on the left of the report. if there isn't one in fontawesome we can just add a small svg file (and it will probably look better, too)

Vincent-Maladiere:
Haha, hacky indeed! Let me try that

Vincent-Maladiere:
I have to say that this last commit creates some weirdness around vertical alignment. I'm struggling to align the h1 and h4 in the band with the text below.

Screenshot 2024-09-11 at 12 25 52

jeromedockes:
> I have to say that this last commit creates some weirdness around vertical alignment. I'm struggling to align the h1 and h4 in the band with the text below.

in the custom.css there is a `sk-landing-container` class that overrides the max-width. in the part you show below the max-width is set by bootstrap's `container` class and is narrower. as both are centered with `auto` margins and the top one is wider, they don't line up.

Vincent-Maladiere:
Yes, but more specifically the issue I have is maintaining the gradient band as it is. I can vertically align the h1 and h4 with the text below, but it creates a white margin on the left and right of the band.

Also, I think putting the button on the right cuts the natural flow of a user landing on the page. It doesn't feel natural to read text on the left and then cross the spread horizontally to see/click the button. I'm -1 for this tradeoff with the vertical scrolling.

GaelVaroquaux:
Side comment: if you want inspiration on a good page (done by professionals with a lot of money) https://spark.apache.org/

It's an awesome tradeoff in terms of information density and vertical scrolling

Vincent-Maladiere:
> Side comment: if you want inspiration on a good page (done by professionals with a lot of money) [spark.apache.org](https://spark.apache.org/)

They did an amazing job indeed

Vincent-Maladiere:
This is now ready for review!

Side note; Anass from `:probabl.` looked at the homepage during his spare time and came up with new ideas. I like his draft, I think this is promising. To be continued :)

image (5)

The sponge "S" logo is very fun

Screenshot 2024-09-11 at 16 50 10

GaelVaroquaux:
I would move the "Our community" text on the side

GaelVaroquaux:
Also: the first line is done with a pair of "md-4" / "md-8" divs. As a consequence, the div where there is the text is much smaller than that where there is the button. I would use "md-6" / "md-6" to have the same size, or "md-8", "md-4" to have a bigger div for the text: it makes reading the text easier

GaelVaroquaux:
How about moving the "Click anywhere on the table ↓" to the side?

Vincent-Maladiere:
> I would move the "Our community" text on the side

What do you mean? Isn't it already on the side?

Vincent-Maladiere:
> How about moving the "Click anywhere on the table ↓" to the side?

Do you mean on the *left* side? Below the code?

GaelVaroquaux:
> > I would move the "Our community" text on the side

> What do you mean? Isn't it already on the side?

The whole text in the right panel

> >     How about moving the "Click anywhere on the table ↓" to the side?

> Do you mean on the left side? Below the code?

Yes indeed

GaelVaroquaux (approved):
This is great. Thanks a lot for all these changes.

Let's merge

jeromedockes:
I like the sponge logo a lot btw :)

GaelVaroquaux:
If we can simplify the sponge enough to have it as a favicon, it could be a good replacement for our current logo.

jeromedockes:
> If we can simplify the sponge enough to have it as a favicon

I guess it could be a uniform yellow S without any of the decorations?

PR #1065, opened 2024-09-10
arrow key navigation in sample table

jeromedockes:
arrow keys or hjkl to move around, Esc to deselect a cell, Ctr-c to copy

also now the column headers can be selected


Vincent-Maladiere (approved):
Very nice

Issue #1050, opened 2024-09-02
Add zero padding on embeddings column names for ordering purposes


Vincent-Maladiere:
### Problem Description

This issue is not super important, but it would help to have it fixed.

The output of `MinHashEncoder` and `GapEncoder` is a dataframe whose columns are in the form `<col>_1`, `<col>_2`, ..., `<col>_10`, `<col>_11`. While intuitive, this is slightly confusing when columns get sorted, e.g. with the `AggJoiner`, because then the ordering becomes `<col>_1`, `<col>_10`, `<col>_11`, `<col>_2`

### Feature Description

I suggest we add left zero padding by accounting for the number of columns:
- between 1 and 9 columns, no padding: 1, ..., 9
- between 10 and 99 columns, a single 0 padding: 01, ..., 99
- and so on

### Alternative Solutions

Keep it as it is.

Shree7676:
@Vincent-Maladiere 
I would like to work upon this issue

Vincent-Maladiere:
Hello @Shree7676 thank you, go ahead 😊

Shree7676:
@Vincent-Maladiere 
I am going through the code of GapEncoder and I am bit confused between columns and labels.
can you provide some code through which i can reproduce.
below is something I explored and the column name here doesn't provide a good picture of above issue.

`
import pandas as pd
from skrub import GapEncoder
data = pd.DataFrame({
    'category': ['apple', 'banana', 'apple', 'orange', 'banana', 'orange', 'banana']
})
encoder = GapEncoder(n_components=7, random_state=42)
encoded_data = encoder.fit_transform(data['category'])
encoded_df = pd.DataFrame(encoded_data)
print(encoded_df.columns.tolist())
`
it provides output as 
`['category: apple, orange, banana', 'category: banana, apple, orange', 'category: orange, apple, banana', 'category: apple, orange, banana (3)', 'category: apple, orange, banana (4)', 'category: apple, orange, banana (5)', 'category: apple, orange, banana (6)']`

Vincent-Maladiere:
Yes, indeed, I may have written this issue too fast. This issue was specific to the `MinHashEncoder`, so I don't think there is a need for the GapEncoder. Sorry for that.

Shree7676:
Thanks @Vincent-Maladiere 
For all the guidance :)

Vincent-Maladiere:
[...]

PR #1069, opened 2024-09-10
added zero padding for column names in MinHashEnocder

Shree7676:
#1050
Adding zero padding for MinHashEncoder
Column names are now able to get zero padding based on the units length of the column 
Added the test case


jeromedockes:
Hi @Shree7676 , thanks a lot for handling this! I like that we pad only as much as is necessary and it's great that you added a test case.

There are always a couple of setup details to take care of in a first Pull Request, but we only need to do it once and then we can merge this one and you'll be able to open many more :) : 

- we use [pre-commit](https://pre-commit.com/#install) to handle some linting and code formatting. By using it you can make sure the code on your local copy will get automatically formatted in the same way as the main branch, and this avoids spurious lines of diff that are only due to formatting changes (as those we see at the beginning of the minhash module). you can install it with `pip install pre-commit` and then activate the hooks that run when you create a git commit by running `pre-commit install` inside the skrub directory (you only need to do this once)
- when making changes that are visible to users, we usually add an entry in [`CHANGES.rst`](https://github.com/skrub-data/skrub/blob/main/CHANGES.rst#minor-changes) (in this case it would be a "minor" change). it is a brief statement about the addition and a link to the corresponding pull request; you can follow the format of the other entries in there.

addressing these 2 points should take care of the 2 Continuous Integration checks that are complaining, and then we can merge. thanks again!

Vincent-Maladiere:
Hi @Shree7676, thank you for this PR! Out of curiosity, have you found the [contributing page](https://skrub-data.org/stable/CONTRIBUTING.html#writing-your-first-pull-request)? It's ok if you didn't, we want to improve the contributing experience in skrub and are eager to get your feedback :)

Shree7676:
Hi @Vincent-Maladiere , thank you for asking! I did go through the contributing page, and I found it clear and to the point. During EuroSciPy, I learned that many projects use tools like pytest for testing, so I knew to look for it in skrub as well. Similarly, through this PR, I learned that most projects also have specific formatting requirements, like pre-commit, which I now know to check before pushing changes.

I think these are general things that beginners, especially those without a computer science background like myself, might not be familiar with right away. If you wanted to make the contributing page more beginner-friendly, it could be helpful to mention tools like pytest and pre-commit as part of the development process. However, if you feel this is already common knowledge for most contributors, it might not be necessary.

Overall, I think the page is well-written and informative. Thanks again for your support!

Shree7676:
Hi @jeromedockes,
Thanks for the guidance :)

Shree7676:
I still have to work on gap encoder but I am thinking to open a separate pr for it.

jeromedockes:
Thanks!

> you feel this is already common knowledge for most contributors

we don't want to assume that, and for new contributors at any level of experience each project has its own quirks and getting started can be a hassle so improving the contributor guide is always time well spent. thanks for your suggestions!

> I still have to work on gap encoder but I am thinking to open a separate pr for it.

that's a good idea, I would indeed recommend doing it in a separate one

skrub/tests/test_minhash_encoder.py
```diff
@@ -254,3 +254,36 @@ def test_get_feature_names_out(df_module):
...
+                "col_08",
+                "col_09",
+                "col_10",
+                "col_11",
+            ],
+        ),  # 2-digit padding
+    ],
+)
+def test_zero_padding_in_feature_names_out(df_module, n_components, expected_columns):
+    """Test that ``get_feature_names_out`` returns the correct zero-padded feature names."""
```
jeromedockes, line 283: the only formatting thing left is that the formatter complains this line is too long. it cannot reformat it itself because it is the docstring and it only modifies code, not docstrings nor comments so we have to fix it manually. maybe we can make it a bit shorter like "Check the feature names are zero-padded"

jeromedockes, line 283:
```suggestion
    """Check that the feature names are zero-padded."""
```

Shree7676:
I think I just messed up was trying to resolve the conflict.
But as I can see some additional unwanted files have changed 
Need Help!!

Shree7676:
for commit number 7dd566f (resolved conflict)
I just added changes.rst file and test_minhash_encoder.py file and then commited as you can see in below screenshot 
but still after git push why do I see other files here
image

Vincent-Maladiere:
Hey @Shree7676, github can be tricky sometimes! As far as I understand, you want to revert your last 6 commits to start from  "Update skrub/tests/test_minhash_encoder.py" (`d2d16808278b1daffb7bb036cf9068e03931ce49`).

You can do it with:
```shell
git reset HEAD~6  # revert to the 6th last commit
git push -f origin add_zero_padding  # push and force the changes
```

CHANGES.rst
```diff
@@ -37,6 +37,9 @@ Minor changes
   ``pandas.info()`` ) in a table.
   :pr:`1056` by :user:`Jérôme Dockès <jeromedockes>`.
 
+* Added zero padding for column names in :class:`MinHashEncoder` to improve column ordering consistency.
+  :pr:`1069` by :user:`Shreekant Nandiyawar <Shree7676>`.
+
```
jeromedockes, line 42:
```suggestion

```

jeromedockes, line 42:
```suggestion

```

CHANGES.rst
```diff
@@ -37,6 +37,8 @@ Minor changes
   ``pandas.info()`` ) in a table.
   :pr:`1056` by :user:`Jérôme Dockès <jeromedockes>`.
 
+* Added zero padding for column names in :class:`MinHashEncoder` to improve column ordering consistency.
+  :pr:`1069` by :user:`Shreekant Nandiyawar <Shree7676>`.
```
jeromedockes, line 41:
```suggestion
  :pr:`1069` by :user:`Shreekant Nandiyawar <Shree7676>`.

```

jeromedockes:
alright we can merge it now! thanks a lot @Shree7676 !! :tada:

jeromedockes (approved):

Shree7676:
Thanks @jeromedockes  and @Vincent-Maladiere for the help

PR #1072, opened 2024-09-12
Column selection in "distributions" tab

jeromedockes:
the columns are now all unselected when opening the page (they were all selected before) and there is a tooltip explaining that the checkboxes are for constructing a copy-pasteable list


jeromedockes:
@Vincent-Maladiere  I think this one is ready. the change is in the "distributions" panel

screenshot_2024-09-13T14:03:30+02:00

Vincent-Maladiere (approved):
LGTM!

PR #1073, opened 2024-09-16
arrow key navigation in jupyter notebooks

jeromedockes:
jupyter notebooks capture arrow key events which prevents the report from seeing them and breaks using arrow keys to move to the next tab or sample table cell. this pr fixes this problem.

please give it a try on your preferred notebook-like environment if you use one and report where it breaks! we probably need to add some environments to the list that we detect and which trigger the special handling of keyboard events added by this pr

### DIFF
--- a/skrub/_reporting/_data/templates/report.js
+++ b/skrub/_reporting/_data/templates/report.js
@@ -227,6 +227,9 @@ if (customElements.get('skrub-table-report') === undefined) {
                 this.activate();
                 event.preventDefault();
             });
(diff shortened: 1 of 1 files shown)


Vincent-Maladiere:
Is this one ready for merging?

jeromedockes:
yes I think so! we checked jupyter notebooks, jupyterlab, vscode, google colab and kaggle. if something else comes up I'll open another PR. thanks!

GaelVaroquaux (approved):
LGTM, merging

PR #1075, opened 2024-09-16
DOC change style of the landing page

glemaitre:
This PR makes a couple of cosmetic changes to the landing page.

Changes are:

- light/dark mode compatible image
- light/dark mode compatible separator
- center single column introduction + tagline
- remove previous tagline from original gradient
- add ending section to redirect toward "getting started"
- remove "getting started" from contributor section
- add a link to the online table report generator
- reduce the number of colors on page (orange, black, white, touch of blue)
- reduce the max-width on the navbar and landing page (not on API and doc pages)


Vincent-Maladiere (commented):
A quick, early, feedback

1. I prefer having only one call to action (the orange button) and would rather not have additional links below feature sections. Also, we wanted to change the button name back to "Getting started" instead of "See example".
2. The "less wrangling, more ML" header occupies much space. Ideally, this would replace the tagline "Prepping model for ML" which is less explicit, or the gradient band entirely.
3. I'm personally not a fan of the new header font Poppins

doc/_static/css/custom.css
```diff
@@ -99,6 +102,87 @@ div.card-body details[open] summary i {
...
+    font-weight: normal;
+}
+
+.code-margin {
+    margin-top: 0.75em !important;
+    margin-bottom: 0.75em !important;
+}
+
+html[data-theme="light"] .encoding-image {
+    background-image: url('../encoding_light.png');
```
glemaitre, line 169: @Vincent-Maladiere Do you have the code that generate the image `_static/gap.png`? I have a placeholder but I would like to generate the original one.

glemaitre:
> Also, we wanted to change the button name back to "Getting started" instead of "See example".

I'm not a big fan of the orange button in terms of design right now. I did not yet think of what could be a drop in replacement that is visually appealing.

> one call to action (the orange button)

While for the first 2 features, I might agree, I think this is a miss opportunity to redirect to the demo that is the right "call for action" for the last feature.

GaelVaroquaux:
@glemaitre : I think that we do need that it is important to keep visual separations between the horizontal panels.

One way to easily achieve this that caters for the fact that the div does not span the full width of the screen is something like:

```css
div.row-padding-between-features {
  border-top: #aaa solid 3px;
  border-image: linear-gradient(to right, #fff 0%, #4085a7 50%, #fff 100%) 1;
}
```
image

GaelVaroquaux:
Also, I pushed very hard to have the "Click anywhere on the table " on the side panel, because I really worry about the amount of vertical scrolling that this page creates

GaelVaroquaux:
@Vincent-Maladiere says: 

> The "less wrangling, more ML" header occupies much space. Ideally, this would replace the tagline "Prepping model for ML" which is less explicit, or the gradient band entirely.

I'm not opposed to replacing the tagline "Prepping tables...". We should make the contrast more pronounced here, but this can be done by playing with the drop shadow of the text (which is there for the contrast). It would have the benefit of reducing the vertical scroll (you know me and vertical scroll...)

I would however really want to keep the gradient band, in order to have a similar visual identity to scikit-learn

GaelVaroquaux:
If we remove the "prepping tables..." tagline, I would like to have " tabular machine learning" in bold in the short text. I want people to see very fast that this is about tables.

GaelVaroquaux:
Another tiny detail, I notice that police and correction are very visible on our front page:
image
I think that for certain people it will trigger the wrong thoughts. Can we just skip a few lines to focus on things like schools.

GaelVaroquaux:
For compatibility with light/dark mode, one option (maybe slightly less "flash" than the current one, would be to use the variable defined by the pydata sphinx theme, for instance:

```css
border-image: linear-gradient(to right, var(--pst-color-background) 0%, #4085a7 50%, var(--pst-color-background) 100%) 1;
}
```
Which would give:
image

versus:
image

glemaitre:
> For compatibility with light/dark mode, one option (maybe slightly less "flash" than the current one, would be to use the variable defined by the pydata sphinx theme, for instance:

I wanted to see if we could still use the blue color (to recall the scikit-learn theme). I don't have yet an opinion on what is best, I'm just trying stuff around :). I'll be happy to converge on anything that people agree.

glemaitre:
On a crazy note, I was wondering if we would be inclined to drop a bit the current visual to gave something like:

image

image

Of course the second scikit-learn logo in only a place holder right now.

GaelVaroquaux:
> On a crazy note...

I suggest that for now we stick to cosmetics on the latest/current layout. Indeed, there is a danger of going back and forth on layouts, pushed by personal preferences. If we want to do a good thinking on the front page (which can certainly be useful), we should tackle this as a design project: start by agreeing on the priorities in the messages, then sketch a few designs, possible ask a wider crowd to rate them (for style but also for ease of understanding and rapidity of finding the information)

GaelVaroquaux:
Can we work on the verticality of the top of the page:
image

The information is piled up vertically, and does not make use of the screen. It induces much scrolling

glemaitre:
> The information is piled up vertically, and does not make use of the screen. It induces much scrolling

Yes I can work on the scrolling mainly by changing the font size and spacing.

image

I'm not certain that the 2 columns at this specific location is the best: the two intro lines are quite long and will certainly be cut as shown above in almost all type of screen. Thus, in term of vertical scrolling, we end up with the same amount of scrolling and a visual less appealing to me (I'm therefore open for changes because this is a bias view).

GaelVaroquaux:
Regarding vertical scrolling, given that you have removed the tagline from the header, we can do the following:

remove the min-height from div.sk-landing-container and use "col-md-4" and "col-md-8" for the columns of the header. This gives a bit less vertical space taken by the header:
image

GaelVaroquaux:
This makes me think that I would like to order the bullet points differently:

* scikit-learn compatible
* Pandas and Polars dataframes inputs and outputs
* Work on heterogenous types (numeric, categorical, dates, text, missing values...)

This is for two reasons: first I think that anchoring in the scikit-learn world (skworld? :smile: ) first is a good thing. Second, I like going from short to long

jeromedockes:
while we're at it can we replace the funky arrow with either a svg image or a rotated fontawesome or unicode right pointing arrow? something like `<span style="display: inline-block; transform: rotate(45deg); font-size: 5rem;">→</span>` (but with the styling in custom.css not inline)

glemaitre:
I move the PR for ready for review. I think that I did answer to most of comments IRL or in this list. From the top of the head, there is still the issue with the fonts that @Vincent-Maladiere was not a fan of. I'm open to changes with alternative. I would prefer to have a font with a "purest" design (not an expert in this but basically a "sans serif" font together with smoothed transition would be my preference).

GaelVaroquaux:
Overall the changes are fine with me.

My thoughts are (both mentioned above and non blocker):

- I find the gradient orange-blue in the separations too flashy for me, and I would prefer using "border-image: linear-gradient(to right, var(--pst-color-background) 0%, #4085a7 50%, var(--pst-color-background) 100%) 1;
}"
- It would be nice to remove the lines with "police" and "correction" from the dataframe displayed in the TableReport

I would really like to hear from @Vincent-Maladiere as he led the rewamp

glemaitre:
> It would be nice to remove the lines with "police" and "correction" from the dataframe displayed in the TableReport

I'll do that in a separate PR since this is an aside Python script that generate the table.

Vincent-Maladiere:
I agree with Gael that starting a design track (soonish?) would be good, with the help of professional UX/UI people.

My feedback, for the non-UI things:
1. I think we should improve the wording of "Try on interactive demo" to "Try it on your dataset" because the TableReport on the right side is already an interactive demo.
2. On mobile, we should justify all headers and text with e.g. `text-align: justify` (I forgot to do it in my previous PR).
3. On mobile, "Click anywhere" should be immediately above the TableReport, instead of "Try interactive demo".
4. On mobile, "our community" and text should be above the profile pics
5. On mobile, the profile pics should be horizontally centered (the right margin is too large)

For the UI things (more opinionated):
6. I would prefer "Arial", for instance, as a font for both titles and text. It's simpler and it works better here IMO.
7. I'm not very comfortable with the big blue horizontal line, can we try something more discreet? Perhaps keeping a grey line of 1px or slightly higher (but not 3px).

doc/_static/css/custom.css
Vincent-Maladiere: Is this resolved? (cf IRL convo)

glemaitre:
I reverted to the original font from the PST theme.

@Vincent-Maladiere I pushed the justify text such that you can check the preview and let me know your thought.

Regarding the separator, I think that there is two approaches here: (i) we think about accessibility and we want to provide a visible, high-contrast bar (ii) we want to go more stylish and in this case I would argue that we don't need separator but we should use vertical spacing to have the feeling of separation.

My original proposal was (ii) but I see the point of (i) (that looks what @GaelVaroquaux was striving for). The proposal of @Vincent-Maladiere is the in-between solution and I'm thinking that we are not going to satisfy either criterion imposed by (i) and (ii). So my judgement call here would to go with (i) (at least for this refactoring - real designer later can propose more optimal solution).

GaelVaroquaux:
The latest version is just right for me

Vincent-Maladiere (commented):
Ok, after discussing IRL with @glemaitre, the justify-center on mobile doesn't render as well as I expected.
Regarding the horizontal bar, I still think i) is not great from a UI perspective. However, I'm open to making a compromise with your current proposal, since we don't have the help of a professional designer.

Last remark: the description of skrub in the hero has too many style effects IMO (italic, bold, mono). I'd be keen on removing the italic and bold effect.

GaelVaroquaux:
> Last remark: the description of skrub in the hero has too many style effects IMO (italic, bold, mono). I'd be keen on removing the italic and bold effect.

I'd like to keep the bold: the word "table" is not visible enough, while it's an absolutely central keyword of skrub.

glemaitre:
I removed the italic and moved the "And many more" within the contributor list as the last elemetn.

Vincent-Maladiere (approved):
LGTM, thank you @glemaitre! :)

glemaitre:
The failure is not linked to the website. Let's @GaelVaroquaux click on the green button :)

GaelVaroquaux:
Done, merged!! Wohoo!

PR #1077, opened 2024-09-19
FEA Add `TextEncoder`

### LINKED ISSUE #1047: Add a FastText encoder
### Problem Description

When encoding long text on small datasets, https://arxiv.org/abs/2312.09634 has shown that embeddings improve prediction performance over string models like `MinHashEncoder`. More recently, [CARTE](https://hal.science/hal-04596816) performed well using [FastText](https://github.com/facebookresearch/fastText) to initialize column names and category embeddings.

### Feature Description

Create an encoder that downloads FastText weights, loads them during `fit`, and applies them during `transform`. Note that FastText dependencies are only ["pybind11>=2.2", "numpy"].

### Alternative Solutions

Instead, we could create a transformer using `SentenceTransformer`, which would download weights from HuggingFace. The issue is that although these models provide more powerful embeddings than FastText, this solution would require installing `torch`, `transformers`, and finally `sentence-transformers`. Also, running these models is markedly slower than using FastText.

### Additional Context

_No response_

Vincent-Maladiere:
closes #1047 

**Reference**
https://github.com/skrub-data/skrub/issues/1047

**What does this PR propose?**

This PR wraps the SentenceTransformers library, in the same fashion as [ragger-duck](https://github.com/probabl-ai/sklearn-ragger-duck/blob/main/ragger_duck/embedding/_sentence_transformer.py) and [embetter](https://github.com/koaning/embetter/blob/main/embetter/text/_sbert.py) does.

Following the good empirical baselines of [arxiv.org/abs/2312.09634](https://arxiv.org/abs/2312.09634), this PR also proposes to:
1. Add a PCA within the estimator, controlled by a ``n_components`` parameter. This also allows the class to stay consistent with the other skrub encoders and perform HP tuning more easily. Following the paper, we set ``n_components=30`` by default.
2. Add a normalization option, with a default "l2". This simplifies cosine distance to dot products and brings embeddings as comparable features to distance-based estimators (e.g., NearestNeighbors). This is useful when other normalized features are used (fuzzy joining).

What the SentenceEncoder does **not**:
- perform weights or embedding quantization (kept for later research)
- set torch threading, which doesn't interact well with joblib and raises an error if called twice.
- choose a different embedding Pooling scheme (only mean pooling is implemented in SentenceTransformer)

See the [implementation of the paper](https://github.com/LeoGrin/lm_tab/blob/reproduce/scripts/launch_computations.py#L161) for more details.

**Todo in another PR**
Enriching an existing example from skrub with the SentenceEncoder.


Vincent-Maladiere:
Pixi is extremely annoying. It has negative returns on productivity as far as I am concerned.

pyproject.toml
```diff
@@ -166,12 +174,12 @@ test = { cmd = "pytest -vsl --cov=skrub --cov-report=xml skrub" }
...
-test = ["test"]
-dev = ["dev", "optional", "doc", "lint", "test"]
-ci-py309-min-deps = ["py309", "min-dependencies", "test"]
-ci-py309-min-optional-deps = ["py309", "min-dependencies", "min-optional-dependencies", "test"]
-ci-py312-latest-deps = ["py312", "test"]
-ci-py312-latest-optional-deps = ["py312", "optional", "test"]
+test = ["test", "transformers"]
+dev = ["dev", "optional", "doc", "lint", "test", "transformers"]
+ci-py309-min-deps = ["py309", "min-dependencies", "test", "transformers"]
+ci-py309-min-optional-deps = ["py309", "min-dependencies", "min-optional-dependencies", "test", "transformers"]
```
glemaitre: Let's avoid to test for transformers for python 3.9

pyproject.toml
```diff
@@ -167,9 +180,10 @@ test = { cmd = "pytest -vsl --cov=skrub --cov-report=xml skrub" }
 lint = ["lint"]
 doc = ["optional", "doc"]
 test = ["test"]
-dev = ["dev", "optional", "doc", "lint", "test"]
+dev = ["dev", "optional", "doc", "lint", "test", "transformers"]
 ci-py309-min-deps = ["py309", "min-dependencies", "test"]
 ci-py309-min-optional-deps = ["py309", "min-dependencies", "min-optional-dependencies", "test"]
+ci-py311-transformers = ["py311", "transformers", "test"]
```
glemaitre, line 186: OK so it seems that we get closer to a working CI. However, with

```
numpy = ">=1.23.5,<2"
pytorch = { version = "*", channel = "pytorch" }
transformers = "*"
sentence-transformers = "*"
```

We get a old `sentence-transformers` (2.0.0). I'm going to try forcing it to have it to API 3.0 and the latest pytorch.

glemaitre: OK I think it works.

glemaitre:
OK, so it seems that we would be able to make the CI works with `sentence-transformers=2.X`. Basically, the latest release have not been created on conda-forge but I'm trying to work something out: https://github.com/conda-forge/sentence-transformers-feedstock

glemaitre:
This CI effort is worth a couple of 🍻

Vincent-Maladiere:
> This CI effort is worth a couple of 🍻 (I still don't get how people deal with the DL stack without being an expert).

Wouhou! I'm buying the first round 😉

Vincent-Maladiere:
There are a lot of codecov warnings on the _sentence_encoder.py file, but this is a display bug; the coverage is close to 100%.

Vincent-Maladiere:
We will be able to remove torchvision from the optional dependency once @glemaitre [PR](https://github.com/conda-forge/sentence-transformers-feedstock/pull/42/files#diff-f3725a55bf339595bf865fec73bda8ac99f283b0810c205442021f29c06eea9aL29) on conda forge is merged. Having a dep on torchvision is not detrimental for this POC, though.

CHANGES.rst
```diff
@@ -14,7 +14,8 @@ It is currently undergoing fast development and backward compatibility is not en
 
 Major changes
 -------------
-
+* The :class:`SentenceEncoder` is now available to encode columns with diverse entries.
```
jeromedockes, line 17: I think it's worth expanding the whatsnew entry as this is a rather high-profile addition. I would say that it allows representing entries in a table with embeddings computed by a deep-learning model

doc/encoding.rst
```diff
@@ -50,6 +50,32 @@ categories, eg to replace :class:`~sklearn.preprocessing.OneHotEncoder`:
        Similarity encoding for learning with dirty categorical variables. 2018.
        Machine Learning journal, Springer.
 
+
+Encoding text with diverse entries
+----------------------------------
+
+When it comes to encoding text with diverse entries, large language models
```
jeromedockes, line 57: "large language model" might have users thinking "i can't run this on my machine", maybe insist on the fact that there are language models of all shapes and sizes on the hugging face hub and that now thanks to skrub you can plug any of them in a scikit-learn pipeline for tabular data

doc/reference/encoding_a_column.rst
```diff
@@ -12,6 +12,7 @@ Encoding a column
    GapEncoder
    MinHashEncoder
    SimilarityEncoder
+   SentenceEncoder
```
jeromedockes, line 15: I wonder about the name. I get that it reflects the name of the underlying library, but for a user who isn't familiar with it or with this implementation detail it might not be obvious what this does. maybe LanguageEncoder, TextEmbedding, ...

pyproject.toml
```diff
@@ -245,6 +259,9 @@ filterwarnings = [
     # should be addressed in pandas
     'ignore:np.find_common_type is deprecated.*:DeprecationWarning',
     'ignore:.*distutils Version classes are deprecated.*',
+    # TODO: add more context but linked to sentence-transformers
```
jeromedockes, line 262: just a reminder to address the TODO

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+        To use this class, you need to install the optional ``transformers``
+        dependencies for skrub. For example, you can use pip:
+
+        .. code::
+
+           pip install skrub[transformers] -U
+
+        This will install the ``sentence-transformers``, ``transformers``,
+        and ``torch`` libraries on your machine. Be aware that this might lead
+        to dependency conflicts.
```
jeromedockes, line 37: not sure I understand the part about conflicts?

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+
+    This class uses a pre-trained model, so calling ``fit`` or ``fit_transform``
+    will not train or fine-tune the model. Instead, the model is loaded from disk,
+    and a PCA is fitted if ``n_components`` is not None.
```
jeromedockes, line 41: PCA is fitted to reduce the dimension of the language model's output

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+    When PCA is disabled, this class is essentially stateless, with loading the
+    pre-trained model from disk being the only difference between ``fit_transform``
+    and ``transform``.
+
+    Be aware that parallelizing this class (e.g., using
+    :class:`~skrub.TableVectorizer` with ``n_jobs`` > 1) may be computationally
+    expensive.
+    This is because a copy of the pre-trained model is loaded into memory
+    for each thread. If memory usage is a concern, check the characteristics of
+    your selected model.
```
jeromedockes, line 52: maybe set n_jobs to 1 and use the transformer library's own mechanism for parallelization within the language model?

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+            Unused. Here for compatibility with scikit-learn.
+
+        Returns
+        -------
+        X_out : pandas or polars DataFrame of shape (n_samples, n_components)
+            The embedding representation of the input.
+        """
+        st = import_optional_dependency("sentence_transformers")
+
+        if not sbd.is_string(column):
```
jeromedockes, line 157: not important but if we do the check before we avoid importing the library in the case where we won't use it and raise an error anyway. but it's true that we usually put imports at the top

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+                    random_state=self.random_state,
+                )
+                X_out = self.pca_.fit_transform(X_out)
+            else:
+                warnings.warn(
+                    f"The matrix shape is {(X_out.shape)}, and its minimum is "
+                    f"{min_shape}, which is too small to fit a PCA with "
+                    f"n_components={self.n_components}. "
+                    "The embeddings will be truncated by keeping the first "
+                    f"{self.n_components} dimensions instead. "
```
jeromedockes, line 198: I wonder about the choice of truncating in this case, have you seen or used that approach before? I guess the default behavior of scikit-learn pca would be to set the number of components to n rows in this case?

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Returns
+        -------
+        X_out : pandas or polars DataFrame of shape (n_samples, n_components)
+            The embedding representation of the input.
+        """
+        check_is_fitted(self, "estimator_")
+
+        if not sbd.is_string(column):
+            raise RejectColumn(f"Column {sbd.name(column)!r} does not contain strings.")
```
jeromedockes, line 231: I should have written clearer documentation about RejectColumn; it is only meant to be used during fit/fit_transform.

the reason is that it is a way to signal to the meta-estimator (like the tablevectorizer) "I'm not the right transformer for this column, pick someone else" .

but if we don't reject the column during fit, then assignments will not change so at transform time it is too late to change our mind: we have accepted the responsibility to handle this column and now we must deal with it. if the data passed during transform is not appropriate, we can raise an exception (like ValueError or anything else, but not RejectColumn because RejectColumn will receive no special treatment at this point), or output NaNs or apply whatever strategy seems most appropriate.

TLDR keep the RejectColumn in fit_transform, but in `transform`, either raise a different exception or output NaNs

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+
+        X_out = self._vectorize(column)
+
+        if hasattr(self, "pca_"):
+            X_out = self.pca_.transform(X_out)
+        elif self.n_components != "all":
```
jeromedockes, line 237: what is "all", didn't we use `None` for keeping all dimensions?

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+        is_null = sbd.to_numpy(sbd.is_null(column))
+        column = sbd.to_numpy(column)
+        unique_x, indices_x = unique_strings(column, is_null)
+
+        return self.estimator_.encode(
+            unique_x,
+            normalize_embeddings=False,
+            batch_size=self.batch_size,
+            show_progress_bar=self.verbose,
+        )[indices_x]
```
jeromedockes, line 256: I'm not familiar with SetenceEncoder; who is taking care of moving the embeddings from gpu to numpy on cpu?

skrub/_sentence_encoder.py
```diff
@@ -0,0 +1,300 @@
...
+
+        Returns
+        -------
+        feature_names_out : list of str
+            Transformed feature names.
+        """
+        check_is_fitted(self)
+        n_digits = len(str(self.n_components_))
+        return [
+            f"{self.input_name_}_{str(i).zfill(n_digits)}"
```
jeromedockes, line 298: not in this pr but we may want a helper in _utils for the column name zfilling

skrub/tests/test_sentence_encoder.py
```diff
@@ -0,0 +1,139 @@
...
+pytest.importorskip("sentence_transformers")
+
+
+@pytest.fixture
+def encoder():
+    """Create a fixture such that we set the common parameters of the encoder.
+
+    Here, we have two constraints:
+
+    - We want a really small model to not blow up the memory of the GitHub Actions
```
jeromedockes, line 19: if it is still a problem with this model we can also mock the sentence transformer with something that has the same interface but returns random values

skrub/tests/test_sentence_encoder.py
```diff
@@ -0,0 +1,139 @@
...
+    x = pd.Series(["oh no"])
+    with pytest.raises(ImportError, match="Missing optional dependency"):
+        st.fit(x)
+
+
+def test_sentence_encoder(df_module, encoder):
+    X = df_module.make_column("", ["hello sir", "hola que tal"])
+    encoder = clone(encoder).set_params(n_components=2)
+    X_out = encoder.fit_transform(X)
+    assert X_out.shape == (2, 2)
```
jeromedockes, line 49: it's a detail but .shape only happens to work because pandas and polars use the same name for it, so if we want to be really nitpicky we would write sbd.shape(X_out)

skrub/tests/test_sentence_encoder.py
```diff
@@ -0,0 +1,139 @@
...
+def test_transform_error_on_float_data(df_module, encoder):
+    """Check that we raise an error when data without any string is passed at
+    transform."""
+    x = df_module.make_column("", [1.0, 2.5, 3.7])
+
+    encoder = clone(encoder).set_params(n_components=None)
+    encoder.fit(df_module.make_column("", ["hello", "world"]))
+
+    with pytest.raises(RejectColumn, match="does not contain strings"):
+        encoder.transform(x)
```
jeromedockes, line 139: very nice suite of tests :)

shall we also have one with the sentenceencoder inside a tablevectorizer to make sure that works?

jeromedockes (commented):
Very cool addition, thanks! I think this will be a big deal for some users.

I wonder if we could showcase it in an example? or is it too computationally expensive to run on circleci?

jeromedockes:
I also wonder about pickling. maybe we should not store the model in an attribute but only store the name, and reload the model as necessary? or if we are worried about loading it many times we can cache it in an attribute that does not get pickled and repopulate the cache when necessary

jeromedockes:
> cache it in an attribute that does not get pickled

this can be done by defining `__getstate__` as shown [here](https://docs.python.org/3/library/pickle.html#handling-stateful-objects)

jeromedockes:
here is a toy example with `functools.cached_property`

`encoder.py`

```python
import functools

class Encoder:
    def __init__(self, name):
        self.name = name

    def transform(self, x):
        return self._estimator(x)

    @functools.cached_property
    def _estimator(self):
        print(f'loading {self.name}')
        return lambda x: f"{self.name}({x})"

    def __getstate__(self):
        state = self.__dict__.copy()
        if '_estimator' in state:
            del state['_estimator']
        return state
```

usage

```
>>> import pickle

>>> import encoder

>>> e = encoder.Encoder('supermodel')
>>> print(e.transform(7))
loading supermodel
supermodel(7)

the model is only loaded once

>>> print(e.transform(8))
supermodel(8)

pickling works

>>> s = pickle.dumps(e)
>>> e1 = pickle.loads(s)

after unpicling we reload the model

>>> print(e1.transform(9))
loading supermodel
supermodel(9)
>>> print(e1.transform(10))
supermodel(10)

the `_estimator` itself could not be pickled

>>> pickle.dumps(e._estimator)
    ...
AttributeError: Can't pickle local object 'Encoder._estimator.<locals>.<lambda>'
```

doc/reference/encoding_a_column.rst
Vincent-Maladiere: TextEmbedding is nice, the only issue is vectors and embeddings are close concepts, so people might not understand its difference with other string-based encoders like MinHash. What about LLMEncoder?

skrub/_sentence_encoder.py
Vincent-Maladiere: This class has no `n_jobs` parameter because we let pytorch set the default number of threads.

jeromedockes: yes sorry I didn't mean in the class but in the doc's recommendation we could say something like "set n_jobs to 1 on the TableVectorizer and let pytorch handle parallelism" . but maybe that's going too much into details

skrub/_sentence_encoder.py
Vincent-Maladiere: Yes, SentenceTransformers proposes [this option](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/util.py#L314) directly in [the encode method](https://github.com/UKPLab/sentence-transformers/blob/master/sentence_transformers/SentenceTransformer.py#L605).

But your question is interesting; my POV is that several arguments can back truncation when the number of rows is insufficient to fit a PCA (< 30, by default, which is fairly low):
1. Embedding dimensions are different across models, setting a default truncation would allow fixing the number of columns of the output.
2. Embedding dimensions are often huge (> 128, up to > 1000) which is tedious to deal with in the context of a dataframe.
3. Users can keep all dimensions with `n_components=None`, but the opposite wouldn't be true; users couldn't limit the number of embedding dimensions on a small dataset if we skipped the dimension reduction.
4. Limiting the number of components rather than using a PCA can even be advantageous with special encoders like [Matryoshka Embeddings](https://sbert.net/examples/training/matryoshka/README.html).

WDYT?

skrub/_sentence_encoder.py
Vincent-Maladiere: > the reason is that it is a way to signal to the meta-estimator (like the tablevectorizer) "I'm not the right transformer for this column, pick someone else" .

Ok that makes sense!

skrub/_sentence_encoder.py
Vincent-Maladiere: Oupsi, that was from a previous iteration. Which makes me wonder about the coverage of our tests.

skrub/_sentence_encoder.py
Vincent-Maladiere: Sentence-encoder does that for us. It outputs numpy arrays.

skrub/tests/test_sentence_encoder.py
Vincent-Maladiere: With the instabilities in torch, transformers, and sentence-transformers, maybe it's best not to trust them fully and run the full tests anyway?

skrub/tests/test_sentence_encoder.py
Vincent-Maladiere: right, I thought that for simple case like this one sbd wasn't necessary. Ok to use it though

skrub/tests/test_sentence_encoder.py
Vincent-Maladiere: Yes, good idea!

skrub/_sentence_encoder.py
Vincent-Maladiere: Doesn't cost too much to change :)

skrub/_sentence_encoder.py
Vincent-Maladiere: It was to answer Gael concern's that the dependencies are constraining, for instance using pytorch < 2.4.1 with numpy > 2 leads to a segfault. Happy to change the wording or remove it entirely.

skrub/_sentence_encoder.py
Vincent-Maladiere: An error is probably best here?

Vincent-Maladiere:
> TextEmbedding is nice, the only issue is vectors and embeddings are close concepts, so people might not understand its difference from other string-based encoders like MinHash. What about LLMEncoder?
> Maybe DeepEncoder? I think if we put "LLM" in the name some users will think they can only use it if they have a big GPU

HuggingFaceEncoder would probably be more explicit

Vincent-Maladiere:
> I also wonder about pickling. maybe we should not store the model in an attribute but only store the name, and reload the model as necessary? or if we are worried about loading it many times we can cache it in an attribute that does not get pickled and repopulate the cache when necessary

Why are you concerned about pickling? Note that loading using names wouldn't work for local models in another context. For example, if a user fine-tunes a sentence-transformer and gives the local path to the SentenceEncoder, pickling the SentenceEncoder and unpickling on, e.g., a remote server would fail because the model weights aren't there anymore. Users would have to ship the weights along with the pickle object and make sure the versions are correct, which is a bit burdensome.

pyproject.toml
Vincent-Maladiere: Right, that was Guillaume's, I'm not entirely sure what is missing here, I will need further testing

GaelVaroquaux:
I typically have a rule to never introduce a major feature without a corresponding example. I find that writing the example helps so much finding usability improvements.

skrub/_sentence_encoder.py
jeromedockes: I would say that is the case for all packages (old versions of scikit-learn are probably not compatible with numpy 2 either) so maybe not necessary to re-state here, but it's true these new ones might be more picky. I'll let you decide no strong opinion

skrub/_sentence_encoder.py
jeromedockes: ok in this case let's keep truncation :) the alternative I had in mind was not to keep all dimensions, but to set the number of components to min(n rows, n columns) with a warning so that the PCA could be fitted

skrub/_sentence_encoder.py
jeromedockes: yes an error is probably best. In the TableVectorizer, this situation will not happen because the string preprocessing will have made sure you are dealing with strings, but of course this can be used outside of the tablevectorizer so the check is useful

skrub/_sentence_encoder.py
jeromedockes: tha'ts great. maybe we could have a short comment saying that it outputs numpy arrays?

skrub/tests/test_sentence_encoder.py
jeromedockes: maybe one day we will add ibis whose tables don't have a shape :)

jeromedockes:
> Downloading the weights might be a bit slow though.

we can use the circleci cache so that downloading models does not happen often (if/when that turns out to be a problem)

jeromedockes:
>  Note that loading using names wouldn't work for local models in another context.

yes that's true, I was assumign the main use-case would be to use a pre-trained model directly from the huggingface hub. the reason I wonder about pickling is (i) model size: I imagine we might end up with quite a few variants of a skrub pipeline, or fitted models for different splits etc., but all of them using the same hugging face model, so including a copy of it in each pickle might not be optimal. but most importantly (ii) if the model has its weights on a gpu, what happens when we pickle it and unpickle on a machine that does not have the same device? if we get an error that could be a problem. the problem is we don't have a way to check this in the CI because the runners don't have a gpu. I'm also ok with saying we will investigate this further in another issue. still I would at least add a test to check that pickling and unpickling in the simple case (same machine, cpu, the only setup we have in the CI) works

skrub/_sentence_encoder.py
Vincent-Maladiere: The only difference with dependencies like scikit-learn is that errors are raised in a manner that is easy to understand, whereas torch plainly segfault and crash the shell session :/

skrub/_sentence_encoder.py
Vincent-Maladiere: Ah ok! This alternative also makes sense. @GaelVaroquaux WDYT?

GaelVaroquaux:
Regarding loading, pickling, and memory usage, have you considered if this is a case for a Borg pattern (refs: https://www.oreilly.com/library/view/python-cookbook/0596001673/ch05s23.html https://stackoverflow.com/questions/1318406/why-is-the-borg-pattern-better-than-the-singleton-pattern-in-python )?

doc/reference/encoding_a_column.rst
GaelVaroquaux: I would go for "TextEncoder".

A detail, but in a list like the above, I would like it to be in a separate sub(sub)section: "Deep Learning", with a tiny text in the beginning that explains the need to install other libraries and points to the part of the documentation where we explain how to do this.

skrub/_sentence_encoder.py
GaelVaroquaux: I think that this case will be very infrequent, no? It fires when the number of components requested is smaller than the number of samples, no? And we will typically use as a number of components 30.
In which case it doesn't matter much which way we go.
But maybe I got confused

skrub/_sentence_encoder.py
Vincent-Maladiere: I added this suggestion

skrub/_sentence_encoder.py
Vincent-Maladiere: So, when the number of samples is smaller than the PCA dimension (`n_components`, which defaults to 30), we have two options for reducing dimensionality:

1. Truncate the embeddings by selecting the first `n_components` dimensions
2. Adjust the `n_components` parameter of the PCA to match `n_samples` and then run the PCA

This situation often arises with very small or demo datasets.

Vincent-Maladiere:
Thanks for the reference! If I understand correctly, this pattern could help maintaining the same state accross different instances, but I wonder how that could help in terms of memory. Say I have a shared model accross different instances, if I run them parallel I guess the RAM consumption will still explode, right? I mean, I have to bring the matrices somewhere in memory to perform my parallel forward passes. I don't see how this could help with pickling, however. @jeromedockes WDYT?

doc/encoding.rst
```diff
@@ -50,6 +50,43 @@ categories, eg to replace :class:`~sklearn.preprocessing.OneHotEncoder`:
...
        Machine Learning journal, Springer.
 
+
+Encoding text with diverse entries
+----------------------------------
+
+When encoding strings with diverse entries and large chunks of text, we can leverage
+language models of various sizes to represent string columns as embeddings.
+Depending on the task and dataset, this approach may lead to significant performance
+improvements, albeit with potential increases in memory usage and latency.
```
jeromedockes, line 60: maybe "computation time" rather than "latency"?

skrub/_text_encoder.py
```diff
@@ -0,0 +1,320 @@
...
+    >>> X = pd.Series([
+    ...   "The professor snatched a good interview out of the jaws of these questions.",
+    ...   "Bookmarking this to watch later.",
+    ...   "When you don't know the lyrics of the song except the chorus",
+    ... ], name='video comments')
+
+    Fitting does not train the underlying pre-trained deep-learning model,
+    but ensure various checks and enable dimension reduction.
+
+    >>> enc.fit_transform(X)
```
jeromedockes, line 127:
```suggestion
    >>> enc.fit_transform(X) # doctest: +SKIP
```

?

so we can run pytest without transformers installed

jeromedockes (commented):
It's looking really good! looking forward to having this in skrub. from our discussion out-of-band:

- @Vincent-Maladiere will add the option to not include model weights in the pickled transformer
- we must take care of not pickling the huggingface token
- a good place to showcase it in the examples might be to expand the gapencoder one to show all text encoding methods

basically the example is the main remaining todo :)

jeromedockes:
>  I don't see how this could help with pickling, however. @jeromedockes WDYT?

Indeed my question was not really about memory usage or sharing state between objects but rather whether the model weights should be included in the pickle file. (i) if yes, then it cannot be unpickled in a machine that does not have the same device (without taking extra steps, AFAIK, I may very well be mistaken on this one), and the pickle may be much bigger than it needs to be. for example I can see several skrub/scikit-learn pipelines re-using the same sentence encoder model (ii) if no, then it cannot be unpickled on a machine that does not have a copy of the sentenceencoder or a way to download it from the huggingface hub.

So neither option is perfect, we have to guess which is least likely to bother many users (and possibly add a parameter to let the user choose, in this pr or a later one)

Vincent-Maladiere:
Last piece missing: updating the encoder example

Vincent-Maladiere:
@GaelVaroquaux, I'd like to merge this PR now and, instead of adding a separate example, include the example in the current GapEncoder one in a separate PR.

The goal would be to compare the TextEncoder, MinHashEncoder, and perhaps some LSA as well, using [the toxicity dataset](https://github.com/surge-ai/toxicity/blob/main/toxicity_en.csv) that Jovan shared yesterday. It's a small text dataset (about 1000 entries), but I think it’s a good choice to illustrate the tradeoff between the TextEncoder's computation time and performance.

GaelVaroquaux:
> I'd like to merge this PR now and, instead of adding a separate example, include the example in the current GapEncoder one in a separate PR.

I would like to have a hard rule that no major feature gets merged without being in at least one example. In my experience this is good practice.

Vincent-Maladiere:
Okay, but I have to raise awareness that adding the dataset plus revamping the example might create a big PR and delay its merging.

GaelVaroquaux:
Fair enough, but my experience is that often going through the process of the example forces to rethink the functionality and thus is really useful.

As long as we don't have the example, we're not sure that it's ready for the user. 

Thanks!!

Vincent-Maladiere:
> going through the process of the example forces to rethink the functionality and thus is really useful.

Got it, I like this point of view, it sounds healthy indeed!

doc/encoding.rst
```diff
@@ -50,6 +50,43 @@ categories, eg to replace :class:`~sklearn.preprocessing.OneHotEncoder`:
...
        Machine Learning journal, Springer.
 
+
+Encoding text with diverse entries
+----------------------------------
+
+When encoding strings with diverse entries and large chunks of text, we can leverage
+language models of various sizes to represent string columns as embeddings.
+Depending on the task and dataset, this approach may lead to significant performance
+improvements, albeit with potential increases in memory usage and computation time.
```
jeromedockes, line 60:
```suggestion
Depending on the task and dataset, this approach may lead to significant improvements
in the quality of predictions, albeit with potential increases in memory usage and computation time.
```

nitpick: "performance" is often understood to mean "computation time"

pyproject.toml
```diff
@@ -246,6 +260,11 @@ filterwarnings = [
...
```
jeromedockes, line 263:
```suggestion
```

skrub/_text_encoder.py
```diff
@@ -0,0 +1,405 @@
...
+          the sentence-transformer model again from HuggingFace Hub.
+          This process could fail if, for example, the machine doesn't have
+          internet access. Additionally, if you use weights stored on disk
+          that are *not* on the HuggingFace Hub (by passing a path to
+          ``model_name``), these weights will not be pickled either.
+          Therefore you would need to copy them to the machine where you
+          unpickle the ``TextEncoder``.
+        - When set to True, the ``_estimator`` property is included in
+          the serialized object. Users deploying fine-tuned models stored on
+          disk are recommended to use this option.
```
jeromedockes, line 119: maybe point out that the machine where we unpickle must have the correct device?

skrub/_text_encoder.py
```diff
@@ -0,0 +1,405 @@
...
+        Used when the PCA dimension reduction mechanism is used, for reproducible
+        results across multiple function calls.
+
+    verbose : bool, default=True
+        Verbose level, controls whether to show a progress bar or not during
+        ``transform``.
+
+    Attributes
+    ----------
+    _estimator : sentence_transformers.SentenceTransformer
```
jeromedockes, line 131: I find it a bit surprising to document private attributes. I would rename it `estimator_` if we want to expose it to users, or remove it from the docstring if we don't

skrub/_text_encoder.py
```diff
@@ -0,0 +1,405 @@
...
+        The torch encoder.
+
+    _cache_folder : Path
```
jeromedockes, line 134: same here

jeromedockes (approved):
LGTM! I added a couple of nitpicks but I think it's ready. very nice addition, thanks!

jeromedockes:
screenshot from the rendered example :sweat_smile: :

screenshot_2024-10-22T12:48:21+02:00

skrub/_text_encoder.py
Vincent-Maladiere: Right, let's remove it from the docstring.

jovan-stojanovic (commented):
Hi @Vincent-Maladiere, thanks for this great addition! Here are some comments:

doc/install.rst
```diff
@@ -41,6 +58,22 @@ Install
...
 
+|
+
+**Deep learning dependencies**
+
+Deep-learning based encoders like :class:`TextEncoder` require installing optional
+dependencies to use them. The following will install
+`torch <https://anaconda.org/pytorch/pytorch>`_,
+`transformers <https://anaconda.org/conda-forge/transformers>`_,
+`torchvision <https://anaconda.org/pytorch/torchvision>`_,
```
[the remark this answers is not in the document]
Vincent-Maladiere: Hey Jovan, thank you for the review! This one is quite tricky; sentence transformers [used to require torchvision in their conda-forge recipe](https://github.com/conda-forge/sentence-transformers-feedstock/blob/d3e5d69710606bc40b76c3a560d44f5113af92f3/recipe/meta.yaml#L29), but a @glemaitre fix has been merged recently, so I think we can now remove that indeed!

doc/install.rst
```diff
@@ -139,6 +188,21 @@ If no errors or failures are found, your environment is ready for development!
...
 
+|
+
+**Deep learning dependencies**
+
+Deep-learning based encoders like :class:`TextEncoder` require installing optional
+dependencies to use them. The following will install
+`torch <https://pypi.org/project/torch/>`_,
+`transformers <https://pypi.org/project/transformers/>`_,
+and `sentence-transformers <https://pypi.org/project/sentence-transformers/>`_.
```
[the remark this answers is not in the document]
Vincent-Maladiere: Yes, because on Pypi, sentence-transformers does not depends on torchvision, which should have been the case for conda-forge as well.

Vincent-Maladiere:
Ping @jovan-stojanovic :)

jovan-stojanovic (approved):
Great!

Vincent-Maladiere:
@GaelVaroquaux can we merge this one?

GaelVaroquaux:
I'm having a look right now. I played with the example on my computer. It's really cool

I've created a small PR to your branch here: https://github.com/Vincent-Maladiere/skrub/pull/1
If you're happy with it, can you merge it please?

Also, there is a conflict with main, which means that we cannot merge

GaelVaroquaux (approved):
LGTM, but please merge my doc PR https://github.com/Vincent-Maladiere/skrub/pull/1 :)

jeromedockes:
Nice!!! thanks @Vincent-Maladiere ! very exciting -- let's try to release it soon :) :rocket:

GaelVaroquaux:
Very exciting indeed! Bravo!

GaelVaroquaux:
Darn, we have a build failure on circleCI on main:
image

I'm restarting the build.

GaelVaroquaux:
Yes indeed. It fails reliably. I fear that it might be related to the PR: some nasty interaction between examples.

We'll need to bisect to know whether it's related to the dependencies that are being installed, or the code that is being run. I've created an issue to track this: https://github.com/skrub-data/skrub/issues/1143

glemaitre:
Memory issues?

Vincent-Maladiere:
I can try to bisect on this tomorrow!

PR #1078, opened 2024-09-19
improve tabular learner for unseen categories

GaelVaroquaux:
When used with trees, tabular_learner created a pipeline that could fail with unknown categories, as noticed by @rcap107 

This fixes the problem


Vincent-Maladiere (approved):
LGTM!

TheooJ (approved):
Thanks @GaelVaroquaux !

PR #1080, opened 2024-09-20
MAINT update the version of pixi to 0.30.0 and the lock files

glemaitre:
Bumping the version of `pixi` on the CI and update the lock files.


Vincent-Maladiere (approved):
LGTM!

PR #1074, opened 2024-09-16
show the pandas index in tablereport

jeromedockes:

PR #1083, opened 2024-09-23
pandas index & multiindex in TableReport

jeromedockes:
supersedes #1074 

adds displaying pandas index, handle correctly multiIndex (both for index and columns) and index/columns level names if any are present

allows controlling how many rows are displayed in the sample table

also some refactoring of how the html table is constructed


jeromedockes:
todo: [ ] keep only one parameter for controlling number of displayed rows

jeromedockes:
I reduced the number of rows from 10 to 5 in the srkub homepage.

BTW, not really related to this PR but in the home page if we want the report to take yet less vertical space one way is to give it a bit more horizontal space (either making the sk-landing-page a bit wider or giving the report more space compared to the code snippet on its right) because then the column filters select will drop to be on the same line as the tabs

jeromedockes:
ok this one is ready for review

here is an example to generate a report with multiple column and index levels:

<details>

```python
# %%
import datetime

import pandas as pd
import numpy as np

from skrub import TableReport

df = pd.DataFrame(
    {
        "A": ["one", "one", "two", "three"] * 6,
        "B": ["A", "B", "C"] * 8,
        "C": ["foo", "foo", "foo", "bar", "bar", "bar"] * 4,
        "D": np.random.randn(24),
        "E": np.random.randn(24),
        "F": [datetime.datetime(2013, i, 1) for i in range(1, 13)]
        + [datetime.datetime(2013, i, 15) for i in range(1, 13)],
    }
)

df = pd.pivot_table(
    df,
    values="E",
    index=["B", "C"],
    columns=["A"],
    aggfunc=["sum", "mean"],
)

report = TableReport(df)
report

# %%
report.open()
```
</details>

skrub/_reporting/_data/templates/dataframe-sample.css
```diff
@@ -3,23 +3,75 @@
...
 }
 
 /* The column containing the selected cell has green stripes. */
 /* The rest of the table has gray stripes added by pure.css */
 
-tr:nth-child(2n -1) .table-cell[data-is-in-active-column] {
-    background-color: var(--mediumgreen);
+.table-cell[data-is-in-active-column] {
+    /* !important to override pure.css */
+    background-color: var(--active-column-color) !important;
```
GaelVaroquaux, line 23: Can you give a precise css path to avoid the important?
Precise css selectors override imprecise ones

skrub/_reporting/_summarize.py
```diff
@@ -1,14 +1,22 @@
...
 
 
-def summarize_dataframe(df, *, order_by=None, with_plots=True, title=None):
+def summarize_dataframe(
+    df,
+    *,
+    order_by=None,
+    with_plots=True,
+    title=None,
+    max_top_slice_size=5,
```
TheooJ, line 17: Maybe ``top_rows`` and ``bottom_rows`` ? Or ``n_top_rows`` ?
Feels like ``max_top_slice_size`` is kind of long

jeromedockes: indeed it is long, but there are quite a few "rows" flying around, and we need to distinguish between the maximum number of displayed rows and the actual number so I wanted to be very explicit here. 

note this is a parameter of a private helper, the parameter exposed to the user is just `n_rows`

GaelVaroquaux (approved):
LGTM, but I would prefer if the "!important" in CSS could be removed

jeromedockes:
> LGTM, but I would prefer if the "!important" in CSS could be removed

yep I removed it already. it's just that pure has a selector that is quite specific already (table class name + nth pseudo-class for alternating rows + elem) and I didn't want to have to come up with an artificially specific one but actually I remembered I wrapped the report in an element with an id which comes in handy for that

PR #1086, opened 2024-09-24
align copy buttons to flex-start + always visible

jeromedockes:
the copy buttons stretch across the box they copy which is fine in the most common case of just 1 line but weird otherwise.

also @Vincent-Maladiere said it would be best to have them always visible and I agree

this pr also improves a bit the colors of default/hover/active states

**before**

screenshot_2024-09-24T10:04:14+02:00

**after**

screenshot_2024-09-24T10:03:56+02:00

### DIFF
--- a/skrub/_reporting/_data/templates/buttons.html
+++ b/skrub/_reporting/_data/templates/buttons.html
@@ -1,6 +1,6 @@
 {% macro copybutton(target) %}
 <button
-    class="copybutton"
+    class="copybutton copybutton-left"
     type="button"
     onclick="copyButtonClick(event)"
     data-target-id="{{ target }}">
--- a/skrub/_reporting/_data/templates/copybutton.css
+++ b/skrub/_reporting/_data/templates/copybutton.css
@@ -35,6 +35,10 @@
     align-self: flex-start;
 }
 
+.copybutton-left {
+    order: -1;
+}
+
 .copybutton {
     background: #e0e0e0;
 }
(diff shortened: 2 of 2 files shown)


Vincent-Maladiere:
Can you try placing the copy button on the left side rather than the right? I know Pydata-theme and others put them on the right, but it would be easier to have it where your eyes and mouse are.

Vincent-Maladiere:
Otherwise +1 for the vertical button. What is your point of view regarding the right padding of the text? There is a tradeoff between readability and vertical space when the text is very long.

jeromedockes:
> Can you try placing the copy button on the left side rather than the right? I know Pydata-theme and others put them on the right, but it would be easier to have it where your eyes and mouse are.

I just pushed a change with the buttons on the left, you can check it once ci runs. I'm 50/50 on this one; I like it on the left but I haven't found a single example of similar buttons being on the left so it might be slightly surprising. OTOH I think the reason they are usually on the right is because they are inside text that is right-aligned, which does not apply to us

jeromedockes:
>  What is your point of view regarding the right padding of the text? There is a tradeoff between readability and vertical space when the text is very long.

what you see is not padding but actual hard returns `"\n"` in the random text I generated. the text in the box preserves white space so those line breaks are respected

if the text has longer lines you will get a horizontal scrollbar too:

screenshot_2024-09-24T11:10:40+02:00

jeromedockes:
in any case if we want to switch between button on the left or the right it's just a question of adding or removing a css class in one place so it will be easy to revisit if we don't like it after all. ATM it's on the left

this is ready for a final review @Vincent-Maladiere

Vincent-Maladiere (approved):
Got it, thank you for the explainers. LGTM!

PR #1090, opened 2024-09-26
tablereport captures keydown events in reveal js (slides)

jeromedockes:
in jupyter and vscode we make sure that the tablereport gets arrow key keydown events which are otherwise captured by the notebook or vscode.

this adds revealjs slides to the list where this is done


TheooJ (approved):
Thanks !

Issue #948, opened 2024-06-17
Display columns with the HTML representation of the fitted TableVectorizer


Vincent-Maladiere:
### Problem Description

When displaying the HTML representation of a fitted TableVectorizer on a Jupyter notebook, it would help to see the names of the columns matching the different fitted transformers. This info would be useful for a quick sanity check and a better understanding of what the TV does, especially for beginners.

### Feature Description

The `ColumnTransformer` from scikit-learn implements this feature. [Reference](https://scikit-learn.org/stable/modules/compose.html#visualizing-composite-estimators)

<img width="383" alt="Screenshot 2024-06-17 at 14 35 39" src="https://github.com/skrub-data/skrub/assets/18560386/4bec9596-5c79-4dde-ad18-2c1e0dd906e8">

Shree7676:
![image](https://github.com/user-attachments/assets/a694aa62-0f3c-4c04-afab-d38717b1a541)

Is this what you are suggesting ? @Vincent-Maladiere

Vincent-Maladiere:
Try the ColumnTransformer from scikit-learn, we would like to see the column names when clicking on the dropdown!

Shree7676:
I would like to work on this

jeromedockes:
fixed in #1093

PR #1093, opened 2024-09-27
Adds column name in html table vectorizer

Shree7676:
#948
cc @jeromedockes 
@ PyData Paris Sprint


Shree7676:
to fix the codecov/patch 
should I just update the test case name as test_sk_visual_block_

skrub/tests/test_table_vectorizer.py
```diff
@@ -714,6 +714,16 @@ def test_bad_specific_cols():
         TableVectorizer(specific_transformers=[(None, [0])]).fit(None)
 
 
+def test_sk_visual_block_():
```
jeromedockes, line 717:
```suggestion
def test_sk_visual_block():
```

skrub/tests/test_table_vectorizer.py
```diff
@@ -714,6 +714,16 @@ def test_bad_specific_cols():
...
+    X = _get_clean_dataframe()
+    vectorizer = TableVectorizer()
```
jeromedockes, line 719: here we could check the repr html of the unfitted estimator to check it works and cover the missing line that codecov is complaining about

```python
unfitted_repr = vectorizer._repr_html_()
assert 'TableVectorizer' in unfitted_repr
```

jeromedockes (approved):
thanks a lot @Shree7676 !!

PR #1097, opened 2024-09-30
Text in matplotlib svg

### LINKED ISSUE #1095: Table Report unable to read emojis
### Describe the bug

The Table Report raises a warning when emojis are in a string column, because when creating the column stats tables, used font (DejaVu Sans) is not able to represent emojis.
We should probably use another font when emojis are in columns, by changing the parameters of plotting functions.

### Steps/Code to Reproduce

```python 
df = pd.DataFrame(["👋 Hello !", "Take a look ⤵️", "Let's go ✊️"])
TableReport(df)
```

### Expected Results

Column stats can represent emojis.

### Actual Results

```python
python3.10/site-packages/skrub/_reporting/_plotting.py:42: UserWarning: Glyph 9994 (\N{RAISED FIST}) missing from font(s) DejaVu Sans.
  fig.savefig(buffer, format="svg", bbox_inches="tight")
python3.10/site-packages/skrub/_reporting/_plotting.py:42: UserWarning: Glyph 10549 (\N{ARROW POINTING RIGHTWARDS THEN CURVING DOWNWARDS}) missing from font(s) DejaVu Sans.
  fig.savefig(buffer, format="svg", bbox_inches="tight")
python3.10/site-packages/skrub/_reporting/_plotting.py:42: UserWarning: Glyph 128075 (\N{WAVING HAND SIGN}) missing from font(s) DejaVu Sans.
  fig.savefig(buffer, format="svg", bbox_inches="tight")
```

Column stats use a particular font that is unable to represent emojis.

### Versions

```shell
skrub version 0.3.1

Dependencies:
      sklearn: 1.5.1
          pip: 24.0
   setuptools: 70.3.0
        numpy: 2.0.0
        scipy: 1.14.0
       Cython: None
       pandas: 2.2.2
   matplotlib: 3.9.1
       joblib: 1.4.2
threadpoolctl: 3.5.0
```

jeromedockes:
Fixes #1095 

previously in the matplotlib plots, we used the default behavior for text which is that matplotlib itself renders the text (the matplotlib output svg contains svg paths rather than text)

this is simple because then matplotlib knows the size taken by the text and can set the correct viewbox so we don't need any processing of the svg afterwards.

but the drawbacks are that:

- it makes the plots bigger (around 40% I would say)
- matplotlib is not as good as a browser at finding the right font for each glyph and rendering the text. at the moment we don't specify a font stack for matplotlib so some glyphs will be missing and replaced by empty boxes, and even if we do the result may be suboptimal. This is reported for emojis in #1095 but it will also occur for many characters outside of the default font on the system where the report is generated -- eg in my machine the default matplotlib config uses dejavu which misses Chinese and Japanese characters.

the alternative, chosen in this PR, is to ask matplotlib to put the labels as text in the svg, in which case the application which displays it (in our case a web browser) is responsible for rendering it. the main drawback is that matplotlib does not know how much space the text will take, so we need to adjust the plot's viewbox and width and height with javascript after the page loads.

not directly related, but this pr also improves a bit how the text is truncated for right-to-left scripts, so that the ellipsis ("...") is displayed on the correct side of the text (in more cases). and it also normalizes whitespace so that we don't have line breaks in matplotlib labels

## main

screenshot_2024-09-30T14:44:34+02:00

## this PR

screenshot_2024-09-30T14:44:48+02:00

<details>
<summary> wider screenshots</summary>

## main

screenshot_2024-09-30T14:44:28+02:00

## this pr

screenshot_2024-09-30T14:43:33+02:00

</details>


jeromedockes:
Ok this one is ready for review. @jovan-stojanovic  could you check that it fixes #1095 for you?

jovan-stojanovic:
Thanks @jeromedockes, works perfectly well (verified on more than 4 million tweets)! Merging when the conflicts are solved

GaelVaroquaux (approved):
Looks great!

On small comment: can you add a comment on the fact that there is direction-changing character which could be picked up as a security risk.

jovan-stojanovic (approved):

jeromedockes:
> On small comment: can you add a comment on the fact that there is direction-changing character which could be picked up as a security risk.

thanks @GaelVaroquaux in the tests I replaced the literal bidirectional characters by unicode escape sequences so that the issue doesn't come up, with a comment explaining why this causes the display to look messed up in the source code

I also added an example in the [script](https://github.com/skrub-data/skrub/pull/1097/files#diff-18a72af08c2a87fdb1d443c7108685ef68fdb2430df18a7721836a866cb21f8aR54) that generates example reports for testing that we can visually check

jeromedockes:
> Thanks @jeromedockes, works perfectly well (verified on more than 4 million tweets)! Merging when the conflicts are solved

thanks a lot for checking @jovan-stojanovic !

GaelVaroquaux:
Great! Merged!

PR #1101, opened 2024-10-07
make sample table sequentially focusable when no cell is active

jeromedockes:
currently before clicking on any of the cells it is not possible to bring focus to the sample table with the keyboard. this PR makes the table sequentially focusable (with Tab); when it receives focus it activates the first cell


glemaitre (approved):
Nice. If you can only solve the merge conflict.

GaelVaroquaux:
I fixed the merge conflict. Launching auto-merge

PR #1102, opened 2024-10-07
make the 'column name' column always visible in summary stats table

jeromedockes:
The column containing the column name in the summary statistics tab panel sticks on the left of the table, so that when we scroll horizontally we can still see (the end of) the column names


GaelVaroquaux (approved):
LGTM. Thanks!!

PR #1098, opened 2024-10-01
base sizes in report on font size of parent element

jeromedockes:
also the tables are compressed a bit more

PR #1106, opened 2024-10-07
add missing changelog entry

jeromedockes:
as mentioned in this [comment](https://github.com/skrub-data/skrub/pull/1098#issuecomment-2387959764) a previous pr was missing a changelog entry

ATM the bot only complaines about the changelog if some tests have been modified, which causes a few false negatives. shall we make it check the changelog for all PRs, given that silencing false positives is easy with the "changelog not needed" label?


GaelVaroquaux (approved):
LGTM. Thanks a lot!

PR #1107, opened 2024-10-08
make column filter select smaller in tablereport

jeromedockes:
reduces the width and height of the column filter in the tablereport. it is a little less obvious what it does now but the upside is there is a smaller chance that it needs to wrap to a different line than the tabs


GaelVaroquaux (approved):
Absolutely useful!!

LGTM

PR #1108, opened 2024-10-08
add functions to patch the _repr_html_ of pandas & polars dataframes

jeromedockes:
With 

```python
skrub.patch_display()
```

the default HTML representations of pandas & polars dataframes are replaced with TableReports

it can be undone with

```python
skrub.unpatch_display()
```

`patch_display` can be called before or after importing the libraries (it will import them if they are installed but not imported yet. as this is only useful for interactive work I don't think it is a problem)

still need to add tests & an example but first I would like opinions on whether we actually want something like this or not

TODO:
- [x] changelog
- [x] tests
- [ ] use in more examples?
- [x] mention in tablereport docstring
- [ ] the drawback is it is slower than the default displays. Limit the number of columns for which we create plots?

[Example](https://output.circle-artifacts.com/output/job/88bb3835-6978-4982-b198-bafef2e78783/artifacts/0/doc/auto_examples/00_getting_started.html#generating-an-interactive-report-for-a-dataframe)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -12,16 +12,19 @@ Ongoing development
 Skrub is a very recent package.
 It is currently undergoing fast development and backward compatibility is not ensured.
 
-Major changes
--------------
-
-Minor changes
--------------
+New features
+------------
 
 * The :func:`patch_display` function has been added. It changes the display of
   pandas and polars dataframes in jupyter notebooks to replace them with a
   :class:`TableReport`. This can be undone with :func:`unpatch_display`.
   :pr:`1108` by :user:`Jérôme Dockès <jeromedockes>`
+
+Major changes
+-------------
+
+Minor changes
+-------------
 
 * The column filter selection dropdown in the tablereport is smaller and its
   label has been removed to save space. :pr:`1107` by :user:`Jérôme Dockès
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -18,6 +18,11 @@ Major changes
 Minor changes
 -------------
 
+* The :func:`patch_display` function has been added. It changes the display of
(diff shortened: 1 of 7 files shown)


skrub/_reporting/_patching.py
```diff
@@ -0,0 +1,83 @@
...
+    with a TableReport.
+
+    It can be undone with ``skrub.unpatch_display()``.
+
+    Parameters
+    ----------
+    pandas : bool, optional (default=True)
+        If False, do not override the displays for pandas dataframes.
+    polars : bool, optional (default=True)
+        If False, do not override the displays for polars dataframes.
```
GaelVaroquaux, line 65: Can you add a "See also" to the TableReport (and reference this function as a see also in the docstring of the TableReport)

skrub/_reporting/_patching.py
```diff
@@ -0,0 +1,83 @@
...
+    This function restores the default HTML displays of pandas and polars
+    DataFrames.
+
+    Parameters
+    ----------
+    pandas : bool, optional (default=True)
+        If False, do not restore the displays for pandas dataframes.
+    polars : bool, optional (default=True)
+        If False, do not restore the displays for polars dataframes.
+    """
```
GaelVaroquaux, line 82: See also, also :)

jeromedockes:
ok this one is ready for review :)

CHANGES.rst
```diff
@@ -18,6 +18,11 @@ Major changes
 Minor changes
 -------------
 
+* The :func:`patch_display` function has been added. It changes the display of
```
GaelVaroquaux, line 21: I think that this entry should not be under "minor change". Maybe a section "new features"?

GaelVaroquaux (approved):

PR #1109, opened 2024-10-08
add column_associations public function

jeromedockes:
adds `column_associations` (the function that computes the pairwise cramer v stats shown in the tablereport) as a function accessible in the public api


Vincent-Maladiere (approved):
Very nice! Should we append this function to the "Generating an HTML report" API reference section?

By the way, should we have an entry of the TableReport in the user guide of the documentation? This is such a major feature, we could talk about it a little. WDYT?

Issue #925, opened 2024-06-03
Shorthand for getting only the preprocessing part of the TableVectorizer


jeromedockes:
### Problem Description

Sometimes we may want to apply the preprocessing/cleaning steps of the TableVectorizer (parsing datetimes, handling pandas extension dtypes, etc.), while handling the actual encoding in separate pipeline steps.
This will probably become more relevant when the Recipe (or whatever its name will be) is introduced: we can use it to build exactly the pipeline we want, but we would still like to apply the default cleaning done by the TableVectorizer

If this sounds like a plausible use-case maybe we could have a shorthand for

```python
TableVectorizer(
    high_cardinality_transformer="passthrough",
    low_cardinality_transformer="passthrough",
    datetime_transformer="passthrough",
    numeric_transformer="passthrough",
    specific_transformers=(),
)
```

maybe 

```python
TableSkrubber()
```

### Feature Description

...

jeromedockes:
some examples of the kind of cleaning the tablevectorizer does:

```python

>>> import pandas as pd
>>> from skrub import TableVectorizer


>>> skrubber = TableVectorizer(
...     high_cardinality_transformer="passthrough",
...     low_cardinality_transformer="passthrough",
...     datetime_transformer="passthrough",
...     numeric_transformer="passthrough",
...     specific_transformers=(),
... )

>>> df = pd.DataFrame({
...     'a': ['2020-01-02', '2020-01-03'],
...     'b': ['2.2', 'nan'],
...     'c': [1.5, pd.NA],
...     'd': [True, False],
...     'e': pd.Series([4.5, 'a'], dtype='category'),
... })
>>> df
            a    b     c      d    e
0  2020-01-02  2.2   1.5   True  4.5
1  2020-01-03  nan  <NA>  False    a
>>> df.dtypes
a      object
b      object
c      object
d        bool
e    category
dtype: object
>>> df['e'].cat.categories
Index([4.5, 'a'], dtype='object')

>>> skrubbed = skrubber.fit_transform(df)
>>> skrubbed
           a    b    c    d    e
0 2020-01-02  2.2  1.5  1.0  4.5
1 2020-01-03  NaN  NaN  0.0    a
>>> skrubbed.dtypes
a    datetime64[ns]
b           float32
c           float32
d           float32
e          category
dtype: object
>>> skrubbed['e'].cat.categories
Index(['4.5', 'a'], dtype='object')
```

GaelVaroquaux:
[...]

PR #1115, opened 2024-10-17
Added DropNullColumn transformer to remove columns that contain only nulls 

### LINKED ISSUE #1110: Add a `DropEmpty` flag to `tabular_learner`
### Problem Description

I am currently working with tables that may contain columns that contain only null/missing values. I also need to perform imputation (with sklearn's SimpleImputer) over all the columns to fill nulls there. 

A simplified version of my pipeline is this:

```py
import polars as pl
from sklearn.pipeline import Pipeline
from sklearn.impute import SimpleImputer
from sklearn.preprocessing import StandardScaler


df_ = pl.DataFrame({"a": [1,2,3], "empty":[None, None, None]})

num_transformer = Pipeline(
    steps=[
        ("imputer", SimpleImputer(strategy="mean", keep_empty_features=False)),
        ("scaler", StandardScaler()),
    ]
)

# Transformer for high cardinality categorical features 
high_card_transformer = Pipeline(
    steps=[
        (
            "imputer",
            SimpleImputer(strategy="most_frequent", keep_empty_features=False),
        ),
        ("onehot_encoder", OneHotEncoder(handle_unknown="ignore", sparse_output=False)),
    ],
)

low_card_transformer = Pipeline(
    steps=[
        (
            "imputer",
            SimpleImputer(strategy="most_frequent", keep_empty_features=False),
        ),
        (
            "onehot_encoder",
            OneHotEncoder(handle_unknown="ignore", sparse_output=False),
        ),
    ],
)
preprocessor = TableVectorizer(
    low_cardinality=low_card_transformer,
    high_cardinality=high_card_transformer,
    numeric=num_transformer,
)

preprocessor.fit_transform(df_)

```

This raises an exception on column `empty` because no values are found. 

### Feature Description

Rather than raising an exception at the end of the pipeline, which makes debugging more complex (the error messages were less than helpful), it would be nice to have an automated check of all columns before any operation is done to screen for empty columns. 

Some possible improvements could be:
- A warning to the user ("Column something something is empty") 
- Dropping the column immediately
- Dropping the column if the fraction of null values is larger than a given threshold

This should also be added to the `tabular_learner` so that if such a situation occurs, it doesn't fail for some weird reason. 

### Alternative Solutions

Setting the SimpleImputer parameter `keep_empty_features=True` seems to fix the problem (or at least, the code does not fail anymore). Null values are replaced by 0. 

### Additional Context

_No response_

rcap107:
fixes #1110

DropNullColumn (provisional name) takes as input a column, and drops it if all the values are nulls or nans. TableVectorizer was also updated with a `drop_null_columns` flag set to `False` by default; if the flag is set to `True`, the DropNullColumn is added as a processing step for all columns. 

I've also added `drop` and `is_all_null` to `_common.py`, though I am not sure if they should go there. Maybe `is_all_null` can stay in the `DropNullColumn` file. 

The test I wrote passes, but I'm not sure if it's good enough. 

The documentation is still missing.


skrub/tests/test_dropnulls.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
+import pandas as pd
+import polars as pl
+import pytest
+
+from numpy.testing import assert_array_equal
+
+from skrub import TableVectorizer
+from skrub._dataframe import _common as ns
+from skrub._dropnull import DropNullColumn
```
TheooJ, line 10: I would call the module `_drop_null.py`

skrub/tests/test_dropnulls.py
```diff
@@ -0,0 +1,58 @@
...
+            "value_null": [
+                None,
+                None,
+                None,
+            ],
+        }
+    )
+
+
+def test_drop_nullcolumn(main_table):
```
TheooJ, line 36: All the tests in this file are going to test DropNullColumn so I'd avoid this test name

skrub/_dataframe/_common.py
```diff
@@ -824,6 +826,25 @@ def _to_categorical_polars(col):
     col = to_string(col)
     return _cast_polars(col, pl.Categorical())
 
+@dispatch
+def is_all_null(col):
```
TheooJ, line 830: You need a test for this in `test_common.py`

skrub/_dataframe/_common.py
```diff
@@ -824,6 +826,25 @@ def _to_categorical_polars(col):
...
+    raise NotImplementedError()
+
+
+@is_all_null.specialize("pandas", argument_type="Column")
+def _is_all_null_pandas(col):
+    return bool(col.isna().all())
```
TheooJ, line 836:
```suggestion
    return col.isna().all()
```

I think it's already a boolean here

skrub/_dataframe/_common.py
```diff
@@ -1187,3 +1208,15 @@ def with_columns(df, **new_cols):
     cols = {col_name: col(df, col_name) for col_name in column_names(df)}
     cols.update({n: make_column_like(df, c, n) for n, c in new_cols.items()})
     return make_dataframe_like(df, cols)
+
+@dispatch
+def drop(obj, col):
```
TheooJ, line 1213: I don’t know if `drop` is necessary, you could directly use skrub selectors: 
`df = s.select(df, ~s.cols(col))`

skrub/_table_vectorizer.py
```diff
@@ -191,6 +192,9 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
         similar functionality to what is offered by scikit-learn's
         :class:`~sklearn.compose.ColumnTransformer`.
 
+    drop_null_columns : bool, default=False
```
TheooJ, line 195: Do we want it to be `True` by default ?

skrub/tests/test_dropnulls.py
```diff
@@ -0,0 +1,58 @@
+import numpy as np
+import pandas as pd
+import polars as pl
+import pytest
+
+from numpy.testing import assert_array_equal
+
+from skrub import TableVectorizer
+from skrub._dataframe import _common as ns
```
TheooJ, line 9:
```suggestion
from skrub import _dataframe as sbd
```

skrub/tests/test_dropnulls.py
TheooJ: (this is the same thing, we need to uniformize this in the codebase)

skrub/tests/test_dropnulls.py
```diff
@@ -0,0 +1,58 @@
...
+    assert dn.fit_transform(main_table["value_null"]) == []
+
+    assert_array_equal(ns.to_numpy(ns.col(main_table, "idx")), np.array([1, 2, 3]))
+
+def test_transform_nullcolumn(main_table):
+    main_table_dropped = ns.drop(main_table, "value_null")
+    main_table_dropped = ns.drop(main_table_dropped, "value_nan")
+    
+    # Don't drop null columns
+    tv = TableVectorizer(drop_null_columns=False)
```
TheooJ, line 49: This test needs to go in the TV test file IMO

TheooJ (commented):
Hi @rcap107 ! I made a first pass and have a few comments :

- Personnally I like the name `DropNullColumn`, I think it’s clear what it does !
- I would rename the file `_drop_null.py`
- Make sure you `pre-commit run --all-files` before pushing, it seems to be what’s breaking the CI for you here
- I think `is_all_null` could be placed in the `DropNullColumn` file if it’s only used there for now, but I could also see it being in `_common.py`

skrub/tests/test_dropnulls.py
TheooJ: (This is the same thing, we need to uniformize the codebase)

rcap107:
Hi @TheooJ, thanks a lot for the comments! I'll address them and update the PR :+1:

skrub/_dataframe/_common.py
rcap107: `col.isna().all()` of type `numpy.bool`, I'm not sure if that makes a difference downstream, I decided to cast it to bool to avoid issues.

skrub/_table_vectorizer.py
rcap107: That should be discussed with others I think

skrub/tests/test_dropnulls.py
rcap107: I can move it :+1:

skrub/tests/test_drop_nulls.py
```diff
@@ -0,0 +1,87 @@
...
+    )
+
+    assert_array_equal(
+        sbd.to_numpy(sbd.col(drop_null_table, "value_almost_nan")),
+        np.array([2.5, np.nan, np.nan]),
+    )
+
+    # assert_array_equal(
+    #     sbd.to_numpy(sbd.col(drop_null_table, "value_almost_null")),
+    #     np.array(["almost", None, None]),
```
rcap107, line 60: Not sure how to write this check so that it works with either pandas or polars

TheooJ: You could use `df_module` as a fixture in the test by adding it to the arguments, then comparing series instead of numpy arrays:

```
df_module.assert_column_equal(
    sbd.col(drop_null_table, "value_almost_null"),
    df_module.make_column("value_almost_null", ["almost", None, None]),
)
```

TheooJ: This also circumvents that depending on the version of pandas, null values are not treated the same

jeromedockes:
the failure in the min-deps environment is not related to this pr; the fix is in #1122

skrub/_dataframe/_common.py
jeromedockes: I think that's a good idea

skrub/_dataframe/_common.py
```diff
@@ -825,6 +826,27 @@ def _to_categorical_polars(col):
...
+@is_all_null.specialize("polars", argument_type="Column")
+def _is_all_null_polars(col):
+    if col.dtype == pl.Null:
+        return True
+    elif col.dtype.is_numeric() and col.is_nan().all():
+        return True
+    # col is non numeric
+    elif col.null_count() == col.len():
+        return True
+
```
jeromedockes, line 848: when it's not all null we want to return False rather than None

skrub/_drop_null.py
```diff
@@ -0,0 +1,54 @@
...
+
+__all__ = ["DropNullColumn"]
+
+
+class DropNullColumn(SingleColumnTransformer):
+    """Drop a single column if it contains only null values."""
+
+    def __init__(self):
+        super().__init__()
+        self._is_fitted = False
```
jeromedockes, line 15:
```suggestion
```

to follow the scikit-learn requirements, estimators must not do anything in their __init__ except store the parameters. here you don't have any parameters to store so you don't need to define __init__

skrub/_drop_null.py
```diff
@@ -0,0 +1,54 @@
...
+            column : Pandas or Polars series. The input column to check.
+            y : None. Ignored.
+
+        Returns:
+            The input column, or an empty list if the column contains only null values.
+        """
+        del y
+
+        self._is_fitted = True
+        return self.transform(column)
```
jeromedockes, line 30: we want the schema of the output to be the same after every call to `transform`. at the moment, the check is done during transform so we decide each time transform is called, depending on the input, if we want to drop it or not. instead, we want to check if the column passed to `fit` is all null, if yes, we always drop during transform (even if we later get a column with non-null values) if no, we never drop (even if we later get a column full of nulls).

so it would look like:

```python
def fit_transform(self, column, y=None):
    self.drop_ = sbd.is_all_null(column)
    return self.transform(column)

def transform(self, column):
    check_is_fitted(self, "drop_")
    return [] if self.drop_ else column

# __sklearn_is_fitted__ not needed anymore
```

skrub/_table_vectorizer.py
```diff
@@ -536,6 +542,9 @@ def add_step(steps, transformer, cols, allow_reject=False):
         cols = s.all() - self._specific_columns
 
         self._preprocessors = [CheckInputDataFrame()]
+        if self.drop_null_columns:
+            add_step(self._preprocessors, DropNullColumn(), cols, allow_reject=True)
```
jeromedockes, line 546: - we may want to insert it after CleanNullStrings? so that if the column becomes full of nulls after converting `"N/A"` to `null` it will be dropped. also it's not important but your transformer never raises a RejectColumn exception so `allow_reject` has no effect you don't need it here and can leave the default

skrub/_dataframe/_common.py
```diff
@@ -825,6 +826,27 @@ def _to_categorical_polars(col):
...
+@is_all_null.specialize("pandas", argument_type="Column")
+def _is_all_null_pandas(col):
+    return bool(col.isna().all())
+
+
+@is_all_null.specialize("polars", argument_type="Column")
+def _is_all_null_polars(col):
+    if col.dtype == pl.Null:
+        return True
+    elif col.dtype.is_numeric() and col.is_nan().all():
```
jeromedockes, line 843: there could be a mix of nan and null which I think would fall through the cracks here.

For simplicity I think we could do:

```python
def is_all_null(col): return all(is_null(col))
```

but if you really want the specialization for polars to get the speedup with null count it could be

```python
def _is_all_null_polars(col):
    if col.dtype == pl.Null:
        return True
    if col.null_count() == col.len():
        return True
    return all(is_null(col))
```

skrub/_drop_null.py
```diff
@@ -0,0 +1,54 @@
+# drop columns that contain all null values
```
jeromedockes, line 1: nitpick: can we have the same name for the class, the module and the test module? eg `DropNullColumns`, `drop_null_columns`, `test_drop_null_columns` (or the singular version if you prefer)

skrub/_table_vectorizer.py
```diff
@@ -542,16 +542,21 @@ def add_step(steps, transformer, cols, allow_reject=False):
...
 
         self._preprocessors = [CheckInputDataFrame()]
-        if self.drop_null_columns:
-            add_step(self._preprocessors, DropNullColumn(), cols, allow_reject=True)
 
-        for transformer in [
+        transformer_list = [
             CleanNullStrings(),
+        ]
+        if self.drop_null_columns:
```
rcap107, line 549: This is ugly, but I'm not sure how to write this better.

skrub/_table_vectorizer.py
rcap107: I added it after CleanNullStrings, but I think I did it in an ugly way, maybe it can be fixed

skrub/_dataframe/_common.py
```diff
@@ -825,6 +826,26 @@ def _to_categorical_polars(col):
...
+@is_all_null.specialize("pandas", argument_type="Column")
+def _is_all_null_pandas(col):
+    return all(is_null(col))
+
+
+@is_all_null.specialize("polars", argument_type="Column")
+def _is_all_null_polars(col):
+    if col.dtype == pl.Null:
+        return True
+    # col is non numeric
```
jeromedockes, line 843: not sure I understand the comment

skrub/_drop_null_column.py
```diff
@@ -0,0 +1,49 @@
...
+
+        Args:
+            column : Pandas or Polars series. The input column to check.
+
+        Returns:
+            The input column, or an empty list if the column contains only null values.
+        """
+        check_is_fitted(
+            self,
+        )
```
jeromedockes, line 41:
```suggestion
        check_is_fitted(self)
```

skrub/tests/test_drop_null_column.py
```diff
@@ -0,0 +1,74 @@
...
+                None,
+            ],
+        }
+    )
+
+
+def test_single_column(drop_null_table, df_module):
+    """Check that null columns are dropped and non-null columns are kept."""
+    dn = DropNullColumn()
+    assert dn.fit_transform(drop_null_table["value_nan"]) == []
```
jeromedockes, line 45: in theory not to rely on polars or pandas specific behavior we would write `sbd.col(drop_null_table, "value_nan")` rather than `drop_null_table["value_nan"]`

skrub/tests/test_drop_null_column.py
```diff
@@ -0,0 +1,74 @@
...
+
+    df_module.assert_column_equal(
+        sbd.col(drop_null_table, "value_almost_nan"),
+        df_module.make_column("value_almost_nan", [2.5, np.nan, np.nan]),
+    )
+
+    df_module.assert_column_equal(
+        sbd.col(drop_null_table, "value_almost_null"),
+        df_module.make_column("value_almost_null", ["almost", None, None]),
+    )
```
jeromedockes, line 60: not sure I understand the 3 tests above, the transformer does not seem to be used? we just compare a column from the table to a new one we construct

skrub/tests/test_drop_null_column.py
```diff
@@ -0,0 +1,74 @@
...
+
+
+def test_drop_null_column(drop_null_table):
```
jeromedockes, line 63: I understand that you put it here to use the fixture defined in this module but I think this test should probably go in `test_table_vectorizer.py`

skrub/_table_vectorizer.py
jeromedockes: I vote for true by default -- there's nothing we can learn from a completely empty column.

if it is False by default, I think it should be set to True in the `tabular_learner`

skrub/_dataframe/_common.py
rcap107: it was a leftover from the previous version, I rewrote the comments

skrub/tests/test_table_vectorizer.py
```diff
@@ -506,8 +528,11 @@ def test_changing_types(X_train, X_test, expected_X_out):
     """
     table_vec = TableVectorizer(
         # only extract the total seconds
-        datetime=DatetimeEncoder(resolution=None)
+        datetime=DatetimeEncoder(resolution=None),
+        # True by default
```
rcap107, line 532: I set this to false to keep the original behavior with no DropNullColumns. Given that the default value is True, should I change the test so that the "default behavior" is what is tested here?

jeromedockes: I think it's ok the way you did it

skrub/_table_vectorizer.py
```diff
@@ -389,7 +394,7 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
     ``ToDatetime()``:
 
     >>> vectorizer.all_processing_steps_
-    {'A': [Drop()], 'B': [OrdinalEncoder()], 'C': [CleanNullStrings(), ToFloat32(), PassThrough(), {'C': ToFloat32()}]}
+    {'A': [Drop()], 'B': [OrdinalEncoder()], 'C': [CleanNullStrings(), DropNullColumn(), ToFloat32(), PassThrough(), {'C': ToFloat32()}]}
```
jeromedockes, line 397: I wonder if we should have an 'if' in there eg call it 'DropColumnIfNull' to make it clearer that this may be a no-op when reading the list of transformations

rcap107: I like the idea, it's definitely clearer. I renamed the object.

skrub/_dataframe/tests/test_common.py
```diff
@@ -557,6 +557,30 @@ def test_to_categorical(df_module):
         assert list(s.cat.categories) == list("ab")
 
 
+def test_is_all_null(df_module):
+    """Check that is_all_null is evaluating null counts correctly."""
+
+    # Check that all null columns are marked as "all null"
+    assert ns.is_all_null(df_module.make_column("all_null", [None, None, None]))
+    assert ns.is_all_null(df_module.make_column("all_nan", [np.nan, np.nan, np.nan]))
```
jeromedockes, line 565:
```suggestion
    assert ns.is_all_null(df_module.make_column("all_nan", [np.nan, np.nan, np.nan]))
    assert ns.is_all_null(df_module.make_column("all_nan_or_null", [np.nan, np.nan, None]))
```

skrub/_drop_column_if_null.py
```diff
@@ -0,0 +1,43 @@
...
+
+
+class DropColumnIfNull(SingleColumnTransformer):
+    """Drop a single column if it contains only Null, NaN, or a mixture of null
+    values. If at least one non-null value is found, the column is kept."""
+
+    def fit_transform(self, column, y=None):
+        """Fit the encoder and transform a column.
+
+        Args:
```
jeromedockes, line 17: can you check the formatting of the other estimators in the library, here the section header would look like 
```
Parameters
----------
```

not super important here as it is a private class, but if we decide to make it public it will be important because sphinx relies on this formatting to produce the reference html documentation

jeromedockes (approved):
thanks a lot @rcap107 !! LGTM once the last couple of nitpicks are addressed :tada:

Vincent-Maladiere:
Hey @rcap107, you haven't addressed my comment in https://github.com/skrub-data/skrub/issues/1110#issuecomment-2408869551. I'm -1 to drop all nan columns in the table vectorizer by default, as these might be useful information after concatenation.

rcap107:
> Hey @rcap107, you haven't addressed my comment in [#1110 (comment)](https://github.com/skrub-data/skrub/issues/1110#issuecomment-2408869551). I'm -1 to drop all nan columns in the table vectorizer by default, as these might be useful information after concatenation.

I'm fine with having the drop flag false by default, I don't see how having a concatenation of empty columns brings in any information after concatenation :thinking: If the point is acting as indicator, then the user should just add the column as constant, which would be more informative than a null value anyway (and it wouldn't trigger the drop). 

I can imagine a situation where both tables A and B have a column `col_1`, and `col_1` is all null in A, but isn't in B. With the current behavior, `col_1` is dropped from A and kept in B, and if I try to concatenate A and B I'd expect the behavior to be "add a column called `col_1` to the final table and fill the rows of A with nulls", unless an exception is raised if the same set of columns must be provided (which may be the case and I'm just not aware of it). 

In any case I can modify the code to have the three choices rather than a binary flag.

Vincent-Maladiere:
>  then the user should just add the column as constant

My intuition is that we should avoid assuming users would "just do" things and try to prevent this kind of mistake.

> I can imagine a situation where both tables A and B have a column col_1, and col_1 is all null in A, but isn't in B ...

Consider a situation where you have a TableVectorizer in a scikit-learn pipeline, trained in a production environment every week. If you built this pipeline assuming you would have access to some column "col_1", but for some reason, the data production system now only produces NaN values for this column (e.g. this column has been deprecated, which happened quite often), then this pipeline would raise a "column is missing" error that would be tricky to understand, because most users wouldn't be familiar with the internals of the TableVectorizer. Having a clear error or a warning would significantly help debugging and maintenance in such a scenario. WDYT?

rcap107:
Yes, I agree that having a more informative error would be better. 

I am changing the code so that it is possible to set different either `drop`, `ignore` or `raise`. I modified the init of DropColumnIfNull. Sidenote, where should the error checking for that value be? In DropNull or in TableVectorizer, or both? 

I could also add a `warn` strategy, but wouldn't it be better to set it so that it triggers when in debug mode? I am not sure how to do that. 

For the default strategy, we can use `ignore` to keep null columns.

Vincent-Maladiere:
> Sidenote, where should the error checking for that value be? In DropNull or in TableVectorizer, or both?

I think the most practical would be raising a specific error in `DropNull`. The way I picture it:
- We add a new argument in `TableVectorizer`, like `all_nulls_cols` whose options are "warn" (default), "ignore", "drop", "raise"
- `DropNull` then takes the same argument and default

Also, "warn" looks like a good default, but that's up for discussion.

WDYT?

> [...] when in debug mode?

I'm not sure to understand what debug mode is?

jeromedockes:
I think "ignore" is a bit ambiguous -- if I ignore a column does it mean I don't use it / leave it out of the output? so I would suggest "keep" instead

rcap107:
> > [...] when in debug mode?
> 
> I'm not sure to understand what debug mode is?
nevermind


> * `DropNull` then takes the same argument and default
> 
> Also, "warn" looks like a good default, but that's up for discussion.

We also need to decide whether it's "warn and drop", or "warn and keep", and be clear about the behavior.

skrub/tests/test_table_vectorizer.py
```diff
@@ -778,11 +777,7 @@ def test_drop_null_column():
 
     # Raise exception if a null column is found
     with pytest.raises(
```
rcap107, line 779: This test is still failing because the TableVectorizer is not raising the correct exception and I don't know how to make it do that.

rcap107:
I have updated the code to have "warn and keep" as the default behavior, I think it's the version that makes the most sense. 

At the moment the only problem I have is that I don't know how to raise the proper exception from TableVectorizer. In  DropNull I am raising RejectColumn, but then I don't know how to propagate it correctly,

skrub/tests/test_table_vectorizer.py
jeromedockes: [here](https://github.com/skrub-data/skrub/pull/1115/files#diff-915f4c6f453724cbd65c06d5c7f55a1aed027f2b5649d15ec4ec3b9edb90ad70R77) raise a ValueError instead of RejectColumn

jeromedockes: rejectcolumn is a way to signify to the tablevectorizer "I'm not the right transformer for this column, don't apply me here".

rcap107: Gotcha, fixed

jeromedockes:
I find it a bit weird to have a DropColumnIfNull that does not drop the column and just raises an exception. maybe it should be named something like CheckNulls or something?

jeromedockes:
> Consider a situation where you have a TableVectorizer in a scikit-learn pipeline, trained in a production environment every week. If you built this pipeline assuming you would have access to some column "col_1", but for some reason, the data production system now only produces NaN values for this column

I'm not sure I understand -- in any case the tablevectorizer chooses the estimators during fit so the schema of the output can change every week in this scenario. eg if a column has one more unique value than the previous week it can change from a one-hot encoding to a gap encoding. or for example if you had `high_cardinality="drop"` those columns would also be dropped or kept depending on their content

jeromedockes:
so if you want consistent output schema (number of columns, names and types) across training the tablevetorizer is not what you want anyway

Vincent-Maladiere:
> I find it a bit weird to have a DropColumnIfNull that does not drop the column and just raises an exception. maybe it should be named something like CheckNulls or something?

Yes, I agree this sounds weird. I recognize dropping NaN columns –which are usually useless– looks good. What about having the transformer drop by default, but allowing the user to pass other options as arguments in the `TableVectorizer`?

> I'm not sure I understand --  in any case the tablevectorizer chooses the estimators during fit [...]

Right, I was pointing out that even outside of the `TableVectorizer` (e.g. in the `ColumnTransformer`), the `DropColumnIfNull` could raise surprising mistakes.

It's true though that this issue broadly applies to the `TableVectorizer`, hence we should encourage people to perform checks in production (with dedicated check functions? There might be some potential here for production usage).

rcap107:
Following today's discussion with @Vincent-Maladiere and @jeromedockes, I reverted the changes back to the old version (with the simple flag)

jeromedockes (approved):
LGTM! thanks again @rcap107 

(the codecov report is bogus)

I'll let @Vincent-Maladiere do a final review & merge

Vincent-Maladiere (commented):
In our IRL discussion, haven't we agreed on a threshold for the null ratio in the column (which is 1.0 by default)?

Apart from that, and to keep in mind for later, the items we discussed were:
1. The ability to freeze TableVectorizer column-to-transformer mapping. It would help to obtain consistent results for retraining in an automated environment, and having more sensible errors to debug.
2.  Decoupling the check/cleaning part of the TableVectorizer (which comes before the vectorizing part) so that it can be used as a standalone object wherever.

jeromedockes:
> In our IRL discussion, haven't we agreed on a threshold for the null ratio in the column (which is 1.0 by default)?

we definitely did; I had understood it would be tackled in a separate PR

+1 for points 1. and 2. -- let's open a separate issue for 1. and there is #925 for 2.

rcap107:
I also understood that the threshold would be added in a separate issue

Vincent-Maladiere (approved):
LGTM then, thanks @rcap107 :)

jeromedockes:
yay :tada: !! thanks @rcap107 !

rcap107:
:tada:

Issue #991, opened 2024-07-10
Regroup developer docs in the main sphinx documentation


GaelVaroquaux:
that's a good point. maybe in a dedicated PR? there will be several things to regroup in there, such as the instructions about making a release

_Originally posted by @jeromedockes in https://github.com/skrub-data/skrub/pull/984#discussion_r1672545065_

rcap107:
This issue should be tackled once we refactor the dev documentation (cfr #1582)

PR #1116, opened 2024-10-17
[ENH] Dispatch aggregate, refactor AggJoiner & AggTarget

TheooJ:
The goal of this PR is to dispatch `aggregate`, currently written in two files, by directly implementing it in `_agg_joiner.py`.

Following discussions with @Vincent-Maladiere and @jeromedockes, `AggJoiner` and `AggTarget` now require the `operations` parameter by default, and will try to apply all operations on all columns — as opposed to now, where columns are separated in categorical and numeric and only some operations are computed on each category.

I’m planning on doing follow ups to completely remove the _pandas.py, _polars.py, _namespace.py files, and on refactoring `AggTarget` with cross-fitting.


skrub/_agg_joiner.py
```diff
@@ -6,49 +6,140 @@
...
-        if op_root not in ALL_OPS:
-            raise ValueError(f"operations options are {ALL_OPS}, got: {operation=!r}.")
 
-    return num_operations, categ_operations
+    # summing strings works in pandas, not in polars
+    num_only_operations = ["sum", "median", "mean", "std"]
+
+    key = atleast_1d_or_none(key)
+    cols_to_agg = atleast_1d_or_none(cols_to_agg)
+    operations = atleast_1d_or_none(operations)
```
Vincent-Maladiere, line 71: All of these checks have been performed upstream in `check_input` method. We should probably clarify our intent with this function; should it only be called by `AggJoiner` and `AggTarget`, or should it be publicly accessible?

skrub/_join_utils.py
```diff
@@ -84,12 +84,15 @@ def check_missing_columns(table, key, table_name):
...
         The table that should contain the columns listed in `key`.
 
-    key : list of str
+    key : list of str or selector
         List of column names, all of which must be found in `table`.
 
     table_name : str
         Name by which to refer to `table` in the error message if necessary.
     """
+    if isinstance(key, s.Selector):
```
Vincent-Maladiere, line 93: This function could be simplified with:
```python
def check_missing_columns(table, key):
    s.make_selector(key).expand(table)
    return
```
or
```python
def check_missing_columns(table, key, table_name):
    s.make_selector(key).expand(table, table_name)
    return
```
@jeromedockes I think adding a `table_name` argument to the `expand` function error message would be a nice addition for debuggability, WDYT?

Vincent-Maladiere (commented):
Hey @TheooJ, very nice effort! Here is a first pass of comments.

On a higher level:

1. Since you use the `CheckInputDataFrame` class in `fit_transform`, you should be able to remove the  `_check_dataframes` method entirely. And since `_check_inputs`  currently calls `_check_dataframes`, you have to place them in reverse order:
```
self._main_check_input = CheckInputDataFrame()
X = self._main_check_input.fit_transform(X)
self._check_inputs(X)
```
Additionally, the `check_inputs` method of the `AggTarget` could be simplified because `CheckInputDataFrame` does most of the checks.

2. You should add a `get_feature_names_out` method that returns `self.all_outputs_` for both `AggJoiner` and `AggTarget`.

skrub/_join_utils.py
jeromedockes: that sounds useful. another way would be to have Cols raise a custom exception with an attribute listing the missing columns, and then whoever is calling it and catching the exception can retreive that and format it however they see fit in their error message. It would allow less generic error messages and avoid adding the table name argument that only one selector type needs

skrub/_agg_joiner.py
```diff
@@ -6,49 +6,140 @@
...
+
+    aggregated = perform_groupby(table_to_agg, key, cols_to_agg, operations)
+
+    new_col_names = [
+        f"{col}{suffix}" if col not in key else col
+        for col in sbd.column_names(aggregated)
+    ]
+    new_col_names = _join_utils.pick_column_names(new_col_names)
+    aggregated = sbd.set_column_names(aggregated, new_col_names)
+    aggregated = s.select(aggregated, sorted(sbd.column_names(aggregated)))
```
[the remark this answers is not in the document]
TheooJ: I kept it because it was in the behavior of `aggregate`. Since operations were applied on columns in 2 steps (for value_counts / hist, and for all others), I assume this was done to put all the aggregation results on a column together.

In the new case I agree it could be unexpected, let's remove it

skrub/_join_utils.py
Vincent-Maladiere: Yes, that looks even better

skrub/_agg_joiner.py
TheooJ: You're right, they're also leftovers from the original aggregate => removing them

I think `aggregate` should only be called by `AggJoiner` and `AggTarget`, users doing groupbys are most likely to directly use pandas / polars directly IMO

skrub/_join_utils.py
TheooJ: I think this is a good idea, and that it deserves its own PR -- I wanted to only change `AggJoiner` / `AggTarget` here

skrub/_agg_joiner.py
TheooJ: The checks are now left the the Joiners, I enforced that the parameters are iterables when they arrive to `aggregate`

TheooJ:
Thanks a lot @Vincent-Maladiere !

To answer your points:
1. I can remove `_check_dataframes` and use `CheckInputDataFrame`, I just need to handle the case where `aux_table="X"` to keep the option of using the placeholder. I agree that `AggTarget`'s `check_inputs` could be simplified, I plan on cleaning it in another PR (it's also missing some tests).
2. Sure ! Is it something to do here or in another PR ? Should we open an issue so that we add this method for other transformers like `Joiner` ?
3. I'm not sure we want this. On the one hand, it might be useful for long-term compatibility when selectors are public, so that people can either use list of cols or selectors directly. On the other, while I think it makes sense to say "compute the mean on all numeric columns", you are not going to join dataframes on all numeric or all string columns with `key`.

Vincent-Maladiere:
1. and 2. can be done in this PR, it's not a huge change.

I agree that 3. use cases are rarer, but it looks weird to support selectors for the cols argument only. I might have keys that I can identify using regexes for instance, when the name of the column changes for whatever reason. Not a strong requirement, but it shouldn't be too costly to add.

TheooJ:
> 1. and 2. can be done in this PR, it's not a huge change.
> 
> 
> I agree that 3. use cases are rarer, but it looks weird to support selectors for the cols argument only. I might have keys that I can identify using regexes for instance, when the name of the column changes for whatever reason. Not a strong requirement, but it shouldn't be too costly to add.

Doing 1 & 2 here, the main issue being checking dtypes

I don't have a strong opinion about 3

TheooJ:
If think the PR is good for a second pass. I probably need to rename it and add some extra things in the changelog.

@Vincent-Maladiere I addressed points 1 & 2: simplifying `AggTarget`'s `_check_dataframes` (it was more of a complete RFC of `AggTarget`) and adding a `get_feature_names_out` method. Point 3 I didn't implement because I'm not sure it's a good idea to indicate users "you should use selectors to choose keys", so I didn't implement it in this PR. Shouldn't be a blocker for the review though and is a quite minor change IMO

skrub/_agg_joiner.py
```diff
@@ -6,57 +6,138 @@
...
+    """Aggregate the `table` by `key` and compute statistics for `cols_to_agg`.
 
     Parameters
     ----------
-    operations : list of str
-        The input operators names.
+    table : DataFrame
+        The input dataframe to aggregate.
+
+    key : iterable of str
```
jeromedockes, line 42:
```suggestion
    key : list of str
```

skrub/_agg_joiner.py
```diff
@@ -6,57 +6,138 @@
...
-            raise ValueError(f"operations options are {ALL_OPS}, got: {operation=!r}.")
+    table_to_agg = s.select(table, s.make_selector(key) | s.make_selector(cols_to_agg))
+
+    # Don't check the ID column, as it's not the one we aggregate on
+    table_to_check = s.select(table_to_agg, ~s.cols(*key))
+    not_numeric_cols = (~s.numeric()).expand(table_to_check)
+
+    num_only_op = list(set(operations).intersection(set(NUM_ONLY_OPS)))
+
+    if (len(not_numeric_cols) > 0) & (len(num_only_op) > 0):
```
jeromedockes, line 73:
```suggestion
    if not_numeric_cols and num_only_op:
```

skrub/_agg_joiner.py
```diff
@@ -6,57 +6,138 @@
...
+
+@dispatch
+def perform_groupby(table, key, cols_to_agg, operations):
+    raise NotImplementedError()
+
+
+@perform_groupby.specialize("pandas", argument_type="DataFrame")
+def _perform_groupby_pandas(table, key, cols_to_agg, operations):
+    pandas_aggfuncs = {
+        "mode": pd.Series.mode,
```
jeromedockes, line 100: I guess `groupby(key).agg('mode')` is not allowed for some reason, could you add a comment explaining why this is necessary?

skrub/_agg_joiner.py
```diff
@@ -6,57 +6,138 @@
...
+            f" following operations: {num_only_op}."
+        )
+
+    aggregated = perform_groupby(table_to_agg, key, cols_to_agg, operations)
+
+    new_col_names = [
+        f"{col}{suffix}" if col not in key else col
+        for col in sbd.column_names(aggregated)
+    ]
+    new_col_names = _join_utils.pick_column_names(new_col_names)
```
jeromedockes, line 86: why can there be some duplicates in the output column names? and if it can happen, shouldn't the deduplication be done when picking the `named_agg` / aliases?

skrub/_agg_joiner.py
```diff
@@ -130,119 +209,73 @@ class AggJoiner(TransformerMixin, BaseEstimator):
...
         """
-        X, self._aux_table = self._check_dataframes(X, self.aux_table)
-
         self._main_key, self._aux_key = _join_utils.check_key(
             self.main_key, self.aux_key, self.key
         )
         _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
         _join_utils.check_missing_columns(self._aux_table, self._aux_key, "'aux_table'")
+        _join_utils.check_missing_columns(
+            self._aux_table, s.make_selector(self.cols), ""
```
jeromedockes, line 257: why `""`?

skrub/_agg_joiner.py
```diff
@@ -130,119 +209,73 @@ class AggJoiner(TransformerMixin, BaseEstimator):
...
-                self.operations,
-            ]
-        else:
-            self._operations = self.operations
+        self._cols = s.make_selector(self.cols) - self._aux_key
 
-        self.num_operations, self.categ_operations = split_num_categ_operations(
-            self._operations
-        )
+        self._operations = atleast_1d_or_none(self.operations)
```
jeromedockes, line 263: I think operations cannot be None

skrub/_agg_joiner.py
```diff
@@ -263,17 +296,24 @@ def fit_transform(self, X, y=None):
         DataFrame
             The augmented input.
         """
-        self._check_inputs(X)
+        if isinstance(self.aux_table, str) and self.aux_table == "X":
+            self.aux_table = X
```
jeromedockes, line 300: we shouldn't modify `self.aux_table` but instead store a new fitted attribute `aux_table_` or private attribute `_aux_table`

skrub/_agg_joiner.py
```diff
@@ -263,17 +296,24 @@ def fit_transform(self, X, y=None):
...
+                " use the MultiAggJoiner instead."
+            )
+        self._aux_table = CheckInputDataFrame().fit_transform(self.aux_table)
         self._main_check_input = CheckInputDataFrame()
         X = self._main_check_input.fit_transform(X)
+        self._check_inputs(X)
 
-        skrub_px, _ = get_df_namespace(self._aux_table)
-        self.aux_table_ = skrub_px.aggregate(
+        self.aux_table_ = aggregate(
```
jeromedockes, line 312: as it is of a different nature than `aux_table` (which is not aggregated) maybe it should be given a more different name eg `aggregated_aux_table`

skrub/_agg_joiner.py
```diff
@@ -263,17 +296,24 @@ def fit_transform(self, X, y=None):
         DataFrame
             The augmented input.
         """
-        self._check_inputs(X)
+        if isinstance(self.aux_table, str) and self.aux_table == "X":
```
jeromedockes, line 299: shouldn't those checks be part of check_inputs?

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+            raise TypeError(f"X must be a dataframe, got {type(X)}")
+
+        self._main_key = atleast_1d_or_none(self.main_key)
+        _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
+
+        # `y` is copied or converted to a df to be compatible with `aggregate`
+
+        # If `y` is already a dataframe
+        if sbd.is_dataframe(y):
+            y_ = sbd.make_dataframe_like(y, sbd.to_column_list(y))
```
jeromedockes, line 485: - why do we need to make a copy? will we modify y in this case?
- if we do need it maybe add a comment to say this line does a shallow copy. maybe we shouild have a little helper function for this in `_utils` or `_dataframe`

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+        # If `y` is a named series, we convert it to a dataframe
+        elif sbd.is_column(y) and sbd.name(y) is not None:
+            y_ = sbd.make_dataframe_like(y, {sbd.name(y): y})
+        else:
+            # If `y` is an unnamed series
+            if sbd.is_column(y):
+                y = sbd.to_numpy(y).reshape(-1, 1)
+            # If `y` is a list
+            elif isinstance(y, list):
+                y = np.atleast_2d(y)
```
jeromedockes, line 495: I'm not sure that's what you want
```
>>> np.atleast_2d([1, 2, 3])
array([[1, 2, 3]])
```

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+        # Check lengths
+        if sbd.shape(y_)[0] != sbd.shape(X)[0]:
+            raise ValueError(
+                f"X and y length must match, got {sbd.shape(X)[0]} and"
+                f" {sbd.shape(y_)[0]}."
+            )
+
+        self._cols = sbd.column_names(y_)
+
+        self._operations = atleast_1d_or_none(self.operations)
```
jeromedockes, line 512: maybe we could put those checks in a function to avoid duplication between here and aggjoiner?

skrub/_agg_joiner.py
```diff
@@ -441,51 +551,39 @@ def fit_transform(self, X, y):
         y_ = self.check_inputs(X, y)
         self._main_check_input = CheckInputDataFrame()
         X = self._main_check_input.fit_transform(X)
-        skrub_px, _ = get_df_namespace(X, y_)
 
         # Add the main key on the target
-        y_[self.main_key_] = X[self.main_key_]
+        y_ = sbd.with_columns(y_, **{k: sbd.col(X, k) for k in self._main_key})
```
Vincent-Maladiere, line 556: Nice 👍

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+        y : DataFrameLike or SeriesLike or ArrayLike
+            The target to check.
+
+        Returns
+        -------
+        y_ : DataFrameLike
+            The transformed target.
+        """
+        if not sbd.is_dataframe(X):
+            raise TypeError(f"X must be a dataframe, got {type(X)}")
```
Vincent-Maladiere, line 476: This is already handled by ``CheckInputDataFrame`` IIUIC
```suggestion
```

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+
+        self._main_key = atleast_1d_or_none(self.main_key)
+        _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
+
+        # `y` is copied or converted to a df to be compatible with `aggregate`
+
```
Vincent-Maladiere, line 482:
```suggestion
```

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+        # If `y` is already a dataframe
+        if sbd.is_dataframe(y):
+            y_ = sbd.make_dataframe_like(y, sbd.to_column_list(y))
+        # If `y` is a named series, we convert it to a dataframe
+        elif sbd.is_column(y) and sbd.name(y) is not None:
+            y_ = sbd.make_dataframe_like(y, {sbd.name(y): y})
+        else:
+            # If `y` is an unnamed series
+            if sbd.is_column(y):
+                y = sbd.to_numpy(y).reshape(-1, 1)
```
Vincent-Maladiere, line 492: Why not imputing a name and using L488 instead?

skrub/_agg_joiner.py
Vincent-Maladiere, line 495: I don't think we should support lists

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+            # If `y` is a list
+            elif isinstance(y, list):
+                y = np.atleast_2d(y)
+            # 1d array that needs to be reshaped
+            elif isinstance(y, np.ndarray) & len(y.shape) == 1:
```
Vincent-Maladiere, line 497:
```suggestion
            elif isinstance(y, np.ndarray) and y.ndim == 1:
```

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+                y = y.reshape(-1, 1)
+
+            cols = {f"y_{i}": y[:, i] for i in range(y.shape[1])}
+            y_ = sbd.make_dataframe_like(X, cols)
```
Vincent-Maladiere, line 501: I guess we could use a single `y_ = sbd.make_dataframe_like(X, cols)` and have all previous conditions to only define `cols`, WDYT?

Btw, is there a typo here? Is this tested?
```suggestion
            y_ = sbd.make_dataframe_like(y, cols)
```

skrub/_agg_joiner.py
```diff
@@ -390,49 +434,115 @@ class AggTarget(TransformerMixin, BaseEstimator):
...
+        if not sbd.is_dataframe(X):
+            raise TypeError(f"X must be a dataframe, got {type(X)}")
+
+        self._main_key = atleast_1d_or_none(self.main_key)
+        _join_utils.check_missing_columns(X, self._main_key, "'X' (the main table)")
+
+        # `y` is copied or converted to a df to be compatible with `aggregate`
+
+        # If `y` is already a dataframe
+        if sbd.is_dataframe(y):
```
Vincent-Maladiere, line 484: I wonder whether we should bring the following lines into a utils function or transformer, similar to `CheckInputDataFrame`, but this time `ToDataFrame`. WDYT?

Vincent-Maladiere (commented):
Nice effort! I think we are almost done here

skrub/_agg_joiner.py
jeromedockes: the first argument of `make_dataframe_like` is only used for dispatch ie to choose the type of dataframe (pandas or polars), so either x or y would work as long as it comes from the correct module

skrub/_agg_joiner.py
Vincent-Maladiere: Yeah, I'm not sure we need this copy

skrub/_agg_joiner.py
Vincent-Maladiere: Ok, not what I would have expected TBH

Vincent-Maladiere:
Sorry, there's some duplicate with Jerome's feedback, we reviewed it simultaneously

skrub/_agg_joiner.py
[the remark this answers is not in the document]
TheooJ: Replied to Jérôme above :)

skrub/_agg_joiner.py
TheooJ: They were supported in the original `AggTarget`, let's remove them here :)

TheooJ: Removing support for lists

skrub/_agg_joiner.py
TheooJ: Do we need a copy since we're going to transform `_aux_table` ?

skrub/_agg_joiner.py
TheooJ: You're right, we don't need this 

I'm going to simplify the checks on y a bit

skrub/_agg_joiner.py
TheooJ: Need to update the changelog for this one

skrub/_agg_joiner.py
TheooJ: Nice catch ! Let me do that

skrub/_agg_joiner.py
TheooJ: I need them before check inputs in case `aux_table` is the placeholder `"X"`, because I assign `_aux_table` then use `CheckInputDataFrame` before `_check_inputs`

But I'm fine moving this whole section that's before `_check_inputs` inside of `_check_inputs`

skrub/_agg_joiner.py
TheooJ: You're right ! Removing this for the `main_key` check also 

thanks :)

skrub/_agg_joiner.py
TheooJ: It could happen in the event where the main dataframe already has column names in the forme of `{col}_{operation}` like `rating_mean`, but I realize now it's handled in `left_join`. I don't believe it's actually possible there -- removing :)

Vincent-Maladiere (approved):
Notwithstanding the discussion regarding the documentation of skrub selector (which may be done in another PR), this LGTM! Congratulations @TheooJ :)

Let's aim to remove `_namespace.py`, `_polars.py`, and `_pandas` in a follow-up PR!

skrub/_agg_joiner.py
```diff
@@ -85,19 +173,10 @@ class AggJoiner(TransformerMixin, BaseEstimator):
         the join operation.
         If `aux_key` is an iterable, we will perform a multi-column join.
 
-    cols : str or iterable of str, default=None
+    cols : str or iterable of str or selector, default=s.all()
```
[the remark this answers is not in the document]
Vincent-Maladiere: It might be a good time to start writing the dev doc in another PR and fixing https://github.com/skrub-data/skrub/issues/991

TheooJ:
After discussion with @GaelVaroquaux, it was decided we won't make selectors public here. Let's open a meta issue on their support and documentation first

> Let's aim to remove `_namespace.py`, `_polars.py`, and `_pandas` in a follow-up PR!

I'm on it :)

jeromedockes:
the codecov failure is spurious, codecov seems to be mixing the report with that of an older commit and is showing a different diff than the actual one

jeromedockes (approved):
very nice, thanks a lot @TheooJ ! I think the aggjoiner and aggtarget are simpler now

PR #1119, opened 2024-10-20
DOC: Add patch_display to the API docs

GaelVaroquaux:
As well as the corresponding function unpatch_display

This is important also to get to work all the ":func:" references (including those added automatically to the docs)


TheooJ (approved):
LGTM :)

PR #1122, opened 2024-10-22
fix regression with set_output in scikit-learn < 1.4

jeromedockes:
OnEachColumn, when its transformer is a scikit-learn transformer exposing set_output, calls set_output('pandas') or set_output('polars') to get the output in the correct dataframe.

in old scikit-learn versions, set_output('polars') is not supported, so OnEachColumn does set_output('pandas') and later does the conversion itself with `polars.from_pandas`.

or at least it's supposed to -- when doing a small change in #973 to handle possible exceptions raised by set_output I forgot to copy the "else" part of the "if scikit-learn < 1.4" block so now it is always using "polars" even in scikit-learn versions that don't support it.

this pr fixes the regression

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -56,9 +56,16 @@ Minor changes
   is now always visible when scrolling the table. :pr:`1102` by :user:`Jérôme
   Dockès <jeromedockes>`.
 
+Bug fixes
+---------
+
 * The :class:`TableReport` could raise an exception when one of the columns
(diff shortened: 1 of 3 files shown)


Vincent-Maladiere (approved):
Thanks @jeromedockes !

jeromedockes:
thanks for the review :) I'll merge it to unblock #1115

Issue #1118, opened 2024-10-20
Consider remerging the API list into one page


GaelVaroquaux:
The API page has recently been broken down into many small pages.

The drawback of such division is that one needs to know on which page a given function is going to be listed to access its documentation (and that's not obvious at all, even myself I get it wrong).
Relatedly, "ctrl-f" is no longer very useful. My typical pattern when I don't find what I'm looking for is to go to the API page and "ctr-f" for it.

I would push for remerging the pages into a single page.

Vincent-Maladiere:
To ensure I understand the issue, you can still ctr-f on the top-level page skrub-data.org/dev/reference/index.html, correct? So, you want to remove the intermediate API pages when clicking on a section, and bring these together instead of the top-level list of links?

GaelVaroquaux:
> To ensure I understand the issue, you can still ctr-f on the top-level page skrub-data.org/dev/reference/index.html, correct?

Good point indeed.

What happened was that I clicked on the section, didn't find what I want, then was lost.

But indeed, the problem is less pronounced than I thought. Maybe we can ignore this issue. I don't know.

> So, you want to remove the intermediate API pages when clicking on a section, and bring these together instead of the top-level list of links?

This was my suggestion.

Vincent-Maladiere:
I guess it makes sense to simplify the hierarchy and remove the intermediary levels. This way people won't get lost and will save a few clicks.

PR #1123, opened 2024-10-22
[DOC] Simplify the API reference hierarchy

Vincent-Maladiere:
Fix https://github.com/skrub-data/skrub/issues/1118

The only drawback I see is that the left panel is no longer collapsible. This is due to the removal of the subdirectories.


jeromedockes (approved):
LGTM, once the conflict is fixed, thanks. I think the conflict is in the meanwhile "column_associations" has been added to the "generating a report" section.

I don't really have an opinion re. this vs the subsections; it is true that the division into categories is somewhat arbitrary so it is not obvious in which of them we need to search for something (in the main branch)

PR #1125, opened 2024-10-22
accept duplicate column names in table report

jeromedockes:
also remove the `first_row_dict` function which is not used anywhere


Vincent-Maladiere (approved):
I don't know all the internals but the output looks good to me :)

jeromedockes:
> I don't know all the internals but the output looks good to me :)

thanks for checking! there's not much to it it's just making sure we retreive a column by giving its index rather than its name so we don't get any problems when it's a pandas dataframe with duplicate column names. not sure what's going on with codecov ci but I checked the added lines are covered by the tests by running them locally

PR #1126, opened 2024-10-26
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.

TODO
- [ ] remove TEMP commit and potentially codespell workflow


jeromedockes:
thanks a lot @yarikoptic ! no objections on my side, seems like a nice addition but I don't have experience with codespell, maybe @glemaitre has an opinion about adding the workflow or not? IIRC in nilearn we use the precommit hook but not the github workflow

glemaitre:
We have `codespell` correction in `scikit-learn` but we did not put it as a `pre-commit`. But it might be OK just to use as a pre-commit hook.

jeromedockes (approved):

Vincent-Maladiere (approved):
LGTM, thanks @yarikoptic !

PR #1135, opened 2024-11-09
MAINT adapt for scikit-learn 1.6

glemaitre:
Adapt the code with changes of related to some changes in the developer API that became public in scikit-learn.

### DIFF
--- a/skrub/_fixes.py
+++ b/skrub/_fixes.py
@@ -1,0 +1,10 @@
+import sklearn
+from sklearn.utils.fixes import parse_version
+
+sklearn_version = parse_version(parse_version(sklearn.__version__).base_version)
+
+
+if sklearn_version < parse_version("1.6"):
+    from sklearn.utils._tags import _safe_tags as get_tags  # noqa
+else:
+    from sklearn.utils import get_tags  # noqa
--- a/skrub/_interpolation_joiner.py
+++ b/skrub/_interpolation_joiner.py
@@ -1,4 +1,5 @@
 import warnings
+from dataclasses import is_dataclass
 
 import joblib
 import numpy as np
@@ -7,11 +8,11 @@ from sklearn.ensemble import (
     HistGradientBoostingClassifier,
     HistGradientBoostingRegressor,
 )
-from sklearn.utils._tags import _safe_tags
 
 from . import _dataframe as sbd
 from . import _join_utils, _utils
 from . import _selectors as s
+from ._fixes import get_tags
 from ._minhash_encoder import MinHashEncoder
 from ._table_vectorizer import TableVectorizer
 
@@ -399,7 +400,14 @@ def _get_assignments_for_estimator(table, estimator):
 
 
 def _handles_multioutput(estimator):
-    return _safe_tags(estimator).get("multioutput", False)
+    tags = get_tags(estimator)
+    if isinstance(tags, dict):
+        # scikit-learn < 1.6
+        return tags.get("multioutput", False)
+    elif is_dataclass(tags):
+        # scikit-learn >= 1.6
+        return tags.target_tags.multi_output
+    return False
 
 
 def _fit(key_values, target_table, estimator, propagate_exceptions):


skrub/_dataframe/_common.py
```diff
@@ -326,7 +326,7 @@ def _concat_horizontal_pandas(*dataframes):
     init_index = dataframes[0].index
     dataframes = [df.reset_index(drop=True) for df in dataframes]
     dataframes = _join_utils.make_column_names_unique(*dataframes)
-    result = pd.concat(dataframes, axis=1, copy=False)
+    result = pd.concat(dataframes, axis=1)
```
glemaitre, line 329: This should remove the deprecation warning. However, I'm wondering if we should opt-in for the new behaviour from now on using `pd.options.mode.copy_on_write = True` if available.

glemaitre:
The remaining pandas error is:

```pytb
E               TypeError: Cannot use .astype to convert from timezone-naive dtype to timezone-aware dtype. Use obj.tz_localize instead or series.dt.tz_localize instead
```

I assume we try to `cast`.

skrub/_datetime_encoder.py
```diff
@@ -69,7 +70,7 @@ def _get_dt_feature_polars(col, feature):
     return getattr(col.dt, feature)()
 
 
-class DatetimeEncoder(SingleColumnTransformer):
+class DatetimeEncoder(TransformerMixin, SingleColumnTransformer):
```
Vincent-Maladiere, line 73: @jeromedockes I just noticed the `SingleColumnTransformer` doesn't inherit from `TransformerMixin`, is there a reason for that?

skrub/_dataframe/_common.py
Vincent-Maladiere: Interesting! In which version was this feature introduced in pandas?

skrub/_fixes.py
```diff
@@ -0,0 +1,19 @@
...
+sklearn_version = parse_version(parse_version(sklearn.__version__).base_version)
+
+
+if sklearn_version < parse_version("1.6"):
+    from sklearn.utils._tags import _safe_tags as get_tags  # noqa
+else:
+    from sklearn.utils import get_tags  # noqa
+
+
+def _check_n_features(estimator, X, *, reset):
```
Vincent-Maladiere, line 13: Should we monkey patch sklearn's `BaseEstimator` with this function in skrub instead of calling it explicitly from `skrub._fixes`?

Vincent-Maladiere (commented):
A bunch of questions I had while reading :)

skrub/_datetime_encoder.py
glemaitre: `OnEachColumn` is the one implementing `fit_transform` and thus inherit from `TransformerMixin`. `SingleColumnTransformer` does not implement `transform` so per se, it makes sense to not inherit from `TransformerMixin`.

skrub/_dataframe/_common.py
glemaitre: It will be introduced in pandas 3.0 that is the next version. That's why the nightly build are catching this change of behaviour. However, I think that it could be cool to opt-in this behaviour to not have surprised between versions.

skrub/_fixes.py
glemaitre: Doing the monkey patch is undoing the change done in scikit-learn. So it means that we are going to have our own version of `BaseEstimator` in skrub. I think this is safer to implement our own `_check_n_features` and drop it at the time that the minimum supported version of scikit-learn is 1.6, we switch back to the scikit-learn version. It looks less invasive.

This is already a kind of pattern that we have in `scikit-learn` when it comes to numpy or scipy (cf. `sklearn/utils/fixes.py`).

skrub/_table_vectorizer.py
```diff
@@ -640,6 +649,7 @@ def _sk_visual_block_(self):
 
     # scikit-learn compatibility
 
+    @available_if(check_version)
```
jeromedockes, line 652: same comment abouit `@available_if`

jeromedockes (commented):
thanks a lot for taking care of this @glemaitre !

jeromedockes:
@glemaitre here is the fix for the time zone failures (pandas changes the representation of time zones in 1.6):

```diff
diff --git a/skrub/_to_datetime.py b/skrub/_to_datetime.py
index 4a08013f..a14d3c8a 100644
--- a/skrub/_to_datetime.py
+++ b/skrub/_to_datetime.py
@@ -28,6 +28,8 @@ def _get_time_zone_pandas(col):
         return None
     if hasattr(tz, "zone"):
         return tz.zone
+    if hasattr(tz, "key"):
+        return tz.key
     return tz.tzname(None)
 ```

File: skrub/_datetime_encoder.py (hunk 2 above)
Comment (glemaitre) [reply skrub/_datetime_encoder.py:73]: Mainly for the tags infrastructure. Otherwise, we need to set manually the tags related to `transformer_tags` ourselves.

File: skrub/_dataframe/_common.py (hunk 1 above)
Comment (glemaitre) [reply skrub/_dataframe/_common.py:329]: So let's do via the pandas version then.

File: skrub/_datetime_encoder.py [hunk 5]
```diff
@@ -323,3 +330,17 @@ def _check_params(self):
             raise ValueError(
                 f"'resolution' options are {allowed}, got {self.resolution!r}."
             )
+
+    if sklearn_below_1_6:
```
glemaitre, line 334: @jeromedockes The issue was that the scikit-learn tests (that are skipped) issue an error at collection because `preserve_dtype` was not defined.

Somehow it was not raising in the past for some reason (but it could have raised if we run the test). So we don't need to inherit from `TransformerMixin`, just make a similar definition of the tags.

jeromedockes (approved):
thanks a lot @glemaitre ! super useful that you help us keep up with scikit-learn :)

jeromedockes (approved):
a big thanks for this effort both on the skrub and scikit learn side @glemaitre 
I'll merge it the coverage should be there once scikit-learn 1.6 is released or on the nightly build but for that we'll need to handle breaking changes in pandas & scipy which I'll deal with in separate prs

PR #1137, opened 2024-11-14
Fix getting started example

TheooJ:
Follow-up to #1116 

Fixing example now that we don't split columns by type in `AggJoiner` anymore


jeromedockes (approved):

jeromedockes:
thanks a lot for the follow-up!

PR #1138, opened 2024-11-14
Faster svg viewbox computation in tablereport

jeromedockes:
The way the computation of the correct viewbox for the plots is done currently is slow when there are many columns.
This makes all the bounding box computation in one go with no changes to the document in between to avoid unnecessary recomputations of the styles and layout

also as it adds a few functions to the report code, now is a good time to put it in a module to avoid polluting the global namespace so this PR adds that too

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -59,7 +59,8 @@ Minor changes
   Moreover, when the text contains line breaks it now appears all on one line.
   Note this only affects the labels in the plots; the rest of the report did not
   have these problems.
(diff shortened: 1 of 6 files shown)


GaelVaroquaux (approved):
LGTM, but I think that it would be useful to add a short changelog entry: it's user-visible enough

jeromedockes:
I marked it as draft because @Vincent-Maladiere  agreed to check it works in a different environment so we should wait for that info before merging

Vincent-Maladiere (approved):
LGTM after testing on my machine

jeromedockes:
thanks a lot!

Issue #1143, opened 2024-11-18
CircleCI no longer runs


GaelVaroquaux:
### Describe the bug

CircleCI no longer complete:

https://app.circleci.com/pipelines/github/skrub-data/skrub/4753/workflows/32a9d8ee-f6cc-45db-8b30-70304424c633/jobs/9236

### Steps/Code to Reproduce

It's in the circle CI build :)

### Expected Results

CircleCI to complete

### Actual Results

```
generating gallery for auto_examples... [ 10%] 00_getting_started.py
Processing column   8 / 8
Processing column   8 / 8
generating gallery for auto_examples... [ 20%] 01_encodings.py
generating gallery for auto_examples... [ 30%] 02_text_with_string_encoders.py
Processing column   2 / 2
Processing column  31 / 31
/home/circleci/project/.pixi/envs/doc/lib/python3.12/site-packages/sphinx_gallery/scrapers.py:173: UserWarning: Glyph 4108 (\N{MYANMAR LETTER TTHA}) missing from font(s) DejaVu Sans.
  fig.savefig(image_path, **these_kwargs)
generating gallery for auto_examples... [ 40%] 03_datetime_encoder.py
generating gallery for auto_examples... [ 50%] 04_fuzzy_joining.py
generating gallery for auto_examples... [ 60%] 05_deduplication.py
generating gallery for auto_examples... [ 70%] 06_ken_embeddings.py
generating gallery for auto_examples... [ 80%] 07_multiple_key_join.py
make: *** [Makefile:32: html] Killed

Exited with code exit status 2
```

### Versions

```shell
on main.
```

GaelVaroquaux:
First thing to do is probably to try renaming the example added in https://github.com/skrub-data/skrub/pull/1077 to ".py1" so that it is not run. That way we'll see if it's related to the newly added dependencies, or to actual interactions between exemples

PR #1145, opened 2024-11-19
FIX failing sphinx-gallery CI

Vincent-Maladiere:
Fix https://github.com/skrub-data/skrub/issues/1143


Vincent-Maladiere:
So, it succeeds when I cut example 07 just before importing the Joiner

Vincent-Maladiere:
Let's now run everything but the final CV

Vincent-Maladiere:
Then the last cell with the CV should crash

Vincent-Maladiere:
... which it did. So, there might be some weird memory usage due to the text-encoder example, and sentence-transformer or pytorch caching shenanigans.

GaelVaroquaux:
> So, it succeeds when I cut example 07 just before importing the Joiner

OK, so it's an interaction...

Vincent-Maladiere:
It failed, that's weird. It means that this is not provoked by the TextEncoder example: https://app.circleci.com/pipelines/github/skrub-data/skrub/4761/workflows/c392d89f-b5dd-4ab2-8d72-3438dd3e4911/jobs/9249 (example 2 is not run there, although example 7 fails)

Vincent-Maladiere:
The simple fit worked. So we're likely having a memory error due to CV on the (Joiner, TV, HGBT) pipeline. But why now?

GaelVaroquaux:
It could be triggered by an import...

GaelVaroquaux:
> But why now?

Memory that is not cleared? Maybe something that you can investigate on your computer.

Are we holding a reference on a big object?

jeromedockes:
thanks a lot for investigating this @Vincent-Maladiere  I know it's not fun :sweat_smile:  . it's good to have in mind that sphinx gallery runs all the examples in the same process so I guess it is possible that some of the memory from the textencoder example has not been freed when the example 7 runs and that the 2 together pass circleci's memory limit

Vincent-Maladiere:
So, when we remove the "transformers" environment (i.e., pytorch, transformers and sentence-transformers dependencies) from the pixi doc environment, the CI runs fine.

>  some of the memory from the textencoder example has not been freed when the example 7 runs and that the 2 together pass circleci's memory limit

That was my hypothesis too, but we have observed that without example 02 (which uses TextEncoder):
- Running the CV on example 07, with "transformers" in the doc environment fails ❌
- Running **a simple fit** on example 07, with "transformers" in the doc environment works ✅
- Running the CV **without "transformers"** in the doc environment works ✅

We only import sentence-transformers or pytorch within the TextEncoder, not outside. However, since I disabled example 02, the CI never imports or uses the TextEncoder. So I don't know how to conclude here.

GaelVaroquaux:
> thanks a lot for investigating this @Vincent-Maladiere I know it's not fun 😅

Indeed. But it is an amazing stress test that reveals how the code lives (and dies) in challenging settings. It makes the software more robust, which the users then love.

> . it's good to have in mind that sphinx gallery runs all the examples in the same process so I guess it is possible that some of the memory from the textencoder example has not been freed when the example 7 runs and that the 2 together pass circleci's memory limit

Quite possible. If that's the case, hopefully we can find how to free memory: it would make the library more lean.

jeromedockes:
> That was my hypothesis too, but we have observed that without example 02 (which uses TextEncoder):
> - Running the CV on example 07, with "transformers" in the doc environment fails ❌
> - Running **a simple fit** on example 07, with "transformers" in the doc environment works ✅
> - Running the CV **without "transformers"** in the doc environment works ✅
>
> We only import sentence-transformers or pytorch within the TextEncoder, not outside. However, since I disabled example 02, the CI never imports or uses the TextEncoder. So I don't know how to conclude here.

if both installing transformers and running the CV take a lot of time,
could it be a time limit?

Vincent-Maladiere:
> Quite possible. If that's the case, hopefully we can find how to free memory: it would make the library more lean.

This is not what's happening. Cf my message above.

Vincent-Maladiere:
I can't reproduce that locally. Let's rerun the CV with high verbosity to see if we can detect something.

Vincent-Maladiere:
Great! So, the example 02 runs fine. The matter is with example 07. It looks like a timeout issue as @jeromedockes guessed

Vincent-Maladiere:
When example 02 runs, the CI crashes before reaching the cross-validation part of example 07.

It looks like there is a timeout at 10min (11 actually) in circle-ci, but the run is not idle for long, and `no_output_timeout` is set to 30m anyway. With circle-ci, the [max run time is 1h for freemium accounts](https://circleci.com/docs/configuration-reference/#jobs), so we're not hitting this limit either. I can't find any other kind of timeout limit.

Could it be RAM consumption? But then, why would the cross-validation in example 07 crash now, whereas we never importthe TextEncoder, or torch. I'm starting to run out of ideas.

jeromedockes:
I think the duration of example 7 is a problem in itself anyways -- it takes 8 minutes on circleci. So I would suggest reducing it by reducing the sizes of the subsamples we use (say 5,000 flights and 10,000 weather points), and replacing the cross-validation with a train/test split. Hopefully that will get the doc build passing again.

In a second time, we can work on improving that example and maybe using a different dataset. Indeed, at the moment 
- it downloads a very large dataset
- the join with weather data does not improve predictions significantly IIRC
- the predictions are not super far from chance level which makes the example less compelling IMO

we may never pin down the exact reason why the circleci job getting killed but I think I can live with that as long as we get it running again :D

jeromedockes:
and I think we owe @Vincent-Maladiere  a beer for this long session of blind debugging with a 20-min feedback delay which can be quite frustrating :sweat_smile:

Vincent-Maladiere:
> we may never pin down the exact reason why the circleci job getting killed but I think I can live with that as long as we get it running again :D

Haha, I couldn't agree more! You're right, let's revamp this example slightly to make the CI run.
I'm also creating an issue to improve this example in the longer term.

jeromedockes:
cool, thanks! and if that doesn't work: I remember that for nilearn at some point I had reached out to the circleci support and had a good experience (and nilearn uses the free plan) so that option exists as well if all else fails

jeromedockes:
nice!! thanks again! I'll merge it before circleci gets a chance to change its mind :laughing:

jeromedockes (approved):

jeromedockes:
the example still looks pretty much the same after reducing the subsample sizes so that's good

Issue #1147, opened 2024-11-19
Add a threshold to DropColumnIfNull


rcap107:
### Problem Description

The current version of `DropColumnIfNull` automatically drops any columns where _all_ values are null. However, it would be good to have a threshold so that all columns that have a fraction of null values larger than a given threshold are dropped instead. 

### Feature Description

The `DropColumnIfNull` object should be modified so that it can take a "threshold" parameter, and then drop columns whose null fraction is above this threshold. 

The value of the threshold should be 1.0 by default (i.e., all columns must be null to drop), and there should be a parameter for it in the TableVectorizer that the user can modify to tweak this.

Something to note is that the user may want to have different thresholds for different columns, which would complicate things. 

### Alternative Solutions

I am not sure whether we should have two separate objects like `DropColumnIfNull` and `DropNullsAboveThreshold` (tentative name), or if `DropColumnIfNull` should stay and get updated with the threshold. 

We might want to have two different objects, one that takes a single value for the threshold, and one that allows for more granularity. Though, at that point it might just be better going with a TableVectorizer with specific transformers for each column :thinking:

Vincent-Maladiere:
Hey @rcap107! We can add this parameter to the current DropNull object and rename it slightly IMO. Then, you could have a single parameter like `drop_null_frac` in the TableVectorizer, which is set to 1.0 by default (and `None` would mean no drop mechanism). Ideally, I would rather avoid having 2 separate input parameters in the TableVectorizer for this.

I guess having a single threshold for all columns is fine for now to keep things simple. WDYT?

rcap107:
Agreed, the first version should definitely be just a single threshold that's used for all columns, and that can take a single default value.

I still think that having an object that does the filtering with different thresholds for each column would be useful, but I don't see that as being a "default" part of the TV. Something for a different PR!

rcap107:
First draft in #1149

jeromedockes:
solved in #1149

PR #1149, opened 2024-11-21
Adding a threshold to DropColumnIfNull

rcap107:
Quick PR to address #1147 

Notes:
- I am using `None` to specify "Do not drop anything", rather than `0.0`. 
- Do we want to change the name of the object? I don't think it's necessary. 

(I also added fixed my account in the changelog)


skrub/_drop_column_if_null.py
```diff
@@ -8,8 +8,23 @@
...
+    nulls or NaNs are dropped. Set the threshold to `None` to keep all columns.
+
+    Parameters
+    ----------
+    threshold : float, or None
+        Threshold of null values past which the column is dropped.
+    """
+
+    def __init__(self, threshold=1.0):
+        if threshold is not None:
```
jeromedockes, line 24: any checking of parameters should go in `fit` (in this case, at the start of `fit_transform`. the `__init__` of scikit-learn-compatible estimators never does anything else than `def __init__(self, **kwargs): self.__dict__.update(kwargs)`

skrub/_drop_column_if_null.py
```diff
@@ -8,8 +8,23 @@
...
+            assert 0.0 <= threshold <= 1.0, "Invalid value for the threshold."
```
jeromedockes, line 25: can we show the value of the threshold in the error message? also use a valueerror rather than an assertionerror. assertions would be used to stop execution if a bug in skrub code has led to a situation that shouldn't be possible. valueerror is how we communicate to the user that the value they provided is inappropriate

skrub/_drop_column_if_null.py
```diff
@@ -25,8 +40,16 @@ def fit_transform(self, column, y=None):
...
 
-        self.drop_ = sbd.is_all_null(column)
-
+        if self.threshold == 1.0:
+            self.drop_ = sbd.is_all_null(column)
+        elif self.threshold is None:
+            self.drop_ = False
+        else:
+            n_count = sum(sbd.is_null(column))
+            if n_count / len(column) > self.threshold:
```
jeromedockes, line 49: maybe this can be simplified to `self.drop_ = n_nulls / len(column) > self.threshold`? also we probably want to handle the len=0 corner case

skrub/_drop_column_if_null.py
```diff
@@ -25,8 +40,16 @@ def fit_transform(self, column, y=None):
         """
         del y
 
-        self.drop_ = sbd.is_all_null(column)
-
+        if self.threshold == 1.0:
```
jeromedockes, line 43: this doesn't match the docstring or the code below which drop the column if the proportion is strictly greater than the threshold. of course, the proportion cannot be strictly greater than 1.0 ...
we could say that we drop if it is greater or equal than the threshold but then we would have no way of expressing "drop if it contains any nulls" because threshold=0.0 would drop all columns

not sure what we should do about this, but in any case the docstring and tests should be very clear on what happens in the cases where the 2 values are equal

jeromedockes (commented):
thanks for starting this @rcap107 !

skrub/_drop_column_if_null.py
jeromedockes: in other words assertions are more of a debugging tool, not a way to validate user input. for example if the script is run with `python -O` or `PYTHONOPTIMIZE=1` they are not executed

rcap107: Noted, thanks for pointing that out. I changed it to a ValueError.

skrub/_drop_column_if_null.py
rcap107: I modified the code to try and address both this and the case when no nulls are found

Vincent-Maladiere: Could we make an exception for 0 and drop when the number of nulls is > 0? Not 100% consistent but this is the high level behaviour I would expect as a user

rcap107: That's how I coded it in the latest commit

CHANGES.rst
```diff
@@ -76,8 +76,9 @@ Minor changes
   is now always visible when scrolling the table. :pr:`1102` by :user:`Jérôme
   Dockès <jeromedockes>`.
 
-* Added a `DropColumnIfNull` transformer that drops columns that contain only null
-  values. :pr:`1115` by :user: `Riccardo Cappuzzo <riccardocappuzzo>`
+* Added a `DropColumnIfNull` transformer that drops columns based on how many
```
jeromedockes, line 79: actually DropColumnIfNull has not been added to skrub's public API (and I'm not sure if it should before we decide what to do with selectors) so here we should document the added parameter to TableVectorizer rather than DroopColumnIfNull

skrub/_drop_column_if_null.py
```diff
@@ -8,8 +8,22 @@
 
 
 class DropColumnIfNull(SingleColumnTransformer):
```
jeromedockes, line 10: should we change the name to something like NullProportionThreshold ? :thinking:

skrub/_drop_column_if_null.py
```diff
@@ -8,8 +8,22 @@
...
-    """Drop a single column if it contains only Null, NaN, or a mixture of null
-    values. If at least one non-null value is found, the column is kept."""
+    """Drop a single column if the fraction of Null or NaN values in the column
+    is larger than the given threshold.
+
+    If the threshold is set to `1.0`, the column is dropped if it contains only
+    nulls or NaNs (this is the default value). If the threshold is set to `None`,
+    all columns are kept. Otherwise, the column is dropped if the fraction
+    of nulls is strictly larger than the threshold
```
jeromedockes, line 17: :+1:

skrub/_drop_column_if_null.py
```diff
@@ -21,12 +35,35 @@ def fit_transform(self, column, y=None):
...
+            The input column, or an empty list if the column is selected to be
+            dropped depending on the threshold.
         """
         del y
 
-        self.drop_ = sbd.is_all_null(column)
+        if self.threshold is not None:
+            if (
+                not (
+                    isinstance(self.threshold, float) or isinstance(self.threshold, int)
```
jeromedockes, line 46: you can use 
```python
import numbers
isinstance(self.threshold, numbers.Number)
```
instead.

it will work with numpy numeric types whereas

```
>>> isinstance(np.float32(0.5), float)
False
```

jeromedockes (commented):
thanks @rcap107 !! looks like we're close :)

skrub/_drop_column_if_null.py
rcap107: Maybe `DropColumnWithNullThreshold`? `DropIfNullAboveThreshold`?

skrub/_drop_column_if_null.py
rcap107: Thanks, I knew that check didn't look right, but I wasn't sure how to write it properly

jeromedockes: you're welcome! also next time you want a disjunction of isinstance checks you can put the types in a tuple
```
>>> isinstance(2, (str, int))
True
>>> isinstance(2.5, (str, int))
False
```

skrub/_drop_column_if_null.py
jeromedockes: maybe!
DropIfTooManyNulls?

Vincent-Maladiere: +1 for DropIfTooManyNulls

skrub/_table_vectorizer.py
```diff
@@ -192,8 +192,12 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
         similar functionality to what is offered by scikit-learn's
         :class:`~sklearn.compose.ColumnTransformer`.
 
-    drop_null_columns : bool, default=True
-        If set to `True`, columns that contain only null values are dropped.
+    null_threshold : float or None, default=1.0
```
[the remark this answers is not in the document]
Vincent-Maladiere: `drop_null_threshold` maybe?

Vincent-Maladiere (approved):
LGTM

Vincent-Maladiere:
That is extremely cool :) let's merge this

PR #1152, opened 2024-11-21
better handling of timedelta in tablereport

### LINKED ISSUE #1132: TimeDelta64DType treated as category by the TableReport
### Describe the bug

This is somewhat niche, but TableReport treats TimeDelta64DType as discrete categories, while it's a continuous quantity as DateTime64DType.

### Steps/Code to Reproduce

```python
import pandas as pd
from skrub import TableReport

TableReport(
    pd.to_timedelta([20, 40], unit="D").to_frame()
)
```

### Expected Results

A histogram distribution

### Actual Results

A value counts distribution

### Versions

```shell
dev
```

jeromedockes:
fixes #1132 

this converts the duration/timedelta columns to numeric so we get a histogram & summary statistics rather than a bar plot

durations and datetimes are also converted to numeric for the column associations detection

the choice of time unit & conversion could probably be improved but this is already better than treating them as categorical

screenshot_2024-11-21T17:16:03+01:00

### DIFF
--- a/skrub/_reporting/_utils.py
+++ b/skrub/_reporting/_utils.py
@@ -112,3 +112,22 @@ class JSONEncoder(json.JSONEncoder):
             if isinstance(value, np.floating):
                 return float(value)
             raise
+
+
+def duration_to_numeric(col):
+    seconds = sbd.total_seconds(col)
+    q = sbd.quantile(sbd.abs(seconds), 0.9)
+    HOUR = 3600
+    DAY = HOUR * 24
+    YEAR = DAY * 365.2425
+    if q < 1e-3:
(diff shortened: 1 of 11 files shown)


skrub/_reporting/_utils.py
```diff
@@ -112,3 +112,22 @@ def default(self, value):
             if isinstance(value, np.floating):
                 return float(value)
             raise
+
+
+def duration_to_numeric(col):
```
Vincent-Maladiere, line 117: I like this simple but robust heuristic

Vincent-Maladiere (approved):
LGTM

jeromedockes:
thanks @Vincent-Maladiere

PR #1157, opened 2024-11-25
Clip the range of histograms when there are outliers

### LINKED ISSUE #1155: Clip range of histograms in TableReport
This has been reported in a few places: when a column contains outliers, the histogram plot produced by the tablereport is not very informative because the range is very wide and almost all the distribution is in the same bin. (it does help realize that there are outliers, which is good)

It would be better to remove outliers before plotting the histogram so we can see more detail, _but_ in this case we need to come up with a way to make it obvious that outliers had to be removed and convey how many there were.

the "outlier detection" would have to be very cheap, for example we could do something similar to what box plots do (clip at some percentile's position offset by a factor of the inter-percentile range)

jeromedockes:
fixes #1155 

this limits the range of data shown in some histograms to avoid having all the data in one bin and seeing no details of the distribution


jeromedockes:
## main branch:

screenshot_2024-11-25T15:22:31+01:00


## this pr:

screenshot_2024-11-25T15:22:41+01:00

skrub/_reporting/_plotting.py
```diff
@@ -115,6 +117,67 @@ def _adjust_fig_size(fig, ax, target_w, target_h):
     fig.set_size_inches((w, h))
 
 
+def _get_range(values, frac=0.2, factor=3.0):
+    min_value, low_p, high_p, max_value = np.percentile(
+        values, [0, frac * 100.0, (1.0 - frac) * 100.0, 100.0]
```
Vincent-Maladiere, line 122: np.quantile to simplify this line?

Vincent-Maladiere (approved):
Works fine when there are left and right outliers (yes, prices can be negative)

Screenshot 2024-11-27 at 14 39 33

without this PR:

Screenshot 2024-11-27 at 14 49 59

skrub/_reporting/_plotting.py
```diff
@@ -115,6 +117,67 @@ def _adjust_fig_size(fig, ax, target_w, target_h):
...
+    )
+    delta = high_p - low_p
+    if not delta:
+        return min_value, max_value
+    margin = factor * delta
```
Vincent-Maladiere, line 127: Interesting heuristic, how did you get these default parameters?

jeromedockes: from the random number generator in my head :sweat_smile:  as mentioned I think the heuristic can definitely be refined or replaced with somehting else

jeromedockes:
thanks for the review @Vincent-Maladiere  :)

skrub/_reporting/_plotting.py
jeromedockes: I thought of using the inter-quantile range because that's what box plots do, but I think we could read more about simple outlier detection methods

PR #1158, opened 2024-11-26
small tablereport display adjustments

jeromedockes:
ATM the keyboard shortcut hints are a bit too eager to disappear when horizontal space is tight; this adjusts a bit the threshold while improving a bit the display in the off chance that it has to wrap before it disappears

also makes the column filter very slightly wider to avoid the default option being ellided in rare cases


jeromedockes:
before

screenshot_2024-11-26T15:26:08+01:00


after

screenshot_2024-11-26T15:25:44+01:00

Issue #1121, opened 2024-10-21
[FEAT] Add LSA encoder


Vincent-Maladiere:
### Problem Description

Latent Semantic Analysis (LSA) consists of a TfidfVectorizer followed by Singular Value Decomposition (SVD). [Scikit-learn mentions it in TruncatedSVD](https://scikit-learn.org/dev/modules/generated/sklearn.decomposition.TruncatedSVD.html), and I wonder why it hasn't been implemented in scikit-learn in the first place, @GaelVaroquaux?

### Feature Description

Create the `LSAEncoder`, a simple pipeline chaining `TfidfVectorizer` and `TruncatedSVD` (or a PCA, both support sparse matrices).

GaelVaroquaux:
Great!!

We need to think about a name. I think LSA is a bit of a technical name that might ring a bell to non technical users.

We brainstormed a bit in terms of name with @jeromedockes and @rcap107 . The name StringEncoder came to mind. It would be close to TextEncoder (https://github.com/skrub-data/skrub/pull/1077), but we feel that the difference is somewhat understandable.

That said, maybe it would be an argument to move the name TextEncoder to SentenceEncoder, which would also be (maybe) a good name because it would be more explicit (link to "SentenceTransformer")

Vincent-Maladiere:
Very interesting! One might wonder why we don't consider the GapEncoder as a string encoder, though. WDYT?

GaelVaroquaux:
> One might wonder why we don't consider the GapEncoder as a string encoder, though. WDYT?

Yes, this was raised, and it is true. I guess that one difference that I make is that the GapEncoder assumes more latent structure (aka dirty-category structure) than open ended strings.

One argument for naming it the "StringEncoder" is that if you really have no prior information on the data or the use of the encoding, it's probably a good default to encode a string. Of course, we'll have to have good "see also" section, and a good discussion in the docs.

Vincent-Maladiere:
Okay, this sounds easy to explain in the doc!

Vincent-Maladiere:
> [Scikit-learn mentions LSA in TruncatedSVD](https://scikit-learn.org/dev/modules/generated/sklearn.decomposition.TruncatedSVD.html), and I wonder why it hasn't been implemented in scikit-learn in the first place

Any thoughts @GaelVaroquaux? I'm curious

GaelVaroquaux:
[...]

PR #1159, opened 2024-11-26
Adding the StringEncoder transformer

rcap107:
This is a first draft of a PR to address #1121

I looked at GapEncoder to figure out what to do. This is a very early version just to have an idea of the kind of code that's needed. 

Things left to do:
- [x] Testing
- [ ] Parameter checking?
- [x] Default value for the PCA? 
- [ ] Docstrings
- [ ] Deciding name of the features

### DIFF
--- a/skrub/_string_encoder.py
+++ b/skrub/_string_encoder.py
@@ -7,7 +7,6 @@ from sklearn.feature_extraction.text import (
     TfidfVectorizer,
 )
 from sklearn.pipeline import Pipeline
-from sklearn.utils.validation import check_is_fitted
 
 from . import _dataframe as sbd
 from ._on_each_column import SingleColumnTransformer
@@ -190,9 +189,3 @@ class StringEncoder(SingleColumnTransformer):
         result = sbd.copy_index(X, result)
 
         return result
-
-    def __sklearn_is_fitted__(self):
-        """
-        Check fitted status and return a Boolean value.
-        """
-        return check_is_fitted(self, "all_outputs_")


rcap107:
Tests fail on minimum requirements because I am using PCA rather than TruncatedSVD for the decomposition, and that raises issues with potentially sparse matrices.

@jeromedockes suggests using directly TruncatedSVD to begin with, rather than adding a check on the version. 

Also, I am using tf-idf as vectorizer, should I use something else? Maybe HashVectorizer? 

(writing this down so I don't forget)

GaelVaroquaux:
I'm very happy to see this progressing.

Can you benchmark it on the experiments from Leo's paper: this is important for modeling choices (eg the hyper-parameters)

GaelVaroquaux:
Actually, let's keep it simple, and use the CARTE datasets, they are good enough: https://huggingface.co/datasets/inria-soda/carte-benchmark

You probably want to instanciate a pipeline that uses TableVectorizer + HistGradientBoosting, but embeds one of the string columns with the StringEncoder (the one that is either higest cardinality, or most "diverse entry" in the sense of https://arxiv.org/abs/2312.09634

rcap107:
> Should we also add this to the text encoder example, along the TextEncoder, MinHashEncoder and GapEncoder? It shows a tiny benchmark on the toxicity dataset.

It's already there, and it shows that StringEncoder has performance similar to that of GapEncoder and runtime similar to that of MinHashEncoder

Vincent-Maladiere:
That's very interesting!

CHANGES.rst
```diff
@@ -14,7 +14,9 @@ It is currently undergoing fast development and backward compatibility is not en
 
 New features
 ------------
-
+* The :class:`StringEncoder` encodes strings using tf-idf and truncated SVD
+  decomposition and provides a cheaper alternative to :class:`TextEncoder`.
```
jeromedockes, line 18: or :class:`GapEncoder`

example_string_encoder.py
```diff
@@ -0,0 +1,33 @@
+# %% test string encoder
+import polars as pl
```
jeromedockes, line 2: I guess this should go in a test file?

examples/02_text_with_string_encoders.py
```diff
@@ -221,6 +224,25 @@ def plot_box_results(named_results):
 
 plot_box_results(results)
 
+# %%
+# |TextEncoder| embeddings are very strong, but they are also quite expensive to
+# train. A simpler, faster alternative for encoding strings is the |StringEncoder|,
```
jeromedockes, line 229: to use -- we don't actually train them

examples/02_text_with_string_encoders.py
```diff
@@ -221,6 +224,25 @@ def plot_box_results(named_results):
...
+# which works by first performing a tf-idf vectorization of the text, and then
```
jeromedockes, line 230: maybe tf-idf (computing vectors of rescaled word counts + wikipedia link)

skrub/_string_encoder.py
```diff
@@ -0,0 +1,132 @@
...
+
+        Returns
+        -------
+        X_out: Pandas or Polars dataframe with shape (len(X), tsvd_n_components)
+            The embedding representation of the input.
+        """
+        del y
+        self.pipe = Pipeline(
+            [
+                ("tfidf", TfidfVectorizer()),
```
jeromedockes, line 96: I think @GaelVaroquaux suggested using a HashingVectorizer instead of tfidfvectorizer (i don't think this would require changes elswhere in your code)

skrub/_string_encoder.py
```diff
@@ -0,0 +1,132 @@
...
+       video comments_0  video comments_1
+    0      8.218069e-01      4.557474e-17
+    1      6.971618e-16      1.000000e+00
+    2      8.218069e-01     -3.046564e-16
+    """
+
+    def __init__(self, n_components=30):
+        self.n_components = n_components
+
+    def _transform(self, X):
```
jeromedockes, line 60: maybe this could be inlined into `transform` rather than having a separate function unless I'm missing where `_transform` is called outside of `transform`

skrub/_string_encoder.py
```diff
@@ -0,0 +1,132 @@
...
+        name = sbd.name(X)
+        if not name:
+            name = "tsvd"
+        self.all_outputs_ = [f"{name}_{idx}" for idx in range(self.n_components)]
+
+        self.pipe.fit(sbd.to_numpy(X))
+
+        self._is_fitted = True
+
+        return self.transform(X)
```
jeromedockes, line 110: I think we could take advantage of `pipe.fit_transform` to avoid tokenizing the input twice during fit

jeromedockes (commented):
nice! :tada:  looking really good @rcap107  . it will be awesome to have this as the cost of the GapEncoder is the main pain point of the tablevectorizer ATM :rocket: 

I left a couple of minor comments, the main one is just a possible performance improvement if we use the vectorizer's `fit_transform` to avoid tokenizing the input twice during fit

the code is very close to ready and we have an example so once that's done what are the next steps? maybe some lightweight comparisons on a couple of datasets, then merge, then more extensive experiments to decide if it can be come the tablevectorizer's default?

example_string_encoder.py
rcap107: ops, I forgot to delete this

skrub/_string_encoder.py
rcap107: I modified the code, but I am not sure the implementation is what you meant

rcap107:
> nice! 🎉 looking really good @rcap107 . it will be awesome to have this as the cost of the GapEncoder is the main pain point of the tablevectorizer ATM 🚀
>
> I left a couple of minor comments, the main one is just a possible performance improvement if we use the vectorizer's `fit_transform` to avoid tokenizing the input twice during fit

I implemented a change, I'm not sure if it's what you meant..

> the code is very close to ready and we have an example so once that's done what are the next steps? maybe some lightweight comparisons on a couple of datasets, then merge, then more extensive experiments to decide if it can be come the tablevectorizer's default?

Sounds good

skrub/_string_encoder.py
jeromedockes: yep that's exactly what I meant thanks!

Vincent-Maladiere:
> the code is very close to ready and we have an example so once that's done what are the next steps? maybe some lightweight comparisons on a couple of datasets, then merge, then more extensive experiments to decide if it can be come the tablevectorizer's default?

I like this plan as well!

GaelVaroquaux:
Before we merge, I would love a bit of experimenting, with a fairly systematic eye on the choice of parameters.

For instance: HashingVectorizer vs TfIdfVectorizer? If we go for hashingVectorizer, should it be followed by a TfIdfTransformer or not? For the vectorizer, what should be the default values for the analyzer, the n_features and the n_gram range. I suspect that we want 'analyzer="char_wb"' and ngram_range=(3, 4) based on my experience.

These choices are actually very important, and we should drive them using systematic experimentation. I know that it is a lot of work, but it makes a huge difference.

rcap107:
So I went and ran a few experiments and I have some preliminary results. I reused some of the code from the main example, so the plots are not very good. 

I modified the StringEncoder to have
- `vectorizer` to choose between `tfidf` and `hashing` (mapped to `TfidfVectorizer` and `HashingVectorizer`). Default is `tfidf`
- `ngram_range`, which can be any pair of integers >= `(1,1`). Default is `(1,1)`.
- `tf_idf_followup`, which is a boolean flag. If active, the `HashingVectorizer` is followed by a `TfidfTransformer`.
- `analyzer`, which can be `word`, `char`, or `char_wb`. Default is `word`.
I also added:
- `n_features`, which is a parameter for `tfidf`.
- `max_features`, which is a parameter for `hashing`.

Then I tested these parameters:
```python
configurations = {
    "ngram_range": [(1, 1), (3, 4)],
    "analyzer": ["word", "char", "char_wb"],
    "vectorizer": ["hashing", "tfidf"],
    "n_components": [30],
    "tf_idf_followup": [True, False],
}
```


Only tfidf


Only hashingvectorizer, no tfidf followup
image

Hashingvectorizer with tfidf followup
image


Summary is:
- tfidf is faster
- hashing + tfidf has the same prediction performance as tfidf, but it's much slower
- hashing alone is worse than tfidf and is much slower

jeromedockes:
thanks for those experiments! I'm surprised the (1, 1 ) ngram with "char" or "char_wb" can perform so well -- isn't it just counting individual characters?

GaelVaroquaux:
> I'm surprised the (1, 1 ) ngram with "char" or "char_wb" can perform so well -- isn't it just counting individual characters?

char_wb makes sense: it's counting words and characters

rcap107:
I ran a few more experiments, because the results looked suspicious, and I think what I have now makes much more sense.

Pareto plots (time is on a log scale):
image

Boxplots, test score on y-axis
image

Boxplots, time on y-axis
image

I don't know why the previous results were so off, I might have been reusing results across runs :thinking: 

What is consistent is that tf-idf is way faster and has similar performance if not better than hashing vectorizer.

GaelVaroquaux:
OK, those results look great. Please do keep the associated data, as one day we may publish a paper on skrub :grin: 

If I can make an editorial decision: let's use the char_wb, ngram_range=(3, 4). It matches what we do in the GapEncoder.

rcap107:
Great, then I'll clean up the script I am using and I'll put it somewhere, so I can find it if we decide to run more in-depth experiments (like by testing more tables, or changing the number of components or idk).

As for the StringEncoder, do I keep the HashingVectorizer at all? Or do I keep only the arguments relative to tf-idf?

jeromedockes:
>> I'm surprised the (1, 1 ) ngram with "char" or "char_wb" can perform so well -- isn't it just counting individual characters?
>
> char_wb makes sense: it's counting words and characters

IIUC correctly char_wb prevents char ngrams from crossing word boundaries but they're still only character ngrams no?

skrub/tests/test_string_encoder.py
```diff
@@ -0,0 +1,142 @@
...
+    result = se.fit_transform(encode_column)
+
+    # Converting dtypes to avoid nullable shenanigans
+    check_df = sbd.pandas_convert_dtypes(check_df)
+    result = sbd.pandas_convert_dtypes(result)
+
+    df_module.assert_frame_equal(check_df, result)
+
+
+def test_hashing(encode_column, df_module):
```
rcap107, line 55: I think this test is failing because the hashing vectorizer is not deterministic, if that's the case, I'm not sure how to handle it in the test other than exposing RNG everywhere. 

I didn't bother because we haven't decided whether to keep the hashing vectorizer in the first place.

rcap107:
For the record, this is the same example with the new defaults 
image

image

Pretty good result I think

jeromedockes:
nice!! do you mind trying the employee_salaries dataset too, as it has columns that are more like categories rather than text?

skrub/tests/test_string_encoder.py
jeromedockes: otherwise in the test you can just check the type and shape of the output and the content of `self.pipe`

rcap107:
> nice!! do you mind trying the employee_salaries dataset too, as it has columns that are more like categories rather than text?

I used R2 as metric, and did not test HashingVectorizer. It seems like when columns are more categories than text, then the GapEncoder is doing a better job than all the alternatives.
image

image

image

Still, StringEncoder is much faster and the performance is pretty good. TextEncoder is not doing quite as well on the other hand. 

Bottom line is, for more "categorical-looking" attributes, use the GapEncoder for better performance, or StringEncoder for decent performance and faster fit time, for text use TextEncoder for better performance or StringEncoder for decent performance and much faster fit time. 

This is from a pretty quick and dirty set of experiments, we could also try something like the heuristics from Leo's paper to choose, but that's going to get way more complicated. 

By the way, where should I put all the code I used for these results? So far I have it in my scratch folder.

skrub/_string_encoder.py
```diff
@@ -0,0 +1,201 @@
...
+            )
+        if not len(self.ngram_range) == 2:
+            raise ValueError(
+                f"`ngram_range` must have length 2, found {len(self.ngram_range)}."
+            )
+
+        if not isinstance(self.n_components, int) and self.n_components > 0:
+            raise ValueError(
+                f"`n_components` must be a positive integer, found {self.n_components}"
+            )
```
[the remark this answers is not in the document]
rcap107: I left in only the first check, the other constraints are in there, but I think they're less strict

GaelVaroquaux:
> IIUC correctly char_wb prevents char ngrams from crossing word boundaries but they're still only character ngrams no?

Good point. I was confusing with the "add_word" strategy of the GapEncoder ( https://github.com/skrub-data/skrub/blob/8a542bbe3674527feed8d4b0194dbedfe62ace20/skrub/_gap_encoder.py#L236 ). I would be interested if we could also explore this option. I seem to remember that it can help markedly, though come at a cost

GaelVaroquaux:
One last thing (I always come up with more :D ):

we'll have to be very careful to summarize the tradeoffs between the different encoders in a few lines (a few lines, something short and clear :D ) at the top of the corresponding section of the docs. It is very important that we convey to the user what we have learned

rcap107:
> One last thing (I always come up with more :D ):
> 
> we'll have to be very careful to summarize the tradeoffs between the different encoders in a few lines (a few lines, something short and clear :D ) at the top of the corresponding section of the docs. It is very important that we convey to the user what we have learned

This is something for a separate PR though

examples/02_text_with_string_encoders.py
```diff
@@ -132,7 +135,7 @@ def plot_gap_feature_importance(X_trans):
 # We set ``n_components`` to 30; however, to achieve the best performance, we would
 # need to find the optimal value for this hyperparameter using either |GridSearchCV|
 # or |RandomizedSearchCV|. We skip this part to keep the computation time for this
-# example small.
+# small example.
```
jeromedockes, line 138: to keep the computation time <missing word> for this ...

skrub/_string_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+class StringEncoder(SingleColumnTransformer):
+    """Generate a lightweight string encoding of a given column using tf-idf \
+        vectorization and truncated SVD.
+
+    First, apply a tf-idf vectorization of the text, then reduce the dimensionality
+    with a truncated SVD decomposition with the given number of parameters.
+
+    New features will be named `{col_name}_{component}` if the series has a name,
```
jeromedockes, line 21: I think you need double backticks

skrub/_string_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+    and `tsvd_{component}` if it does not.
+
+    Parameters
+    ----------
+    n_components : int, default=30
+        Number of components to be used for the PCA decomposition. Must be a
```
jeromedockes, line 27: to keep the number of acronyms under control maybe we should stick to "SVD" not "PCA". also we could have the expanded acronyms in parentheses the first time we mention them and links to their wikipedia pages in  a Notes section

skrub/_string_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+        positive integer.
+    vectorizer : str, "tfidf" or "hashing"
+        Vectorizer to apply to the strings, either `tfidf` or `hashing` for
```
jeromedockes, line 30: also here not sure what was your desired formatting -- single backticks will be italic, double for monospace

skrub/_string_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+        scikit-learn TfidfVectorizer or HashingVectorizer respectively.
+
+    ngram_range : tuple of (int, int) pairs, default=(3,4)
+        Whether the feature should be made of word or character n-grams.
```
jeromedockes, line 34: looks like the docs for n_gram_range and analyzer got swapped

skrub/_string_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+        Option ‘char_wb’ creates character n-grams only from text inside word
+        boundaries; n-grams at the edges of words are padded with space.
+
+    analyzer : str, "char", "word" or "char_wb", default="char_wb"
+        The lower and upper boundary of the range of n-values for different
+        n-grams to be extracted. All values of n such that min_n <= n <= max_n
+        will be used. For example an `ngram_range` of `(1, 1)` means only unigrams,
```
jeromedockes, line 41: same comment about rst vs markdown

skrub/_string_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+        `(1, 2)` means unigrams and bigrams, and `(2, 2)` means only bigrams.
+
+    See Also
+    --------
+    MinHashEncoder :
+        Encode string columns as a numeric array with the minhash method.
+    GapEncoder :
+        Encode string columns by constructing latent topics.
+    SimilarityEncoder :
+        Encode string columns as a numeric array with n-gram string similarity.
```
jeromedockes, line 51:
```suggestion
```

skrub/_string_encoder.py
```diff
@@ -0,0 +1,186 @@
...
+        if self.vectorizer == "tfidf":
+            self.pipe = Pipeline(
+                [
+                    (
+                        "tfidf",
+                        TfidfVectorizer(
+                            ngram_range=self.ngram_range, analyzer=self.analyzer
+                        ),
+                    ),
+                    ("tsvd", TruncatedSVD(n_components=self.n_components)),
```
jeromedockes, line 128: as in the textencoder, I think we need to handle the case where we end up with the smaller dimension of the tfidf < self.n_components (could happen for example if fitting on a column with few unique words and setting a large n_components and using the word analyzer). in that case we can do the same as textencoder ie keep `tfidf[:, :self.n_components]`

jeromedockes: (adding that logic might require you to move the svd out of the pipeline)

jeromedockes (commented):
mostly corner cases remaining :tada:  :)

rcap107:
I tested TableVectorizer with `drop` for `high_cardinality` and the result is pretty bad. I also tested GapEncoder with `add_words=True`, but it does not seem to help here.

image

image

image

I also tried to use OrdinalEncoder as @Vincent-Maladiere suggested, but I was not able to run the experiments :thinking: 
```python
from sklearn.preprocessing import OrdinalEncoder

ordinal_encoder = OrdinalEncoder(handle_unknown="use_encoded_value")

ordinal_encoder_pipe = clone(gap_pipe).set_params(
    **{"tablevectorizer__high_cardinality": ordinal_encoder}
)
ordinal_encoder_results = cross_validate(ordinal_encoder_pipe, X, y, scoring="r2")
results.append(("OrdinalEncoder", ordinal_encoder_results))
```
This gives me: 
```sh
    res = func(*args, **kwargs)
  File "/home/rcappuzz/Projects/skrub/skrub/_on_each_column.py", line 547, in _fit_transform_column
    raise ValueError(
ValueError: Transformer OrdinalEncoder.fit_transform failed on column 'division'. See above for the full traceback.
```
Plus a very long traceback I couldn't figure out.

Vincent-Maladiere:
Nice! So what is the conclusion regarding the `StringEncoder(1, 1)`? How can it perform so well against `drop` and `OrdinalEncoder`, when it only considers individual characters?

Vincent-Maladiere:
I'm happy that the string encoder looks like a great baseline for short, messy columns and long, free-form text as well.

rcap107:
> we'll have to be very careful to summarize the tradeoffs between the different encoders in a few lines (a few lines, something short and clear :D ) at the top of the corresponding section of the docs. It is very important that we convey to the user what we have learned This is something for a separate PR though
> I'd rather not. IMHO the docs need to be reorganized as we add complexity to the package. Also, the evidence for this recommendation comes from this PR.

I updated the doc page on the Encoders, but it was only to add the `StringEncoder` and a short summary of the different methods. Looking at the page, I think it would be better to expand on it with more detail for all encoders and maybe an explanation of the parameters, but that's something that would take way more effort (and definitely something for a separate PR).

rcap107:
> Nice! So what is the conclusion regarding the `StringEncoder(1, 1)`? How can it perform so well against `drop` and `OrdinalEncoder`, when it only considers individual characters?

image

My feeling is that OrdinalEncoder is just not that good if there is no order in the feature to begin with, while strings that are similar to each other usually are related no matter how they are sliced. 

I think an interesting experiment would be having a dictionary replacement where all strings in the starting table are replaced by random alphanumeric strings and check the performance of the encoders on that. In that case, I can imagine StringEncoder would not do so well compared to OrdinalEncoder.

doc/encoding.rst
```diff
@@ -10,10 +10,22 @@ Encoding or vectorizing creates numerical features from the data,
 converting dataframes, strings, dates... Different encoders are suited
 for different types of data.
 
-.. _dirty_categories:
+Summary
+.......
+:class:`StringEncoder` should be used in most cases when working with high-cardinality
```
rcap107, line 15: StringEncoder is not rendered properly, not sure how to fix this

jeromedockes: you need to add it to the reference documentation index:

https://github.com/skrub-data/skrub/blob/main/doc/reference/index.rst?plain=1#L43


so that sphinx can create a page and a link for it

skrub/_string_encoder.py
```diff
@@ -0,0 +1,200 @@
...
+                f"{min_shape}, which is too small to fit a truncated SVD with "
+                f"n_components={self.n_components}. "
+                "The embeddings will be truncated by keeping the first "
+                f"{self.n_components} dimensions instead. "
+            )
+            # self.n_components can be greater than the number
+            # of dimensions of result.
+            # Therefore, self.n_components_ below stores the resulting
+            # number of dimensions of result.
+            result = X_out[:, : self.n_components].toarray()
```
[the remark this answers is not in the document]
rcap107: Sure, not in this PR though

skrub/_string_encoder.py
```diff
@@ -0,0 +1,195 @@
...
+        result = sbd.make_dataframe_like(X, dict(zip(self.all_outputs_, result.T)))
+        result = sbd.copy_index(X, result)
+
+        return result
+
+    def __sklearn_is_fitted__(self):
+        """
+        Check fitted status and return a Boolean value.
+        """
+        return check_is_fitted(self, "all_outputs_")
```
jeromedockes, line 195: it doesn't hurt to have it but this method is not needed: by default scikit-learn's `check_is_fitted` will look for attributes with name ending with an underscore, so it would return the same result whether we override the default behavior by defining __sklearn_is_fitted__ or not

Vincent-Maladiere (commented):
OpenML downloads still fail and break the CI

jeromedockes (approved):
LGTM!! Thanks a lot @rcap107 this is a great addition :rocket:

Vincent-Maladiere (approved):
Looks good! Thanks @rcap107 :)

GaelVaroquaux:
I'm doing a last review of this PR. It's great, I love it. A tiny comment: I think that I prefer the comparison plot in log; if people agree, I'll implement the change
image

jeromedockes:
fine by me!

Vincent-Maladiere:
+1

Issue #1165, opened 2024-11-29
Add a dark mode to table reports


rouk1:
### Problem Description

Table report only has a light theme.

### Feature Description

The implementation could be in three parts:

- extract all colors to css variables
- use the created variables everywhere
- create a new set of variables (a dark theme), for a first round dark mode could be current color with inverted luminance
- use a media query to switch between variables sets

Implementation could look like:

```css
@media (prefers-color-scheme: dark) {
  :host {
    /* backgrounds */
    --color-background-primary: #1a1a1a;
    --color-background-secondary: #222;
    --color-background-plots: #121212;

    /* text */
    --color-text-primary: #fff;
    --color-text-secondary: #a2a2a2;

    /* shadow */
    --color-shadow: rgb(0 0 0/58%);
  }
}

@media (prefers-color-scheme: light) {
  :host {
    /* backgrounds */
    --color-background-primary: #fff;
    --color-background-secondary: #fafafb;
    --color-background-plots: #fdfdfd;

    /* text */
    --color-text-primary: #2f3037;
    --color-text-secondary: #616271;

    /* shadow */
    --color-shadow: rgb(0 0 0/19%);
  }
}
```

> [!NOTE]  
> Naming css variables is always tricky, my proposal is to use a consistent pattern: --what-where-variation-interactivity
> `what` could be colors for now but may scale to sizes
> `where` defines in which css rules the variable should be used (background for background, text for font color, ...)
> `variation` values may be variable according to context a title may not have exactly the same color as a subtitle for example
> `interactivity` value on hover for example

jeromedockes:
Thanks a lot for opening this @rouk1 !  Your plan sounds great. I'm looking forward to having someone who knows what they are doing (as opposed to me :sweat_smile: ) improve the display part of the reports :)

after the dark mode was mentioned in the skore issue I had started a branch (which was not ready for a PR yet) so just in case there's some information in that I opened #1166 

>  for a first round dark mode could be current color with inverted luminance

that sounds good, though note not much thought has gone into the current colors so once everything is organized with a reasonable set of variables we can also redefine all colors for both modes. 

> Naming css variables is always tricky, my proposal is to use a consistent pattern: --what-where-variation-interactivity

Indeed that part is rather chaotic in the tablereport ATM...

jeromedockes:
done in #1201

PR #1166, opened 2024-11-29
Add a dark theme for the tablereport

jeromedockes:
rough draft to start reorganizing the colors & add a dark variant.
I'm opening the PR to have a trace of it but we don't have to use it we can start over from scratch based on #1165

### DIFF
--- a/skrub/_reporting/_data/templates/base.css
+++ b/skrub/_reporting/_data/templates/base.css
@@ -28,23 +28,109 @@
     --radius: var(--space-s);
     --line-height: 1.5;
 
+    --blue: #3584e4;
+    --darkblue: #0f4e9b;
     --lightgreen: #eafaea;
     --mediumgreen: #d6f5d6;
     --darkgreen: 	#c1f0c1;
     --verydarkgreen: #84e184;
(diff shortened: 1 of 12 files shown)


jeromedockes:
the colors are very random and need to be changed   but I think at this point the report should be readable both in light & dark modes

screenshot_2024-11-29T09:47:32+01:00

jeromedockes:
I would rather not try to match the exact theme with a finer granularity than dark/light. IMO we should hard code a full set of colors for light mode and for dark mode and just match the mode rather than reuse colors from a vscode theme. because vscode is just one of the many places report can be displayed, and also I assume that as in any editor there is a myriad of themes and each user can come up with their own. so I think if we try to reuse colors from vscode themes we can end up having to maintain many variants, spend time checking them every time we change something, and most importantly risk having some parts unreadable due to low contrast in some environments

GaelVaroquaux:
> IMO we should hard code a full set of colors for light mode and for dark mode and just match the mode rather than reuse colors from a vscode theme.

I would agree. It's just about checking that it is visible and legible in all settings.

jeromedockes:
closing in favor of #1201

PR #1169, opened 2024-11-29
MAINT compatibility sklearn 1.6.0

glemaitre:
Alternative to #1135
closes #1135 

Vendor the version of `sklearn-compat` (https://github.com/sklearn-compat/sklearn-compat).
This is still an RC but I wanted to give it a try.

We discussed with @jeromedockes to vendor the file instead of depending on the package.


glemaitre:
We will probably need to release 0.4.1 for this change when scikit-learn 1.6.0 will be released next week.

skrub/_datetime_encoder.py
```diff
@@ -323,3 +324,11 @@ def _check_params(self):
             raise ValueError(
                 f"'resolution' options are {allowed}, got {self.resolution!r}."
             )
+
+    def _more_tags(self):
+        return {"preserves_dtype": []}
+
+    def __sklearn_tags__(self):
+        tags = super().__sklearn_tags__()
+        tags.transformer_tags = TransformerTags(preserves_dtype=[])
```
adrinjalali, line 333: does this class not inherit from `TransformerMixin`? Why not 

```suggestion
        tags.transformer_tags.preserves_dtype = []
```

to not override all the other tags from the parent classes?

glemaitre: We get into trouble with the `set_output` for this one: https://github.com/skrub-data/skrub/pull/1135#discussion_r1837807800

glemaitre: Actually this is not mentioned in the comment. @jeromedockes can probably offer the detailed perspective.

adrinjalali: I understand that comment as "we don't need it, we don't have to use it". Now with tags, you kinda better use them, and if there's something you don't like from that mixin, you can always override.

glemaitre: > Now with tags, you kinda better use them, and if there's something you don't like from that mixin, you can always override.

I recall that the discussion was around the `set_output` (we need @jeromedockes for the details but it might be related to avoiding any data copy or dtype conversion - keep the feature of the container library in short). We are fine having the tags, but inheriting from the `TransformerMixin` brings the all the infrastructure of `set_output`. Overwriting the inheritance is not easy while redefining the `Tags` that as in `TransformerMixin` is easy.

adrinjalali: If wrapping with `set_output` is adding copies, we should fix that.

jeromedockes: it does not add copies, at least when the output is already in the type of container the configuration is asking for. (AFAICT)

skrub/_sklearn_compat.py
```diff
@@ -0,0 +1,572 @@
+"""Ease developer experience to support multiple versions of scikit-learn.
```
jeromedockes, line 1: can we

- have the URL of where this file came from here
- exclude it from codecov, pre-commit formatting etc. maybe we should have something like a `vendor/` submodule?

jeromedockes:
Hi @glemaitre I just realized that after we discussed it again and decided the PR was ready to merge I went again and ... merged the wrong PR! :facepalm:  -- I merged #1135  instead of this one. I know both work but this one with scikit-learn compat was your recommendation right?

I guess the best way forward is to revert the merge? WDYT

jeromedockes (approved):

PR #1172, opened 2024-12-03
remove use of matplotlib.rc_context

### LINKED ISSUE #1171: Interaction between `TableReport` and `matplotlib`
Given the following code blocks:

```python
# %%
import skrub

# %%
from sklearn.datasets import load_iris
data, target = load_iris(return_X_y=True, as_frame=True)
TableReport(data)

# %%
data["sepal length (cm)"].plot.hist()
```

We expect (in a notebook or vs code) that running those cell will end-up showing the report and the histogram plot. However, the later cell requires a `plt.show()` to show the plot. This is not the behaviour when not patching the display.

Also it seems that if we intend a first plot, then the interaction with matplotlib does not happen:

```python
# %%
import skrub

# %%
from sklearn.datasets import load_iris
data, target = load_iris(return_X_y=True, as_frame=True)

# %%
data["sepal length (cm)"].plot.hist()

# %%
skrub.TableReport(data)

# %%
data["sepal length (cm)"].plot.hist()
```

I'm wondering if this is something expected?

jeromedockes:
fixes #1171

matplotlib.rc_context seems to prevent the inline display of plots in jupyter notebooks in some cases
this removes the use of this context manager

the context manager does not seem to do anything fancier than the try/finally statement in this PR, but the difference is here we only set the key we are interested in `svg.fonttype` whereas upon exit `rc_context` updates all params except the backend:

https://github.com/matplotlib/matplotlib/blob/v3.9.3/lib/matplotlib/__init__.py#L1142-L1196

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -32,6 +32,11 @@ Changes
 
 Bug fixes
 ---------
+* Generating a ``TableReport`` could have an effect on the matplotib
+  configuration which could cause plots not to display inline in jupyter
+  notebooks any more. This has been fixed in skrub in :pr:`1172` by
+  :user:`Jérôme Dockès <jeromedockes>` and the matplotlib issue can be tracked
+  [here](https://github.com/matplotlib/matplotlib/issues/25041).
 
 
 Release 0.4.0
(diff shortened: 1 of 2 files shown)


glemaitre (approved):
LGTM. Thanks @jeromedockes

jeromedockes:
I was going to merge it now but still needs a changelog entry so putting back in draft mode

Issue #1175, opened 2024-12-03
Better control over the TableReport's progress display


jeromedockes:
### Problem Description

In some cases the TableReport progress display clutters the output.

- it uses carriage return to update the line it prints. in some cases (eg when stdout is a file, when rendering notebooks with `nbconvert` or generating slides with `quarto`) this just results in many lines which the user may want to avoid. then we may want it to just print an increasing row of "." or similar
- sometimes we may want it to be completely silent
- it should print progress information to stderr, not stdout

### Feature Description

- write progress info to the standard error stream
- maybe add a `verbose` parameter?

glemaitre:
Indeed, I was going to show on such case when creating a jupyter book:

![Uploading image.png…]()

reshamas:
I am working on this: Pearson's correlation coefficient.
skrub/_column_associations.py

via: PyLadiesCon 2024 online sprint

rcap107:
We might want to add rich (https://rich.readthedocs.io/en/stable/) as an optional dependency to deal with this

PR #1182, opened 2024-12-06
Add a verbosity parameter to TableReport to control the printing of messages in stdout when opening summary report

priscilla-b:
Towards #1175


skrub/_reporting/_summarize.py
```diff
@@ -45,6 +46,11 @@ def summarize_dataframe(
         Maximum number of rows from the end of the dataframe to show in the
         sample table.
 
+    verbose : int, default = 1
+        A switch that determines if messages show in the standard output
+        when report summary is being opened. Defaults to 1 to show messages.
+        Set it to 0 to avoid displaying messages in standard output.
```
jeromedockes, line 52:
```suggestion
        Whether to print progress information while the report is being generated.

        * verbose = 1 prints how many columns have been processed so far.
        * verbose = 0 silences the output.
```

Just a nitpick suggestion to say that what we print is information about the progress, and it happens when creating the report even if we do not open it. Please feel free to reword it!

jeromedockes (commented):
Thank you very much @priscilla-b ! I took the liberty to push a small change to align the identation of the paragraph for the 'verbose' parameter in the docstring (the tools that check and generate html from the docstrings are quite picky about whitespace :sweat_smile: ) and the tests are passing now, which is great!

skrub/_reporting/_table_report.py
```diff
@@ -35,6 +35,10 @@ class TableReport:
         mapping with the keys ``display_name`` (the name shown in the menu,
         e.g. ``"First 10 columns"``) and ``columns`` (a list of column names).
         See the end of the "Examples" section below for details.
+    verbose : int, default = 1
+       A switch that determines if messages show in the standard output
+       when report summary is being opened. Defaults to 1 to show messages.
+       Set it to 0 to avoid displaying messages in standard output.
```
jeromedockes, line 41:
```suggestion
        Whether to print progress information while the report is being generated.

        * verbose = 1 prints how many columns have been processed so far.
        * verbose = 0 silences the output.
```

jeromedockes:
The last thing we need to take care of will be to edit the changelog, so that when we release the next version of skrub users can easily find out about this nice addition!

could you add an entry to [this part of the changelog](https://github.com/skrub-data/skrub/blob/396e9bfbf6722b552cd71aeb844b35b1a9ec0b74/CHANGES.rst?plain=1#L20), you can see the formatting in other entries below in the same file. 

in particular it needs to contain a sentence like

```
:pr:`123` by :user:`My Name <mygithubusername>`
```

to link the changelog entry to the PR and your GitHub account

skrub/_reporting/_summarize.py
priscilla-b: Thank you  very much @jeromedockes. Was wondering why the tests were failing.

Also on the suggested changes to the documentation, I think we generalized it because setting  verbose = 1 also prints that extra message (flush = True) and not just the columns processed

jeromedockes: I see,  that makes sense! note the `print(flush=True)` is just to start a new line (move the terminal's cursor one line down so that the next printed thing doesn't get mangled with the report's output), it doesn't really print any content so we can think of it as wrapping up the display of progress.

priscilla-b: Right, that makes sense. Committing your suggestion.

jeromedockes: thanks! I'll also commit it in the TableReport's docstring in that case :)

jeromedockes:
all green, merging! :tada:  thanks @priscilla-b  :)

priscilla-b:
>  as a follow up we could also expose that option in the [`patch_display`](https://skrub-data.org/stable/reference/generated/skrub.patch_display.html) function (which is used in the upcoming course I mentioned yesterday)?

Sure, going to look into this @jeromedockes

jeromedockes:
great, thanks! I'm on the discord even when I'm not in the skrub room so don't hesitate to ping me there if you have any questions

PR #1185, opened 2024-12-07
Updating contributing doc to add more detail on how to start 

rcap107:
I added more detail on how to prepare the dev environment and reordered some of the sections to hopefully make the process clearer.


CONTRIBUTING.rst
```diff
@@ -124,15 +124,32 @@ See the relevant sections above on how to do this.
...
-Follow the steps in the :ref:`installation_instructions` > "From Source" section
-to set up your environment, install the required development dependencies, and
-run the tests.
+To contribute, you will first have to run through some steps:
 
-When starting to work on a new issue, it's recommended to create a new branch:
+- Set up your environment by forking the repository ((`Github doc on
+  forking and
+  cloning <https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo>`__)).
+- While at the root of your local copy of skrub, install the required
```
jeromedockes, line 132: maybe a short paragraph about creating a virtual env?

rcap107: good point

CONTRIBUTING.rst
```diff
@@ -129,8 +129,12 @@ To contribute, you will first have to run through some steps:
 - Set up your environment by forking the repository ((`Github doc on
   forking and
   cloning <https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo>`__)).
-- While at the root of your local copy of skrub, install the required
-  development dependencies by running
+- Create a new virtual environment either with
```
rcap107, line 132: should I add more detail?

jeromedockes: as they're quite short I would say yes, this one someone following along can just copy-paste without needing to follow the links

CONTRIBUTING.rst
```diff
@@ -124,15 +124,36 @@ See the relevant sections above on how to do this.
...
+   git checkout -b my-branch-name-eg-fix-issue-123
+   # make some changes
+   git add ./the/file-i-changed
+   git commit -m "my message"
+   git push --set-upstream origin my-branch-name-eg-fix-issue-123
 
-   git switch -c branch_name
+At this point, if you visit again the `pull requests
+page <https://github.com/skrub-data/skrub/pulls>`__ github should show a
+banner asking if you want to open a pull request from your new branch.
```
jeromedockes, line 156: also I don't think the "Implementation guidelines" below is very helpful

CONTRIBUTING.rst
```diff
@@ -124,15 +124,36 @@ See the relevant sections above on how to do this.
...
+  forking and
+  cloning <https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/working-with-forks/fork-a-repo>`__)).
+- Create a new virtual environment either with
+  `venv <https://docs.python.org/3/library/venv.html>`__ or
+  `conda <https://docs.conda.io/projects/conda/en/latest/user-guide/tasks/manage-environments.html>`__,
+  then activate it.
+- While at the root of your local copy of skrub and within the new env,
+  install the required development dependencies by running
+  ``pip install --editable ".[dev, lint, test, doc]"``.
+- Run all the tests by running ``pytest -s skrub/tests``.
```
jeromedockes, line 139: this is not something we need to do right away right? especially as it can take a while

CONTRIBUTING.rst
```diff
@@ -124,15 +124,36 @@ See the relevant sections above on how to do this.
...
+- Run ``pre-commit install`` to activate some checks that will run every
+  time you do a ``git commit`` (mostly, formatting checks).
```
jeromedockes, line 141: I think one thing that can be confusing with pre-commit is it can modify some files and cancel the commit, and then you need to stage its modification and run `commit` again. I'm not sure what would be the best place to add a little explanation about that without making this section too long :thinking:

CONTRIBUTING.rst
```diff
@@ -197,6 +218,7 @@ Once you are satisfied with your changes, you can run all the tests to make sure
 that your change did not break code elsewhere:
 
 .. code:: sh
+
```
jeromedockes, line 221: maybe another important pytest feature we could mention is running a specific test either with `pytest the_file.py::the_test` or `pytest the_file.py -k 'test_name_pattern'` to avoid waiting for all the other tests to run.


Also, I've noticed that in some cases in windows `pytest` can fail (I guess that depends on how it was installed) while `python -m pytest` seems a bit more robust. in the CI it is good to use `pytest` to avoid adding the current working directory to the python path but for local development I think `python -m pytest` is a good option; WDYT?

jeromedockes: also we could add a link to the pytest documentation

CONTRIBUTING.rst
rcap107: should it be removed?

CONTRIBUTING.rst
rcap107: I added a section on that a bit further down

CONTRIBUTING.rst
rcap107: added notes

README.rst
```diff
@@ -32,7 +32,14 @@ The goal of skrub is to bridge the gap between tabular data sources and machine-
 
 skrub provides high-level tools for joining dataframes (``Joiner``, ``AggJoiner``, ...),
 encoding columns (``MinHashEncoder``, ``ToCategorical``, ...), building a pipeline
-(``TableVectorizer``, ``tabular_learner``, ...), and more.
+(``TableVectorizer``, ``tabular_learner``, ...), and explore interactively your data (``TableReport``).
+
+.. figure::
```
rcap107, line 37: This is not necessarily the final position, I can move it around if needed

CONTRIBUTING.rst
```diff
@@ -124,15 +124,51 @@ See the relevant sections above on how to do this.
...
+- Run ``pre-commit install`` to activate some checks that will run every
+  time you do a ``git commit`` (mostly, formatting checks).
+
+If you want to make sure that everything runs properly, you can run all
+the tests with the command ``pytest -s skrub/tests``; note that this may
+take a long time. Some tests may raise warnings such as:
 
-When starting to work on a new issue, it's recommended to create a new branch:
+.. code:: sh
```
Vincent-Maladiere, line 152: The generated documentation does not render this (see circle CI output).

CONTRIBUTING.rst
Vincent-Maladiere: @jeromedockes Do you think it's too generic? Otherwise, I believe it summarizes our first principles quite well for a new contributor.

> Avoid using binary extensions, Cython, or other compiled languages.

This one for example is not obvious and it's good to have it highlighted there.

CONTRIBUTING.rst
```diff
@@ -229,6 +276,21 @@ the docstrings. Check for possible problems by running
 
    pytest skrub/path/to/file
 
+
+Formatting and pre-commit checks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Formatting the code well helps with code development and maintenance,
+which is skrub requires that all commits follow a specific set of
```
Vincent-Maladiere, line 284: missing word?

CONTRIBUTING.rst
```diff
@@ -229,6 +276,21 @@ the docstrings. Check for possible problems by running
 
    pytest skrub/path/to/file
 
+
+Formatting and pre-commit checks
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Formatting the code well helps with code development and maintenance,
```
Vincent-Maladiere, line 283: I don't think we need to motivate why we use formatting TBH

CONTRIBUTING.rst
```diff
@@ -273,18 +328,32 @@ actions are taken.
...
+  correct.
+
+Integration
+^^^^^^^^^^^
+
+Community consensus is key in the integration process. Expect a minimum
+of 1 to 3 reviews depending on the size of the change before we consider
+merging the PR.
+
+Please be mindful that maintainers are volunteers, so review times may vary.
```
Vincent-Maladiere, line 356: > Please be mindful that maintainers are volunteers

I don't think this is technically correct nowadays 😁

Vincent-Maladiere (approved):
A couple of remarks

Vincent-Maladiere:
Not very important, but this looks quite inaccurate to me:

> Please be mindful that maintainers are volunteers, so review times may vary.

rcap107:
> Not very important, but this looks quite inaccurate to me:
> 
> > Please be mindful that maintainers are volunteers, so review times may vary.

I removed that line

jeromedockes (approved):
Thanks a lot @rcap107 ! we'll keep improving this documentation as it is super important, as we said separating 2 sections one for first-time setup and one for more advanced details. this is already a significant improvement, another nice outcome from the sprint!

Issue #1181, opened 2024-12-06
Create examples for the column associations metrics


reshamas:
### Describe the issue linked to the documentation

https://skrub-data.org/stable/reference/generated/skrub.column_associations.html

Metrics:  
- Cramer’s V statistic
- other metrics

Working example from pandas:
- https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.corr.html
- can also find examples in skrub

jeromedockes:
to recap what we said yesterday: the `column_associations` function is missing an "Examples" section in its docstring. It would be great to add a small example section to show generating the associations for a toy example dataframe, so users can see how to call it and what would be the shape and content of the output. something like

```
Examples
--------
>>> import pandas as pd
>>> from skrub import column_associations
>>> df = ...
>>> column_associations(df)
       ...
```

jeromedockes:
I'll mark this one as completed: when more association metrics are added they will show up in the output of the example added in #1186 

When we add more metrics we will also update the documentation to include a short paragraph and wikipedia link for each.

To be continued in #1176  ...

PR #1186, opened 2024-12-07
DOC add example for Cramer V for column_associations

reshamas:
add example for Cramer V

Will be rendered here in documentation:  
https://skrub-data.org/stable/reference/generated/skrub.column_associations.html

Towards #1181

### DIFF
--- a/skrub/_column_associations.py
+++ b/skrub/_column_associations.py
@@ -18,8 +18,8 @@ def column_associations(df):
 
     The result is returned as a dataframe with columns:
 
-    ['left_column_name', 'left_column_idx', 'right_column_name',
-    'right_column_idx', 'cramer_v']
+    `['left_column_name', 'left_column_idx', 'right_column_name',
+    'right_column_idx', 'cramer_v']`
 
     As the function is commutative, each pair of columns appears only once
     (either col_1, col_2 or col_2, col_1 but not both). The results are sorted
@@ -180,6 +180,31 @@ def _compute_cramer(table, n_samples):
 
     This returns the symmetric matrix with shape (n cols, n cols) where entry
     i, j contains the statistic for column i x column j.
+
+    Returns
+    -------
+    dataframe
+        of computed associations.
+
+    Notes
+    -----
+    Cramér's V is a measure of association between two nominal variables,
+    giving a value between 0 and +1 (inclusive).
+    * `Cramer V <https://en.wikipedia.org/wiki/Cramér%27s_V>`_
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import pandas as pd
+    >>> import skrub
+    >>> pd.set_option('display.precision', 4)
+    >>> rng = np.random.default_rng()
+    >>> df = pd.DataFrame({f"c_{i}": rng.random(size=20)*10 for i in range(5)})
+    >>> df["c_str"] = [f"val {i}" for i in range(df.shape[0])]
+    >>> df.shape()
+    >>> df.head()
+    >>> associations = skrub.column_associations(df)
+    >>> associations
     """
     marginal_0 = table.sum(axis=-2)
     marginal_1 = table.sum(axis=-1)
--- a/skrub/_column_associations.py
+++ b/skrub/_column_associations.py
@@ -41,6 +41,26 @@ def column_associations(df):
     -------
     dataframe
         The computed associations.
+
+    Notes
+    -----
+    Cramér's V is a measure of association between two nominal variables,
+    giving a value between 0 and +1 (inclusive).
+    * `Cramer V <https://en.wikipedia.org/wiki/Cramér%27s_V>`_
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import pandas as pd
+    >>> import skrub
+    >>> pd.set_option('display.precision', 4)
+    >>> rng = np.random.default_rng()
+    >>> df = pd.DataFrame({f"c_{i}": rng.random(size=20)*10 for i in range(5)})
+    >>> df["c_str"] = [f"val {i}" for i in range(df.shape[0])]
+    >>> df.shape
+    >>> df.head()
+    >>> associations = skrub.column_associations(df)
+    >>> associations
     """
     return _stack_symmetric_associations(_cramer_v_matrix(df), df)
 
@@ -181,30 +201,6 @@ def _compute_cramer(table, n_samples):
     This returns the symmetric matrix with shape (n cols, n cols) where entry
     i, j contains the statistic for column i x column j.
 
-    Returns
-    -------
-    dataframe
-        of computed associations.
-
-    Notes
-    -----
-    Cramér's V is a measure of association between two nominal variables,
-    giving a value between 0 and +1 (inclusive).
-    * `Cramer V <https://en.wikipedia.org/wiki/Cramér%27s_V>`_
-
-    Examples
-    --------
-    >>> import numpy as np
-    >>> import pandas as pd
-    >>> import skrub
-    >>> pd.set_option('display.precision', 4)
-    >>> rng = np.random.default_rng()
-    >>> df = pd.DataFrame({f"c_{i}": rng.random(size=20)*10 for i in range(5)})
-    >>> df["c_str"] = [f"val {i}" for i in range(df.shape[0])]
-    >>> df.shape()
-    >>> df.head()
-    >>> associations = skrub.column_associations(df)
-    >>> associations
     """
     marginal_0 = table.sum(axis=-2)
     marginal_1 = table.sum(axis=-1)
--- a/skrub/_column_associations.py
+++ b/skrub/_column_associations.py
@@ -46,6 +46,7 @@ def column_associations(df):
     -----
     Cramér's V is a measure of association between two nominal variables,
     giving a value between 0 and +1 (inclusive).
+
     * `Cramer V <https://en.wikipedia.org/wiki/Cramér%27s_V>`_
 
     Examples
--- a/skrub/_column_associations.py
+++ b/skrub/_column_associations.py
@@ -54,13 +54,36 @@ def column_associations(df):
     >>> import pandas as pd
     >>> import skrub
     >>> pd.set_option('display.precision', 4)
-    >>> rng = np.random.default_rng()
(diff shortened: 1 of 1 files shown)


jeromedockes:
Hi @reshamas thanks for opening this! such examples directly in the reference documentation are super useful to understand how to use a library.

Your example is great! I think `>>> df.shape()` needs to be replaced with `>>> df.shape` to fix the error. to run the doctests on your machine, which can be easier than waiting for the CI to run you can use the command `pytest skrub/_column_associations.py`

jeromedockes:
Also I see you add the example section to the `_compute_cramer` function. This one is a "private" helper function used only within the `_column_associations` module itself. you can tell because its name starts with an underscore, and it does not have an entry in the skrub documentation. Could we move the Examples section that you wrote to the docstring of the `column_associations` function [here](https://github.com/skrub-data/skrub/blob/6ac0aebdeccc13e871d25aba0582f64996073da9/skrub/_column_associations.py#L43)? thanks!

GaelVaroquaux:
Super good idea to have done this as a docstring example

skrub/_column_associations.py
```diff
@@ -41,6 +41,26 @@ def column_associations(df):
...
+    giving a value between 0 and +1 (inclusive).
+    * `Cramer V <https://en.wikipedia.org/wiki/Cramér%27s_V>`_
+
+    Examples
+    --------
+    >>> import numpy as np
+    >>> import pandas as pd
+    >>> import skrub
+    >>> pd.set_option('display.precision', 4)
+    >>> rng = np.random.default_rng()
```
rcap107, line 57: Doctests will complain if the output of `df.head()` is not exactly the same every time, so this seed should be set to some value like `np.random.default_rng(33)`

skrub/_column_associations.py
```diff
@@ -41,6 +41,26 @@ def column_associations(df):
...
+    >>> df = pd.DataFrame({f"c_{i}": rng.random(size=20)*10 for i in range(5)})
+    >>> df["c_str"] = [f"val {i}" for i in range(df.shape[0])]
+    >>> df.shape
```
rcap107, line 60: This line should be followed by another line that's the output of this command

```py
>>> df.shape
(20, 6)
```

skrub/_column_associations.py
```diff
@@ -41,6 +41,26 @@ def column_associations(df):
...
+    >>> df.head()
```
rcap107, line 61: The same for this line and the one past `associations`

rcap107:
Hi @reshamas, thanks for contributing! I added a few comments that should fix the tests.

rcap107:
You might also want to run `pytest skrub/tests/test_docstrings.py` to check everything runs properly.

skrub/_column_associations.py
reshamas: add seed

skrub/_column_associations.py
reshamas: added output for shape

skrub/_column_associations.py
reshamas: added output for associations

reshamas: It looks like the output is too wide (> 80 chars):
```
       left_column_name  left_column_idx right_column_name  right_column_idx  cramer_v
    0               c_3                3             c_str                 5    0.8215
    1               c_1                1               c_4                 4    0.8215
    2               c_0                0               c_1                 1    0.8215
    3               c_2                2             c_str                 5    0.7551
    4               c_0                0             c_str                 5    0.7551
    5               c_0                0               c_3                 3    0.7551
    6               c_1                1               c_3                 3    0.6837
    7               c_0                0               c_4                 4    0.6837
    8               c_4                4             c_str                 5    0.6837
    9               c_3                3               c_4                 4    0.6053
    10              c_2                2               c_3                 3    0.6053
    11              c_1                1             c_str                 5    0.6053
    12              c_0                0               c_2                 2    0.6053
    13              c_2                2               c_4                 4    0.5169
    14              c_1                1               c_2                 2    0.4122
```

rcap107: It's something to do with different configurations for different pandas versions, tests fail on the min reqs configuration :thinking:

rcap107: Adding  `>>> pd.set_option('expand_frame_repr', False)` seems to print it properly, however now there's a problem with the ordering of the rows, which for whatever reason changes between pandas 1.5 and pandas 2.2.

Not sure how we can get around this, @jeromedockes any idea?

jeromedockes:
sorry for the delay I'll look at it now

skrub/_column_associations.py
jeromedockes: so I think this should be ok for the display:
```
    >>> pd.set_option('display.width', 200)
    >>> pd.set_option('display.max_columns', 10)
    >>> pd.set_option('display.precision', 4)
```

and then at the end of the example

```
    >>> pd.reset_option('display.width')
    >>> pd.reset_option('display.max_columns')
    >>> pd.reset_option('display.precision')
```

skrub/_column_associations.py
```diff
@@ -41,6 +41,55 @@ def column_associations(df):
...
+    (20, 6)
+    >>> df.head()
+          c_0     c_1     c_2     c_3     c_4  c_str
+    0  4.4364  4.0114  6.9271  7.0970  4.8913  val 0
+    1  5.6849  0.7192  7.6430  4.6441  2.5116  val 1
+    2  9.0810  9.4011  1.9257  5.7429  6.2358  val 2
+    3  2.5425  2.9678  9.7801  9.9879  6.0709  val 3
+    4  5.8878  9.3223  5.3840  7.2006  2.1494  val 4
+    >>> associations = skrub.column_associations(df)
+    >>> associations
```
jeromedockes, line 73:
```suggestion
    >>> associations # doctest: +SKIP
```

Enough battling with the display options :) because there are some ties in the cramer_v the ordering of the rows can change. as this docstring will be revisited soon as we add the pearson correlation we can just skip this one

jeromedockes:
I think we should have a little function in the `datasets` module like `toy_dataframe()` that we can use in the reference examples so simplify the setup code and so we get the same every time

also we probably can set pandas display options before running the doctests and perhaps reset them before and after running each module's doctests

I don't want this stuff to block this PR though, because it really helps clarify what the output of the `column_associations` contains. I suggest we skip this one doctest output, as we do in a few other places where ensuring the output is exactly consistent would make the example too complicated, defeating its documentation purpose.

skrub/_column_associations.py
```diff
@@ -18,14 +18,14 @@ def column_associations(df):
 
     The result is returned as a dataframe with columns:
 
-    ['left_column_name', 'left_column_idx', 'right_column_name',
-    'right_column_idx', 'cramer_v']
+    `['left_column_name', 'left_column_idx', 'right_column_name',
+    'right_column_idx', 'cramer_v']`
```
jeromedockes, line 22:
```suggestion
    ``['left_column_name', 'left_column_idx', 'right_column_name',
    'right_column_idx', 'cramer_v']``
```

rcap107 (approved):

skrub/_column_associations.py
```diff
@@ -18,14 +18,14 @@ def column_associations(df):
...
 
     As the function is commutative, each pair of columns appears only once
-    (either col_1, col_2 or col_2, col_1 but not both). The results are sorted
+    (either `col_1`, `col_2` or `col_2`, `col_1` but not both). The results are sorted
```
jeromedockes, line 25:
```suggestion
    (either ``col_1``, ``col_2`` or ``col_2``, ``col_1`` but not both).
    The results are sorted
```

skrub/_column_associations.py
jeromedockes: sphinx uses restructured text which unlike markdown wants 2 backticks for formatting code :shrug: one backtic gives italic font -- I don't think I well ever get used to it

skrub/_column_associations.py
```diff
@@ -173,13 +222,14 @@ def _contingency_table(encoded):
...
 def _compute_cramer(table, n_samples):
-    """Compute the Cramer V statistic given a contingency table.
+    """Compute the Cramer's V statistic given a contingency table.
 
     The input is the table computed by ``_contingency_table`` with shape
     (n cols, n cols, n bins, n bins).
 
     This returns the symmetric matrix with shape (n cols, n cols) where entry
     i, j contains the statistic for column i x column j.
+
```
jeromedockes, line 232:
```suggestion
```

jeromedockes: for some reason numpy doc does not like an empty line at the end of a docstring -- an annoying quirk

skrub/_column_associations.py
```diff
@@ -41,6 +41,55 @@ def column_associations(df):
...
     dataframe
         The computed associations.
+
+    Notes
+    -----
+    Cramér's V is a measure of association between two nominal variables,
+    giving a value between 0 and +1 (inclusive).
+
+    * `Cramer's V <https://en.wikipedia.org/wiki/Cramér%27s_V>`_
+
```
jeromedockes, line 51: thanks for adding the explanation and the link that's super useful!

jeromedockes:
the rendred output looks great: https://output.circle-artifacts.com/output/job/c186c9ef-59a1-47b8-81f2-3b901a436f1b/artifacts/0/doc/reference/generated/skrub.column_associations.html

jeromedockes (approved):
Looks great!! thanks a ton @reshamas . this function was moved to the public api in the last release as we realized it can be very useful on its own ... but it was lacking a proper documentation and examples! battling with pandas output formatting details across multiple versions of pandas and python is not fun but thanks a lot for pushing through :)

reshamas:
Thank you @jeromedockes and @rcap107. It was a very good learning experience.

rcap107:
> Thank you @jeromedockes and @rcap107. It was a very good learning experience.

Likewise, thanks for contributing!

jeromedockes:
yep, thanks a lot! I know you had also started working on adding Pearson correlations to the output,  so if you have any questions about that don't hesitate to reach us here on github or on the skrub discord server: https://discord.gg/ABaPnm7fDC

PR #1192, opened 2024-12-09
[DOC] add link to bluesky account instead of twitter

jeromedockes:
Ping @rcap107 who has become our official social media / PR person despite himself :stuck_out_tongue_closed_eyes:

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -175,9 +175,9 @@ html_theme_options = {
     "header_links_before_dropdown": 4,
     "icon_links": [
         {
-            "name": "Twitter",
-            "url": "https://twitter.com/skrub_data",
-            "icon": "fa-brands fa-twitter",
+            "name": "Bluesky",
+            "url": "https://bsky.app/profile/skrub-data.bsky.social",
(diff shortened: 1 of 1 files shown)


rcap107:
"volunteered" 

Not sure if we should replace the icon entirely, or keep both bsky and twitter, @GaelVaroquaux what do you think?

jeromedockes:
note the pydata sphinx theme does not seem to like more than 2 icons -- but we already have that problem with the current number of links

skrub-screenshot

Vincent-Maladiere:
What happens if we have 4 icons? I'm slightly uncomfortable leaving Twitter entirely, some people could see a politically motivated move

GaelVaroquaux:
>  I'm slightly uncomfortable leaving Twitter entirely, some people could see a politically motivated move

We're not leaving twitter. We'll still have a presence there. The question is: which icon to put on our page. We have more followers on bluesky, so I'm in favor of putting the bluesky one.

jeromedockes:
superseded by #1212

PR #1193, opened 2024-12-09
Fix nightly builds

jeromedockes:
adapt to the upcoming releases of pandas and scipy

mostly adjusting tests after pandas introduces the datetime64[s] and datetime64[us] types.
but also in deduplicate, I think guess_n_clusters was computing silhoutte scores for n_cluster ranging from to to n * (n -1) / 2 instead of 2 to n , where n is the number of unique values. now this results in an error from scipy but i guess it was a mistake anyway


glemaitre:
@jeromedockes Could you push a commit with the `[deps nightly]`

jeromedockes:
@jovan-stojanovic  I don't remember did you want to have another look at the deduplicate n clusters computation?

jeromedockes:
I think it would be good to have this in before we do a release this week

glemaitre (approved):
LGTM.

PR #1196, opened 2024-12-11
FIX better display of object columns with multi-line repr in tablereport

### LINKED ISSUE #1195: TableReport doesn't like python objects like nested sklearn estimators
### Describe the bug

In the example below, `TableReport` uses the string representation of the sklearn pipeline as a category.

Screenshot 2024-12-11 at 16 54 33

### Steps/Code to Reproduce

```python
from skrub import TableVectorizer, tabular_learner

TableReport(
    pd.DataFrame(
        dict(
            a=[
                tabular_learner("regressor"),
                tabular_learner("classifier"),
            ]
        )
    )
)
```

### Expected Results

A nonnested representation of the object? In the worst-case scenario, its Python id? Not sure

### Actual Results

See screenshot above

### Versions

```shell
'0.4.dev0'
```

jeromedockes:
fixes #1195

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -40,6 +40,10 @@ Bug fixes
   :user:`Jérôme Dockès <jeromedockes>` and the matplotlib issue can be tracked
   [here](https://github.com/matplotlib/matplotlib/issues/25041).
 
+* The labels on bar plots in the ``TableReport`` for columns of object dtypes
+  that have a repr spanning multiple lines could be unreadable. This has been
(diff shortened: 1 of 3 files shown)


jeromedockes:
using the example from the issue, on main branch:

screenshot_2024-12-11T18:10:50+01:00


with the proposed changes:


screenshot_2024-12-11T18:10:16+01:00

TheooJ (approved):
LGTM !

PR #1201, opened 2024-12-13
feat: TableReport dark theme

rouk1:
This PR introduces a dark mode for the TableReport.
I tried to use as much as possible css variables all centralized in a `_variables.css` file.

# UI preview

https://github.com/user-attachments/assets/7201da44-306b-4d8a-a816-1c94aabf52cf


jeromedockes:
Hi @rouk1 ! thanks so much for doing this :rocket:  the TableReport looks really nice in the proposed dark mod:star_struck: 
also having more systematic naming for all css variables will definitely help with maintainability

I noticed that some aspects that are not related to colors are affected by the PR as well; here are a first few comments

- I noticed `standalone-report.css` has been removed; this is where we set the font when we ask the report for a full html page (when we ask for a fragment it uses the font of the parent element). as a result the report now uses the default browser's font but I would like to keep a sans-serif font which I think tends to be a bit more compact and more readable for short text
- that's really a detail but a line has appeared that separates the selected tab and its panel, I think it is a bit clearer that it is the selected tab when they are connected
screenshot_2024-12-16T16:04:49+01:00

- the selected column's card is not very well separated from the background anymore (I think due to removing the shadow)

screenshot_2024-12-16T16:05:05+01:00

- the "Copied!" message from the copy to clipboard button used to be in reverse video, I think that helps notice it and separate it from the background

screenshot_2024-12-16T16:05:41+01:00

- the copy buttons on the sample from the dataframe used to be bigger and in the top-left corner, now they are centered

screenshot_2024-12-17T10:27:15+01:00

- also that sample text box used to have a max width and scroll horizontally which I think can be useful if the text is very long

- in the summary statistics column the buttons are now smaller and have more padding which makes the icons hard to read IMO

screenshot_2024-12-16T16:07:24+01:00

In the summary statistics there used to be a thick gray line separating the first column when it overlaps the rest, I think it helps realize that this column is sticky and some content is hidden behind it due to horizontal scrolling, that has been removed

screenshot_2024-12-16T16:07:48+01:00


in the toggle tip shown below, there used to be a transition delay so that if we hover the ( i ) button and then move the mouse to hover over the text it doesn't disappear; now as soon as the mouse leaves the button the text disappears. also there used to be a ring around it when we click it

screenshot_2024-12-16T16:08:17+01:00

not sure if it is intentional but the background color of the header in the associations panel is gone, I don't have a strong opinion as to whether it is better or worse

screenshot_2024-12-16T16:14:30+01:00


most of these changes don't seem to be related to dark mode but it is true we had discussed that this is an opportunity to refactor the css a bit. so I'm not sure which of these are intentional or not, maybe it would help to have another quick call also with @glemaitre  to discuss this? it might be  easier interactively while looking at an open report.

now regarding the colors and the dark mode itself, here are a couple of comments:

- the card in the first panel was green to show its association with the selected column, but that does not apply to the "distributions" panel so using a neutral gray color there was intentional
- I think the orange in dark mode looks good but the reason we had used a different color is that we show histograms of numeric columns in blue and bar plots for categorical columns in orange so it might be good to use a third color for the card header to not give the impression there is a connection with the plot?
- the axis lines and ticks in dark mode are not visible; I think we can find the previous pr one way to make sure they are

screenshot_2024-12-16T16:15:34+01:00

- I think the white text on orange background does not have a very high contrast and is a bit hard to read


thanks again for all this work and LMK if you would like to schedule a call or I can also come to montparnasse sometime!

rouk1:
Thanks for the review @jeromedockes, I'll try to work on your feedbacks on friday.

jeromedockes:
cool, thanks!!

rouk1:
Hello @jeromedockes  !

I think I've addressed your feedback. Please let me know if you want something else or if I miss something.

## tackled feedbacks

- standalone-report.css has been re-added (using new css vars nomenclature)
- selected tab border bottom color is now equal to background color
- fix column summaries shadow
- copy button tooltip is now inverted and bold
- copy text box as a max width of 100% hence it will never be larger than it's container (a horizontal scroller will appear at need)
- summary statistics is now bigger and icons fits it
- grey line that indicates that first two columns are sticky is visible to me
- toggled tip content is now part oh the hover, hence it does not disappear when hovering it
- tip button outline is now blue
- association table header now uses default tertiary background color
- card header are now in highlight color only whena ssociated with a table column
- primary highlight color is now raisin instead of orange, it provides high contrast and is not related to plot colors
- use 4 spaces in CSS to avoid large diff

> the axis lines and ticks in dark mode are not visible; I think we can find the previous pr one way to make sure they are

Not sure on how I may address this.

## UI preview

https://github.com/user-attachments/assets/dbf7a7de-abc9-474e-a2a4-8bb443abaa60

rouk1:
Made a few change...
Now javascript is used to set a `light` or `dark` class to the `#report` element.
The report will be dark if the containing body as `data-jp-theme-light="false"` or `data-vscode-theme-kind="vscode-dark"`, light if the shadowRoot body as `data-jp-theme-light="true"` or `data-vscode-theme-kind="vscode-light"` and if none of this attributes are set fallback to the user preferred color scheme.

skrub/_reporting/_data/templates/report.js
```diff
@@ -1,4 +1,5 @@
 if (customElements.get('skrub-table-report') === undefined) {
+    console.log("coucou");
```
jeromedockes, line 2: :wave:

rouk1: whoooooups

skrub/_reporting/_data/templates/dataframe-sample.css
```diff
@@ -11,41 +11,60 @@
 }
 
 .table-with-selectable-cells tbody {
-    border: var(--border-s) solid var(--mediumg);
+    border: var(--border-s) solid var(--color-border-secondary);
+}
+
+.table-with-selectable-cells .table-cell {
+    transition: background-color var(--animation-duration)
```
jeromedockes, line 18: I am a bit on the fence about this animation -- I have the impression it makes the interaction feel a little bit more sluggish; also we see some of the cells' color fading in when we load the page (or insert the report). could you say a bit more about the motivation? 

I know on the main branch it is not perfect either because the focus ring disappears when we press the mouse button down on a new cell but the colors only update when we release the button so there is a delay between the 2 : was adding the animation meant to make that less noticeable?

skrub/_reporting/_data/templates/report.js
```diff
@@ -93,7 +93,30 @@ if (customElements.get('skrub-table-report') === undefined) {
                 this.exchange.add(new HideOn(elem, this.exchange));
             });
 
-            adjustAllSvgViewBoxes(this.shadowRoot.getElementById('report'));
+            const report = this.shadowRoot.getElementById("report");
+            report.classList.add(this.detectTheme());
+            adjustAllSvgViewBoxes(report);
+        }
+
+        detectTheme() {
```
jeromedockes, line 101: as this does not use the skrubTableReport object and could potentially be useful to other components if we ever add any, I wonder if it would make sense for `detectTheme` to be a top-level function, outside of this class? there are a few at the bottom of this module. WDYT?

skrub/_reporting/_data/templates/report.js
```diff
@@ -386,7 +409,7 @@ if (customElements.get('skrub-table-report') === undefined) {
             while (!stop(i, j)) {
                 const cell = this.elem.querySelector(`[data-spans__${i}__${j}]`);
                 if (cell !== null && cell.id !== startCellId && !cell.hasAttribute(
-                        "data-excluded-by-column-filter") && cell.dataset.role !==
+                    "data-excluded-by-column-filter") && cell.dataset.role !==
```
jeromedockes, line 412: as we discussed briefly at some point it could be nice to configure some automatic formatting for the js and css files so that we don't get spurious diff lines (but that is for another pr)

jeromedockes (approved):
Thanks a lot @rouk1  !

I pushed a few small changes, mainly

- to set the matplotlib config before plotting so that all the lines, ticks etc. have the same color as the text
- to revert 1 or 2 details not related to the color theme eg the size of some buttons in pixels instead of relative to font size could cause them to be too big or small in some cases (image below)
- to also set the color of the full page when we generate a standalone report using the same media query.

I think the PR is ready to merge now. Let me know if you want to make some more changes otherwise I'll merge it (and we can always do further tweaks in other prs).

<details>
<summary>image</summary>

screenshot_2025-01-31T10:52:35+01:00
screenshot_2025-01-31T10:52:29+01:00

</details>

skrub/_reporting/_data/templates/dataframe-sample.css
rouk1: To be honest I always add animation when changing color on hover, it's more of an habit than motivated by objective arguments. To me having a few animations here and there make the widget more pleasing to the eye, but that is a matter of taste. Not that as it uses css transition user who do not like animations can disable it browser wide.

Feel free to remove if it doesn't make sense.

skrub/_reporting/_data/templates/report.js
rouk1: Why not but for now you have only one : )

skrub/_reporting/_data/templates/report.js
rouk1: Having a prettier pre commit hook could be a quick win.

rouk1:
> I think the PR is ready to merge now. Let me know if you want to make some more changes otherwise I'll merge it (and we can always do further tweaks in other prs).

LGTM then and feel free to ping me for future enhancements :)

skrub/_reporting/_data/templates/dataframe-sample.css
jeromedockes: thanks for explaining :) let's keep it then and we'll see if we get any feedback

jeromedockes:
> LGTM then and feel free to ping me for future enhancements :)

cool, will do! and thanks so much for the tablereport dark theme :star_struck:

jeromedockes (approved):

Issue #1176, opened 2024-12-03
Add more association metrics to `column_associations`


jeromedockes:
### Problem Description

ATM `skrub.column_associations` reports some variation of Cramer's V statistic. It could also compute some more common metrics such as Pearson and Spearman correlation when both columns are numeric. Those could also be reported in the TableReport.

we would have to make sure computing them does not significantly slow down the function, or make them optional

Vincent-Maladiere:
Closed by #1203, I guess we can create a new issue if the need for more associations metrics arises.

PR #1203, opened 2024-12-15
Add Pearsons correlation to _column_associations

reshamas:
Towards #1176 

Add Pearsons correlation to column associations.


reshamas:
This is the output, which I excluded for now, due to errors:

```
       left_column_name  left_column_idx right_column_name  right_column_idx  cramer_v  pearson
    0               c_3                3             c_str                 5    0.8215      NaN
    1               c_1                1               c_4                 4    0.8215   0.1597
    2               c_0                0               c_1                 1    0.8215   0.1123
    3               c_2                2             c_str                 5    0.7551      NaN
    4               c_0                0             c_str                 5    0.7551      NaN
    5               c_0                0               c_3                 3    0.7551   0.3212
    6               c_1                1               c_3                 3    0.6837  -0.1887
    7               c_0                0               c_4                 4    0.6837  -0.3202
    8               c_4                4             c_str                 5    0.6837      NaN
    9               c_3                3               c_4                 4    0.6053  -0.0150
    10              c_2                2               c_3                 3    0.6053   0.1757
    11              c_1                1             c_str                 5    0.6053      NaN
    12              c_0                0               c_2                 2    0.6053  -0.0578
    13              c_2                2               c_4                 4    0.5169  -0.2885
    14              c_1                1               c_2                 2    0.4122  -0.4986
```

Vincent-Maladiere (commented):
Hey @reshamas, thank you for starting this!

In skrub, we use methods defined in `skrub._dataframe._common.py` to dispatch operations on polars or pandas dataframes. We do this to avoid copying dataframes from one backend to the other, and later to enable features like lazy evaluation. Therefore, the input of the function `skrub._column_associations.column_associations()` could be a pandas or a polars DataFrame.

So, here, we would like to use something like `sbd.corr()` ("sbd" stands for "skrub dataframe"), instead of pandas corr. Since this method doesn't exist yet, you'll have to add it to `skrub._dataframe._common.py` 😁

Alternatively, you could turn your dataframe into a numpy array with `skb.to_numpy()` and then use `np.corrcoef`. I slightly prefer the former suggestion because it adds `corr()` in skrub, which I think would be also useful later.

If you go with the first option, I suggest converting the output to a numpy 2d array, so that you can directly use the `_stack_symmetric_associations()` function to melt the symmetric matrix into a long format. You need to select only numeric columns when passing your dataframe to `_stack_symmetric_associations()`.

Finally, you can apply the merge operation with `skrub._join_utils.left_join`.

We know that developing skrub is fairly involved at the moment, and we plan on making a simple documentation for all our internals. In the meantime, I'm here to help you to move this PR forward!

Vincent-Maladiere:
Hey @reshamas, do you plan to continue working on this?

jeromedockes:
I don't think so :thinking:

Vincent-Maladiere:
Ok, it's gone!

skrub/_column_associations.py
```diff
@@ -70,32 +83,35 @@ def column_associations(df):
     2  9.0810  9.4011  1.9257  5.7429  6.2358  val 2
     3  2.5425  2.9678  9.7801  9.9879  6.0709  val 3
     4  5.8878  9.3223  5.3840  7.2006  2.1494  val 4
+    >>> # Compute the associations
     >>> associations = skrub.column_associations(df)
     >>> associations # doctest: +SKIP
-       left_column_name  left_column_idx right_column_name  right_column_idx  cramer_v
```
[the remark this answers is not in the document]
Vincent-Maladiere: Ah yes, I missed that! Will do

jeromedockes:
works great :tada: 

screenshot_2025-03-31T18:22:29+02:00

skrub/_reporting/_data/templates/column-associations.html
```diff
@@ -22,6 +23,7 @@
                     >
                     {{ association["cramer_v"] | format_number }}
                 </td>
+                <td> {{ association["pearson_corr"] | format_number }} </td>
```
jeromedockes, line 26:
```suggestion
                <td>
                    {%- if not (association['pearson_corr']) | is_null -%}
                    {{ association["pearson_corr"] | format_number }}
                    {%- endif -%}
                </td>
```

maybe we could do this, so that we don't print the "nan"s


screenshot_2025-03-31T18:28:57+02:00


screenshot_2025-03-31T18:28:50+02:00

Vincent-Maladiere:
very interesting to see in your iris dataset example that these correlations are not concordant (third line: 0.395 vs 0.818, compared to the two lines above where the Cramer's is much higher)

skrub/_reporting/_data/templates/column-associations.html
```diff
@@ -8,6 +8,7 @@
                 <th scope="col">Column 1</th>
                 <th scope="col">Column 2</th>
                 <th scope="col"><a href="https://en.wikipedia.org/wiki/Cram%C3%A9r%27s_V">Cramér's V</a></th>
+                <th scope="col">Pearson's Correlation</a></th>
```
jeromedockes, line 11:
```suggestion
                <th scope="col"><a href="https://en.wikipedia.org/wiki/Pearson_correlation_coefficient">Pearson's Correlation</a></th>
```

to be consistent with the other column we can add the link even though most people know what pearson correlation is. if you prefer not to add the link you need to remove the closing `</a>` tag, as there is no corresponding opening `<a>`

skrub/_column_associations.py
```diff
@@ -33,6 +34,10 @@ def column_associations(df):
     associations between the values of 2 columns or between their missingness
     patterns may be captured.
 
+    To compute the Pearson's Correlation Coefficient, only numeric columns are
+    considered. The correlation is computed using the Pearson method used in
+    pandas.
```
jeromedockes, line 39:
```suggestion
    pandas or polars, depending on the input dataframe.
```

skrub/_column_associations.py
```diff
@@ -50,6 +55,14 @@ def column_associations(df):
 
     * `Cramer's V <https://en.wikipedia.org/wiki/Cramér%27s_V>`_
 
+    Pearson's Correlation Coefficient is a measure of the linear correlation
+    between two variables, giving a value between -1 and +1 (inclusive).
+
+    * `Pearson's Correlation Coefficient
+    <https://en.wikipedia.org/wiki/Pearson_correlation_coefficient>`_
+    * `pandas.DataFrame.corr
```
jeromedockes, line 63: if we have the link to pandas doc would it make sense to ahve the polars one?

jeromedockes:
> very interesting to see in your iris dataset example that these correlations are not concordant (third line: 0.395 vs 0.818, compared to the two lines above where the Cramer's is much higher)

yeah that's true... the cramer thing is quite rough and does aggressive binning and subsampling... it will be interesting to investigate a bit more

jeromedockes:
> > very interesting to see in your iris dataset example that these correlations are not concordant (third line: 0.395 vs 0.818, compared to the two lines above where the Cramer's is much higher)
> 
> yeah that's true... the cramer thing is quite rough and does aggressive binning and subsampling... it will be interesting to investigate a bit more

it should be a bit more similar to spearman's I guess

jeromedockes:
I guess polars and pandas have a different way of handling nans, pandas drops them whereas in polars they result in NaN correlations. WDYT would be the right behavior here? I think the pandas one because we already have other ways of indicating when there are nans so even if it should be interpreted cautiously the non-nan correlation is more informative. it will require a bit more work so could also be done in a separate pr

jeromedockes:
and finally, have you checked the computation time for largish dataframes, does it change a lot the total time of the report?

skrub/_column_associations.py
```diff
@@ -70,32 +83,35 @@ def column_associations(df):
...
+        df,
+        metric_name="pearson_corr",
+    )
+    on = [
+        "left_column_name",
+        "right_column_name",
+        "left_column_idx",
+        "right_column_idx",
+    ]
+    return _join_utils.left_join(
```
jeromedockes, line 109: that's a great way to handle it :)

skrub/_column_associations.py
```diff
@@ -247,3 +263,30 @@ def _compute_cramer(table, n_samples):
...
+
+
+def _compute_pearson(df):
+    """Compute the Pearson correlation coefficient for all pairs of columns.
+
+    This returns the symmetric matrix with shape (n cols, n cols) where entry
+    i, j contains the statistic for column i x column j.
+
+    """
+    # Replace categorical columns with np.nan to get a correlation matrix of shape
```
jeromedockes, line 275: I think we could simplify this a bit

- select numeric columns only
- use sbd.pearson_corr on those
- use pandas stack().reset_index() to convert to tall format
- or in polars case insert a column like `with_column(left_col=<the column names>)` then use `df.unpivot(index=left_col)` (only we would use a random name instead of `left_col` to have no name clash)
- then do the left join as you are doing in the caller


we would no longer use stack_symmetric_associations for this one I guess, but we would no longer need to do the filling with np.nan or the loops over columns. WDYT?

Vincent-Maladiere: Yes, I hesitated with this approach, but I saw avoiding manually dispatching between polars and pandas as an advantage, and reusing `stack_symmetric_associations` also makes things easier to understand IMO. Note that you wouldn't need to select numeric because `sbd.pearson_corr` does it for you. I agree that adding NaNs columns is a bit convoluted, though. WDYT?

Vincent-Maladiere:
> I guess polars and pandas have a different way of handling nans, pandas drops them whereas in polars they result in NaN correlations. WDYT would be the right behavior here? I think the pandas one because we already have other ways of indicating when there are nans so even if it should be interpreted cautiously the non-nan correlation is more informative. it will require a bit more work so could also be done in a separate pr

Right, I guess polars `.drop_nulls().corr()` do the trick.

jeromedockes:
> Right, I guess polars .drop_nulls().corr() do the trick.

I think it's a bit more complicated than that because you want to drop nulls for each pair of columns. when computing the correlation between columns A and B I don't want to drop rows that have a null in column C but have values for A and B:

```
>>> df = pd.DataFrame({'a': [None, 1, 2, 3], 'b': [0, None, 2, 3], 'c': [0, 1, None, 3]})
>>> df.corr()
     a    b    c
a  1.0  1.0  1.0
b  1.0  1.0  1.0
c  1.0  1.0  1.0
>>> df.dropna().corr()
    a   b   c
a NaN NaN NaN
b NaN NaN NaN
c NaN NaN NaN
```

so it might be easier to leave that for a different PR

Vincent-Maladiere:
Yes right, I realized that just now. Note a huge deal to compute corr pairwise, though

Vincent-Maladiere:
Ok, I did some benchmarking with both solutions (pairwise or converting to pandas), some conclusions:
- The pairwise solution doesn't scale well with the number of columns.
- The "converting to pandas" solution follows closely the performances of pandas `corr`, and doesn't lead to an bottleneck, computing time-wise.
- Pandas `corr` scales well with the number of rows. At 1e6 rows and 100 columns, it takes ~4s so we might want to subsample at some point
- However, Pandas `corr` gets very slow when the number of columns > 1000, and takes up to 120s for 10k columns. When the number of columns is this high, we might want to avoid computing Pearson's correlation on it.
WDYT?

**Pairwise**
```python
@pearson_corr.specialize("polars", argument_type="DataFrame")
def _pearson_corr_polars(df):
    col_names = df.columns
    n_cols = len(col_names)

    out = np.eye(n_cols)

    # Compute pairwise correlations for upper triangle
    for i, col1 in enumerate(col_names):
        for j in range(i + 1, n_cols):
            col2 = col_names[j]
            out[i, j] = out[j, i] = df.select([col1, col2]).drop_nulls().corr()[0, 1]

    return pl.DataFrame(
        out, schema={col: pl.Float64 for col in col_names}, orient="col"
    )
```

Screenshot 2025-04-01 at 11 51 14


**Converting to pandas**
```python
@pearson_corr.specialize("polars", argument_type="DataFrame")
def _pearson_corr_polars(df):
    return pl.from_pandas(_pearson_corr_pandas(df.to_pandas()))
```

Screenshot 2025-04-01 at 11 59 28

Row-wise performance:
Screenshot 2025-04-01 at 11 58 38

Vincent-Maladiere:
Out of curiosity, I will run the same benchmarking for Cramer's V

Vincent-Maladiere:
As I feared, dispatching was not straightforward, and I ran into tons of edge cases

Vincent-Maladiere:
No worries, I learned a lot while doing it

jeromedockes (approved):
great, thank you very much @reshamas  and @Vincent-Maladiere ! seeing correlations is one of the  first things we want so this will be appreciated

PR #1208, opened 2025-01-06
[DOC] Remove API change banner on landing page

TheooJ:
As discussed, removing the banner mentioning that the API is subject to change without notice

### DIFF
--- a/doc/announcement.html
+++ b/doc/announcement.html
@@ -1,3 +1,0 @@
-<div class="sidebar-message">
-    We are releasing <code>skrub</code> to gather early feedback, but <strong>the API is subject to change without notice</strong>.
-</div>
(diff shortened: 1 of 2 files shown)


Vincent-Maladiere (approved):
Thanks @TheooJ! I was wondering if the missing `announcement.html` in main would affect the stable version of the documentation, but it seems harmless

TheooJ:
I believe we're fine because CircleCI pushes to https://github.com/skrub-data/skrub-data.github.io, which is then used for the website, and the stable doc is only changed during releases. For more info we should ask @jeromedockes, the maestro behind all this

Vincent-Maladiere:
So I was wondering that because the `annoucements.html` content is fetched from the main GitHub branch and not from the `github.io` repo, for the stable and dev doc alike. This is a bit odd, but that's Sphinx.
I changed the URL of `announcements.html` in `conf.py` locally to see if something breaks, but the banner is just not displayed anymore without errors, so I assume that's fine.

Vincent-Maladiere:
See: https://github.com/skrub-data/skrub-data.github.io/blob/main/dev/index.html#L107

Vincent-Maladiere:
Let's merge it then!

PR #1214, opened 2025-01-13
small follow-up after #1190

jeromedockes:
In the code I suggested in #1190 for checking the file's encoding, an assert is used to check we have the correct encoding. but assert is not meant to be used this way and the check will not be executed if python is running with the -O flag. This pr slightly changes the check to remove the problematic use of assert


Vincent-Maladiere (approved):
Ah I didn't catch it either, LGTM!

PR #1215, opened 2025-01-14
ENH introduce `SKRUB_DATA_DIRECTORY` envar to control the data directory

thomass-dev:
Closes https://github.com/skrub-data/skrub/issues/1216.

Adding a new way to control the location of the data directory, using envar. 

It can be useful when you want to set the data cache directory from outside the code, especially in CI.
In addition, fix the expansion of `~` when passing programmatically an explicit path to `get_data_home`.


thomass-dev:
Can i instantiate a logger from stdlib? 
I've seen the use of `loguru` in the benchmark folder, so i reproduced, but i see that it's not a dependency.

glemaitre:
Whoops, it is actually late and I'm confusing the `sk***` pacakge. So we don't use logger which is more what I thought. I would expect that we use verbose + print.

@jeromedockes could confirm

glemaitre:
we will need a small test to check that we write in the good directory as well. I think that we only need to test the `get_data_home` for that.

skrub/datasets/_utils.py
```diff
@@ -1,5 +1,18 @@
+from os import environ
 from pathlib import Path
 
+from loguru import logger
+
+DATA_HOME_ENVAR_NAME = "SKRUB_DATA_DIRECTORY"
+DATA_HOME_ENVAR = environ.get(DATA_HOME_ENVAR_NAME)
```
jeromedockes, line 7: I would prefer to do this in the fetching function rather than upon import. for one thing it will make testing easier

skrub/datasets/_utils.py
```diff
@@ -1,5 +1,18 @@
...
+
+if DATA_HOME_ENVAR and (path := Path(DATA_HOME_ENVAR)).is_absolute():
```
jeromedockes, line 9: when the variable is set but is a relative path we silently ignore it?

skrub/datasets/_utils.py
```diff
@@ -1,5 +1,18 @@
...
+    DATA_HOME_DEFAULT = path
+else:
+    DATA_HOME_DEFAULT = Path("~").expanduser() / "skrub_data"
```
jeromedockes, line 12: nitpick: Path.home() is a bit simpler IMO

skrub/datasets/_utils.py
```diff
@@ -1,5 +1,18 @@
...
+
+logger.debug(f"Setting `DATA_HOME_DEFAULT` to '{DATA_HOME_DEFAULT}'.")
```
jeromedockes, line 14: we don't really use logging in the rest of the library. at some point we should probably do it in a systematic way, but for now you can just use `print` or maybe write to `sys.stderr`

jeromedockes (commented):
thanks!

skrub/datasets/_utils.py
thomass-dev: Fixed. I wanted to limit the diff.

skrub/datasets/_utils.py
thomass-dev: Yes, i supposed that the docstr + the log displaying the content of `DATA_HOME_DEFAULT` were enough explicit.
Would you like me to change it?

skrub/datasets/_utils.py
thomass-dev: I knew I'd get a comment like that ^^. 

Indeed the test is a bit more complicated, because we need to reload the module to reset the value of the global variables. But my intention was to set these values only once during the import, to avoid side-effect if the user changes the envar at runtime.

thomass-dev: You decide.

skrub/datasets/_utils.py
jeromedockes: > I wanted to limit the diff.

I see; that's appreciated!

skrub/datasets/_utils.py
jeromedockes: I agree with you but that's kind of a separate issue

skrub/datasets/_utils.py
```diff
@@ -10,6 +23,9 @@ def get_data_home(data_home=None):
     By default the data directory is set to a folder named 'skrub_data' in the
     user home folder.
 
+    You can even customize the default data directory by setting in your environment
```
jeromedockes, line 26: users do not use that function directly only through the fetcher functions so they will not see this. maybe in addition we could mention the env variable in the "getting started" example: 

https://github.com/skrub-data/skrub/blob/main/examples/00_getting_started.py#L17

I guess we _could_ mention it as well in the fetchers' docstrings but not sure that's necessary, or it could be done in another PR as at the moment they don't even mention the location of the data directory

jeromedockes:
the tests failure are not related to your PR but to #1217  :/

skrub/datasets/_utils.py
thomass-dev: Fixed.

thomass-dev:
@jeromedockes i have remove the `print`, because it makes no sense to print the `data_home` each time you call the function.

thomass-dev:
The PR is up-to-date.

thomass-dev:
The PR is up-to-date and ready for review.

jeromedockes (approved):
thanks a lot @thomass-dev ! sorry you needed to update it after fixing a bunch of conflicts and thanks for addressing all comments

jeromedockes:
oops sorry I forgot one last detail: if you could add a brief entry with your name and github handle in the changelog that would be great:

https://github.com/skrub-data/skrub/blob/1831731caae31cf4e0b4a183004f2f2093c76b9b/CHANGES.rst?plain=1#L12

thanks!

thomass-dev:
Thanks @jeromedockes , it should be okay?

jeromedockes:
yes, perfect. thanks!

Issue #1217, opened 2025-01-15
Migrate datasets from OpenML to Figshare


Vincent-Maladiere:
### Problem Description

After IRL discussions with @jeromedockes, we observed that we're still experiencing a lot of CI disruptions and errors when fetching datasets from OpenML. Figshare seems to be a more reliable alternative. 

### Feature Description

Move all skrub datasets to Figshare.

jeromedockes:
also, we should mock downloads in unit tests (less urgent)

joaquinvanschoren:
OpenML is currently unreachable because of a [cyberattack that hit TU Eindhoven](https://www.tue.nl/en/news-and-events/cyberattack). The service has been very reliable but this event is sadly out of our control. OpenML itself is not affected and we're in contact with the university IT team to bring it back soon. We have redundancy but sadly all within the tue network.

In the meantime, we are preparing to set up a secondary server (on an entirely different network) in the Dutch supercomputing center so that such an outage won't happen again.

jeromedockes:
Hi @joaquinvanschoren , thanks very much for explaining this outage -- I hope this attack gets resolved quickly and without major consequences for the university! Indeed OpenML has been generally reliable and we're very grateful for it. 

However for the skrub datasets we don't really need the great features of OpenML because they are just a handful of fixed, pre-defined datasets so we basically just need a place to store a few small parquet files (and some of the datasets are already stored like this). So I still think it makes sense to have a copy of those datasets on figshare if their license allows it. Of course in any case skrub depends on scikit-learn so skrub users will always have easy access to any OpenML dataset through scikit-learn's `fetch_openml`. we could consider fetching from openml by default and falling back to figshare if it's not available :thinking:

glemaitre:
I would be more in favor to have the redundancy instead of migrating. We also experience issue in the past with scikit-learn: https://github.com/scikit-learn/scikit-learn/issues/28297

The resolution of the issue took several months and I would say that the resolution of the ticket was not easy. At least, it is a plus on my side when trying to resolve an issue with OpenML because @joaquinvanschoren and the team do a great job.

ogrisel:
Note that we have been experiencing the HTTP 403 errors on figshare again on the scikit-learn CI this week:

https://github.com/scikit-learn/scikit-learn/pull/30636#issuecomment-2604394010

thomass-dev:
And why don't you put these static files in a dedicated python package?
[...]

PR #1218, opened 2025-01-19
Add redundancy to OpenML datasets with Figshare

Vincent-Maladiere:
Addresses https://github.com/skrub-data/skrub/issues/1217


Vincent-Maladiere:
Datasets to upload as CSV, in a zip when using multi-tables :

- Midwest_survey
   - [ ] Figshare
   - [ ] OSF
- drug-directory
   - [ ] Figshare
   - [ ] OSF
- employee_salaries
   - [ ] Figshare
   - [ ] OSF
- drug-directory
   - [ ] Figshare
   - [ ] OSF
- world bank + happiness records, [cf this skrub example](https://skrub-data.org/stable/auto_examples/04_fuzzy_joining.html)
   - [ ] Figshare
   - [ ] OSF
 - toxicity tweets
   - [ ] Figshare
   - [ ] OSF
 - credit fraud
   - [ ] Figshare
   - [ ] OSF
- movie-lens 
   - [ ] Figshare
   - [ ] OSF 
 - bike-sharing? (cf [this skrub example](https://skrub-data.org/stable/auto_examples/03_datetime_encoder.html#sphx-glr-auto-examples-03-datetime-encoder-py))
   - [ ] Figshare
   - [ ] OSF 
 - vgsales? (cf this [skrub example](https://skrub-data.org/stable/auto_examples/06_ken_embeddings.html#the-data)
   - [ ] Figshare
   - [ ] OSF 
 - flights/airports + weather data
   - [ ] Figshare
   - [ ] OSF

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
     if name is not None:
         data_dir = data_dir / name
     return data_dir
+
+
+def load_simple_dataset(dataset_name, data_home=None):
+    """TODO
+    """
+    bunch = _load_dataset_files(dataset_name, data_home)
```
jeromedockes, line 89: total nitpick: maybe we can use variable names a bit more informative than "bunch" (if it was a plain dict we prob. wouldn't call it "dict")

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+    bunch["X"] = bunch.pop(dataset_name)
```
jeromedockes, line 90: that might fail if `dataset_name` is not in `bunch`

jeromedockes: also maybe we should add X but keep the full dataframe ie the `dataset_name` key; wdyt?

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+            fraud.tar.gz
+            fraud/
+                baskets.csv
+                products.csv
+                metadata.json
+    """
+    data_home = get_data_home(data_home)
+    dataset_dir = data_home / dataset_name
+    datafiles_dir = dataset_dir / dataset_name
+    datafiles_dir.mkdir(parents=True, exist_ok=True)
```
jeromedockes, line 111: I think we could create the `datafiles_dir` only once we have successfully extracted the archive's content, that way we can simply check its existence (see below)

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+
+    archive_path = dataset_dir / f"{dataset_name}.zip"
+    metadata = DATASET_INFO[dataset_name]
+
+    diff_checksum = False
```
jeromedockes, line 116: maybe this section could be simplified a bit by doing:
- if the archive exists but has the wrong checksum, remove it and the files_dir (if it exists)
- if the files_dir does not exist, extract_archive

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+    return bunch
+
+
+def _extract_archive(dataset_dir, archive_path, diff_checksum):
+    dataset_name = dataset_dir.name
+
+    if not archive_path.exists() or diff_checksum:
+        _download_archive(dataset_name, archive_path)
+
+    shutil.unpack_archive(archive_path, dataset_dir, format="zip")
```
jeromedockes, line 146: maybe we could have the same strategy here as for the download: extract to a temp dir and rename it once that finishes successfully. this way we now that if the files dir exists it (very probably) has the correct contents

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+    metadata = DATASET_INFO[dataset_name]
+
+    for idx in range(1, retry + 1):
+        for target_url in metadata["urls"]:
+            print(
+                f"Downloading {dataset_name!r} from {target_url} (retry {idx}/{retry})"
+            )
+            try:
+                error_flag = False
+                _stream_download(archive_path, target_url, timeout, chunk_size)
```
jeromedockes, line 166: maybe checking that the download was successful ie that the checksum is correct should be the responsibility of `stream_download`. because the temp file should only be renamed if it has the correct checksum, and should be removed otherwise. also it will simplify a bit this function as we would not need the "error flag" any more and just return if `_stream_download` succeeds

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+            except Exception as e:
+                error_flag = True
+                warnings.warn(repr(e), category=FutureWarning)
```
jeromedockes, line 169: note: once our oldest python version is 3.11 we should remove these warnigns and raise an [excpetion group](https://docs.python.org/3.11/library/exceptions.html#exception-groups) with all the failures; maybe worth a todo note

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+
+            if not error_flag:
+                if _sha256(archive_path) != metadata["sha256"]:
+                    raise OSError(
```
jeromedockes, line 173: ideally, when the dataset gets updated, we will upload the new file to a new url this way the old skrub version can still keep working with the old files. so the most likely reason for a wrong checksum would be that somehow the download was incomplete but did not raise an exception, and in this case the exception should come from the downloader and we can still retry (or maybe move on to the next available url)

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+                        f"File {archive_path.stem!r} checksum verification has failed, "
+                        "which means the remote file has been updated.\n"
+                        "Please update your skrub version."
+                    )
+                return
+        time.sleep(delay)
+        delay *= 5
```
jeromedockes, line 180: I think 5 is a lot if we start from 1, if the download is not working we probably don't want to hang the program for too long maybe we can do 1, 3, 10 or something like that

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+
+    shutil.unpack_archive(archive_path, dataset_dir, format="zip")
+
+
+def _download_archive(
+    dataset_name,
+    archive_path,
+    retry=3,
+    delay=1,
+    timeout=30,
```
jeromedockes, line 154: time out is the allowed time before receiving the first response from the server (not the whole download) so I think 30 is too long; 3 would already be plenty

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+    # fails, the temporary file is removed manually in the except block.
+    temp_file = NamedTemporaryFile(
+        mode="wb",
+        prefix=archive_path.stem + ".part_",
+        dir=archive_path.parent,
+        delete=False,
+    )
+
+    try:
+        temp_file_path = Path(temp_file.name)
```
jeromedockes, line 209: doesn't seem to be used?

skrub/datasets/_utils.py
```diff
@@ -54,3 +81,155 @@ def get_data_dir(name=None, data_home=None):
...
+        response = requests.get(target_url, timeout=timeout, stream=True)
+        for chunk in response.iter_content(chunk_size):
+            temp_file.write(chunk)
+
+    except (Exception, KeyboardInterrupt):
+        Path(temp_file.name).unlink()
```
jeromedockes, line 215: here I don't think we have closed the file so on windows trying to remove it will raise an exception. maybe the easiest thing is to close the file right after creating it and then open it again in a `with` statement

jeromedockes (commented):
nice! a first few comments :)

Vincent-Maladiere:
The CI now fails because I tried to remove pyarrow from all tests env, will iterate on that

jeromedockes:
I think pyarrow will still be needed to run some tests & the examples due to the ken embeddings so let's deal with that in another pr

Vincent-Maladiere:
I think something is off with circle CI

skrub/datasets/_fetching.py
```diff
@@ -1112,49 +176,26 @@ def fetch_credit_fraud(load_dataframe=True, data_directory=None):
...
     -------
     bunch : sklearn.utils.Bunch
-        A dictionnary-like object, whose fields are:
-        - product : pd.DataFrame
-        - baskets : pd.DataFrame
-        - source_product : str
-        - source_baskets : str
-        - path_product : str
-        - path_baskets : str
+        A dictionary-like object with the following keys:
```
jeromedockes, line 187:
```suggestion
        A dictionary-like object with the following keys:

```

jeromedockes:
there seems to be a problem with the videogame sales dataset (example 6 -- see the negative r2 scores), I'll look into it

skrub/datasets/_utils.py
```diff
@@ -54,3 +163,187 @@ def get_data_dir(name=None, data_home=None):
...
+        The name of the dataset to load. The name must be a key of `DATASET_INFO`.
+
+    data_home : path, default=None
+        The directory where to download and unpack a zip file. If None, 'skrub_data'
+        is used.
+
+    Returns
+    -------
+    bunch : sklearn.utils.Bunch
+        A dictionary-like object with the following keys:
```
jeromedockes, line 194:
```suggestion
        A dictionary-like object with the following keys:
        
```

jeromedockes:
I uploaded the zip files to osf and updated the urls in this PR so it should be mostly ok now

jeromedockes:
@glemaitre if you have time you may want to have a look at this pr, too

Vincent-Maladiere:
+1 for the changes. I think this PR is ready for a final review before merging (@glemaitre if you have some spare time) :)

jeromedockes (approved):

PR #1222, opened 2025-01-27
prepare 0.5.0

Vincent-Maladiere:


CHANGES.rst
```diff
@@ -6,8 +6,8 @@ Release history
 
 .. currentmodule:: skrub
 
-Ongoing development
-===================
+Release 0.5.0
+=============
```
jeromedockes, line 10: can you remove the paragraph below?

(actually I think we can also remove it from main when doing the post-release PR)

LGTM otherwise

jeromedockes: also you can remove the empty sections bug fixes and maintenance

PR #1228, opened 2025-02-03
try to guess the appropriate color theme based on parent element's color

jeromedockes:
I wonder if looking at the text color on the report's parent element and trying to choose the most appropriate theme based on the result would be a good heuristic (when the other heuristics we have for vscode & jupyter notebook don't give a result), or if it is better to fall back on the system preference as we currently do...

this detects for example when the user set the theme in the skrub online documentation manually by clicking the button, to something different than their system or browser default. But I don't know if this is something that happens in practice; I guess dark theme users will have it set at the OS level?

@rouk1  WDYT?

### DIFF
--- a/skrub/_reporting/_data/templates/report.js
+++ b/skrub/_reporting/_data/templates/report.js
@@ -774,9 +774,11 @@ if (customElements.get('skrub-table-report') === undefined) {
     }
     SkrubTableReport.register(CopyButton);
 
-    function detectTheme(parentId) {
+    function detectTheme(refElementId) {
+
         const shadowRootBody = document.querySelector('body');
 
+        // Check VSCode theme
         const themeKindAttr = shadowRootBody.getAttribute('data-vscode-theme-kind');
         const themeNameAttr = shadowRootBody.getAttribute('data-vscode-theme-name');
 
@@ -784,7 +786,6 @@ if (customElements.get('skrub-table-report') === undefined) {
             const themeKind = themeKindAttr.toLowerCase();
             const themeName = themeNameAttr.toLowerCase();
 
-            // Check VSCode theme
             if (themeKind.includes("dark") || themeName.includes("dark")) {
                 return "dark";
             }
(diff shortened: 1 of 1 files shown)


rouk1:
Indeed dark theme is setted at OS level.

To handle the switch in sphinx we should detect the value of the `data-theme` attribute of the html tag and swap our class accordingly.
If you want to react dynamically to this change we'll have to add a [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver) and edit our theme class name on change.

skrub/_reporting/_data/templates/report.js
```diff
@@ -791,6 +791,16 @@ if (customElements.get('skrub-table-report') === undefined) {
...
+        // Guess based on parent element's color
+        const parentElem = document.getElementById(parentId);
+        const color = window.getComputedStyle(parentElem, null).getPropertyValue('color');
+        const match = color.match(/^rgb\s*\(\s*(\d+)\s*,\s*(\d+)\s*,\s*(\d+)\s*\)\s*$/i);
+        if(match){
+            const [r, g, b] = [match[1], match[2], match[3]];
+            const value = Math.max(r, g, b);
+            return value > 127 ? 'dark' : 'light';
+        }
+
```
rouk1, line 803: I would have converted the color to hsl and use only the luminance to decide is this is bright or dark.
But I'm not sure that this will work all the time.

jeromedockes:
thanks for checking @GaelVaroquaux . with the high contrast theme I can reproduce with main but not with this PR

this PR:

screenshot_2025-02-03T11:54:18+01:00

main branch:


screenshot_2025-02-03T11:53:50+01:00

rouk1:
Maybe we can enhance the follwing snippet of `report.js`.

```javascript
// Check VSCode theme
if (shadowRootBody.getAttribute('data-vscode-theme-kind') === 'vscode-dark') {
    return 'dark';
} else if (shadowRootBody.getAttribute('data-vscode-theme-kind') === 'vscode-light') {
    return 'light';
}
```

Indeed high contrast theme does set kind to `vscode-high-contrast`which does not help us ^^

```html
<body style="overflow: hidden;" role="document" class="vscode-high-contrast" data-vscode-theme-kind="vscode-high-contrast" data-vscode-theme-name="Dark High Contrast" data-vscode-theme-id="Default High Contrast">
``` 

The previous snippet could become (not tested)

```js
// Check VSCode theme
const themeKind = shadowRootBody.getAttribute('data-vscode-theme-kind').toLowerCase();
const themeName = shadowRootBody.getAttribute('data-vscode-theme-name').toLowerCase();
if (themeKind.includes("dark") || themeName.includes("dark")) {
    return "dark";
}
if (themeKind.includes("light") || themeName.includes("light")) {
    return "light";
}
```

cc @jeromedockes @GaelVaroquaux

Vincent-Maladiere:
I can confirm that I only have issues with the high-contrast theme, using the main branch!

Vincent-Maladiere (approved):
Looks good!

skrub/_reporting/_data/templates/report.js
```diff
@@ -800,6 +801,27 @@ if (customElements.get('skrub-table-report') === undefined) {
...
+                // If the text is very bright we have a dark theme
+                return 'dark';
+            }
+            if (luma < 75 ) {
+                // If the text is very dark we have a light theme
+                return 'light';
+            }
+            // Otherwise fall back to the next heuristic.
+        }
+
```
rouk1, line 824: 👍🏻

PR #1231, opened 2025-02-04
enh: vscode theme detection

rouk1:
This PR tries to enhance how the report detects vscode theme.
It should address the issue with high contrast theme discovered [here](https://github.com/skrub-data/skrub/pull/1228#issuecomment-2630589233).

Here is what I get now in high contrast:

Screenshot 2025-02-04 at 11 20 44

### DIFF
--- a/skrub/_reporting/_data/templates/report.js
+++ b/skrub/_reporting/_data/templates/report.js
@@ -777,11 +777,15 @@ if (customElements.get('skrub-table-report') === undefined) {
     function detectTheme() {
         const shadowRootBody = document.querySelector('body');
 
+        const themeKind = shadowRootBody.getAttribute('data-vscode-theme-kind').toLowerCase();
+        const themeName = shadowRootBody.getAttribute('data-vscode-theme-name').toLowerCase();
+
         // Check VSCode theme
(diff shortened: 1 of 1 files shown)


skrub/_reporting/_data/templates/report.js
```diff
@@ -777,11 +777,15 @@ if (customElements.get('skrub-table-report') === undefined) {
     function detectTheme() {
         const shadowRootBody = document.querySelector('body');
 
+        const themeKind = shadowRootBody.getAttribute('data-vscode-theme-kind').toLowerCase();
```
jeromedockes, line 780: I think we need to handle the case where the `getAttribute` returns `null` (when we are not in VSCode)? otherwise `toLowerCase` might fail

jeromedockes (commented):
Thanks for adding this @rouk1 !

skrub/_reporting/_data/templates/report.js
rouk1: You're right, I added a guard.

jeromedockes (approved):

jeromedockes:
Thank you @rouk1

PR #1233, opened 2025-02-05
Skrub expressions: flexible pipelines with multiple tables, previews and easy hyperparameter tuning

jeromedockes:
This pull request adds the skrub expressions, which provide a new way to build complex machine-learning estimators.

- they can handle multiple inputs of any type and perform arbitrary transformations (not limited to a sequence of scikit-learn transformers)
- The design matrix and targets (which get split for cross-validation) don't have to be available from the start, they can be intermediate results inside the pipeline.
- it is easy to specify ranges for any value (eg estimator hyperparameters) used in the transformation, and tune them with scikit-learn's grid-search or randomized-search
- estimators can be built interactively, while inspecting the results of intermediate steps

The expressions represent a computation and can be evaluated to produce a result (the estimator's output). The simplest expressions are variables, which represent an input to the ML estimator. They are created with `var`

```
>>> import skrub

>>> a = skrub.var("a")
>>> b = skrub.var("b")
>>> b
<Var 'b'>
```

Variables can be combined with operators, functions, constants etc. to build up more complex expressions.

```
>>> c = a + b
>>> c
<BinOp: add>
```

Variables can be given a value, used to display preview results and as a source of data for fitting & validating estimators:

```
>>> a = skrub.var("a", 7)
>>> b = skrub.var("b", 1)
>>> c = a + b
>>> c
<BinOp: add>
Result:
―――――――
8
```

We can use the attributes, methods, operators of the underlying objects:

```
>>> (skrub.var('a', '   hello...')[:-3] + ', world!').strip().capitalize()
<CallMethod 'capitalize'>
Result:
―――――――
'Hello, world!'
```
Skrub functionality is accessed through the special attribute `.skb`

```
>>> c.skb.eval({"a": 10, "b": 5})
15
>>> c.skb.draw_graph()
```

graph

We can obtain a fittable estimator out of the expression.

```
>>> estimator = c.skb.get_estimator()
>>> estimator.fit_transform({"a": 2, "b": 1})
3
```

In the example above the transformation is stateless but expressions can contain scikit-learn estimators:

```
>>> import skrub.datasets

>>> dataset = skrub.datasets.fetch_employee_salaries()

>>> features = skrub.X(dataset.X).skb.apply(
...     skrub.TableVectorizer(high_cardinality=skrub.StringEncoder())
... )
>>> features
<Apply TableVectorizer>
Result:
―――――――
      gender_F  gender_M  ...  date_first_hired_total_seconds  year_first_hired
0          1.0       0.0  ...                    5.277312e+08            1986.0
1          0.0       1.0  ...                    5.900256e+08            1988.0
2          1.0       0.0  ...                    6.274368e+08            1989.0
3          0.0       1.0  ...                    1.399248e+09            2014.0
4          0.0       1.0  ...                    1.173053e+09            2007.0
...        ...       ...  ...                             ...               ...
9223       1.0       0.0  ...                    1.446509e+09            2015.0
9224       1.0       0.0  ...                    5.966784e+08            1988.0
9225       0.0       1.0  ...                    9.885888e+08            2001.0
9226       0.0       1.0  ...                    1.157414e+09            2006.0
9227       0.0       1.0  ...                    1.327882e+09            2012.0

[9228 rows x 143 columns]
```

(A transformer can easily be applied to a subset of columns with the `cols` parameter)

Any value can be replaced by a choice, which is then used to build a hyperparameter grid:

```
>>> from sklearn.ensemble import HistGradientBoostingRegressor, RandomForestRegressor

>>> hgb = HistGradientBoostingRegressor(
...     learning_rate=skrub.choose_float(0.01, 0.8, log=True, name="lr")
... )
>>> rf = RandomForestRegressor(
...     n_estimators=skrub.choose_int(8, 128, log=True, name="N trees")
... )
>>> regressor = skrub.choose_from({"hgb": hgb, "rf": rf}, name="regressor")

>>> y = skrub.y(dataset.y)
>>> pred = features.skb.apply(regressor, y=y)
>>> pred
<Apply HistGradientBoostingRegressor>
Result:
―――――――
      current_annual_salary
0              67678.566435
1              96062.630076
2             107967.887714
3              54766.663259
4              88773.361530
...                     ...
9223           73232.819304
9224          154673.199482
9225          101625.833270
9226          143800.018771
9227           74987.836900

>>> print(pred.skb.describe_param_grid())
- regressor: 'hgb'
  lr: choose_float(0.01, 0.8, log=True, name='lr')
- regressor: 'rf'
  N trees: choose_int(8, 128, log=True, name='N trees')

>>> search = pred.skb.get_randomized_search(fitted=True, n_jobs=8)
>>> search.results_
   mean_test_score regressor        lr  N trees
0         0.914610       hgb  0.185372      NaN
1         0.911956       hgb  0.206369      NaN
2         0.909182       hgb  0.309576      NaN
3         0.902986        rf       NaN    113.0
4         0.901762        rf       NaN     92.0
5         0.901231       hgb  0.055610      NaN
6         0.900955        rf       NaN     88.0
7         0.900353        rf       NaN     40.0
8         0.891114        rf       NaN     14.0
9         0.768370       hgb  0.014172      NaN
```

(we can also get plotly parallel coordinate plots)

There's a lot more to say so please refer to the examples & reference documentation.

~very much WIP, I'm opening the PR so the first example renders and we can discuss it.~

There's still some work to do especially on the examples, but IRL we decided to merge it before the upcoming sprint and improve the examples in follow-up pull requests.


GaelVaroquaux:
Very much WIP, but already a cool example to look at https://output.circle-artifacts.com/output/job/34c076e7-2a4d-4616-946d-bfb5c27d2167/artifacts/0/doc/auto_examples/10_expressions.html#sphx-glr-auto-examples-10-expressions-py
:star_struck:

examples/10_expressions.py
```diff
@@ -0,0 +1,478 @@
...
+# use it to compute previews of the results. This helps inspect what the
+# results will look like, catch errors early, and provide better tab-completion
+# on attribute and item names.
+#
+# Instead of writing ``products = skrub.var("products")`` as before, let us
+# start over, this time passing a placeholder value to make our expression a
+# little bit more helpful:
+
+# %%
+products = skrub.var("products", placeholder=products_df)
```
rcap107, line 232: I don't think this part is explained well. This section is called "Eager previews", but until now it's not clear that expressions are evaluated lazily. I think it should be made more explicit that  `with_total.skb.eval({"products": products_df})` does not print anything because it hasn't been executed yet. 

You could say something like "if we try to print `with_total` at this stage, we do not get useful information. In order to obtain a preview of the current step, we need to add a `placeholder` value, which Skrub uses to compute a preview of the results". 

Two more things:
- A preview of the results means that it contains the results up until this point, or in this particular operation, or a subset of the overall results (a sampling for example)? 
- Is the `placeholder` variable used for something else? If it's only used for previews, maybe it would be better to be more explicit about it. Someone might get the idea that "you could put any placeholder df here to activate the preview". I'm not sure what the best variable name would be here, but I fear `placeholder` may lead to misunderstandings :thinking:

jeromedockes: Thanks for reading through this @rcap107 . I'll try to improve this part based on your comment. 

> with_total.skb.eval({"products": products_df}) does not print anything 

do you mean `with_total` does not print anything? because `eval` runs the computation, so the command above does produce a result

> A preview of the results means that it contains the results up until this point, or in this particular operation, or a subset of the overall results (a sampling for example)?

the result of evaluating the expression with the bindings provided by the placeholder values

```
>>> a = skrub.var('a', 1)
>>> e = a + a + a
>>> e.skb.preview
3
```

> but I fear placeholder may lead to misunderstandings

yes I agree it's not a good name. these values are used

- for the previews
- to provide data for cross-validation, fitting the estimator, or the hyperparam search.

In all cases they can be replaced by passing a dictionary so 

```
pred.skb.cross_validate() # uses the "placeholder" data
pred.skb.cross_validate({'customers': cust, 'orders': orders, 'labels': labels}) # uses the provided values instead
```

maybe we should call it something more generic like `value` or `data`?

jeromedockes: @rcap107 WDYT of `default_value` instead of `placeholder`? it conveys better that this value _can_ be replaced by something else, whereas a placeholder _must_ be replaced?

GaelVaroquaux:
I've been playing with this PR. Thanks a lot, it's fun!

Any way to remove the pydot dependency?

GaelVaroquaux:
Also (sorry, comment coming one by one): it would be great to have less vertical space here
image

examples/expressions/10_expressions_intro.py
```diff
@@ -0,0 +1,299 @@
...
+    learning_rate=skrub.choose_float(0.01, 0.9, log=True, name="learning_rate")
+)
+predictions = baskets.skb.apply(hgb, y=fraud_flags)
+predictions
+
+# %%
+# We can ask for a full report of the pipeline and inspect the results at every
+# step::
+#
+#     predictions.skb.full_report()
```
GaelVaroquaux, line 259: It would be very useful to have the "full_report" appear by default in a notebook/vscode. This is how people work

GaelVaroquaux (commented):
A few comments

skrub/_expressions/_evaluation.py
```diff
@@ -0,0 +1,711 @@
...
+    def __init__(self, target, generator):
+        self.target_id = id(target)
+        self.generator = generator
+
+
+class CircularReferenceError(ValueError):
+    pass
+
+
+class _ExprTraversal:
```
GaelVaroquaux, line 62: Overall, I think that this call needs a bit more comments. It is quite abstract code, and I fear that people reading it will get lost

skrub/_expressions/_estimator.py
```diff
@@ -0,0 +1,448 @@
...
+        return getattr(self.best_estimator_, name)(environment)
+
+    @property
+    def best_estimator_(self):
+        if not hasattr(self, "search_"):
+            attribute_error(self, "best_estimator_")
+        return ExprEstimator(self.search_.best_estimator_.expr)
+
+    def get_cv_results_table(self, return_metadata=False, detailed=False):
+        import pandas as pd
```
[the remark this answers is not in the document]
jeromedockes: honestly I'm not sure it would be a huge effort, the main difficulty would be replacing panda's logic for detecting datetime formats (used in todatetime and thus the tablevectorizer)

skrub/_expressions/_evaluation.py
```diff
@@ -0,0 +1,711 @@
...
+            return (
+                "Only one node can be marked with `mark_as_y()`. "
+                "2 different objects were marked as y:\n"
+                f"first object that used `.mark_as_y()`:\n{first!r}\n"
+                f"second object that used `.mark_as_y()`:\n{second!r}"
+            )
+        if conflict["reason"] == "name":
+            name = conflict["name"]
+            return (
+                f"Choice and node names must be unique. The name {name!r} was used "
```
GaelVaroquaux, line 673: Here we have a problem: as I work interactively in a vscode/jupyter, I modify a cell, and rerun it multiple times.

As a consequence, I end up defining nodes with the same name, get an error which prevents me from working:
```
ValueError: Choice and node names must be unique. The name 'encoder' was used for 2 different objects:
first object using the name 'encoder':
choose_from({'MinHash': MinHashEncoder(), 'LSA': StringEncoder()}, name='encoder')
second object using the name 'encoder':
choose_from({'MinHash': MinHashEncoder(), 'LSA': StringEncoder()}, name='encoder')
```

That's a big blocker

examples/expressions/10_expressions_intro.py
jeromedockes: yes I agree. dumping everything in a directory was faster for prototyping. we should move it to a single page and improve the interface, though that will require a bit of time spent on the front-end side. also, there is the challenge of the page becoming quite large if it contains many tablereports

GaelVaroquaux (commented):
Another couple of comments, one of them is important

skrub/_expressions/_evaluation.py
jeromedockes: but I guess that means you're running something like `table = table.skb.apply(encoder)` several times? that would always be a problem, even if it was `table = encoder.fit_transform(table)` for example? or I'm guessing the content of the cell wrong

GaelVaroquaux:
Still playing with this PR, here is a suggestion:

fetch_credit_fraud should come with an argument to return either a train split (of the multiple tables), a test split, or the whole data.
By default it would return train, which would amount to 2/3rd of the data.

From a didactic point of view, this would help explain the problem of running these type of prediction engines in real-life / operation: new data comes in the future and on multiple tables. We would start the example with only the train data, and bring in the test data after.

From a practical point of view, it would reduce by default the dataset size by 2/3rd, and thus the corresponding compute time and memory footprint, which will be appreciated :grin:

skrub/_expressions/_evaluation.py
GaelVaroquaux: That was me being stupid

jeromedockes:
> Any way to remove the pydot dependency?

I think we could probably find a javascript graph plotting library to vendor and do the plotting inside of the display

GaelVaroquaux:
Train-test split suggestion implemented as a PR to this PR: https://github.com/jeromedockes/skrub/pull/2

jeromedockes:
thanks @glemaitre ! @Vincent-Maladiere do you want to take a look? I think the jitter should only be added when values has a floating-point type

jeromedockes:
@glemaitre for context the issue the jitter tries to solve is that when many lines go through the same points we have a hard time realizing there are many in the plot

jeromedockes:
@glemaitre  please LMK if the latest commit fixes the pb

glemaitre:
Yes it does (it was what I did locally ;)

GaelVaroquaux:
This is so cool @jeromedockes . I'm always happy to look at this PR moving forward.

If you want, you can add a "section" to the front page, below the "Interactive Data Exploration" section. But if you prefer, we can do this in a follow up PR.

GaelVaroquaux:
Small comment: I think that I'd prefer having the API docs of the Expressions symbols not in a separate page, but in the main page of the API docs. Right now, it's too easy to miss it while scrolling through the API page

jeromedockes:
thanks @GaelVaroquaux !!

> If you want, you can add a "section" to the front page, below the "Interactive Data Exploration" section. But if you prefer, we can do this in a follow up PR.

good idea. I was thinking maybe we can manage to get this PR in before the sprint and in this case this would be a good thing to do in the sprint. but if it's not merged by then I'll add it in this PR. WDYT?

> Small comment: I think that I'd prefer having the API docs of the Expressions symbols not in a separate page, but in the main page of the API docs. Right now, it's too easy to miss it while scrolling through the API page

ok that's what I had at first but I thought the table of contents was getting really long ... I'll put it back but maybe the organization of this page could also be something to work on in the sprint. for example maybe all fetchers could be in a separate page

GaelVaroquaux:
> >  If you want, you can add a "section" to the front page, below the "Interactive Data Exploration" section. But if you prefer, we can do this in a follow up PR.

> I was thinking maybe we can manage to get this PR in before the sprint and in this case this would be a good thing to do in the sprint.

Sounds right. Let's do this!

jeromedockes:
circleCI is failing to checkout the repository, I don't think this is specific to skrub because the same problem was happening on [hidimstat](https://github.com/mind-inria/hidimstat) this morning

rcap107 (approved):

GaelVaroquaux:
Wohooo!

Issue #907, opened 2024-05-21
Add features to the `DatetimeEncoder`


koaning:
### Problem Description

Skrub currently encodes some features in the datetime encoder, but there are a few that feel missing. 

- Seasonal patterns during the year
- Seasonal patterns during the day
- Holidays

It feels like these might be great candidates to consider for our `DatetimeEncoder`. 

### Feature Description

The seasonal patterns can both be generated using the `SplineTransformer` in scikit-learn under the hood that uses the `periodic` setting. There's a demo of this technique [here](https://youtu.be/cEpiqu3QCW0). In one case we'd model the features over the ordinal day of the year, while in the other case we'd use time of day.

The holidays might be a bit trickier because we'd need to rely on a 3rd party library to capture all of them. Then again, polars does support [some business day features](https://docs.pola.rs/py-polars/html/reference/expressions/api/polars.Expr.dt.add_business_days.html#polars.Expr.dt.add_business_days), so we might just be able to leverage something there. 

### Alternative Solutions

I'm planning on making a first version of such a component so that it becomes easier to shoot at it. I can imagine that maybe we don't want to support all of these features but maybe just a subset. It can also be the case that business/holiday features should go into another estimator.

jeromedockes:
Thanks for opening this! Both hoidays and temporal patterns sound useful

**holidays:** As you say there is the problem that we would need another dependency to get the holidays. However maybe at first we could have the users provide their own holidays, similarly to what is done in the polars `add_business_day` for which you provided a link. We could also consider starting by adding an example showing how to use one of the joiner transformers to add that feature after constructing a table that lists all holidays with the python-holidays package. 

**seasonal patterns:** in your experience, are those features mostly useful for linear models, or do they also improve the performance of gradient boosting? And what dimensionality do you think is typically useful?


Another addition I would like to see for the DatetimeEncoder is the option to output some of its current features as Categorical dtypes, or as one-hot encoded categories. In particular we may want to encode the day of the week and probably the month as categories rather than as a floating-point numbers (as is done at the moment).

Also note that important changes are made to the DatetimeEncoder in #902 (among others, adding support for polars and making it accept a single column rather than a dataframe). The first version you have in mind would be in the form of changes to the DatetimeEncoder, or a stand-alone prototype?

jeromedockes:
[...]

PR #1235, opened 2025-02-07
Adding a periodic encoder to the DatetimeEncoder

rcap107:
Draft for #907 

Main points:
- Adding transformers for hours in a day and days in a year
- Adding the ordinal day (day in a year) as feature
- Implement transformers using `SplineTransformer` from scikit-learn

Questions:
- Should I also implement a CircularTransformer? 
- The main reason I am considering the CircularTransformer is because I am not sure what would be better for encoding days in a year. Should I just not bother, and keep SplineTransformer with a limited number of knots? 
- What should the default parameters be?
- If we decide to encode the year, there is the problem of having both leap and non-leap years in a dataset. Is that a problem? I don't think it would affect the features that much. 
- Any more features that should be added in this PR? 

Of course, tests and examples are all missing.


skrub/_datetime_encoder.py
```diff
@@ -294,6 +308,20 @@ def fit_transform(self, column, y=None):
...
+        # Adding transformers for periodic encoding
+        self._required_transformers = {}
+
+        for _case in self.add_periodic:
+            if _case == "day":
+                _t = PeriodicEncoder(kind="spline", period=24)
+                self._required_transformers[_case] = _t
+            if _case == "year":
+                # TODO: In theory we should check that the year is leap
+                _t = PeriodicEncoder(kind="circular", period=366)
```
jeromedockes, line 323: isn't it a bit surprising that the encoding is different for day and year? maybe it could be splines with 4 splines or something like that?

skrub/_datetime_encoder.py
```diff
@@ -294,6 +308,20 @@ def fit_transform(self, column, y=None):
...
+                self._required_transformers[_case] = _t
```
jeromedockes, line 324: I think we should also offer to encode the position within the month and within the week

skrub/_datetime_encoder.py
```diff
@@ -316,7 +344,19 @@ def transform(self, column):
...
             extracted = sbd.to_float32(extracted)
             all_extracted.append(extracted)
-        return sbd.make_dataframe_like(column, all_extracted)
+
+        _new_features = []
+        for _case, t in self._required_transformers.items():
+            if _case == "day":
+                _feat = _get_dt_feature(column, "day")
+                _feat = sbd.rename(_feat, "day")
+                _new_features.append(t.fit_transform(_feat))
```
jeromedockes, line 353: we're refitting the encoders every time we call transform. rather, we want to fit them during fit_transform and then call only `transform` not `fit_transform`

skrub/_datetime_encoder.py
```diff
@@ -316,7 +344,19 @@ def transform(self, column):
             extracted = _get_dt_feature(column, feature).rename(f"{name}_{feature}")
             extracted = sbd.to_float32(extracted)
             all_extracted.append(extracted)
-        return sbd.make_dataframe_like(column, all_extracted)
+
+        _new_features = []
+        for _case, t in self._required_transformers.items():
+            if _case == "day":
+                _feat = _get_dt_feature(column, "day")
+                _feat = sbd.rename(_feat, "day")
```
jeromedockes, line 352: why do we remove the column name?

skrub/_datetime_encoder.py
```diff
@@ -316,7 +344,19 @@ def transform(self, column):
             extracted = _get_dt_feature(column, feature).rename(f"{name}_{feature}")
             extracted = sbd.to_float32(extracted)
             all_extracted.append(extracted)
-        return sbd.make_dataframe_like(column, all_extracted)
+
+        _new_features = []
+        for _case, t in self._required_transformers.items():
+            if _case == "day":
```
jeromedockes, line 350: it seems the other `_case` is not handled?

skrub/_datetime_encoder.py
```diff
@@ -325,10 +365,80 @@ def _check_params(self):
...
         return {"preserves_dtype": []}
 
     def __sklearn_tags__(self):
         tags = super().__sklearn_tags__()
         tags.transformer_tags = TransformerTags(preserves_dtype=[])
         return tags
+
+
+class PeriodicEncoder(SingleColumnTransformer):
+    def __init__(self, kind: str = "spline", period=24, n_splines=None, degree=3):
```
jeromedockes, line 392: it seems a bit strange to have "n_splines" and "degree" parameters if the kind can be circular. maybe we should have a separate splineencoder and circular encoder classes. or just the spline one at first

skrub/_datetime_encoder.py
```diff
@@ -256,10 +261,19 @@ class DatetimeEncoder(SingleColumnTransformer):
     timezone used during ``fit`` and that we get the same result for "hour".
     """  # noqa: E501
 
-    def __init__(self, resolution="hour", add_weekday=False, add_total_seconds=True):
+    def __init__(
```
jeromedockes, line 264: should we not generate the values when we are doing the periodic encoding? eg if we have spline encoding of the time of day, maybe we shouldn't by default output the "hour" feature

skrub/_datetime_encoder.py
rcap107: Not yet, I didn't bother implementing it in case we decided to only go with SplineTransformer for the time being

skrub/_datetime_encoder.py
jeromedockes: day of the week and day of the month are ordinally encoded, but for some models that's not a good representation. the splines are a way to do a kind of soft one-hot encoding

skrub/_datetime_encoder.py
rcap107: Maybe users still need the additional columns as features for some reason? Though I guess that by the point they get to vectorizing the table they're already done with exploring the data which is what keeping those features would help with :thinking: 

I'll remove the old features

skrub/_datetime_encoder.py
jeromedockes: sure but we should also retain the name of the original input column. so if the column was `"when"` we would get `"when_day_spline_1"`

skrub/_datetime_encoder.py
jeromedockes: i'm not sure what would the best interface to make the skrub code simple and to make it easy for users to specify which features they want / figure out which features they will get :thinking:

skrub/_datetime_encoder.py
jeromedockes: yeah that's a good idea to share a first draft early

rcap107:
I finished the base implementation and split the `PeriodicEncoder` into `SplineEncoder` and `CircularEncoder`

I am working locally on the examples, and now I need to write all the tests

skrub/_datetime_encoder.py
```diff
@@ -105,11 +115,39 @@ class DatetimeEncoder(SingleColumnTransformer):
...
+        will be day 365 on non-leap years.
+
+    add_periodic : bool, default=False
+        Add periodic features with different granularities. By default, use
+        trigonometric (circular) encoding of features. Spline encoding and
+        custom encoders are also supported.
+
+    year_encoding : str, ``CircularEncoder``, ``SplineEncoder`` or None, default="circular"
+        Select the strategy used to encode days in a year. By default, use a
+        CircularEncoder with period=365. If None, no encoding is performed.
```
GaelVaroquaux, line 130: Maybe use the : class:  sphinx primitive here (and in other places)

skrub/_datetime_encoder.py
```diff
@@ -311,12 +456,40 @@ def transform(self, column):
...
+            # filling nulls only to the feature passed to the periodic encoder
+            _transformed = t.transform(self._fill_nulls(_feat))
+
+            _new_features.append(_transformed)
+
+        X_out = sbd.make_dataframe_like(column, all_extracted)
+        X_out = sbd.concat_horizontal(X_out, *_new_features)
+
+        # Censoring all the null features
+        X_out = sbd.where_row(X_out, self.not_nulls, _null_mask)
```
jeromedockes, line 483: I think the circular and spline encoders could do that themselves on the numpy output and we wouldn't need the where_row

jeromedockes (commented):
thanks a lot @rcap107 ! here is a first few comments

skrub/_datetime_encoder.py
```diff
@@ -26,6 +29,8 @@
     "nanosecond",
 ]
 
+_DEFAULT_ENCODING_PERIODS = {"year": 4, "month": 30, "weekday": 7, "hour": 24}
```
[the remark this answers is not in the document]
rcap107: At the beginning I had 365 and the spline as default, but having 365 features for each year was too much, and I didn't have any better idea than setting it to a random number (4 seasons I guess)

Now that circular is the default, we can set it to 366

skrub/_datetime_encoder.py
```diff
@@ -254,12 +292,71 @@ class DatetimeEncoder(SingleColumnTransformer):
...
+    >>> encoder[-1].extracted_features_
+    ['login_year', 'login_month', 'login_day', 'login_hour', 'login_total_seconds',
+    'login_weekday_spline_0', 'login_weekday_spline_1', 'login_weekday_spline_2',
+    'login_weekday_spline_3', 'login_weekday_spline_4', 'login_weekday_spline_5',
+    'login_weekday_spline_6']
+
+    Finally, it is possible to change the period of the encoders by creating a new
+    ``SplineEncoder`` or ``CircularEncoder`` with the desired parameters:
+    >>> from skrub import SplineEncoder
+    >>> se = SplineEncoder(period=3)
```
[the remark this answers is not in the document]
rcap107: so that the docstring includes only 3 features 

I could change it to `CircularEncoder(period=12)` if you think it confuses the reader

skrub/_datetime_encoder.py
rcap107: this
```
X_out[~self.not_nulls] = np.nan
```
works with pandas, but doesn't work with polars, is there another way?

skrub/_datetime_encoder.py
```diff
@@ -325,10 +498,198 @@ def _check_params(self):
...
+        is set to be equal to ``period``.
+
+    degree : int, default=3
+        Degree of the polynomial used as the spline basis. Must be a non-negative
+        integer.
+    """
+
+    def __init__(self, period=24, n_splines=None, degree=3):
+        self.period = period
+        self.n_splines = n_splines
```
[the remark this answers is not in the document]
rcap107: yes, correct

skrub/_datetime_encoder.py
```diff
@@ -325,10 +498,198 @@ def _check_params(self):
...
+    This encoder will generate two features corresponding to the sine and cosine
+    of the feature, based on the given period as output.
+
+    Parameters
+    ----------
+    period : int, default = 24
+        Period to be used as basis of the trigonometric function.
+    """
+
+    def __init__(self, period=24):
```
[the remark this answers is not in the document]
rcap107: makes sense, I'll remove it

GaelVaroquaux:
The SplineEncoder and the CircularEncoder are not documented in the API page, and thus they don't render correctly in the docs:
https://output.circle-artifacts.com/output/job/cc1b279e-769f-4515-bd08-30d0ebebd0f6/artifacts/0/doc/reference/generated/skrub.DatetimeEncoder.html
image

skrub/_datetime_encoder.py
```diff
@@ -325,10 +498,187 @@ def _check_params(self):
...
+    n_splines : int or None, default=None
+        Number of splines (features) to be generated. If set to None, ``n_splines``
+        is set to be equal to ``period``.
+
+    degree : int, default=3
+        Degree of the polynomial used as the spline basis. Must be a non-negative
+        integer.
+    """
+
+    def __init__(self, period, n_splines=None, degree=3):
```
GaelVaroquaux, line 536: In scikit-learn design, it is very frowned upon to have a parameter without a default value

skrub/_datetime_encoder.py
```diff
@@ -325,10 +498,187 @@ def _check_params(self):
...
     def _more_tags(self):
         return {"preserves_dtype": []}
 
     def __sklearn_tags__(self):
         tags = super().__sklearn_tags__()
         tags.transformer_tags = TransformerTags(preserves_dtype=[])
         return tags
+
+
+class SplineEncoder(SingleColumnTransformer):
```
GaelVaroquaux, line 515: From a design perspective, this class seems to me really like a super thin wrapper on the scikit-learn SplineTransformer.

Would it not be possible to get rid of it, and use the SplineTransformer? I would like to minimize almost-redundant functionnality

skrub/_datetime_encoder.py
```diff
@@ -325,10 +498,187 @@ def _check_params(self):
...
+    This encoder will generate two features corresponding to the sine and cosine
+    of the feature, based on the given period as output.
+
+    Parameters
+    ----------
+    period : int, default = 24
+        Period to be used as basis of the trigonometric function.
+    """
+
+    def __init__(self, period):
```
GaelVaroquaux, line 623: Same comment about scikit-learn really liking defaults. Put 365 for instance

skrub/_datetime_encoder.py
```diff
@@ -325,10 +498,187 @@ def _check_params(self):
...
+        self.period = period
+
+    def fit_transform(self, X, y):
```
GaelVaroquaux, line 626:
```suggestion
    def fit_transform(self, X, y=None):
```

skrub/_datetime_encoder.py
```diff
@@ -26,6 +29,8 @@
     "nanosecond",
 ]
 
+_DEFAULT_ENCODING_PERIODS = {"year": 366, "month": 30, "weekday": 7, "hour": 24}
```
GaelVaroquaux, line 32: Naive question: years and months don't actually have a periodicity of 366 and 30. How do we deal with this?
Is a solution to use the actually average length (365.25 days/year, I believe)?

GaelVaroquaux:
I don't understand why, when I look at the generated example gallery, it looks like the relevant examples did not run:
* https://output.circle-artifacts.com/output/job/cc1b279e-769f-4515-bd08-30d0ebebd0f6/artifacts/0/doc/auto_examples/03_datetime_encoder.html#sphx-glr-auto-examples-03-datetime-encoder-py
* https://output.circle-artifacts.com/output/job/cc1b279e-769f-4515-bd08-30d0ebebd0f6/artifacts/0/doc/auto_examples/01_encodings.html#sphx-glr-auto-examples-01-encodings-py

I would really like to see the examples running. I don't have a feeling for merging a PR without the examples running.

skrub/_datetime_encoder.py
```diff
@@ -105,11 +115,39 @@ class DatetimeEncoder(SingleColumnTransformer):
...
+        will be day 365 on non-leap years.
+
+    add_periodic : bool, default=False
+        Add periodic features with different granularities. By default, use
+        trigonometric (circular) encoding of features. Spline encoding and
+        custom encoders are also supported.
+
+    year_encoding : str, :class:``~CircularEncoder``, :class:``~SplineEncoder`` or None, default="circular"
+        Select the strategy used to encode days in a year. By default, use a
+        CircularEncoder with period=365. If None, no encoding is performed.
```
GaelVaroquaux, line 130: If I follow things right, this is used only if "add_periodic" is True above. We should mention this.

skrub/_datetime_encoder.py
```diff
@@ -105,11 +115,39 @@ class DatetimeEncoder(SingleColumnTransformer):
...
+
+    month_encoding : str, :class:``~CircularEncoder``, :class:``~SplineEncoder`` or None, default="circular"
+        Select the strategy used to encode days in a month. By default, use a
```
GaelVaroquaux, line 133: Same comment here: this is used only if "add_periodic" is True above

skrub/_datetime_encoder.py
```diff
@@ -105,11 +115,39 @@ class DatetimeEncoder(SingleColumnTransformer):
...
 
+    add_day_of_year : bool, default=False
+        Add the day of year (ordinal day) as an integer in the range 1 to 365 (or
+        366 in the case of leap years). January 1st will be day 1, December 31st
+        will be day 365 on non-leap years.
+
+    add_periodic : bool, default=False
+        Add periodic features with different granularities. By default, use
+        trigonometric (circular) encoding of features. Spline encoding and
+        custom encoders are also supported.
```
GaelVaroquaux, line 126: Could we have a simpler API that removes the "*_encoding" arguments below and changes this argument to accepting False, "circular", "spline"?

It would reduce the flexibility of the object, but my hunch is that it would cover most of the usecases. And it would remove the inter-dependency between arguments (aka "add_periodic=False" means that the arguments below are ignored), which make user mistake more likely and hyper-parameter selection easier)

rcap107:
> I don't understand why, when I look at the generated example gallery, it looks like the relevant examples did not run:
> 
> * https://output.circle-artifacts.com/output/job/cc1b279e-769f-4515-bd08-30d0ebebd0f6/artifacts/0/doc/auto_examples/03_datetime_encoder.html#sphx-glr-auto-examples-03-datetime-encoder-py
> * https://output.circle-artifacts.com/output/job/cc1b279e-769f-4515-bd08-30d0ebebd0f6/artifacts/0/doc/auto_examples/01_encodings.html#sphx-glr-auto-examples-01-encodings-py
> 
> I would really like to see the examples running. I don't have a feeling for merging a PR without the examples running.

I haven't gotten to the examples yet

skrub/_datetime_encoder.py
rcap107: I removed it after @jeromedockes suggested that having a default parameter that has nothing to do with the value to encode (e.g., 24 for month or year) would not be useful

I can add it back

skrub/_datetime_encoder.py
rcap107: Should I just fold the logic of both encoders back in the main DatetimeEncoder then? CircularEncoder is just wrapping a call to np.sin/np.cos

 I was thinking that it could be useful to have periodic encoders for non-datetime features, but maybe I'm wrong

skrub/_datetime_encoder.py
rcap107: The period must be an integer

I don't have a good answer, but my gut feeling is that choosing between period 365 and 366 (and even between 28 and 31 for months) is not  going to make a noticeable difference downstream

skrub/_datetime_encoder.py
rcap107: Users might want to gridsearch at least some parameters for different variables (like the number of splines for encoding a year)

Either way, I can remove everything and just hardcode the defaults

If we go with the simplified implementation, there is really no reason to have the separate encoders, is there

GaelVaroquaux:
> Should I just fold the logic of both encoders back in the main DatetimeEncoder then? CircularEncoder is just wrapping a call to np.sin/np.cos

My gut feeling would be to do this

skrub/_datetime_encoder.py
jeromedockes: yep sorry about that @rcap107 . Indeed I had said there was no sensible default here, but yeah you can add it back

rcap107:
I had a IRL discussion with @jeromedockes about the encoder and there are some issues that we might want to discuss in person. 

The main point of discussion is whether we want the user to have the flexibility to choose different parameters for the circular/spline encoders, or if it's better to keep it simple and have only `None`, `circular`, `spline`. 

My opinion is that the flexibility is an advantage, and if we decide to remove everything we will have to either hardcode variables that a user might want to tweak (specifically, the number of splines), or expose parameters to change them anyway. 

If we keep the code "as-is" (bar some refactoring), meaning with `CircularEncoder` and `SplineEncoder` and all the specific parameters:
- The implementation would let the user choose (and maybe search over) specific values for the period and especially the number of splines. Also, the code is already there. 
- There are multiple parameters (`year_encoding` etc.) that bloat the docstring and that are relevant only when `add_periodic=True`. I don't have a better idea on how to write this. 
- The encoders are "there", so if someone needs a periodic encoder for something other than a datetime, they're still available (no idea if this is a realistic use case).

If instead we simplify everything, the result is folding the logic of `CircularEncoder` and `SplineEncoder` back in the `DatetimeEncoder`, and removing the `*_encoding` parameters with just `add_periodic` that can be either a bool, `circular` or `spline`. 
- This would simplify the documentation and the user-experience (I am not fully sold on this, but I get the point).
- It also means that either 1) the flexibility goes out of the window and we hardcode every parameter, or 2) we try to retain some of that flexibility by exposing the number of splines as a parameter. 
- 1. is not a problem for periods, but a user might want to tweak the number of splines for a year (or avoid having 365 features for each year feature).
- 2. Would just functionally be the same as providing a custom `SplineEncoder`, but with exposing a lower level parameter that would also be irrelevant whenever `add_periodic=False`. 
- Sidenote, if we go with 1) I would rather avoid starting a fractal discussion of the number of splines to go with for each case. 
- We run the risk of having someone come in and open an issue asking to provide more customization for the Encoders, which may require moving the logic back out to separate objects. 

Ultimately, I will be implementing fixes for either version, what I _really_ want to avoid is picking one version, writing all the code, and then have to rewrite it again right before merging.

glemaitre:
Here, I have an example that would benefit greatly from this feature when I compare a linear model with a gradient boosting:

https://skore.probabl.ai/0.7/auto_examples/use_cases/plot_employee_salaries.html#sphx-glr-auto-examples-use-cases-plot-employee-salaries-py

It is really close from the bike rental example from scikit-learn:

https://scikit-learn.org/dev/auto_examples/applications/plot_cyclical_feature_engineering.html#sphx-glr-auto-examples-applications-plot-cyclical-feature-engineering-py

examples/03_datetime_encoder.py
```diff
@@ -128,20 +128,34 @@
 # |DatetimeEncoder| is used on the correct column(s).
 pprint(table_vec_weekday.transformers_)
 
+###############################################################################
```
GaelVaroquaux, line 131:
```suggestion
###############################################################################
# 
# Feature engineering for linear models
# ....................................................................
# 
```

examples/03_datetime_encoder.py
```diff
@@ -128,20 +128,34 @@
...
+# The |DatetimeEncoder| can generate additional periodic features using either
+# B-Splines (|SplineTransformer|) or trigonometric functions. To do so, set the
+# ``periodic encoding`` parameter ``circular`` or ``spline``. In this
+# example, we use ``spline``.
```
GaelVaroquaux, line 135:
```suggestion
# The |DatetimeEncoder| can generate additional periodic features. These are
# particularly useful for linear models. This is controlled by ``periodic encoding``
# parameter which can be  ``circular`` or ``spline``, for trigonometric functions
# or B-Splines . In this example, we use ``spline``.
```

examples/03_datetime_encoder.py
```diff
@@ -80,13 +87,13 @@
...
-# finer than hours. This is because we don't want to extract minutes, seconds
-# and lower units, as they are unimportant here.
+# don't want to extract features with a resolution finer than hours. This is
+# because we don't want to extract minutes, seconds and lower units, as they
+# are unimportant here.
 
 from skrub import DatetimeEncoder
 
-date_enc = DatetimeEncoder().fit_transform(date)
+date_enc = DatetimeEncoder(resolution="hour").fit_transform(date)
```
GaelVaroquaux, line 96: Why did we have to add this?

I always try to have something in an example only if it is needed for didactic reasons

examples/03_datetime_encoder.py
```diff
@@ -148,14 +160,21 @@
...
-pipeline = make_pipeline(table_vec, HistGradientBoostingRegressor())
-pipeline_weekday = make_pipeline(table_vec_weekday, HistGradientBoostingRegressor())
-pipeline_periodic = make_pipeline(table_vec_periodic, HistGradientBoostingRegressor())
+pipeline = make_pipeline(table_vec, StandardScaler(), SimpleImputer(), RidgeCV())
+pipeline_weekday = make_pipeline(
+    table_vec_weekday, StandardScaler(), SimpleImputer(), RidgeCV()
+)
+pipeline_periodic = make_pipeline(
+    table_vec_periodic, StandardScaler(), SimpleImputer(), RidgeCV()
+)
```
GaelVaroquaux, line 177: Please add a small text that says that we'll compare different encoding of datetime, for the linear model

examples/03_datetime_encoder.py
rcap107: To make it clear that's how we can specify the resolution, it's either that or adding a line to the previous paragraph saying that "hour" is the default resolution of the encoder

GaelVaroquaux:
> To make it clear that's how we can specify the resolution, it's either that or adding a line to the previous paragraph saying that "hour" is the default resolution of the encoder

I would rather have a one-liner comment just above

skrub/_datetime_encoder.py
```diff
@@ -105,11 +115,22 @@ class DatetimeEncoder(SingleColumnTransformer):
         Add the total number of seconds since the Unix epoch (00:00:00 UTC on 1
         January 1970).
 
+    add_day_of_year : bool, default=False
+        Add the day of year (ordinal day) as an integer in the range 1 to 365 (or
+        366 in the case of leap years). January 1st will be day 1, December 31st
+        will be day 365 on non-leap years.
+
+    periodic_encoding : str or None, default=None
```
jeromedockes, line 123:
```suggestion
    periodic_encoding : 'circular', 'spline', or None, default=None
```

skrub/_datetime_encoder.py
```diff
@@ -105,11 +115,22 @@ class DatetimeEncoder(SingleColumnTransformer):
...
+        Add periodic features with different granularities. Add periodic features
+        using either trigonometric (``circular``) or ``spline`` encoding.
+
     Attributes
     ----------
     extracted_features_ : list of strings
         The features that are extracted, a subset of ["year", …, "nanosecond",
-        "weekday", "total_seconds"]
+        "weekday", "total_seconds"]. If ``add_periodic=True``, the extracted
```
jeromedockes, line 131: I think this part is outdated there is no "add_periodic" anymore

skrub/_datetime_encoder.py
```diff
@@ -181,7 +202,7 @@ class DatetimeEncoder(SingleColumnTransformer):
     0         2024.0             4.0          14.0            1.713053e+09
     1         2024.0             5.0          15.0            1.715731e+09
     >>> encoder.extracted_features_
-    ['year', 'month', 'day', 'total_seconds']
+    ['birthday_year', 'birthday_month', 'birthday_day', 'birthday_total_seconds']
```
jeromedockes, line 205: I think this should stay as before (feature description without the column name) , see https://github.com/skrub-data/skrub/pull/1235/files#r1954645619

skrub/_datetime_encoder.py
jeromedockes: > I'll remove the old features

it seems like they are still here?

jeromedockes (commented):
this is shaping up very nicely :) I didn't have time to finish the review but I'll get back to it tomorrow

jeromedockes:
the failure is unrelated, I guess the model we were using in a test is not available on huggingface anymore

skrub/_datetime_encoder.py
```diff
@@ -284,16 +333,50 @@ def fit_transform(self, column, y=None):
...
+
+        col_name = sbd.name(column)
+        self.extracted_features_ = [
+            f"{col_name}_{_feat}" for _feat in self._partial_features
+        ]
+
+        # Iterating over all attributes that end with _encoding to use the default
+        # parameters
+        if self.periodic_encoding is not None:
+            enc_attr = ["year", "month", "weekday", "hour"][: idx_level + 1]
```
jeromedockes, line 360: here aren't we encoding 

- the year when really want to be encoding the day_of_year (or the month with a period of 12)
- the month when we really want to be encoding the day of the month

or am I missing something?

CHANGES.rst
```diff
@@ -12,19 +12,28 @@ Ongoing development
...
 - Progress messages when generating a ``TableReport`` are now written to stderr instead of stdout.
   :pr:`1236` by :user:`Priscilla Baah<priscilla-b>`
 
-- Optimize the :class:`StringEncoder`: significant memory reduction and 1.5x speed-up.
+- Optimize the :class:`StringEncoder`: memory reduction and speed-up in some circumstances.
+  :pr:`1248` by :user:`Gaël Varoquaux <gaelvaroquaux>`
+
+Bug fixes
+---------
+- Fixed a bug that caused the :class:`StringEncoder` to not work in presence of null values.
```
jeromedockes, line 36: that was a different PR

jeromedockes:
when i run this:

```python
import pandas as pd

login = pd.to_datetime(
    pd.Series(
        ["2024-05-13T01:05:36", "2024-09-02T21:46:02"], name="login")
)

DatetimeEncoder(periodic_encoding='spline').fit_transform(login)
```

I get this error

```
  File "<string>", line 19, in __PYTHON_EL_eval
  File "/home/jerome/workspace/scratch/2025/03/datetime-encoder.py", line 9, in <module>
    DatetimeEncoder(periodic_encoding='spline').fit_transform(login)
  File "/home/jerome/workspace/backedup_repositories/skrub/skrub/_on_each_column.py", line 154, in fit_transform
    return f(self, X, y=y)
           ^^^^^^^^^^^^^^^
  File "/home/jerome/workspace/backedup_repositories/skrub/skrub/_datetime_encoder.py", line 382, in fit_transform
    n_splines=_DEFAULT_ENCODING_SPLINES[enc_feature],
              ~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^
KeyError: 'day'
```

skrub/_datetime_encoder.py
```diff
@@ -254,12 +290,38 @@ class DatetimeEncoder(SingleColumnTransformer):
...
+    >>> encoder.fit_transform(login)
+       login_year  ...  login_hour_circular_1
+    0      2024.0  ...              -1.000000
+    1         NaN  ...                    NaN
+    2      2024.0  ...              -0.965926
+
+    Added features can be explored using ``DatetimeEncoder.all_outputs_``:
+    >>> encoder[-1].all_outputs_
+        ['login_year', 'login_total_seconds', 'login_month_circular_0', 'login_month_circular_1',
+        'login_day_circular_0', 'login_day_circular_1', 'login_hour_circular_0', 'login_hour_circular_1']
```
jeromedockes, line 309: how come there is only 0 and 1 here?

skrub/_datetime_encoder.py
```diff
@@ -284,16 +346,57 @@ def fit_transform(self, column, y=None):
                 f"Column {sbd.name(column)!r} does not have Date or Datetime dtype."
             )
         if self.resolution is None:
-            self.extracted_features_ = []
+            self._partial_features = []
```
jeromedockes, line 349: It seems like `self._partial_features` and `self.extracted_features_` are the same thing. could we keep only extracted_features?

skrub/_datetime_encoder.py
```diff
@@ -284,16 +346,57 @@ def fit_transform(self, column, y=None):
...
+            self._partial_features.append("weekday")
+        if self.add_day_of_year:
+            self._partial_features.append("day_of_year")
+
+        self.extracted_features_ = [f"{feat}" for feat in self._partial_features]
+
+        col_name = sbd.name(column)
+
+        # Adding transformers for periodic encoding
+        self._required_transformers = {}
```
jeromedockes, line 367: could we have a name more specific than "required", for example `self._periodic_encoders`?

skrub/_datetime_encoder.py
```diff
@@ -284,16 +346,57 @@ def fit_transform(self, column, y=None):
...
+        if self.periodic_encoding is not None:
+            encoding_level = list(_DEFAULT_ENCODING_PERIODS.keys())[1 : idx_level + 1]
+            if self.add_day_of_year:
+                encoding_level = encoding_level + ["day_of_year"]
+            if self.add_weekday:
+                encoding_level += ["weekday"]
```
jeromedockes, line 373: could you add a test to cover this line? thanks

skrub/_datetime_encoder.py
```diff
@@ -284,16 +346,57 @@ def fit_transform(self, column, y=None):
...
+        self.extracted_features_ = [f"{feat}" for feat in self._partial_features]
+
+        col_name = sbd.name(column)
+
+        # Adding transformers for periodic encoding
+        self._required_transformers = {}
+        if self.periodic_encoding is not None:
+            encoding_level = list(_DEFAULT_ENCODING_PERIODS.keys())[1 : idx_level + 1]
+            if self.add_day_of_year:
+                encoding_level = encoding_level + ["day_of_year"]
```
jeromedockes, line 371: IIUC when we add day of year we get the spline/circular encoding of both the month and the day of year, that's probably redundant right? do you think that day_of_year will be rarely used? or maybe we should always exclude the month and include the day of year in the peridoic encoders?

skrub/_datetime_encoder.py
```diff
@@ -284,16 +346,57 @@ def fit_transform(self, column, y=None):
...
+            if self.add_weekday:
+                encoding_level += ["weekday"]
+            for enc_feature in encoding_level:
```
jeromedockes, line 374: maybe encoding_levels, plural?

skrub/_datetime_encoder.py
```diff
@@ -26,6 +28,21 @@
...
+    "day_of_year": 366,
+    "month": 12,
+    "day": 30,
+    "hour": 24,
+    "weekday": 7,
+}
+_DEFAULT_ENCODING_SPLINES = {
+    "day_of_year": 12,
+    "month": 12,
+    "day": 30,
```
jeromedockes, line 41: I would put something smaller like 4 here, dedicating 30 dimensions to the day of the month is probably not a good investment. alternatively we could also not encode the day of the month as you had suggested in a previous comment

skrub/_datetime_encoder.py
```diff
@@ -325,10 +457,187 @@ def _check_params(self):
...
+        -------
+        transformed : DataFrame
+            The extracted features.
+        """
+
+        del y
+
+        self.transformer_ = self._periodic_spline_transformer()
+
+        X_out = self.transformer_.fit_transform(sbd.to_numpy(X).reshape(-1, 1))
```
jeromedockes, line 521: I have the impression it would be easier to do the handling of null values here, on the numpy arrays, but maybe I'm wrong

jeromedockes (commented):
very cool @rcap107 , we're down to nitpicks now :rocket:

skrub/_datetime_encoder.py
rcap107: I don't know what the best solution would be, maybe we could encode the day of the year with 12 splines, and avoid encoding the month? 

I can see day_of_year being used on its own (encoding or not)

skrub/_datetime_encoder.py
rcap107: Yes looking at the code you might be right, but I don't think there's a point in moving the check here now that it's being addressed in the main transform

I'll keep it in mind for the next time though

skrub/_datetime_encoder.py
rcap107: I feel like t his decision and a few more would need to be tested to know  :sweat_smile:  I'll set it to 4 for now

skrub/_datetime_encoder.py
jeromedockes: yeah, I would say month & day of year once encoded bring the same information (position within the year); I guess day of year in theory has a slightly better resolution (although that difference is probably not visible after smoothing by using only 12 splines). so I would say keep just 1 of the 2, whichever is easiest. and users who want periodic encoding and also the raw day of year are more of a special case I think, so they can use a columntransformer or feature union (for now, soon the expressions API), or we can think of a better way to accomodate that later if needed

PR #1239, opened 2025-02-14
DOC Update lock-file to have sphinx-gallery 0.19 minigallery duplicate fix

lesteve:
I thought OK, let's give pixi a go. Similar PR in scikit-learn https://github.com/scikit-learn/scikit-learn/pull/30822.

I ran `pixi update -e doc -e dev` and it seems to have updated sphinx-gallery and other dependencies, let me know if that's not "the right way".

For an example where the duplication happens: https://skrub-data.org/stable/reference/generated/skrub.AggJoiner.html#gallery-examples

image


lesteve:
Good news: this does fix the duplication bug [rendered doc](https://output.circle-artifacts.com/output/job/5fe85d54-e894-4ec8-9049-0214d0c29034/artifacts/0/doc/reference/generated/skrub.AggJoiner.html#gallery-examples)
image

Bad news: the red CI is telling me that I may not have used "the right way" :sweat_smile:

lesteve:
Looks like the CI issue is because you have `pixi-version: v0.30.0` and I have `pixi 0.41.3` locally. The updated pixi.lock has version 6 and I am guessing that `pixi 0.30.0` does not know how to deal with it.

jeromedockes:
Thanks so much for doing this @lesteve ! I tried updating the pixi version in the CI let's see if that helps

Vincent-Maladiere:
The rendering looks good!

jeromedockes:
Allright then! thanks again it's really appreciated, @lesteve

jeromedockes (approved):

lesteve:
Great, thanks for fixing the CI for me!

PR #1242, opened 2025-02-18
Better display for grid search

Vincent-Maladiere:
This draft patches sklearn `BaseSearchCV` to bring an informative progress bar during hp search. Improving the search experience will become even more important with the recipe.

To avoid introducing a new object in skrub, this draft suggests keeping the patch in a private module, which could be used internally by the recipe later.

Usage:
```python
from sklearn.model_selection import RandomizedSearchCV
from sklearn.datasets import fetch_california_housing
from scipy.stats import loguniform

# A single import to patch sklearn BaseSearchCV
import skrub._expressions._patch_sk_search


X, y = fetch_california_housing(return_X_y=True)

search = RandomizedSearchCV(
    HistGradientBoostingRegressor(max_iter=10),
    n_iter=100,
    param_distributions={"learning_rate": loguniform(1e-3, 1)},
    cv=3,
)
search.fit(X, y)
```
Screenshot 2025-02-18 at 16 28 01

For now, it displays an orange progress-bar with the best score and best hyper-parameters. As the hyper-parameter dictionary can be large, it can be accessed by scrolling to the right. WDYT?

### DIFF
--- a/skrub/_expressions/_patch_sk_search.py
+++ b/skrub/_expressions/_patch_sk_search.py
@@ -1,0 +1,298 @@
+import time
+from collections import defaultdict
+from itertools import product
+
+import numpy as np
+from joblib import Parallel, delayed
+from sklearn.base import BaseEstimator, clone, is_classifier
+from sklearn.metrics._scorer import (
+    _MultimetricScorer,
+)
+from sklearn.model_selection._search import BaseSearchCV
+from sklearn.model_selection._split import check_cv
+from sklearn.model_selection._validation import (
+    _fit_and_score,
+    _insert_error_scores,
+    _warn_or_raise_about_fit_failures,
+)
+from sklearn.utils.validation import _check_method_params, indexable
+
+
+def progress_bar(
+    idx,
+    total,
+    score_name,
+    best_score,
+    best_parameters,
+    bar_length=30,
+):
+    fraction = idx / total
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
I'm not comfortable with adding to skrub subclass of scikit-learn core functionality. These things should live in scikit-learn, not in skrub in terms of division of labor.

Vincent-Maladiere:
Sure, but this thing has a 0% chance of getting merged in sklearn one day, and in the meantime, the UX of the grid search is not great. How could we improve this?

GaelVaroquaux:
> Sure, but this thing has a 0% chance of getting merged in sklearn one day, and in the meantime, the UX of the grid search is not great. How could we improve this?

We sit in the same organizations than many core scikit-learn devs, we need to work with them to get this upstream. If we don't do this, we are building a fragile stack that will not work long term.

Vincent-Maladiere:
I created an issue in sklearn: https://github.com/scikit-learn/scikit-learn/issues/30852

jeromedockes:
> I created an issue in sklearn: https://github.com/scikit-learn/scikit-learn/issues/30852

nice! thanks @Vincent-Maladiere

Vincent-Maladiere:
Let's wait for scikit-learn to add this feature using callbacks.

PR #1248, opened 2025-02-26
ENH: optimize StringEncoder

GaelVaroquaux:
For memory and speed:

- Significant memory improvement (I did not measure)
- 1.5 speedup on one task

### DIFF
--- a/skrub/_string_encoder.py
+++ b/skrub/_string_encoder.py
@@ -157,7 +157,7 @@ class StringEncoder(SingleColumnTransformer):
             # Therefore, self.n_components_ below stores the resulting
             # number of dimensions of result.
             result = X_out[:, : self.n_components].toarray()
-            result = result.copy()
+            result = result.copy()  # To avoid a reference to X_out
         del X_out  # optimize memory: we no longer need X_out
 
         self._is_fitted = True
--- a/skrub/_string_encoder.py
+++ b/skrub/_string_encoder.py
@@ -182,12 +182,15 @@ class StringEncoder(SingleColumnTransformer):
             The embedding representation of the input.
         """
 
-        X = sbd.fill_nulls(X, "")
-        X_out = self.vectorizer_.transform(X)
+        X_filled = sbd.fill_nulls(X, "")
+        X_out = self.vectorizer_.transform(X_filled).astype("float32")
+        del X_filled  # optimizes memory: we no longer need X
         if hasattr(self, "tsvd_"):
             result = self.tsvd_.transform(X_out)
         else:
             result = X_out[:, : self.n_components].toarray()
+            result = result.copy()
+        del X_out  # optimize memory: we no longer need X_out
 
         return self._post_process(X, result)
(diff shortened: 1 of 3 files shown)


GaelVaroquaux:
FYI, the examples can be seen here: https://output.circle-artifacts.com/output/job/f25f287d-9bba-4153-977a-bbd5fbc2f3ca/artifacts/0/doc/auto_examples/02_text_with_string_encoders.html#sphx-glr-auto-examples-02-text-with-string-encoders-py
And I'm quite happy that with this PR StringEncoder is now faster than MinHashEncoder on this example :v:

skrub/_string_encoder.py
```diff
@@ -133,12 +133,17 @@ def fit_transform(self, X, y=None):
...
 
-        if (min_shape := min(X_out.shape)) >= self.n_components:
-            self.tsvd_ = TruncatedSVD(n_components=self.n_components)
+        if (min_shape := min(X_out.shape)) > self.n_components:
+            self.tsvd_ = TruncatedSVD(
+                n_components=self.n_components, algorithm="arpack"
+            )
             result = self.tsvd_.fit_transform(X_out)
+        elif X_out.shape[1] == self.n_components:
+            result = X_out.toarray()
```
jeromedockes, line 146: does the 'arpack' algorithm not like it when p == n_components? or is skipping the tsvd in that case an optimization?  if the latter, I guess the case where the number of discovered ngrams is exactly equal to `n_components` might be too rare to warrant it? and we might want a test for that branch

skrub/_string_encoder.py
```diff
@@ -152,6 +157,8 @@ def fit_transform(self, X, y=None):
             # Therefore, self.n_components_ below stores the resulting
             # number of dimensions of result.
             result = X_out[:, : self.n_components].toarray()
+            result = result.copy()
```
jeromedockes, line 160: I suppose the `copy()` is that otherwise the reference to the slice would prevent `X_out` from being garbage collected; might be worth a short comment

jeromedockes (commented):
LGTM but it would be nice to cover the shape[1] == n_components case in a test, LMK if you don't have time now I can push a small commit for it

GaelVaroquaux:
> LGTM but it would be nice to cover the shape[1] == n_components case in a test, LMK if you don't have time now I can push a small commit for it

It would be helpful. Thanks!

Let's also wait for a systematic evaluation before merging

skrub/_string_encoder.py
GaelVaroquaux: It's both, actually.

I think that needed this for the tests to pass. At least my first implementation hit a corner case in the tests (good tests!)

jeromedockes (approved):
LGTM, thanks!

Vincent-Maladiere (approved):
Looks good!

GaelVaroquaux:
@rcap107 : super useful. Thanks!!!

PR #1254, opened 2025-03-17
Small rewording of changelog and tweaked two tests

rcap107:
I modified the changelog to account for the results of the StringEncoder study. 

I also also modified the test to also include a null value, which is a case that wasn't tested before.


rcap107:
I'm blind, I didn't see the actual test

PR #1255, opened 2025-03-18
Skip Plotting in Table Report Based on Max Columns Given

### LINKED ISSUE #1245: Skip plotting in TableReport when there are many columns
### Problem Description

If the table has 100 columns, it takes a long time to generate the histograms and I probably won't look a them all anyway.
There could be a user-controllable threshold (say, 30 columns) above which no plots are generated and we get only the scrollable sample and  summary statistics. Not sure if column associations should still be computed. 

### Feature Description

below the threshold, plots are generated for all columns.
above the threshold, no plots are generated.

```python
skrub.TableReport(df, max_n_plots=30)

skrub.patch_display(max_n_plots=40)
```

None means always plot. (and `0` means never plot because there are always at least 0 columns)

### Alternative Solutions

_No response_

### Additional Context

_No response_

priscilla-b:
fixes #1245

### DIFF
--- a/skrub/_reporting/_data/templates/report.js
+++ b/skrub/_reporting/_data/templates/report.js
@@ -778,11 +778,7 @@ if (customElements.get('skrub-table-report') === undefined) {
         constructor(elem, exchange) {
             super(elem, exchange);
             this.elem.querySelector("[data-role='dismiss-button']")
-                .addEventListener("click", (e) => this.hide(e));
-        }
-
-        hide(event) {
-            this.elem.style.display = "none";
+                .addEventListener("click", (e) => this.hide());
         }
 
     }
--- a/skrub/_reporting/_data/templates/report.js
+++ b/skrub/_reporting/_data/templates/report.js
@@ -774,6 +774,22 @@ if (customElements.get('skrub-table-report') === undefined) {
     }
     SkrubTableReport.register(CopyButton);
 
+    class AlertDismissable extends Manager {
+        constructor(elem, exchange) {
+            super(elem, exchange);
+            this.elem.querySelector("[data-role='dismiss-button']")
+                .addEventListener("click", (e) => this.hide(e));
+        }
+
+        hide(event) {
+            this.elem.style.display = "none";
+        }
+
+    }
+
+    SkrubTableReport.register(AlertDismissable);
+
+
     function detectTheme(refElementId) {
 
         const shadowRootBody = document.querySelector('body');
--- a/skrub/_reporting/_patching.py
+++ b/skrub/_reporting/_patching.py
@@ -11,7 +11,7 @@ def _stashed_name(method_name):
     return f"_skrub_{method_name}"
 
 
-def _patch(cls, method_name, verbose):
+def _patch(cls, method_name, verbose, max_plot_columns):
     if (original_method := getattr(cls, method_name, None)) is None:
         return
     stashed_name = _stashed_name(method_name)
@@ -20,7 +20,10 @@ def _patch(cls, method_name, verbose):
     setattr(
         cls,
         method_name,
-        lambda df: getattr(TableReport(df, verbose=verbose), method_name)(),
+        lambda df: getattr(
+            TableReport(df, verbose=verbose, max_plot_columns=max_plot_columns),
+            method_name,
+        )(),
     )
 
 
@@ -52,7 +55,7 @@ def _get_to_patch(pandas, polars):
     return to_patch
 
 
-def patch_display(pandas=True, polars=True, verbose=1):
+def patch_display(pandas=True, polars=True, verbose=1, max_plot_columns=None):
     """Replace the default DataFrame HTML displays with ``skrub.TableReport``.
 
     This function replaces the HTML displays (what is shown when an object is
@@ -72,6 +75,10 @@ def patch_display(pandas=True, polars=True, verbose=1):
 
         * verbose = 1 prints how many columns have been processed so far.
         * verbose = 0 silences the output.
+    max_plot_columns : int, default=None
+        Maximum number of columns for which histogram plots should be generated.
+        If the number of columns in the dataframe is greater than this value,
+        the plots will not be generated. If None, all columns will be plotted.
 
     See Also
     --------
@@ -82,7 +89,10 @@ def patch_display(pandas=True, polars=True, verbose=1):
         Directly create a report from a dataframe.
     """
     _change_display(
-        _patch, _get_to_patch(pandas=pandas, polars=polars), verbose=verbose
+        _patch,
+        _get_to_patch(pandas=pandas, polars=polars),
+        verbose=verbose,
+        max_plot_columns=max_plot_columns,
     )
 
 
--- a/skrub/_reporting/_patching.py
+++ b/skrub/_reporting/_patching.py
@@ -76,7 +76,7 @@ def patch_display(pandas=True, polars=True, verbose=1, max_plot_columns=None):
         * verbose = 1 prints how many columns have been processed so far.
         * verbose = 0 silences the output.
     max_plot_columns : int, default=None
-        Maximum number of columns for which histogram plots should be generated.
+        Maximum number of columns for which plots should be generated.
         If the number of columns in the dataframe is greater than this value,
         the plots will not be generated. If None, all columns will be plotted.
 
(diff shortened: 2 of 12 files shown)


skrub/_reporting/_table_report.py
```diff
@@ -145,9 +151,14 @@ def _summary_without_plots(self):
 
     @property
     def _any_summary(self):
-        if "_summary_with_plots" in self.__dict__:
-            return self._summary_with_plots
-        return self._summary_without_plots
+        if self.max_plot_columns is None:
+            summary = self._summary_with_plots
+        elif self.max_plot_columns >= self.dataframe.shape[1]:
```
jeromedockes, line 156:
```suggestion
        elif self.max_plot_columns >= sbd.shape(self.dataframe)[1]:
```

this change will also require an additional import at the top of the file:

```python
from .. import _dataframe as sbd
```

the way you wrote it with `dataframe.shape` works perfectly, because pandas and polars happen to both expose the `shape` attribute with the same meaning. but to be consistent with the rest of skrub and in case some day we add another backend, for code that could be dealing with either a polars or a pandas dataframe we try to use the backend-agnostic utilities provided in `skrub._dataframe`

skrub/_reporting/_table_report.py
```diff
@@ -145,9 +151,14 @@ def _summary_without_plots(self):
 
     @property
     def _any_summary(self):
-        if "_summary_with_plots" in self.__dict__:
-            return self._summary_with_plots
-        return self._summary_without_plots
+        if self.max_plot_columns is None:
```
jeromedockes, line 154: this is a great way of handling it! I think we should rename `_any_summary` and make it a regular method rather than a property. maybe `self._any_summary -> self._get_summary()`?

jeromedockes (commented):
this is great, thank you very much @priscilla-b ! especially when we are inspecting intermediate steps in a pipeline, we often end up with dataframes that have a lot of columns and this will allow checking them without waiting a long time to generate lots of plots we won't look at :rocket: 

In fact it is so useful that I think the default should not be `None` but maybe 30 or something like that. but that would require adding clear messages in the report itself telling users what to do if they really want the plots, so it could be left for another pull request.

GaelVaroquaux:
> In fact it is so useful that I think the default should not be None but maybe 30 or something like that.

+1

> but that would require adding clear messages in the report itself telling users what to do if they really want the plots, so it could be left for another pull request.

Indeed a message visible somewhere (like the tab being greyed out and when the mouse goes over it a box appears giving the message). But probably for another PR because it requires markedly more work.

skrub/_reporting/_table_report.py
priscilla-b: I thought of this but wasn't sure the benefits of having it as a method than a property, so decided to maintain the existing implementation

skrub/_reporting/_table_report.py
priscilla-b: Yes, that makes sense

jeromedockes:
very nice @priscilla-b ! thanks for adding the message and doing the extra work to make it dismissable!

screenshot_2025-03-28T12:06:00+01:00

screenshot_2025-03-28T12:06:18+01:00

skrub/_reporting/_data/templates/report.js
```diff
@@ -774,6 +774,22 @@ if (customElements.get('skrub-table-report') === undefined) {
...
+    class AlertDismissable extends Manager {
+        constructor(elem, exchange) {
+            super(elem, exchange);
+            this.elem.querySelector("[data-role='dismiss-button']")
+                .addEventListener("click", (e) => this.hide(e));
+        }
+
+        hide(event) {
+            this.elem.style.display = "none";
+        }
```
jeromedockes, line 786:
```suggestion
            this.elem.querySelector("[data-role='dismiss-button']")
                .addEventListener("click", (e) => this.hide());
        }
```

jeromedockes: I think here we can rely on the parent classe's `hide()` to hide the whole element and it simplifies things slightly?

skrub/_reporting/_patching.py
```diff
@@ -52,7 +55,7 @@ def _get_to_patch(pandas, polars):
     return to_patch
 
 
-def patch_display(pandas=True, polars=True, verbose=1):
+def patch_display(pandas=True, polars=True, verbose=1, max_plot_columns=None):
```
jeromedockes, line 58: here also we could put 30 instead of None as the default

skrub/_reporting/_summarize.py
```diff
@@ -105,6 +105,7 @@ def summarize_dataframe(
     summary["n_constant_columns"] = sum(
         c["value_is_constant"] for c in summary["columns"]
     )
+    summary["plots_skipped"] = all(not c["plot_names"] for c in summary["columns"])
```
jeromedockes, line 108: I think we can define this based on the argument rather than inspecting the results, something like `summary["plots_skipped"] = not with_plots`. I think it is a bit simpler and slightly more robust (for example in the weird edge case where all columns are constant there will be no plots but not because they are skipped just because there is nothing to plot). WDYT?

skrub/_reporting/_table_report.py
jeromedockes: sorry for the nitpick: could we rename `get_summary` to `_get_summary`? this is not something for users and we want the freedom to change the contents of the summary whenever we want so the leading underscore indicates it's not part of the public API

skrub/_reporting/_table_report.py
```diff
@@ -143,11 +150,29 @@ def _summary_without_plots(self):
...
+        elif self.max_plot_columns >= sbd.shape(self.dataframe)[1]:
+            summary = self._summary_with_plots
+        else:
+            summary = self._summary_without_plots
+
+        # Add a flag to indicate if the plots were automatically skipped because
+        # there are too many columns
+        # Will be accessed in the HTML template to display a message to the user
+        summary["plots_auto_skipped"] = (
+            summary["plots_skipped"] and self.max_plot_columns == 30
```
jeromedockes, line 172: I'm not sure why we compare `max_plot_columns` to 30 here? for example if the user sets it to 20 it would still be skipped?

maybe the template can look directly at `plots_skipped` which you add in summarize, or we could create a flag in the conditional blocks above that we then use eg

```python
...
    summary = self._summary_with_plots
    plots_skipped = False
else:
    summary = self._summary_without_plots
    plots_skipped = True

summary['plots_auto_skipped'] = plots_skipped
```

jeromedockes (commented):
thanks! I know it's still in draft mode but took a quick look anyway :)

this really speeds up the report, now when there are many columns the bottleneck is definitely the computation of column associations

skrub/_reporting/_summarize.py
priscilla-b: Yes I think I did this at first, but I thought we were just reverting the argument and not really checking the presence of the actual plots. But yeah, what I did makes it more complicated, and should just keep it simple.

skrub/_reporting/_table_report.py
priscilla-b: I was thinking that if the user set the param to skip the plots directly, they wouldn't need to be shown a message explaining why the plots were skipped.
I wanted to show the message only when the plots were automatically skipped, but couldn't really figure out a simple implementation for checking if the default value is being used or not

jeromedockes: aah ok I understand now, thanks. I didn't realize the goal was to avoid showing the message when the user intentionally skipped plotting by setting `max_plot_columns`.

I think it might be good to show the message whenever plotting was skipped, even if it is intentional, especially since you made it easy to dismiss the message but both options make sense

jeromedockes:
>  It seems like the tiles without distributions are somewhat redundant with the 'stats' panel 

that's a good point. Still I would like to have this PR for the next release so maybe we should finish the few remaining details for the current version and open a separate issue for follow-ups?

priscilla-b:
Hi Jerome,

Yes I will complete this and reopen by close of day.

Vincent-Maladiere:
Right, let's do this in two steps then 👍

jeromedockes (approved):
LGTM! Thanks again @priscilla-b  :tada:

PR #1258, opened 2025-03-19
FIX accept input_features parameter in TableVectorizer.get_feature_names_out

### LINKED ISSUE #1256: `TableVectorizer.get_feature_names_out()` not compatible with scikit-learn `Pipeline`
With the following code:

```python
from sklearn.linear_model import LogisticRegression
 
X, y = make_classification(
    n_samples=1000, n_features=20, n_informative=10, n_classes=2, random_state=42
)
lr = make_pipeline(
    TableVectorizer(), LogisticRegression(max_iter=100)
).fit(pd.DataFrame(X), y)
lr.get_feature_names_out()
```

will fail with:

```pytb
Cell In[7], line 1
----> 1 lr.get_feature_names_out()

File ~/Documents/scikit-learn-workspace/src/skrub/.pixi/envs/dev/lib/python3.11/site-packages/sklearn/pipeline.py:820, in Pipeline.get_feature_names_out(self, input_features)
    814     if not hasattr(transform, "get_feature_names_out"):
    815         raise AttributeError(
    816             "Estimator {} does not provide get_feature_names_out. "
    817             "Did you mean to call pipeline[:-1].get_feature_names_out"
    818             "()?".format(name)
    819         )
--> 820     feature_names_out = transform.get_feature_names_out(feature_names_out)
    821 return feature_names_out

TypeError: TableVectorizer.get_feature_names_out() takes 1 positional argument but 2 were given
```

because `TableVectorizer.get_feature_names_out()` does not expose the `input_features` parameter expected by the scikit-learn `Pipeline`.

glemaitre:
closes #1256 

This PR makes sure to expose an `input_features` parameter in the `get_feature_names_out` to be compatible with scikit-learn.

Right now, the parameter is ignored. One question would be: do we want to allow overwriting the column names. In this case, we would need to change more code because we would need to overwrite the name of the columns of the output dataframe in `transform` to have a consistency between `get_feature_names_out()` and the `feature_names_in_` of the subsequent steps in a scikit-learn pipeline.

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -676,8 +676,13 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
         tags.input_tags.allow_nan = True
         return tags
 
-    def get_feature_names_out(self):
+    def get_feature_names_out(self, input_features=None):
         """Return the column names of the output of ``transform`` as a list of strings.
+
+        Parameters
+        ----------
+        input_features : array-like of str or None, default=None
+            Ignored.
 
         Returns
         -------
--- a/skrub/tests/test_table_vectorizer.py
+++ b/skrub/tests/test_table_vectorizer.py
@@ -10,8 +10,9 @@ from numpy.testing import assert_array_almost_equal, assert_array_equal, assert_
 from pandas.testing import assert_frame_equal
 from scipy.sparse import csr_matrix
 from sklearn.base import clone
+from sklearn.linear_model import Ridge
 from sklearn.pipeline import make_pipeline
-from sklearn.preprocessing import FunctionTransformer, OneHotEncoder
+from sklearn.preprocessing import FunctionTransformer, OneHotEncoder, StandardScaler
 from sklearn.utils._testing import skip_if_no_parallel
 from sklearn.utils.fixes import parse_version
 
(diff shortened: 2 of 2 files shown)


GaelVaroquaux (approved):
LGTM, but could we have a tiny changelog entry. it's useful for users

glemaitre:
> LGTM, but could we have a tiny changelog entry. it's useful for users

Indeed :)

jeromedockes:
Thanks a lot @glemaitre !!

Note we have a bunch of estimators defining `get_feature_names_out` and none of them accept `input_features` so it might be worth creating another issue with a TODO list

> do we want to allow overwriting the column names. In this case, we would need to change more code because we would need to overwrite the name of the columns of the output dataframe in transform to have a consistency between get_feature_names_out() and the feature_names_in_ of the subsequent steps in a scikit-learn pipeline.

I would say no, in any case a preprocessing check in the tablevectorizer raises an exception if the columns of the input are different in transform than in fit. but you know more about `get_feature_names_out` and how it is used in scikit-learn

jeromedockes (approved):

GaelVaroquaux:
> Note we have a bunch of estimators defining get_feature_names_out and none of them accept input_features so it might be worth creating another issue with a TODO list

+1

Can we use scikit-learn's common tests to detect this?

Issue #1257, opened 2025-03-19
Some issues with TableReport I noticed while working with many columns


rcap107:
### Problem Description

I am looking at a dataset with ~300 columns and ~13k rows, and the large number of columns is showing some limitations of the current TableReport:

1. It took almost 40s to generate the report (which is a known issue and should be fixed by limiting the number of columns)
2. If there are too many columns, the top bar disappears, so I cannot change tab to something else (in this case, there are some informative columns I would like to take a look at, but they're hidden in the middle of a lot of noise)
3. The "Distribution" tab is particularly problematic because it moves all over the page, and if I open a drop down the view will immediately go down to the bottom row, so that I have to scroll all the way up to see the values and to change the tab 
4. The Associations table has only the first 20 pairs, while I'd expect to have a scroll bar if there are a lot of high scores (see image)
5. It should be possible to select columns directly from the "Table" tab, ideally by navigating through them and selecting with a keystroke. Clicking on the checkbox for each column of interest is very clunky. 

![Image](https://github.com/user-attachments/assets/6b32f1a2-824a-4bf2-a6d8-f0c4c34567a1)

I am using VSCode for this.

### Feature Description

Point 1 should already have been solved in a different issue. 
Point 2 and 3 could be addressed either by freezing the tabs to the top of the screen, or by adding a "go back to top" button. 
The issue with the drop down menu is problematic especially if you're looking at a column in the middle of the list, because then going back to top wouldn't help much, and singling out a column is not easy to do. 

In general, working only with a few columns in arbitrary spots in the table is very clunky, because I can only do it from the table tab one at a time, or from the distributions tab, hoping that there is a pre-made filter that can select what I want. 

I think there should be a way of choosing columns to work with, select only those and hide the rest, then keep the selection for all tabs.

### Alternative Solutions

A way around some of the issues I am having would be pre-processing the dataframe in pandas/polars *before* generating the TableReport, but that would defeat the point of using the TableReport as a first view at the table to direct the preprocessing in the first place.

jeromedockes:
thanks @rcap107 that's super useful. indeed #1 will be taken care of in #1255 

I'm not sure I understand #2 "the top bar disappears"?

rcap107:
> I'm not sure I understand [#2](https://github.com/skrub-data/skrub/pull/2) "the top bar disappears"?

[...]

PR #1259, opened 2025-03-19
Stack cards horizontally in distribution tab

GaelVaroquaux:
To avoid vertical scrolling in IDEs.

Related to #1257

### DIFF
--- a/skrub/_reporting/_data/templates/_variables.css
+++ b/skrub/_reporting/_data/templates/_variables.css
@@ -58,7 +58,7 @@
     --color-background-highlight-primary: #d6f5d6;
     --color-background-active: #ffffaa;
     --color-background-info: #0f4e9b;
-    --color-background-card: #fcfffc;
+    --color-background-card: #f5fff5;
     --color-background-card-header: #d6f5d6;
 
     /* buttons */
--- a/skrub/_reporting/_data/templates/_variables.css
+++ b/skrub/_reporting/_data/templates/_variables.css
@@ -58,8 +58,8 @@
     --color-background-highlight-primary: #d6f5d6;
     --color-background-active: #ffffaa;
     --color-background-info: #0f4e9b;
-    --color-background-card: #fffdfa;
-    --color-background-card-header: #ffebc5;
+    --color-background-card: #fcfffc;
+    --color-background-card-header: #d6f5d6;
 
     /* buttons */
     --color-background-button-primary: #e0e0e0;
@@ -82,7 +82,7 @@
     /* borders */
     --color-border-primary: #aaa;
     --color-border-secondary: #bfbfbf;
-    --color-border-highlight: #9fce9f;
+    --color-border-card: #9fce9f;
 
     /* shadows */
     --color-shadow: hsl(0, 0%, 63%);
@@ -96,10 +96,10 @@
     --color-background-tertiary: #2b2b2b;
     --color-background-quaternary: #353535;
     --color-background-highlight-primary: #512ca4;
-    --color-background-active: #2a9d8f;
+    --color-background-active: #1A6154;
     --color-background-info: #0f4e9b;
-    --color-background-card: #2a2238;
-    --color-background-card-header: #483b60;
+    --color-background-card-header: #392662;
+    --color-background-card: #2c2833;
 
     /* buttons */
     --color-background-button-primary: #212121;
@@ -112,17 +112,17 @@
     /* texts */
     --color-text-primary: white;
     --color-text-secondary: #d4d4d4;
-    --color-text-success: #2a9d8f;
+    --color-text-success: #6ED8CC;
     --color-text-info: white;
-    --color-text-warning: #e76f51;
-    --color-text-critical: #e63946;
+    --color-text-warning: #E8B254;
+    --color-text-critical: #F4A4A4;
     --color-text-link: #8c8cff;
     --color-text-visited: #ffadff;
 
     /* borders */
     --color-border-primary: #666;
     --color-border-secondary: #444;
-    --color-border-highlight: #264653;
+    --color-border-card: #6549a1;
 
     /* shadows */
     --color-shadow: #0f0f0f;
--- a/skrub/_reporting/_data/templates/_variables.css
+++ b/skrub/_reporting/_data/templates/_variables.css
@@ -72,7 +72,7 @@
     /* texts */
     --color-text-primary: black;
     --color-text-secondary: #555;
-    --color-text-success: green;
+    --color-text-success: #006600;
     --color-text-info: white;
     --color-text-warning: #ff6600;
     --color-text-critical: #dd0000;
--- a/skrub/_reporting/_data/templates/_variables.css
+++ b/skrub/_reporting/_data/templates/_variables.css
@@ -86,7 +86,7 @@
 
     /* shadows */
     --color-shadow: hsl(0, 0%, 63%);
-    --color-shadow-keyboard: rgb(255, 255, 255, 0.5);
+    --color-shadow-keyboard: var(--color-border-secondary);
 }
 
 #report.dark {
@@ -126,5 +126,5 @@
 
     /* shadows */
     --color-shadow: #0f0f0f;
-    --color-shadow-keyboard: rgb(255, 255, 255, 0.5);
+    --color-shadow-keyboard: var(--color-border-secondary);
 }
--- a/skrub/_reporting/_data/templates/_variables.css
+++ b/skrub/_reporting/_data/templates/_variables.css
@@ -58,6 +58,8 @@
     --color-background-highlight-primary: #d6f5d6;
     --color-background-active: #ffffaa;
     --color-background-info: #0f4e9b;
+    --color-background-card: #fffdfa;
+    --color-background-card-header: #ffebc5;
 
     /* buttons */
     --color-background-button-primary: #e0e0e0;
@@ -96,6 +98,8 @@
     --color-background-highlight-primary: #512ca4;
     --color-background-active: #2a9d8f;
     --color-background-info: #0f4e9b;
+    --color-background-card: #2a2238;
+    --color-background-card-header: #483b60;
 
     /* buttons */
     --color-background-button-primary: #212121;
--- a/skrub/_reporting/_data/templates/base.css
+++ b/skrub/_reporting/_data/templates/base.css
@@ -191,9 +191,7 @@ dd {
 /* Special-case the article (container) */
 article.wrapper {
     padding-bottom: 0pt;
-}
-
-.flow > * + * {
+}.flow > * + * {
     margin-top: var(--spacing-m);
 }
 
--- a/skrub/_reporting/_data/templates/base.css
+++ b/skrub/_reporting/_data/templates/base.css
@@ -184,6 +184,11 @@ dd {
     padding-block-end: var(--spacing-m);
 }
 
+/* Special-case the article (container) */
+article.wrapper {
+    padding-bottom: 0pt;
+}
+
 .flow > * + * {
     margin-top: var(--spacing-m);
 }
@@ -249,6 +254,10 @@ dd {
 
 .clickable {
     cursor: pointer;
+}
+
+.float-left {
+    float: left;
 }
 
 :is(.ok, .warning, .critical) {
--- a/skrub/_reporting/_data/templates/base.css
+++ b/skrub/_reporting/_data/templates/base.css
@@ -166,8 +166,8 @@ dd {
     margin-top: var(--spacing-m);
 }
 
-.padding-b-m {
-    padding-bottom: var(--spacing-m);
+.padding-b-s {
+    padding-bottom: var(--spacing-s);
 }
 
 .horizontal-scroll {
@@ -190,8 +190,10 @@ dd {
 
 /* Special-case the article (container) */
 article.wrapper {
-    padding-bottom: 0pt;
-}.flow > * + * {
+    padding-block-end: 0;
+}
+
+.flow > * + * {
     margin-top: var(--spacing-m);
 }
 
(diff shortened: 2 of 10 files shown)


skrub/_reporting/_data/templates/base.css
```diff
@@ -251,6 +256,10 @@ dd {
     cursor: pointer;
 }
 
+.float-left {
```
jeromedockes, line 259: why do we float the bar & buttons? it results in wasting space and looks a bit weird IMO

screenshot_2025-03-19T19:10:05+01:00

skrub/_reporting/_data/templates/column-summaries.css
```diff
@@ -4,9 +4,18 @@
 /* Collection of column summary cards (as shown in the "column summaries" tab) */
 /* --------------------------------------------------------------------------- */
 
+/* Used as a container to enable horizontal scroll of the column
+ * summaries*/
+.horizontal-scroll {
```
jeromedockes, line 9: there is already a horizontal-scroll class in base.css that makes an element scrollable horizontally. the padding at the bottom is not really related to that so should maybe be applied separately

skrub/_reporting/_data/templates/column-summaries.css
```diff
@@ -4,9 +4,18 @@
...
+    overflow-x: scroll;
```
jeromedockes, line 10: you probably want `overflow-x: auto` because otherwise the scrollbar is always shown even when it is disabled because the screen is larger than the content (but anyway we can just use the existing horizontal-scroll class)

jeromedockes (commented):
thanks a lot @GaelVaroquaux ! I think this can be nice in notebooks and vs code. 

the downside as you mentioned in the issue it does reduce the number of cards we can see at once and force us to scroll horizontally to see them. 

when in a full page I think it makes a suboptimal use of space:

screenshot_2025-03-19T19:27:16+01:00


here I would rather not have to scroll

skrub/_reporting/_data/templates/base.css
jeromedockes: also moving the buttons to the right means there is much less space for the tooltip

screenshot_2025-03-19T19:33:52+01:00


which was one reason for having it on the left

GaelVaroquaux:
This was meant to be a quick PR so that people told me if the idea was a good one.

This tool is going to be used mostly in IDEs and notebook, so the vertical scaling is crucial (I'm not the only saying this). This is why I moved the button to the side. I'll do it in a cleaner way than the float :).

Yes, we can special case the "open", but it makes the code more complicated and I wonder if we are wasting time on a feature that is almost not going to be used.

skrub/_reporting/_data/templates/column-summaries.css
GaelVaroquaux: Fair point, I hadn't seen it (my bad, I should have "git grep"ed better).

I've refactor for a cleaner solution

skrub/_reporting/_data/templates/column-summaries.css
GaelVaroquaux: OK, I thought that it was a good idea to always show it, but I think that you are right not to

skrub/_reporting/_data/templates/base.css
GaelVaroquaux: Seriously, vertical space.

Look, this is the user experience, even with the button on this side
image

The table report takes a full screen. I cannot see the line that I have run to generate it. It's really annoying. It's a major usability problem, and TBH a reason for me not wanting to use it.

We can improve the tooltip, but it really does not matter compared the waste of vertical space

GaelVaroquaux: Also, I think that the problem that you flagged cannot happen with the new code that I have committed.

skrub/_reporting/_data/templates/tabs.css
```diff
@@ -91,5 +91,4 @@ button.tab[data-has-warning]:not(:hover):not([data-is-selected]) {
 .report-bottom-border {
     width: calc(100% - 2 * var(--radius));
     border-bottom: var(--border-s) solid var(--color-border-secondary);
-    margin: var(--spacing-m) auto 0 auto;
```
jeromedockes, line 94: we can replace `--spacing-m` with `--spacing-s` to make it smaller but having the bottom border touch the content looks a bit weird:

screenshot_2025-03-20T10:09:09+01:00

otherwise we can also just remove that border

rcap107:
I really like this feature, it's far more convenient to use in VSCode now. 
Screenshot_20250320_101040

I don't have issues with vertical space either, I had to zoom in a lot to get to the point where the line of code was cut, so that's not a problem on my side. 

One thing I noticed is that I was working on the code by looking at the TableReport, then writing something and executing the cell, then going back to the TR to see what else needed doing. That mean I was scrolling up all the time to see the columns and the TR tabs, and this feature addresses the endless scrolling.

skrub/_reporting/_data/templates/column-summaries.css
```diff
@@ -18,21 +20,13 @@
     margin-top: var(--spacing-l);
 }
 
-.column-summary-group > .card {
-    margin: 0;
+/* A bit of space between the bar and the buttons */
+.column-selection-buttons {
+    padding-left: 10pt;
```
jeromedockes, line 25:
```suggestion
    padding-left: var(--spacing-m);
```

skrub/_reporting/_data/templates/column-summaries.css
```diff
@@ -18,21 +20,13 @@
...
-
-@media screen and (min-width: 60rem) {
-    .column-summary-group {
-        display: flex;
-        flex-direction: row;
-        flex-wrap: wrap;
-        gap: var(--spacing-l);
-    }
+.column-summary-group > .card {
+    margin: 0;
```
jeromedockes, line 29: the shadows being abruptly cut off looks strange/unintentional so removing the margin is fine but then I think we should also remove the shadows

skrub/_reporting/_data/templates/column-summaries.css
```diff
@@ -4,9 +4,11 @@
...
 
+/* The cards next to each other and horizontal scrolling. */
 .column-summary-group {
     display: flex;
-    flex-direction: column;
+    flex-direction: row;
+    flex-wrap: nowrap;
     gap: var(--spacing-l);
     min-width: min-content;
     max-width: max-content;
```
jeromedockes, line 14:
```suggestion
    width: max-content;
```
(note this will make it use less vertical space)

jeromedockes (commented):
LGTM once the suggestion about relaxing the flex container's width is added

jeromedockes:
for the buttons, indeed there is not the issue  I pointed out anymore. one drawback of having them on the side is that they move when we click them. 

we could also consider just removing those buttons and the bar because I don't know how many people would use them (either because they don't need it or haven't figured out what it does)

GaelVaroquaux:
Could we have the buttons and the bar appear only when the selection is not empty? Ie when clicking on the tick box of the cards?

GaelVaroquaux:
For the button:

> we could also consider just removing those buttons and the bar because I don't know how many people would use them (either because they don't need it or haven't figured out what it does)

I think that what would be really cool would be to have the button and the bar visible only when the bar is not empty. Ie, clicking on the tickbox would have them show up. Do you know easily how to add this, @jeromedockes ?

skrub/_reporting/_data/templates/tabs.css
GaelVaroquaux: I see this padding needed only the statistics and associations panel. I've added it

skrub/_reporting/_data/templates/column-summaries.css
GaelVaroquaux: I like the shadows: 
image

GaelVaroquaux:
> I suggest we keep as it is in this PR and experiment with that in a separate one

I agree! :)

I do like the idea of making this not visible by default. It would also decrease the visual crowding, which would be great

GaelVaroquaux:
OK, I think that this is ready for merge

GaelVaroquaux:
> to prevent cropping the shadows isn't it easier to keep the margin that used to be around the cards?

That does not work, because it applies to the enclosing element and thus the shadow stop inside the enclosed element, leaving the white margin outside.

jeromedockes:
> That does not work, because it applies to the enclosing element and thus the shadow stop inside the enclosed element, leaving the white margin outside.

I'm not sure I understood 100% what you mean but I still see the shadows being cropped with the current pr's version:

screenshot_2025-03-20T14:19:04+01:00

we can avoid it either by having a margin around the cards

```diff
diff --git a/skrub/_reporting/_data/templates/column-summaries.css b/skrub/_reporting/_data/templates/column-summaries.css
index a037513a..bbcbc341 100644
--- a/skrub/_reporting/_data/templates/column-summaries.css
+++ b/skrub/_reporting/_data/templates/column-summaries.css
@@ -9,7 +9,7 @@
     display: flex;
     flex-direction: row;
     flex-wrap: nowrap;
-    gap: var(--spacing-l);
+    gap: 0;
     width: max-content;
 }
 
@@ -25,9 +25,14 @@
 }
 
 .column-summary-group > .card {
-    margin: 0;
+    margin: var(--spacing-s) var(--spacing-l) var(--spacing-l) var(--spacing-l);
 }
 
+.column-summary-group > .card + .card {
+    margin-left: 0;
+}
+
+
 /* Individual column summary cards */
 /* ------------------------------- */
 
```

screenshot_2025-03-20T14:20:22+01:00


or padding on the parent

```diff
diff --git a/skrub/_reporting/_data/templates/column-summaries.css b/skrub/_reporting/_data/templates/column-summaries.css
index a037513a..7aac0e7d 100644
--- a/skrub/_reporting/_data/templates/column-summaries.css
+++ b/skrub/_reporting/_data/templates/column-summaries.css
@@ -11,6 +11,7 @@
     flex-wrap: nowrap;
     gap: var(--spacing-l);
     width: max-content;
+    padding: var(--spacing-s) var(--spacing-l) var(--spacing-l) var(--spacing-l);
 }
 
 /* Increase the gap between buttons & columns to make it clearer the buttons are
```

screenshot_2025-03-20T14:24:09+01:00

jeromedockes:
just so we have the full discussion here: there are also some comments in

https://github.com/GaelVaroquaux/skrub/pull/3

jeromedockes:
not a big deal but in the first tab also, removing the bottom padding means the column summary card overlaps the border

screenshot_2025-03-21T09:49:16+01:00

GaelVaroquaux:
> not a big deal but in the first tab also, removing the bottom padding means the column summary card overlaps the border
> 
> screenshot_2025-03-21T09:49:16+01:00

Fun thing: I'm not getting this, it must be a firefox vs chrome difference. Hence the importance of testing on many browsers

GaelVaroquaux:
Card coloring implemented.

This should now be good for review
image

jeromedockes:
I think we are losing some contrast (unless we also change the test), https://webaim.org/resources/contrastchecker/ complains about these (there may be more):

screenshot_2025-03-21T15:42:42+01:00
screenshot_2025-03-21T15:41:15+01:00

the bar plots & background are also quite close

screenshot_2025-03-21T15:41:36+01:00

and this is a detail but buttons don't match the background anymore, not sure if that's good or bad

screenshot_2025-03-21T15:42:03+01:00

the background color makes it harder to have a good contrast and I'm not sure the benefit it brings makes up for that.

 
maybe we need to tone them down or change the text color. to avoid delaying the horizontal scroll thing we can also merge it like this and revisit the colors in another pr

GaelVaroquaux:
Ok, can you take over.
I really must be working on the grant proposal

jeromedockes:
ok the contrast between the text and background should be ok now, @GaelVaroquaux lmk if it's ok for you

jeromedockes:
I don't think the changes to the bottom border are a big improvement, now it is not centered and it touches the scrollbar. but it's kind  of a detail

jeromedockes:
(closed by mistake)

GaelVaroquaux:
> I don't think the changes to the bottom border are a big improvement,

Agreed

skrub/_reporting/_data/templates/column-summaries.css
```diff
@@ -121,14 +110,9 @@
 
 .columns-in-sample-tab .card-header {
     position: relative;
-    background-color: var(--color-background-highlight-primary);
-    border-block-end: var(--border-s) solid var(--color-border-highlight);
```
GaelVaroquaux, line 125: @jeromedockes : do you think that this is the faulty change, regarding the border?

jeromedockes: no I think it is [this one](https://github.com/skrub-data/skrub/pull/1259/files#diff-bf4fee1dc6f544b4435e8bc709320ea16fff8cd09fbbe5f5e25a88d072f9c81fL94)

if it's ok for you I'll restore the deleted rule it but making the vertical margin much smaller

GaelVaroquaux: Yes, but much smaller :).

Thanks!

skrub/_reporting/_data/templates/_variables.css
```diff
@@ -80,11 +82,11 @@
     /* borders */
     --color-border-primary: #aaa;
     --color-border-secondary: #bfbfbf;
-    --color-border-highlight: #9fce9f;
+    --color-border-card: #9fce9f;
 
     /* shadows */
     --color-shadow: hsl(0, 0%, 63%);
-    --color-shadow-keyboard: rgb(255, 255, 255, 0.5);
+    --color-shadow-keyboard: var(--color-border-secondary);
```
jeromedockes, line 89: that one is not related to this pr but I noticed it and figured let's fix it here and not do another pr for 2 lines of diff

jeromedockes (approved):
should be ok now!

GaelVaroquaux:
@rcap107 : can you check the visuals (eg the rendered docs) and merge if happy?

PR #1264, opened 2025-03-31
ENH add periodic encoding in tabular learner

glemaitre:
This PR adds periodic encoding by default when the estimator is a linear model

#### TODO

- [x] add tests
- [x] update the docstring of the `tabular_learner`
- [x] update the user guide
- [x] update the potential example impacted

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,11 @@ Changes
 - Optimize the :class:`StringEncoder`: significant memory reduction and 1.5x speed-up.
   :pr:`1248` by :user:`Gaël Varoquaux <gaelvaroquaux>`
 
+- The estimator returned by :func:`tabular_learner` now uses spline encoding of
+  datetime features when the supervised learner is not a model based on decision
+  trees such as random forests or gradient boosting. :pr:`1264` by
+  :user:`Guillaume Lemaitre <glemaitre>`.
+
 Bug fixes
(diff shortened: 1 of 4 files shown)


jeromedockes:
should it be just for linear models? or would most non-tree-based models (and perhaps even tree-based ones) benefit from it?

GaelVaroquaux:
> should it be just for linear models? or would most non-tree-based models (and perhaps even tree-based ones) benefit from it?

We're very certain that it should be for linear model, and less certain about the others.

I would be in favor to add it for all non-tree models. But if this delays the merge because of the need to find a consensus, I would rather that we move forward with only linear models.

glemaitre:
I don't mind to have an `else` statement such that all "other models" will compute it.

jeromedockes:
That would be very interesting to investigate in benchmarks. In the meanwhile my hunch would be that spline or circular encoding is the safe bet when we don't know much about the learner: it might be a bit wasteful when the learner doesn't need it, but it shouldn't tank the quality of predictions, and it can really improve the predictions when the learner does need it. that's just a guess though I don't have experiments or a reference to back it up

jeromedockes (approved):
great, thank you very much @glemaitre !

glemaitre:
Lol thanks @jeromedockes ant @Vincent-Maladiere

PR #1266, opened 2025-03-31
Adding the SimpleCleaner object

rcap107:
Addresses #925 

This PR adds moves the code used to build the pre-processors used by the TableVectorizer to a private function, and implements the Skrubber object. The Skrubber is a lightweight object that executes some of the transformations done by the TableVectorizer, mainly to do a "first pass" over datasets, prior to the actual transformation.

The Skrubber is equivalent to a TableVectorizer with all encoders set to "passthrough", with the major distinction of not converting numeric columns to float32. 

We might want to reconsider the name, because calling it "Skrubber" may give it more importance in the context of the library than I think it deserves (this is really just a lightweight object that does something very simple).

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -281,8 +281,7 @@ class SimpleCleaner(TransformerMixin, BaseEstimator):
         self._pipeline = make_pipeline(*all_steps)
         result = self._pipeline.fit_transform(X)
         input_names = all_steps[0].feature_names_out_
-        to_steps = {col: [] for col in input_names}
-        self.all_processing_steps_ = {col: [] for col in to_steps}
+        self.all_processing_steps_ = {col: [] for col in input_names}
         for step in all_steps[1:]:
             for col, transformer in step.transformers_.items():
                 self.all_processing_steps_[col].append(transformer)


jeromedockes:
thanks a lot @rcap107 this is really great and something we've wanted for a while! I think the code part is good to go so it is mostly a question of getting the docstring past the linter, so we can probably have this in the upcoming release :rocket:

Vincent-Maladiere:
Very nice! Regarding the name, what about something like `SimpleCleaner`?

jeromedockes:
you will also need to add it to `doc/reference/index.rst`

Vincent-Maladiere:
Also, to better illustrate the capabilities of this transformer, could you list its features in the docstring, in a bulletpoint way? Some transformations seem missing in the docstring, like `CleanCategories`.

skrub/_table_vectorizer.py
```diff
@@ -112,6 +112,216 @@ def _check_transformer(transformer):
...
+        Maps the name of each column to a list of all the processing steps that were
+        applied to it.
+    Notes
+    -----
+    The ``SimpleCleaner`` performs the following set of transformations on each column:
+    - ``CleanNullStrings()``: replace strings used to represent null values
+    with actual null values.
+    - ``DropIfTooManyNulls()``: drop the column if it contains too many null values.
+    - ``ToDatetime()``: parse datetimes represented as strings and return them as
+    ``datetime64``.
```
jeromedockes, line 182:
```suggestion
    actual datetimes with the correct dtype.
```

(they are called `Datetime` in polars)

skrub/_table_vectorizer.py
```diff
@@ -112,6 +112,216 @@ def _check_transformer(transformer):
...
+    - ``CleanCategories()``: process categorical columns depending on the dataframe
+    library (Pandas or Polars) to force consistent typing and avoid issues downstream.
+    - ``ToStr()``: convert columns to strings, unless they are numerical or datetime.
```
jeromedockes, line 185:
```suggestion
    - ``ToStr()``: convert columns to strings, unless they are numerical, categorical, or datetime.
```

skrub/_table_vectorizer.py
```diff
@@ -112,6 +112,216 @@ def _check_transformer(transformer):
...
+
+    The ``SimpleCleaner`` object should only be used for preliminary observations on
```
jeromedockes, line 187:
```suggestion
    The ``SimpleCleaner`` object should only be used for preliminary sanitizing of
```

skrub/_table_vectorizer.py
```diff
@@ -112,6 +112,216 @@ def _check_transformer(transformer):
...
+    the data, while the ``TableVectorizer`` should instead be used to transform the
+    data. This is because the ``SimpleCleaner`` does not convert null values to a
+    consistent representation, which can lead to issues when the data is passed to
+    downstream models.
```
jeromedockes, line 191: I think that part is a bit hard to understand. Maybe something like:

the simplecleaner does not perform any transformations on numeric columns. the tablevectorizer on the other hand converts them all to float32 and ensures null values are represented with nans which can be handled correctly by downstream scikit-learn estimators.

jeromedockes (commented):
just a few nitpicks (+ the formatting of the docstring to fix).

looks great otherwise, thanks! :rocket:

jeromedockes (approved):
awesome, thanks  a lot @rcap107 . This is super useful when you want more control over the encoding than what the tablevectorizer allows, but still want to benefit from its preprocessing

GaelVaroquaux:
Sorry, I was too slow on this, but regarding name, I would like to strongly suggest that "SimpleCleaner" is not a good name: the "Simple" is too much. It's not informative, and should not be there. "Cleaner" is enough, and if we ever do a sophisticated cleaner, it will come with a longer name.

Maybe that was merged a bit too fast, because we now have a released version and we need to change the name

Issue #1253, opened 2025-03-15
Normalizing the Gap, Text, and String Encoders


GaelVaroquaux:
In many (but not all learners), the Euclidean product and Euclidean distance between samples is what matters. For instance, the kNN will retrieve the closest samples in the l2-norm sense.

I believe that it such settings it makes sense to normalize the various string encoders (in particular Gap, Text, and String encoders), so that the average distance between their encoding has the same mean value. Indeed, if one has a large value, it dominates in the choice of a nearest neighbor.

The mean distance is quadratic and thus costly to compute. A good proxy (that is also meaningful) is the average norm.

I did a small script investigating the problem:

```python
# %%
# Our data
from skrub import datasets
data = datasets.fetch_employee_salaries()
df = data.X[['employee_position_title', 'year_first_hired']]


# %%
# Look at distances induced by the different encoders
import numpy as np
import skrub as skb
from sklearn.preprocessing import StandardScaler
from sklearn.metrics import pairwise_distances

encoders = [skb.StringEncoder(),
            skb.TextEncoder(),
            skb.GapEncoder(),
        ]

encoder_distance = {}
encoder_norm = {}

for encoder in encoders:
    # Vectorize the data with the TableVectorizer
    tab_vec = skb.TableVectorizer(
            high_cardinality=encoder,
            numeric=StandardScaler())
    X = tab_vec.fit_transform(df)

    # compute the pairwise distance distribution
    # looking only at the data first hired, or at the
    # other columns
    X_year = X[['year_first_hired']]
    X_other = X.drop(columns='year_first_hired')
    year_dist = pairwise_distances(X_year, metric='euclidean')
    # keep only the lower diagonal
    year_dist = year_dist[*np.tril_indices_from(year_dist)]

    # same thing for the other columns
    other_dist = pairwise_distances(X_other, metric='euclidean')
    other_dist = other_dist[*np.tril_indices_from(other_dist)]

    encoder_distance[encoder.__class__.__name__] = other_dist

    # Look at the norm (easier to compute, because not quadratic)

    year_norm = np.sum((X_year - np.mean(X_year, axis=0))**2) / X.shape[0]
    X_other = X_other.values  # convert to numpy
    other_norm = np.sum((X_other - np.mean(X_other, axis=0))**2) / X.shape[0]
    encoder_norm[encoder.__class__.__name__] = other_norm

# %%
from matplotlib import pyplot as plt
m_dist = np.mean(year_dist)
[...]

PR #1274, opened 2025-04-03
Add L2 Block Normalization

Vincent-Maladiere:
Adresses https://github.com/skrub-data/skrub/issues/1253

Screenshot 2025-04-02 at 17 54 38

### DIFF
--- a/skrub/_gap_encoder.py
+++ b/skrub/_gap_encoder.py
@@ -1,6 +1,7 @@
 """
 Implements the GapEncoder: a probabilistic encoder for categorical variables.
 """
+
 from __future__ import annotations
 
 from copy import deepcopy
@@ -18,6 +19,7 @@ from sklearn.utils.extmath import row_norms, safe_sparse_dot
 from sklearn.utils.validation import _num_samples, check_is_fitted
 
 from . import _dataframe as sbd
+from ._block_normalizer import BlockNormalizerL2
 from ._on_each_column import RejectColumn, SingleColumnTransformer
 from ._utils import unique_strings
 
@@ -430,6 +432,10 @@ class GapEncoder(TransformerMixin, SingleColumnTransformer):
             raise ValueError("analyzer should be one of ['word', 'char', 'char_wb'].")
 
     def fit(self, X, y=None):
+        _ = self.fit_transform(X)
+        return self
+
+    def fit_transform(self, X, y=None):
         """
         Fit the GapEncoder on `X`.
 
@@ -467,7 +473,7 @@ class GapEncoder(TransformerMixin, SingleColumnTransformer):
         unq_X, unq_V, lookup = self._init_vars(X, is_null)
         n_batch = (len(X) - 1) // self.batch_size + 1
         n_samples = len(X)
-        del X
+
         # Get activations unq_H
         unq_H = self._get_H(unq_X)
         converged = False
@@ -516,7 +522,15 @@ class GapEncoder(TransformerMixin, SingleColumnTransformer):
 
         # Update self.H_dict_ with the learned encoded vectors (activations)
         self.H_dict_.update(zip(unq_X, unq_H))
-        return self
+
+        # Transform and normalize the output.
+        result = self._transform(X, is_null)
+
+        normalizer = BlockNormalizerL2()
+        result = normalizer.fit_transform(result)
+        self.normalizer_ = normalizer
+
+        return self._post_process(X, result)
 
     def get_feature_names_out(self, n_labels=3):
         """
@@ -723,10 +737,8 @@ class GapEncoder(TransformerMixin, SingleColumnTransformer):
         self._check_input_type(X, err_type=ValueError)
         is_null = sbd.to_numpy(sbd.is_null(X))
         result = self._transform(sbd.to_numpy(X), is_null)
-        names = self.get_feature_names_out()
-        result = sbd.make_dataframe_like(X, dict(zip(names, result.T)))
-        result = sbd.copy_index(X, result)
-        return result
+        result = self.normalizer_.transform(result)
+        return self._post_process(X, result)
 
     def _check_input_type(self, X, err_type=RejectColumn):
         if sbd.is_categorical(X) or sbd.is_string(X):
@@ -763,6 +775,12 @@ class GapEncoder(TransformerMixin, SingleColumnTransformer):
         result = self._get_H(X)
         # Restore H
         self.H_dict_ = pre_trans_H_dict_
+        return result
+
+    def _post_process(self, X, result):
+        names = self.get_feature_names_out()
+        result = sbd.make_dataframe_like(X, dict(zip(names, result.T)))
+        result = sbd.copy_index(X, result)
         return result
 
 
--- a/skrub/_block_normalizer.py
+++ b/skrub/_block_normalizer.py
@@ -1,0 +1,57 @@
+import numpy as np
+from sklearn.base import (
+    BaseEstimator,
+    OneToOneFeatureMixin,
+    TransformerMixin,
+    check_is_fitted,
+)
+
+from . import _dataframe as sbd
+
+
+class BlockNormalizerL2(OneToOneFeatureMixin, TransformerMixin, BaseEstimator):
+    def fit(self, X):
+        _ = self.fit_transform(X)
+        return self
+
+    def fit_transform(self, X):
+        self._check_all_numeric(X)
+
+        # Compute column-wise norm by filtering out nonfinite values.
+        X = self._validate_data(X=X, accept_sparse=False, force_all_finite=False)
+
+        self.avg_norm_ = avg_norm(X)
+
+        return X / self.avg_norm_
+
+    def transform(self, X):
+        check_is_fitted(self, "avg_norm_")
+        X = self._validate_data(
+            X=X, reset=False, accept_sparse=False, force_all_finite=False
+        )
+        return X / self.avg_norm_
+
+    def _check_all_numeric(self, X):
+        msg = "BlockNormalizer only accept numeric columns, but {col} is {dtype}."
+        if hasattr(X, "__dataframe__"):
+            for col in sbd.to_column_list(X):
+                if not (sbd.is_numeric(col) or sbd.is_bool(col)):
+                    raise ValueError(msg.format(col=col.name, dtype=sbd.dtype(col)))
+
+
+def avg_norm(X):
+    n_features = X.shape[1]
+    norm = 0
+    mask_finite = np.isfinite(X)
+    for jdx in range(n_features):
+        m = mask_finite[:, jdx]
+        x = X[:, jdx][m]
+        norm += ((x - x.mean()) ** 2).sum(axis=None)
+    n_sample_finite = mask_finite.sum() / n_features
+    norm /= n_sample_finite
+
+    # Avoid division by very small or zero values.
+    if norm < 10 * np.finfo(norm.dtype).eps:
+        norm = 1
+
+    return np.sqrt(norm)
--- a/skrub/_string_encoder.py
+++ b/skrub/_string_encoder.py
@@ -9,6 +9,7 @@ from sklearn.feature_extraction.text import (
 from sklearn.pipeline import Pipeline
 
 from . import _dataframe as sbd
+from ._block_normalizer import BlockNormalizerL2
 from ._on_each_column import SingleColumnTransformer
 
 
@@ -158,6 +159,10 @@ class StringEncoder(SingleColumnTransformer):
             result = result.copy()  # To avoid a reference to X_out
         del X_out  # optimize memory: we no longer need X_out
 
+        normalizer = BlockNormalizerL2()
+        result = normalizer.fit_transform(result)
+        self.normalizer_ = normalizer
+
         self._is_fitted = True
         self.n_components_ = result.shape[1]
 
@@ -191,6 +196,8 @@ class StringEncoder(SingleColumnTransformer):
             result = X_out[:, : self.n_components].toarray()
             result = result.copy()
         del X_out  # optimize memory: we no longer need X_out
+
+        result = self.normalizer_.transform(result)
 
         return self._post_process(X, result)
 
--- a/skrub/_text_encoder.py
+++ b/skrub/_text_encoder.py
@@ -9,6 +9,7 @@ from sklearn.decomposition import PCA
 from sklearn.utils.validation import check_is_fitted
 
 from . import _dataframe as sbd
+from ._block_normalizer import BlockNormalizerL2
 from ._on_each_column import RejectColumn, SingleColumnTransformer
 from ._utils import import_optional_dependency, unique_strings
 from .datasets._utils import get_data_dir
@@ -258,6 +259,10 @@ class TextEncoder(SingleColumnTransformer, TransformerMixin):
                 # number of dimensions of X_out.
                 X_out = X_out[:, : self.n_components]
 
+        normalizer = BlockNormalizerL2()
+        X_out = normalizer.fit_transform(X_out)
+        self.normalizer_ = normalizer
+
         self.n_components_ = X_out.shape[1]
 
         cols = self.get_feature_names_out()
@@ -293,6 +298,8 @@ class TextEncoder(SingleColumnTransformer, TransformerMixin):
             X_out = self.pca_.transform(X_out)
         elif self.n_components is not None:
             X_out = X_out[:, : self.n_components]
+
+        X_out = self.normalizer_.transform(X_out)
 
         cols = self.get_feature_names_out()
         X_out = sbd.make_dataframe_like(column, dict(zip(cols, X_out.T)))


Vincent-Maladiere:
This is now ready for review. Items to review/discuss:
- The name `BlockNormalizerL2` is not great; how should we name this class, @GaelVaroquaux?
- The `BlockNormalizerL2` class is public in our API, but we may not want to bother users with this. If we keep it public, should we illustrate this with an example? Like the plot from Gael above.
- The `TextEncoder`, `StringEncoder` and `GapEncoder` now enable this normalization by default.

Vincent-Maladiere:
Updates after IRL discussion with @GaelVaroquaux
- We don't want to introduce a normalizer object, even a private one
- We don't want to add new hyperparameters for encoders; normalization should be enforced
- We also don't want to replace `fit` with `fit_transform` in the `GapEncoder`. The issue is that we don't have access to the vectors at the end of `fit`, so we have to find some proxy, like the average sample length.

CHANGES.rst
```diff
@@ -37,6 +37,9 @@ Changes
 - The `packaging` dependency was removed.
   :pr:`1307` by :user:`Jovan Stojanovic <jovan-stojanovic>`
 
+- :class:`TextEncoder`, :class:`StringEncoder` and :class:`GapEncoder` outputs during
+  transform are now normalized by computing the total standard deviation during
```
jeromedockes, line 41: maybe be a bit more verbose to make it clearer that the whole thing is rescaled by a global constant, and this is not row or column normalization

skrub/_dataframe/_common.py
```diff
@@ -640,6 +640,8 @@ def is_numeric(col):
 
 @is_numeric.specialize("pandas", argument_type="Column")
 def _is_numeric_pandas(col):
+    if pd.api.types.is_object_dtype(col):
```
jeromedockes, line 643: why is this change necessary? in code that calls that function to find non-numeric columns for example to try converting them to numbers this could cause these columns to be missed? how does it relate to normalization?

jeromedockes: maybe that extra conversion could be done in total_std_norm to make sure we don't affect other code

skrub/_total_std_norm.py
```diff
@@ -0,0 +1,108 @@
+import numpy as np
+import sklearn
+from sklearn.utils.fixes import parse_version
+from sklearn.utils.validation import check_array
+
+from . import _dataframe as sbd
+
+
+def total_std_norm(X):
```
jeromedockes, line 9: what does std stand for

skrub/_total_std_norm.py
```diff
@@ -0,0 +1,108 @@
...
+    .. math::
+
+        \forall i,j, \quad \tilde{X}_{ij} = \frac{X_{ij}}{||X||_B}
+    """
+
+    if hasattr(X, "__dataframe__"):
+        msg = "X must only have numeric columns, but {col} is {dtype}."
+        for col in sbd.to_column_list(X):
+            if not (sbd.is_numeric(col) or sbd.is_bool(col)):
+                raise ValueError(msg.format(col=col.name, dtype=sbd.dtype(col)))
```
jeromedockes, line 52: in theory it should be `sbd.name(col)` . maybe a little simplification could be to remove the loop and use a selector like `~(s.boolean() | s.numeric())` but not super important

jeromedockes: BTW, as this is a private function applied to the output of skrub encoders, do we need these checks? I guess if the encoder outputs non-numeric values (which doesn't happen) there's not much the user can do about it anyway?

skrub/_total_std_norm.py
```diff
@@ -0,0 +1,108 @@
...
+        accept_sparse=False,
+        dtype="numeric",
+        ensure_2d=True,
+        copy=False,
+        **{all_finite_key: False},
+    )
+
+    # Sanity check: replace inf values with nan
+    mask_finite = np.isfinite(X)
+    X[~mask_finite] = np.nan
```
jeromedockes, line 81: if we're going to do this, it should be made very clear with a warning in the docstring that this function will replace inf with nan in the input array. I wouldn't expect a function that computes a norm to modify the array I give it. If we can avoid it at all that would be best

jeromedockes, line 81: if we're going to do this, it should be made very clear with a warning in the docstring that this function will replace inf with nan in the input array. I wouldn't expect a function that computes a norm to modify the array I give it. If we can avoid it at all that would be best

skrub/_total_std_norm.py
```diff
@@ -0,0 +1,108 @@
...
+
+    norm = np.sqrt(np.nanvar(X, ddof=0, axis=0).sum())
```
jeromedockes, line 83: do we have a reference we can point to, or can you elaborate on this value in a comment?

skrub/_total_std_norm.py
```diff
@@ -0,0 +1,108 @@
...
+    if norm < 10 * np.finfo(norm.dtype).eps:
+        norm = 1
+
+    return norm
+
+
+def accumulate_std_norm(X, mean1, var1, n1):
+    total_combined_var = 0
+    if X.ndim == 1:
+        X = X.reshape(-1, 1)
```
jeromedockes, line 95: as this is private, if in our use of this function in the encoders the ndim == 1 doesn't occur I think we can just not handle it (and the missed coverage line will go away)

skrub/_total_std_norm.py
jeromedockes: do we need to handle the dataframe case? if this is to be applied to numpy intermediate results inside the encoders, we could say that this function operates on numpy arrays. using it on dataframes should be avoided to avoid the copy into a numpy array

gap, text and string encoders do their computation in numpy, then compute the norm, then rescale and convert to dataframe. doing the rescaling after the first conversion to dataframe would add an unnecessary copy, no?

jeromedockes (commented):
thanks @Vincent-Maladiere ! this looks great , I would just like a bit more rationale / explanations on this rescaling in the doc or in a comment. 

should this also be done for the minhash encoder?

skrub/_dataframe/_common.py
Vincent-Maladiere: I had a test failing because of this, and it was unrelated to `total_std_norm`, still not sure why. We have a similar logic for booleans in `_is_bool_pandas`, why wouldn't we want a similar behavior for numeric?

skrub/_total_std_norm.py
Vincent-Maladiere: Yes we don't need these extra checks anymore, I initially wrote it to be a public class.

skrub/_total_std_norm.py
Vincent-Maladiere: I need to update the docstring below slightly

skrub/_total_std_norm.py
Vincent-Maladiere: As you said, this function is only meant to be used internally. Having encoders outputting `inf` is not very plausible anyway, so I'll remove this altogether.

skrub/_total_std_norm.py
Vincent-Maladiere: Not really, it's a Gael prototype.

Vincent-Maladiere:
@jeromedockes, notwithstanding the tests, are the documentation/comments better?

Vincent-Maladiere:
@glemaitre we have segfaults with sentence-transformers 🍻

skrub/_dataframe/_common.py
jeromedockes: ok maybe we want it I just want to investigate a bit more because it might affect other parts of skrub. (also it will typically cause a copy of the whole column, making what was a simple dtype check potentially expensive). for bool I guess the reason was booleans with None in an object dtype is how bools with missing values get represented

skrub/_total_std_norm.py
```diff
@@ -0,0 +1,100 @@
+import numpy as np
+
+
+def total_standard_deviation_norm(X):
```
jeromedockes, line 4: I think we should avoid talking about a "norm" as it isn't one, WDYT?

Vincent-Maladiere: I don't have a strong opinion, but on the surface it looks like a norm (e.g. the [RMS](https://pytorch.org/docs/stable/generated/torch.nn.RMSNorm.html#torch.nn.RMSNorm), with some differences: we apply the reduce on axis=0 instead of axis=1, and we sum instead of averaging). We can ask @GaelVaroquaux

Vincent-Maladiere:
@jeromedockes you were right, `sbd.is_numeric` is not raising issues anymore, so I reverted my change. Still not sure what happened 🤔

Vincent-Maladiere:
@jeromedockes, this is now ready for review :)

skrub/_total_std_norm.py
GaelVaroquaux: No strong opinion. But I like the term "scaler"

skrub/_total_std_scaler.py
```diff
@@ -0,0 +1,102 @@
...
+    of all columns of X.
+
+    Parameters
+    ----------
+    X : np.ndarray of shape (n_samples, n_features)
+        The matrix to compute the norm from.
+
+    Returns
+    -------
+    scaler : float,
```
GaelVaroquaux, line 21: I think that this should be called "scale" and not "scaler". "scaler" gives me the impression that the value is an object that performs the scaling.

I think that this modification of the name should also be applied to the name of the attribute when it is store on an object: "self.scale_" and not "self.scaler_", or maybe "self.scaling_factor_".

skrub/_gap_encoder.py
```diff
@@ -456,7 +479,14 @@ def fit(self, X, y=None):
         self._input_name = sbd.name(X)
         self._random_state = check_random_state(self.random_state)
         is_null = sbd.to_numpy(sbd.is_null(X))
-        X = sbd.to_numpy(X)
+        result = self._fit_transform(sbd.to_numpy(X), is_null)
+
+        self.scaler_ = total_standard_deviation_scaler(result)
```
GaelVaroquaux, line 484: To me, the name "self.scaler_" makes me think that it is a scaling object, ie a transformer, and not a float.

I would prefer this value to be called "self.scale_" or maybe "self.scaling_factor_".

skrub/_gap_encoder.py
```diff
@@ -456,7 +479,14 @@ def fit(self, X, y=None):
...
+        result /= self.scaler_
+
+        return self._post_process(X, result)
+
+    def _fit_transform(self, X, is_null):
```
GaelVaroquaux, line 489: I am confused: why do we need a "_fit_transform" method in addition to the "fit_transform" method?

Having multiple nested methods has two drawbacks. First it can make make reading the codebase harder. Second, it creates an extra frame in the stack and the risk is that this frame holds references on objects, thus increasing the memory usage.

skrub/_gap_encoder.py
```diff
@@ -430,8 +435,26 @@ def _check_analyzer(self):
...
+        y : None
+            Unused, only here for compatibility.
+
+        Returns
+        -------
+        GapEncoder
+            The fitted GapEncoder instance (self).
         """
-        Fit the GapEncoder on `X`.
+        _ = self.fit_transform(X)
```
GaelVaroquaux, line 452: I worry that calling fit_transform when we only need fit leads to additional costs, as it is possible to do "fit" without the transform step

skrub/_gap_encoder.py
```diff
@@ -765,6 +809,12 @@ def _transform(self, X, is_null):
         self.H_dict_ = pre_trans_H_dict_
         return result
 
+    def _post_process(self, X, result):
```
GaelVaroquaux, line 812: If factoring out a method does not lead to reusing it across multiple places, I would rather not factor out the code.

GaelVaroquaux (commented):
Thanks a lot for the PR and the rework based on my feedback.

I did a full review and have a few comments.

skrub/_gap_encoder.py
Vincent-Maladiere: We actually use it in `fit_transform` and `transform`

Vincent-Maladiere:
Here are the benchmarks; the GapEncoder is amazingly slow during `transform`. Therefore, we have to find a solution.

Screenshot 2025-05-26 at 15 01 54
Screenshot 2025-05-26 at 15 01 57

<details><summary>reproducer on the main branch</summary>

```python

# %%
# Benchmarking GapEncoder.fit vs transform on the main branch. We want to make sure that
# transform << fit so that using fit_transform for fit doesn't add a major computation
# bottleneck.
#
# We define a synthetic text data generate. We can vary:
# - the number of samples
# - the average sequence length
# - the number of unique sample
#
import pandas as pd
import numpy as np
from tqdm import tqdm
import skrub
from time import perf_counter
from collections import defaultdict
from matplotlib import pyplot as plt
from functools import cache
from itertools import product


@cache
def make_synthetic_text(n_samples=1000, avg_ch_size=20, unique_ratio=0.7):
    n_samples_new = int(n_samples * unique_ratio)
    max_seqlen = 500

    ch_ids = np.random.randint(ord("A"), ord("z"), size=(n_samples_new, max_seqlen))
    seq_size = np.random.normal(
        avg_ch_size, scale=avg_ch_size / 4, size=n_samples_new
    ).astype("int32")

    samples = []
    for idx in range(n_samples_new):
        samples.append("".join(map(chr, ch_ids[idx, : seq_size[idx]])))
    samples.extend(np.random.choice(samples, size=n_samples - n_samples_new))

    return pd.Series(samples, name="synthetic_text")


def run_bench(results, n_samples, encoder, avg_ch_size, unique_ratio):

    encoder = getattr(skrub, encoder)()

    samples = make_synthetic_text(n_samples, avg_ch_size, unique_ratio)

    tic = perf_counter()
    encoder.fit(samples)
    results["fit"].append(perf_counter() - tic)

    tic = perf_counter()
    encoder.transform(samples)
    results["transform"].append(perf_counter() - tic)


params = dict(
    encoders=["StringEncoder", "GapEncoder"],
    avg_ch_size=[20],
    unique_ratio=[.7],
)
all_params = list(product(*params.values()))
all_results = []
all_n_samples = 10 ** np.array([2, 3, 4, 5])
for param_set in all_params:
    results = defaultdict(list)
    for n_samples in tqdm(all_n_samples):
        print(param_set)
        run_bench(results, n_samples, *param_set)
    all_results.append(results)

# %%

log_n_samples = np.log10(all_n_samples).astype("int32")
for idx, (encoder, _, _) in enumerate(all_params):
    fig, ax = plt.subplots()
    for method in "fit", "transform":
        ax.plot(log_n_samples, all_results[idx][method], label=method)
    ax.set_xticks(
        ticks=log_n_samples, labels=[f"1e{p}" for p in log_n_samples]
    )
    ax.legend()
    ax.grid()
    ax.set_xlabel("samples")
    ax.set_ylabel("seconds")
    ax.set_title(encoder)
    plt.show()

# %%

```

</details>

Vincent-Maladiere:
I plotted the sequence length as produced by different analyzers against the row-wise L2 norm of the GapEncoder. Scaling these vectors by the average length seen during training could make sense.

Screenshot 2025-06-05 at 14 51 53

<details><summary>reproducer</summary>

```python
# Compute the text length for some columns.
# Plot the norm of the raw vectors vs the text length. 

import numpy as np
from matplotlib import pyplot as plt
import seaborn as sns
from sklearn.feature_extraction.text import CountVectorizer

from skrub import GapEncoder


X = fetch_employee_salaries().X
categ_cols = ['employee_position_title', 'division', 'department_name']

for i, analyzer in enumerate(["char", "char_wb", "word"]):
    fig, axes = plt.subplots(ncols=3, figsize=(6, 3))
    for j, col in enumerate(categ_cols):
        gap = GapEncoder(analyzer=analyzer)
        Xt = gap.fit_transform(X[col].head(100))
        l2_norm = np.linalg.norm(Xt, axis=1)
        n_tokens = np.array(
            CountVectorizer(analyzer=analyzer)
            .fit_transform(X[col].head(100))
            .sum(axis=1)
        ).ravel()

        ax = axes[j]
        sns.scatterplot(x=n_tokens, y=l2_norm, ax=ax, color=["blue", "red", "green"][i])
        sns.lineplot(x=n_tokens, y=n_tokens, ax=ax, linestyle="--", color="grey", alpha=.4)

        ax.set_title(col)
        if i == 0 and j == 0:
            ax.set_ylabel('L2 Norm of Encoded Vectors')
            ax.set_xlabel('Text Length')
    fig.suptitle(analyzer)
    fig.tight_layout()
```

</details>

GaelVaroquaux:
> Scaling these vectors by the average length seen during training could make sense

Great! It seems like a very pragmatic solution and I really like what I see.

Thanks for finding this!!

Vincent-Maladiere:
These experiments are invariant with the number of components!

When `n_components=20`

Screenshot 2025-06-05 at 15 15 28

When `n_components=100`

Screenshot 2025-06-05 at 15 16 25

GaelVaroquaux:
Nice! Let's see what this gives on the histograms

Vincent-Maladiere:
I ended-up with this heuristics:

```
scaling_factor = unq_V.sum(axis=1).mean() / 2
```

which works reasonably well:

`employee_position_title` column

Screenshot 2025-06-05 at 16 05 25

toxicity (longer strings):

Screenshot 2025-06-05 at 16 04 49

GaelVaroquaux:
Yes, the histograms show that it works well. Let's go with that!

skrub/_gap_encoder.py
```diff
@@ -467,7 +467,7 @@ def fit(self, X, y=None):
         unq_X, unq_V, lookup = self._init_vars(X, is_null)
         n_batch = (len(X) - 1) // self.batch_size + 1
         n_samples = len(X)
-        del X
```
GaelVaroquaux, line 470: Can we not get this "del X" back?

It would help decreasing memory pressure

skrub/_gap_encoder.py
```diff
@@ -687,6 +690,12 @@ def partial_fit(self, X, y=None):
         )
         # Update self.H_dict_ with the learned encoded vectors (activations)
         self.H_dict_.update(zip(unq_X, unq_H))
+
+        # XXX: Maybe we should accumulate unq_V or the scaling factor somehow?
+        scaling_factor = _gap_scaling_factor(unq_V)
+        if not hasattr(self, "scaling_factor_") or scaling_factor != 0:
+            self.scaling_factor_ = scaling_factor
```
GaelVaroquaux:
```suggestion
        if not hasattr(self, "scaling_factor_"):
            scaling_factor = _gap_scaling_factor(unq_V)
            if scaling_factor != 0:
                self.scaling_factor_ = scaling_factor
```
Let us not compute the scaling factor if we are not using it

skrub/_scaling_factor.py
```diff
@@ -0,0 +1,106 @@
+import numpy as np
```
GaelVaroquaux, line 6: It's always good to have a top-level docstring for a file that describes briefly what the file does. It helps people reading the codebase

GaelVaroquaux (changes requested):
A few tiny comments and we are good to go

skrub/_gap_encoder.py
Vincent-Maladiere: Indeed, I forgot

GaelVaroquaux:
Ping me when the comments are addressed, and I merge

Vincent-Maladiere:
They are! Could you look at the file-level docstring of _scaling_factor.py?

GaelVaroquaux (approved):
LGTM.

Thank you very much!!

GaelVaroquaux:
Merged!!

Vincent-Maladiere:
Wuhuuu!!

PR #1275, opened 2025-04-03
Changing the name of SimpleCleaner to Cleaner

rcap107:
Refactoring first, now I need to add the deprecation warning and add back SimpleCleaner as a subclass of Cleaner

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,10 +42,10 @@ New features
   :pr:`1203` by :user:`Reshama Shaikh <reshamas>` and
   :user:`Vincent Maladiere <Vincent-Maladiere>`.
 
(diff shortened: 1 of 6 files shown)


CHANGES.rst
```diff
@@ -14,6 +14,8 @@ New features
 
 Changes
 -------
+- The :class:`SimpleCleaner` has been deprecated and will be renamed to `Cleaner`.
```
jeromedockes, line 17:
```suggestion
- The :class:`SimpleCleaner` has been renamed to `Cleaner`. Use of the name `SimpleCleaner` is deprecated and will result in an error in some future release of skrub.
```

PR #1283, opened 2025-04-06
Fix rst in docstrings

GaelVaroquaux:
Fix formatting problems: https://skrub-data.org/stable/reference/generated/skrub.column_associations.html#skrub.column_associations


Vincent-Maladiere (approved):
Looks good, thanks @GaelVaroquaux!

PR #1303, opened 2025-04-11
Added link to Learning Materials

gabrielapgomezji:


skrub/_expressions/_skrub_namespace.py
```diff
@@ -1387,7 +1390,7 @@ def cross_validate(self, environment=None, **kwargs):
         if environment is None:
             environment = self.get_data()
 
-        return cross_validate(self.get_estimator(), environment, **kwargs)
+        return pd.DataFrame(cross_validate(self.get_estimator(), environment, **kwargs))
```
rcap107, line 1393: This change is tripping tests and I think is already implemented in #1315, so it should probably be removed and done in the other PR

PR #1304, opened 2025-04-11
Skipping associations compute if the number of cols > specified

### LINKED ISSUE #1287: Avoid computing column associations when the number of columns is large
### Problem Description

Computing associations is a very expensive operation that does not scale well then the number of columns becomes very large, so there should be a parameter that disables associations past a given number of columns.

### Feature Description

This feature would be very similar to the current parameter that prevents plotting distributions. 

### Alternative Solutions

_No response_

### Additional Context

_No response_

victoris93:
Closes #1287.

I made changes to skip assocations computation. Haven't  modified to the tests yet, it's coming right up.
It would make sense to remove the `Associations` tab from the html table altogether.


victoris93:
Final changes:
* `max_association_columns = 30`
* If associations are not computed (`n_cols > max_association_columns`), the Associations tab is excluded from the report

Vincent-Maladiere (commented):
Hi @victoris93, thank you for this PR!

@jeromedockes, we probably should have a single parameter to control both         `max_plot_columns` and `max_association_columns`, WDYT? We can do it in another PR, though.

victoris93:
@Vincent-Maladiere we briefly touched upon this in our last conversation w/ @jeromedockes; imo it's an elegant simplification but from the user standpoint it's a big assumption that they would always prefer to avoid both the plotting and the association computation and for the same n. of columns. But totally your call ofc.

jeromedockes:
I don't have a strong opinion. @victoris93 is right that it's an assumption that a user would want to tie the 2 together. OTOH my guess would be that most users will leave those parameters to the default value all the time :thinking:

skrub/_reporting/_table_report.py
```diff
@@ -134,31 +140,38 @@ def __init__(
...
+        else:
+            with_associations = False
 
-    @functools.cached_property
-    def _summary_without_plots(self):
         return summarize_dataframe(
-            self.dataframe, with_plots=False, title=self.title, **self._summary_kwargs
+            self.dataframe,
+            with_plots=with_plots,
+            with_associations=with_associations,
```
Vincent-Maladiere, line 168: We could simplify the lines above like:
```suggestion
            with_plots=self.max_plot_columns >= self.n_columns,
            with_associations=self.max_association_columns >= self.n_columns,
```

Vincent-Maladiere (commented):
I agree that very few users will actually tweak these parameters. That said, having both parameters could make sense for users who, for example, have many correlated columns but would rather not display the distributions of all of them. Therefore we can keep things as they are :)

skrub/_reporting/_table_report.py
```diff
@@ -134,31 +140,38 @@ def __init__(
         self.dataframe = dataframe
         self.verbose = verbose
         self.max_plot_columns = max_plot_columns
+        self.max_association_columns = max_association_columns
+        self.n_columns = sbd.shape(self.dataframe)[1]
+
+        if self.max_plot_columns is None:
+            self.max_plot_columns = self.n_columns
+        if self.max_association_columns is None:
+            self.max_association_columns = self.n_columns
```
jeromedockes, line 149: I would rather still store them as `None` so it is easy to check if the report was created with max_plot_columns set to None or to the number of columns in the dataframe

skrub/_reporting/_table_report.py
```diff
@@ -134,31 +140,38 @@ def __init__(
         self.dataframe = dataframe
         self.verbose = verbose
         self.max_plot_columns = max_plot_columns
+        self.max_association_columns = max_association_columns
+        self.n_columns = sbd.shape(self.dataframe)[1]
+
+        if self.max_plot_columns is None:
```
victoris93, line 146: Not sure I understand the comment about None value handling. Would you like None values to be handled differently than what I have here?

skrub/_reporting/_table_report.py
```diff
@@ -134,31 +140,38 @@ def __init__(
...
 
     def __repr__(self):
         return f"<{self.__class__.__name__}: use .open() to display>"
 
     @functools.cached_property
-    def _summary_with_plots(self):
-        return summarize_dataframe(
-            self.dataframe, with_plots=True, title=self.title, **self._summary_kwargs
-        )
+    def _lightify_summary(self):
```
jeromedockes, line 155: the reason why we had 2 cached properties is that the json does not need the plots so we might want either a full summary or a plot-less one, and both would be cached.

Now if we don't have just 2 kinds of report but several parameters we should just give up on that it wasn't really necessary, and more a relic from an early version where the tablereport was also able to produce output in text format.

We can have just

```
@functools.cached_property
def _summary(self):
    n_columns = sbd.shape(self.dataframe)[1]
    with_plots = self.max_plot_columns is None or self.max_plot_columns >= n_columns
    with_associations = ...
    return summarize_dataframe(self.dataframe, with_plots=with_plots, ...)
```

(and no more `_lightify_summary` nor `_get_summary` )

skrub/_reporting/_table_report.py
Vincent-Maladiere: It's almost nothing, having something like:
```python
self.max_plot_columns = self.n_columns if max_plot_columns is None else max_plot_columns
```

Vincent-Maladiere: Ah to follow @jeromedockes suggestion, we could have `self.max_plot_columns = max_plot_columns`, and then derive `self.max_plot_columns_` using your logic. `self.max_plot_columns_` would be the one actually used. I won't comment further as this gets too nitpicky

jeromedockes:
I think the review comments (at least mine) are getting unclear :sweat_smile:  sorry about that @victoris93 . As the PR is working fine I suggest we merge it and any further changes can be done in another PR, WDYT?

@victoris93 was there anything you still wanted to change before I merge it?

victoris93:
@jeromedockes np pb, I think I got it. Just a sec, the last commit should address the comments, almost there

jeromedockes:
ok thanks! If you do get rid of `_get_summary()` you will need to replace `_get_summary()` with `_summary` everywhere in `_reporting/tests/test_table_report.py`

victoris93:
@jeromedockes @Vincent-Maladiere should be good now I think

skrub/_reporting/_table_report.py
```diff
@@ -134,32 +140,31 @@ def __init__(
...
+        self.n_columns = sbd.shape(self.dataframe)[1]
+
+        self.max_plot_columns_ = (
+            self.n_columns if max_plot_columns is None else max_plot_columns
+        )
+        self.max_association_columns_ = (
+            self.n_columns
+            if max_association_columns is None
+            else max_association_columns
+        )
```
jeromedockes, line 153:
```suggestion
```

skrub/_reporting/_table_report.py
```diff
@@ -134,32 +140,31 @@ def __init__(
...
 
     def __repr__(self):
         return f"<{self.__class__.__name__}: use .open() to display>"
 
     @functools.cached_property
-    def _summary_with_plots(self):
+    def _summary(self):
```
jeromedockes, line 159:
```suggestion
    def _summary(self):
        with_plots = (
            self.max_plot_columns is None or self.max_plot_columns >= self.n_columns
        )
        with_associations = (
            self.max_association_columns is None
            or self.max_association_columns >= self.n_columns
        )

```

jeromedockes (commented):
thanks a lot! LGTM with the suggested changes

skrub/_reporting/_table_report.py
```diff
@@ -134,32 +140,31 @@ def __init__(
...
         return summarize_dataframe(
-            self.dataframe, with_plots=True, title=self.title, **self._summary_kwargs
+            self.dataframe,
+            with_plots=self.max_plot_columns_ >= self.n_columns,
+            with_associations=self.max_association_columns_ >= self.n_columns,
+            title=self.title,
+            **self._summary_kwargs,
```
jeromedockes, line 165:
```suggestion
            self.dataframe,
            with_plots=with_plots,
            with_associations=with_associations,
            title=self.title,
            **self._summary_kwargs,
```

skrub/_reporting/_table_report.py
```diff
@@ -134,32 +140,31 @@ def __init__(
         self.dataframe = dataframe
         self.verbose = verbose
         self.max_plot_columns = max_plot_columns
+        self.max_association_columns = max_association_columns
+        self.n_columns = sbd.shape(self.dataframe)[1]
+
```
jeromedockes, line 145:
```suggestion
```

jeromedockes:
> oops I wanted to run the pre-commit checks before you merged but it was too late, sry about that. I hope it's ok.

there was just an extra blank line but I removed it before merging; the pre-commit checks are passing without complaints :)  thanks again, and whenever you have time don't hesitate to ping us if you want help in picking the next thing to work on!

GaelVaroquaux:
Thanks Victoria!!

PR #1310, opened 2025-04-11
Implemented adaptive squashing

dholzmueller:
WIP

### DIFF
--- a/skrub/adaptive_squashing.py
+++ b/skrub/adaptive_squashing.py
@@ -1,0 +1,132 @@
+import numbers
+
+import numpy as np
+
+from . import _dataframe as sbd
+from ._on_each_column import RejectColumn, SingleColumnTransformer
+
+
+class AdaptiveSquashingTransformer(SingleColumnTransformer):
+    def __init__(
+        self,
+        squash_threshold=3.0,
+        lower_quantile=0.25,
+        upper_quantile=0.75,
+    ):
+        self.squash_threshold = squash_threshold
+        self.lower_quantile = lower_quantile
+        self.upper_quantile = upper_quantile
+
+    def fit_transform(self, X, y=None):
+        """Fit the transformer and transform a column.
+
+        Parameters
+        ----------
+        X : Pandas or Polars series
+            The column to transform.
+        y : None
+            Unused. Here for compatibility with scikit-learn.
+
+        Returns
+        -------
+        X_out: Pandas or Polars series with shape (len(X),)
+            The transformed version of the input.
+        """
+        del y
+
+        if not isinstance(self.squash_threshold, numbers.Number):
+            raise ValueError(
+                "squash_threshold needs to be a number, but got"
+                f" type(squash_threshold)={type(self.squash_threshold)}"
+            )
+        if not isinstance(self.lower_quantile, numbers.Number):
+            raise ValueError(
+                "lower_quantile needs to be a number, but got"
+                f" type(lower_quantile)={type(self.lower_quantile)}"
+            )
+        if not isinstance(self.upper_quantile, numbers.Number):
+            raise ValueError(
+                "upper_quantile needs to be a number, but got"
+                f" type(upper_quantile)={type(self.upper_quantile)}"
+            )
+        if self.squash_threshold <= 0 or not np.isfinite(self.squash_threshold):
+            raise ValueError(
+                "squash_threshold needs to be a positive finite float, but got"
+                f" {self.squash_threshold}"
+            )
+        if not (0 <= self.lower_quantile < self.upper_quantile <= 1):
+            raise ValueError(
+                "need 0 <= lower_quantile < upper_quantile <= 1, but got"
+                f" lower_quantile={self.lower_quantile} and"
+                f" upper_quantile={self.upper_quantile}"
+            )
+
+        self.col_name_ = sbd.name(X)
+
+        if not sbd.is_numeric(X):
+            raise RejectColumn(f"Column {self.col_name_} is not numeric.")
+
+        if not self.col_name_:
+            self.col_name_ = "adaptive_squashed"
+
+        eps = 1e-30  # todo: expose?
+
+        values = sbd.to_numpy(X).astype(np.float32)
+        finite_values = values[np.isfinite(values)]
+        if len(finite_values) > 0:
+            self.median_ = np.median(finite_values)
+            quantiles = [
+                np.quantile(finite_values, q)
+                for q in [self.lower_quantile, self.upper_quantile]
+            ]
+            if quantiles[1] == quantiles[0]:
+                min = np.min(finite_values)
+                max = np.max(finite_values)
+                if min == max:
+                    self.scale_ = 0
+                else:
+                    self.scale_ = 2.0 / (max - min + eps)
+            else:
+                self.scale_ = 1.0 / (quantiles[1] - quantiles[0] + eps)
+        else:
+            self.median_ = 0.0
+            self.scale_ = 1.0
+
+        self._is_fitted = True
+        self.n_components_ = 1
+
+        return self.transform(X)
+
+    def transform(self, X):
+        """Transform a column.
+
+        Parameters
+        ----------
+        X : Pandas or Polars series
+            The column to transform.
+
+        Returns
+        -------
+        result: Pandas or Polars column with shape (len(X),)
+            The scaled and squashed representation of the input.
+        """
+
+        values_np = sbd.to_numpy(X).astype(np.float32)
+        result = np.copy(values_np)
+        isfinite = np.isfinite(values_np)
+        scaled_finite = self.scale_ * (values_np[isfinite] - self.median_)
+        result[isfinite] = scaled_finite / np.sqrt(
+            1 + (scaled_finite / self.squash_threshold) ** 2
+        )
+        isinf = np.isinf(values_np)
+        result[isinf] = np.sign(values_np[isinf]) * self.squash_threshold
+
+        # this will use the column name of the column that was passed in fit_transform,
+        # not the name of X
+        return self._post_process(X, result, self.col_name_)
+
+    def _post_process(self, X, result, name):
+        result = sbd.make_column_like(X, result, name)
+        result = sbd.copy_index(X, result)
+
+        return result
--- a/skrub/tests/test_adaptive_squashing.py
+++ b/skrub/tests/test_adaptive_squashing.py
@@ -1,0 +1,205 @@
+import warnings
+
+import numpy as np
+import pytest
+
+from skrub import _dataframe as sbd
+from skrub._dataframe._common import _set_index
+from skrub._on_each_column import RejectColumn
+from skrub.adaptive_squashing import AdaptiveSquashingTransformer
+
+
+@pytest.mark.parametrize(
+    "values",
+    [
+        [0.6 * i for i in range(20)],
+        [0.0] * 10 + [2.0],
+        [2.0, -1.0] + [0.0] * 10 + [-1.0, 1.0],
+        [],
+    ],
+)
+@pytest.mark.parametrize("dtype", [np.float32, np.float64, np.int32])
+@pytest.mark.parametrize(
+    "params",
+    [
+        dict(squash_threshold=3.0, lower_quantile=0.25, upper_quantile=0.75),
+        dict(squash_threshold=5.0, lower_quantile=0.1, upper_quantile=0.6),
+    ],
+)
+def test_adaptive_squashing_output(df_module, values, dtype, params):
+    with warnings.catch_warnings():
+        # this is to filter the warning created by df_module.make_column
+        # due to pd.Series.convert_dtypes()
+        warnings.filterwarnings(
+            "ignore",
+            category=RuntimeWarning,
+            message="invalid value encountered in cast",
+        )
+        n_train = len(values) // 2
+
+        index = np.arange(len(values))
+
+        X = {"train": values[:n_train], "test": values[n_train:]}
+        X_df = dict()
+        indexes = {"train": index[:n_train], "test": index[n_train:]}
+
+        for part in ["train", "test"]:
+            if dtype in [np.float32, np.float64]:
+                X[part] = X[part] + [np.nan, np.inf, -np.inf]
+
+            X[part] = np.asarray(X[part], dtype=dtype)
+            name = part
+            X_df[part] = df_module.make_column(name, X[part])
+            # try to have different indexes for train and test
+            indexes[part] = np.arange(len(X_df[part])) + (
+                100_000 if part == "test" else 0
+            )
+            X_df[part] = _set_index(X_df[part], indexes[part])
+
+        tfm = AdaptiveSquashingTransformer(**params)
+        ft_out = tfm.fit_transform(X_df["train"])
+
+        t_out = tfm.transform(X_df["test"])
+        tfm_out = {"train": ft_out, "test": t_out}
+
+        eps = 1e-30
+        finite_values = X["train"][np.isfinite(X["train"])].astype(np.float32)
+        if len(finite_values) > 0:
+            median = np.median(finite_values)
+            quantiles = [
+                np.quantile(finite_values, q)
+                for q in [params["lower_quantile"], params["upper_quantile"]]
+            ]
+            if quantiles[1] == quantiles[0]:
+                min = np.min(finite_values)
+                max = np.max(finite_values)
+                if min == max:
+                    scale = 0
+                else:
+                    scale = 2.0 / (max - min + eps)
+            else:
+                scale = 1.0 / (quantiles[1] - quantiles[0] + eps)
+        else:
+            median = 0.0
+            scale = 1.0
+
+        for part in ["train", "test"]:
+            squash_threshold = params["squash_threshold"]
+            values_np = X[part].astype(np.float32)
+            result = np.copy(values_np)
+            isfinite = np.isfinite(values_np)
+            scaled_finite = scale * (values_np[isfinite] - median)
+            result[isfinite] = scaled_finite / np.sqrt(
+                1 + (scaled_finite / squash_threshold) ** 2
+            )
+            isinf = np.isinf(values_np)
+            result[isinf] = np.sign(values_np[isinf]) * squash_threshold
+
+            if np.any(np.abs(result) > params["squash_threshold"] + 1e-8):
+                raise RuntimeError("Test target should not exceed squash_threshold")
+
+            # todo: we test for name='train' in both cases here. Is this desired?
+            result_df = sbd.to_float32(
+                _set_index(df_module.make_column("train", result), indexes[part])
+            )
+            df_module.assert_column_equal(result_df, tfm_out[part])
+
+
+@pytest.mark.parametrize(
+    "config",
+    [
+        (
+            "number",
+            dict(squash_threshold=None, lower_quantile=0.25, upper_quantile=0.75),
+        ),
+        (
+            "positive",
+            dict(squash_threshold=0.0, lower_quantile=0.25, upper_quantile=0.75),
+        ),
+        (
+            "finite",
+            dict(squash_threshold=np.inf, lower_quantile=0.25, upper_quantile=0.75),
+        ),
+        (
+            "number",
+            dict(squash_threshold=3.0, lower_quantile="0.25", upper_quantile=0.75),
+        ),
+        (
+            "number",
+            dict(squash_threshold=3.0, lower_quantile=0.25, upper_quantile="0.75"),
+        ),
+        (
+            "number",
+            dict(squash_threshold=3.0, lower_quantile=0.25, upper_quantile="0.75"),
+        ),
+        ("need", dict(squash_threshold=3.0, lower_quantile=-0.2, upper_quantile=0.75)),
+        ("need", dict(squash_threshold=3.0, lower_quantile=0.25, upper_quantile=1.1)),
+        ("need", dict(squash_threshold=3.0, lower_quantile=0.8, upper_quantile=0.75)),
+    ],
+)
+def test_adaptive_squashing_error_msgs(df_module, config):
+    with warnings.catch_warnings():
+        # this is to filter the warning created by df_module.make_column
+        # due to pd.Series.convert_dtypes()
+        warnings.filterwarnings(
+            "ignore",
+            category=RuntimeWarning,
+            message="invalid value encountered in cast",
+        )
+        msg_match, params = config
+        X = df_module.make_column("col", [-1.0, 0.0, 1.0])
+        tfm = AdaptiveSquashingTransformer(**params)
+        with pytest.raises(ValueError, match=msg_match):
+            tfm.fit_transform(X)
+
+
+def test_adaptive_squashing_no_col_name(df_module):
+    with warnings.catch_warnings():
+        # this is to filter the warning created by df_module.make_column
+        # due to pd.Series.convert_dtypes()
+        warnings.filterwarnings(
+            "ignore",
+            category=RuntimeWarning,
+            message="invalid value encountered in cast",
+        )
+        X = df_module.make_column(None, [-1.0, 0.0, 1.0])
+        tfm = AdaptiveSquashingTransformer()
+        X_out = tfm.fit_transform(X)
+        if sbd.name(X_out) is None:
+            raise RuntimeError("Column name None did not get converted")
+
+
+def test_adaptive_squashing_non_numeric(df_module):
+    with warnings.catch_warnings():
+        # this is to filter the warning created by df_module.make_column
+        # due to pd.Series.convert_dtypes()
+        warnings.filterwarnings(
+            "ignore",
+            category=RuntimeWarning,
+            message="invalid value encountered in cast",
+        )
+        X = df_module.make_column(None, ["a", "b"])
+        tfm = AdaptiveSquashingTransformer()
+        with pytest.raises(RejectColumn, match="not numeric"):
+            tfm.fit_transform(X)
+
+
+def test_adaptive_squashing_known_values(df_module):
+    with warnings.catch_warnings():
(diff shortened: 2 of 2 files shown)


dholzmueller:
Here is an implementation of the robust scaling + smooth clipping from my paper (https://arxiv.org/abs/2407.04491). It could be used as a robust numerical preprocessing methods for neural network based methods.

Some considerations:
- which name do we want to use for the class? Currently it is AdaptiveSquashingTransformer.
- I did not check the rendering of the documentation. Do we need to link the class from somewhere in the documentation?
- If the column passed in transform() has a different name than in fit_transform(), the output column still gets the name of the column passed in fit_transform(). I copied this behavior from StringEncoder, is that good? (See also the todo in the corresponding tests.)

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,196 @@
...
+            The column to transform.
+
+        Returns
+        -------
+        result: Pandas or Polars column with shape (len(X),)
+            The scaled and squashed representation of the input.
+        """
+
+        values_np = sbd.to_numpy(X).astype(np.float32)
+        result = np.copy(values_np)
```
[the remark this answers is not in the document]
dholzmueller: Are you sure that it is a copy of X? to_numpy() in pandas does not guarantee to make a copy unless you set copy=True, which is not set as far as I can see. Also `del X` is not possible because `X` is used afterwards.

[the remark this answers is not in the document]
jeromedockes: I think that in `np.ndarray.astype` the default for `copy` is `True`

`del X` probably wouldn't change memory usage because the caller likely holds a reference to X as well. I'm guessing the `del y` used elsewhere in this PR is not to decrease the reference count but to make it clear that y is intentionally not used

dholzmueller:
> Also, I think having a smallish example that shows when and where to use your class would be great. WDYT?

Sounds good, where would you put it?

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,196 @@
...
+        to determine the scaling factor to rescale the data.
+    upper_quantile_alpha : float, default=0.75
+        Value of the upper quantile that will be used
+        to determine the scaling factor to rescale the data.
+
+    References
+    ----------
+    For a detailed description of the method, see
+    `Better by default: Strong pre-tuned MLPs and boosted trees on tabular data
+    <https://arxiv.org/abs/2407.04491>`_ by Holzmüller, Grinsztajn, Steinwart (2024).
```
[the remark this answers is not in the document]
dholzmueller: Yep, I added one.

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,196 @@
...
+                f" upper_quantile_alpha={self.upper_quantile_alpha}"
+            )
+
+        self.col_name_ = sbd.name(X)
+
+        if not sbd.is_numeric(X):
+            raise RejectColumn(f"Column {self.col_name_} is not numeric.")
+
+        if not self.col_name_:
+            self.col_name_ = "adaptive_squashed"
```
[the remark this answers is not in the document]
dholzmueller: done

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,196 @@
...
+     for edge cases in which the two quantiles are equal.
+    2) apply a soft-clipping function to limit the data
+     to the interval [-max_absolute_value, max_absolute_value]
+     in an injective way.
+
+    Infinite values will be mapped to the corresponding boundaries of the interval.
+    NaN values will be preserved.
+
+    The output feature will be named ``{col_name}`` if the series has a name,
+    and ``adaptive_squashed`` if it does not.
```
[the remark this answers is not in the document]
dholzmueller: ok, removed it

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,196 @@
...
+
+class AdaptiveSquashingTransformer(SingleColumnTransformer):
+    """Preprocess a numerical column by robust centering, scaling, \
+    and soft clipping to a defined interval.
+
+    This class implements the robust scaling and smooth clipping transformation
+    proposed for tabular neural networks by
+    David Holzmüller, Léo Grinsztajn, and Ingo Steinwart,
+    Better by default: Strong pre-tuned MLPs and boosted trees on tabular data,
+    Advances in Neural Information Processing Systems, 2024.
```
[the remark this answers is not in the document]
dholzmueller: done

dholzmueller:
I addressed everything except from the example now.

dholzmueller:
I removed the test that tested if a column with name `None` after transformation still had name `None`, since it failed for polars. I assume this is not relevant?

CHANGES.rst
```diff
@@ -27,6 +27,12 @@ Lemaitre <glemaitre>`, :user:`Adrin Jalali <adrinjalali>`, :user:`Olivier Grisel
 <ogrisel>` and others participated through many discussions in defining the
 requirements and the public API.
 
+- Added the `AdaptiveSquashingTransformer` that
```
jeromedockes, line 30: I think there is a formatting issue:

screenshot_2025-04-15T13:32:29+02:00


the lines below the `-` need to be indented to still be part of that bullet point, and the last line is too indented which turns it into a block quote. yes, restructured text is an acquired taste. a couple of things that can help:

- https://docutils.sourceforge.io/docs/user/rst/quickref.html
- in the pr page in the checks at the bottom click on "check the rendered docs here!"
- or locally go in the `/doc/` folder and there `make html-noplot` to render the doc without actually running the examples. you may need to first `pip install ".[doc]"`

jeromedockes: also you can do 
```
:class:`AdaptiveSquashingTransformer`
```

that will turn it into a link to the documentation

doc/reference/index.rst
```diff
@@ -210,6 +210,7 @@ Cleaning a dataframe
    :nosignatures:
 
    deduplicate
+   AdaptiveSquashingTransformer
```
jeromedockes, line 213: how about `HolzmüllerTransformer` or `AdaptiveScaler`

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,207 @@
...
+       in an injective way.
+
+    Infinite values will be mapped to the corresponding boundaries of the interval.
+    NaN values will be preserved.
+
+    Parameters
+    ----------
+    max_absolute_value : float, default=3.0
+        Maximum absolute value that the transformed data can take.
+    lower_quantile_alpha : float, default=0.25
```
jeromedockes, line 31: could it be `lower_quantile`?

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,207 @@
...
+
+        Returns
+        -------
+        X_out: Pandas or Polars series with shape (len(X),)
+            The transformed version of the input.
+        """
+        del y
+
+        if not isinstance(self.max_absolute_value, numbers.Number):
+            raise ValueError(
```
jeromedockes, line 117: I guess a TypeError would be more appropriate if it is the wrong type

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,207 @@
...
+                f" lower_quantile_alpha={self.lower_quantile_alpha} and"
+                f" upper_quantile_alpha={self.upper_quantile_alpha}"
+            )
+
+        self.col_name_ = sbd.name(X)
+
+        if not sbd.is_numeric(X):
+            raise RejectColumn(f"Column {self.col_name_} is not numeric.")
+
+        eps = 1e-30  # todo: expose?
```
jeromedockes, line 148:
```suggestion
        eps = 1e-30
```

no one will set it so we shouldn't expose it but just make sure we picked a good value. if needed you can use stuff like `np.finfo('float32').tiny`

jeromedockes (commented):
a first few comments, I'll add more. thanks this is great!! :rocket:

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,207 @@
...
+
+        values = sbd.to_numpy(X).astype(np.float32)
+        finite_values = values[np.isfinite(values)]
+        if len(finite_values) > 0:
+            self.median_ = np.median(finite_values)
+            quantiles = [
+                np.quantile(finite_values, q)
```
jeromedockes, line 155: not important but wouldn't `np.quantile(finite_values, [...])` work? (you can ask for multiple quantiles in one call)

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,207 @@
...
+                else:
+                    self.scale_ = 2.0 / (max - min + eps)
+            else:
+                self.scale_ = 1.0 / (quantiles[1] - quantiles[0] + eps)
+        else:
+            self.median_ = 0.0
+            self.scale_ = 1.0
+
+        self._is_fitted = True
+        self.n_components_ = 1
```
jeromedockes, line 172:
```suggestion
```

I don't think we need those

skrub/_adaptive_squashing.py
```diff
@@ -0,0 +1,207 @@
...
+            1 + (scaled_finite / self.max_absolute_value) ** 2
+        )
+        isinf = np.isinf(values_np)
+        values_np[isinf] = np.sign(values_np[isinf]) * self.max_absolute_value
+
+        # this will use the column name of the column that was passed in fit_transform,
+        # not the name of X
+        return self._post_process(X, values_np, self.col_name_)
+
+    def _post_process(self, X, result, name):
```
jeromedockes, line 203: if this function is only called in 1 place we can just inline those 2 lines and not define a function for it

skrub/_adaptive_squashing.py
dholzmueller: This was my first draft, but I wanted to distinguish between the quantile (value in \bbR) and the probability level (in [0, 1]) corresponding to the quantile (could be called quantile_level or quantile_alpha).

Vincent-Maladiere:
> Sounds good, where would you put it?

@dholzmueller, good question, I'm not sure. What would be the best way to illustrate its usefulness? Do you have a small benchmark or experiment that could fit into an example? Otherwise we might need to be creative

skrub/_adaptive_squashing.py
jeromedockes: that makes sense. OTOH it wouldn't make sense to pass the actual quantile value as an argument so people can probably guess what it means. or this parameter is called `q` in numpy and pandas, maybe this could be an option too

doc/reference/index.rst
dholzmueller: Well, I don't think users would like to write an ü. AdaptiveScaler is missing the squashing/clipping aspect. What about SquashScaler/ScaleSquasher?

skrub/_adaptive_squashing.py
dholzmueller: maybe q_lower and q_upper?

doc/reference/index.rst
jeromedockes: that's true. ok we can go for Squash(ing)Scaler for now and we still have until the next release to change our mind

dholzmueller:
We could try it on a dataset with an MLP and see if it improves something. Don't know what would be the best one to do this.

doc/reference/index.rst
Vincent-Maladiere: what does @GaelVaroquaux think?

dholzmueller:
Updates:
- renamed the class to SquashingScaler
- made it work for dataframes and not only series by using OnEachColumn with the old class
- renamed the args to lower_quantile and upper_quantile
- incorporated minor suggestions
- added an example using the SquashingScaler

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,341 @@
...
+
+        # this will use the column name of the column that was passed in fit_transform,
+        # not the name of X
+        result = sbd.make_column_like(X, values_np, self.col_name_)
+        result = sbd.copy_index(X, result)
+
+        return result
+
+
+class SquashingScaler(BaseEstimator, TransformerMixin):
```
jeromedockes, line 208: even though it wouldn't make a difference here normally we would put the transformermixin first because we want methods to be looked up there before looking in baseestimator

```suggestion
class SquashingScaler(TransformerMixin, BaseEstimator):
```

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,341 @@
...
+            The column to transform.
+        y : None
+            Unused. Here for compatibility with scikit-learn.
+
+        Returns
+        -------
+        self: SquashingScaler
+            The current object.
+        """
+        self.tfm_ = SingleColumnSquashingScaler(
```
jeromedockes, line 317: maybe we could make the transformer private by renaming it to `self._transformer` and exposing its learned attributes directly by copying them to the outer object's dict with something like

```python
if isinstance(self._transformer, OnEachColumn):
    self.median_ = [t.median_ for t in self._transformer.transformers_.values()]
    self.scale_ = [t.median_ for t in self._transformer.transformers_.values()]
else:
    self.median_ = self._transformer.median_
    self.scale_ = self._transformer.scale_
```

jeromedockes: (in this case we could add a section like

```
    Attributes
    ----------
    median_ : float or list of float
        median of the training data

    scale_ : ...
```

just below the 'parameters' section in the docstring )

jeromedockes (commented):
Great, thanks a lot @dholzmueller . I just have a small comment on the fitted attributes but LGTM otherwise :)

dholzmueller:
@GaelVaroquaux was suggesting to implement SquashingScaler directly instead of with the indirection through SingleColumnSquashingScaler. I'll need to check how easy that is...

GaelVaroquaux:
> that's an option too, I'll let you decide!

Users (and non advanced programmers) struggle with indirections.

Your point Jerome is about memory overhead? Do you think that 1) they will be really a problem 2) they end up being avoided, because the output of this transformer is likely going to be fed to a scikit-learn learner?

jeromedockes:
> Your point Jerome is about memory overhead?

yes because we go from columnar format, to one contiguous array, then back to columnar (dataframe format). that is the case even if the output dataframe is then passed to an estimator.

> 1) they will be really a problem

not necessarily but just in case it is easy, it is always nice to avoid an extra copy, especially when it requires a much larger contiguous buffer than any column in the input dataframe does. 

the main reason I mention it is because the fitting of each column as currently implemented uses python control flow anyway (to handle corner cases such as a column is all empty, the quantiles are equal, or the column is constant), so I suspect it will not be easy to express in a "vectorized" way on one numpy array. so if we end up with a loop over columns anyway, let's avoid building the full array. 

if it makes things more complicated you can ignore my comment

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,341 @@
...
+
+import numpy as np
+from sklearn.base import BaseEstimator, TransformerMixin
+from sklearn.utils.validation import check_is_fitted
+
+from . import _dataframe as sbd
+from ._on_each_column import OnEachColumn, RejectColumn, SingleColumnTransformer
+
+
+class SingleColumnSquashingScaler(SingleColumnTransformer):
```
GaelVaroquaux, line 11: I seems to me that the logic could be implemented well at the level of multiple columns. Nan handling can be done using np.nanmean and np.nanquantile.

I think that this would be beneficial

Vincent-Maladiere:
Conclusion of the last skrub meeting: this estimator should not use `OnEachColumn` because it adds an extra layer of abstraction on the base object. Therefore, we have to implement the necessary logic in this transformer.

Vincent-Maladiere:
Update: this is now ready for review. This transformer behaves like any scikit-learn preprocessor, taking a pandas or polars dataframe or numpy array as input, and yielding a numpy array. This can be used within the TableVectorizer or ApplyOnCols without additional logic.

dholzmueller:
- I think the old behavior was to produce the original datatype (dataframe or numpy array), but I don't know what we actually want.
- Do we want to support applying the SquashingScaler to pd.Series? I think it could be convenient in some cases.

Vincent-Maladiere:
> I think it could be convenient in some cases.

Yes, totally. I was trying to minimize the differences between sklearn and skrub scalers APIs, but we definitely can.

examples/10_squashing_scaler.py
```diff
@@ -0,0 +1,79 @@
+"""
+SquashingScaler: Robust numerical preprocessing for neural networks
+===================================================================
+
+The following example illustrates the use of the :class:`~skrub.SquashingScaler`, a
+transformer that can rescale and squash numerical features to a range that works well
+with neural networks and perhaps also other related models. Its basic idea is to
+rescale the features based on quantile statistics, to be robust to outliers, and then
```
rcap107:
```suggestion
rescale the features based on quantile statistics (to be robust to outliers), and then
```

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,304 @@
...
+from sklearn.base import BaseEstimator, OneToOneFeatureMixin, TransformerMixin
+from sklearn.preprocessing import RobustScaler
+from sklearn.utils.validation import FLOAT_DTYPES, check_is_fitted
+
+from skrub._sklearn_compat import validate_data
+
+
+def _infinite_sign(X):
+    """Replace infinite values with NaN and return their sign."""
+    if (inf_sign := np.isinf(X)).any():
```
rcap107: I find the name `inf_sign` confusing, I read it as "this is the sign used as infinte", not as the location of the inf 

I'd rather have `inf_mask` instead, I think it conveys better what this is supposed to do.

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,304 @@
...
+
+def _set_zeros(X, zero_cols):
+    """Set the finite values of the specified columns to zero."""
+    mask = np.isfinite(X)
+    mask[:, ~zero_cols] = False
+    X[mask] = 0.0
+    return X
+
+
+def _soft_clip(X, max_absolute_value, inf_sign):
```
rcap107, line 30: here the docstring shows `mask_inf` instead of `inf_sign`, and I prefer `mask_inf` out of the two

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,304 @@
...
+
+def _infinite_sign(X):
+    """Replace infinite values with NaN and return their sign."""
+    if (inf_sign := np.isinf(X)).any():
+        sign = np.sign(X)
+        X = np.where(inf_sign, np.nan, X)
+        # 0 when X is finite, 1 when X is +inf, -1 when X is -inf
+        inf_sign = inf_sign.astype(X.dtype) * sign
+
+    return X, inf_sign
```
rcap107:
```suggestion
def _infinite_sign(X):
    """Replace infinite values with NaN and return their sign."""
    if (mask_inf := np.isinf(X)).any():
        sign = np.sign(X)
        X = np.where(mask_inf, np.nan, X)
        # 0 when X is finite, 1 when X is +inf, -1 when X is -inf
        inf_sign = mask_inf.astype(X.dtype) * sign

    return X, inf_sign
```

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,304 @@
...
+        self.scale_ = 2 / (np.nanmax(X, axis=0) - np.nanmin(X, axis=0) + eps)
+        return self
+
+    def transform(self, X):
+        check_is_fitted(self, ["median_", "scale_"])
+        return self.scale_ * (X - self.median_)
+
+
+class SquashingScaler(OneToOneFeatureMixin, TransformerMixin, BaseEstimator):
+    r"""Perform robust centering and scaling of followed by soft clipping.
```
rcap107: Maybe the sentence is missing something? 

```suggestion
    r"""Perform robust centering and scaling followed by soft clipping.
```

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,304 @@
...
+    This method has been introduced as the robust scaling and smooth clipping transform
+    in
+    `Better by default (Holzmüller et al., 2024) <https://arxiv.org/abs/2407.04491>`_ .
+
+    Examples
+    --------
+    >>> import pandas as pd
+    >>> import numpy as np
+    >>> from skrub import SquashingScaler
+    >>> X = pd.DataFrame(dict(a=[-100.0, 0.0, 8.0, 9.0, 10.0, 20.0, 100.0, np.nan]))
```
rcap107: maybe add inf to the example to show what happens

rcap107: there should also be an example showing that when the input column is 0, the transform is also converting it to 0 regardless of the values (if I'm understanding what the code is doing)

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,304 @@
...
+            self,
+            X=X,
+            reset=True,
+            dtype=FLOAT_DTYPES,
+            accept_sparse=False,
+            ensure_2d=True,
+            ensure_all_finite=False,
+        )
+        X, inf_sign = _infinite_sign(X)
+
```
rcap107, line 255: please add a comment to explain this group of lines

skrub/tests/test_squashing_scaler.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+def test_squashing_scaler_non_numeric(df_module):
+    X = df_module.make_column(None, ["a", "b"])
+    tfm = SquashingScaler()
+    with pytest.raises(ValueError, match="could not convert"):
+        tfm.fit_transform(X)
+
+
+def test_squashing_scaler_known_values(df_module):
```
rcap107, line 138: please add a comment explaining what this is for

rcap107 (commented):
Overall, I think that the code in this PR is not documented well enough, as there are a lot of parts that "do stuff" in a non-transparent way, and that would benefit from being commented more. 

Besides being a problem to maintain (I had to try and reverse engineer most of the code and the tests to figure out how it worked), I think that the scaling could have some unexpected results that may surprise users, and that are not documented or explained properly in the docstrings or in the code itself.

skrub/_squashing_scaler.py
Vincent-Maladiere: Ok, will do

rcap107: thanks a lot 🙏

skrub/_squashing_scaler.py
Vincent-Maladiere: It's more general than that: if the column is constant, then the scaler fills it with zeros. I can add an example for all 3 cases.

rcap107: That would be very useful, thanks

Vincent-Maladiere:
Hey @rcap107 thanks for reviewing. I iterated on the documentation; let me know what you think

skrub/_squashing_scaler.py
```diff
@@ -0,0 +1,337 @@
...
+        scaling factor for the robust scaling step. The quantiles are computed from the
+        finite values in the input column. If the two quantiles are equal, the scaling
+        factor is computed from the 0th and 100th percentiles (i.e., the minimum and
+        maximum values of the finite values in the input column).
+
+    Notes
+    -----
+    This transformer is applied to each column independently. It uses two stages:
+
+    1. The first stage centers the median of the data to zero and multiply the data by a
```
rcap107:
```suggestion
    1. The first stage centers the median of the data to zero and multiplies the data by a
```

rcap107:
> Hey @rcap107 thanks for reviewing. I iterated on the documentation; let me know what you think

It's much better, thanks a lot! I fixed a small typo, but aside from that I think we can merge this

Vincent-Maladiere:
Great! Waiting for other people to give their feedback

dholzmueller:
Interface looks good to me! Did you test what happens if you pass in an all-nan column?

GaelVaroquaux:
I'm waiting for @Vincent-Maladiere 's reply to @dholzmueller 's above, and then merge

Vincent-Maladiere:
> Interface looks good to me! Did you test what happens if you pass in an all-nan column?

Good question. We get a column full of nans, since they are left untouched. Is that what you would expect?

Reproducer:
```python
import numpy as np
import skrub
import pandas as pd

skrub.SquashingScaler().fit_transform(
    pd.DataFrame(dict(a=[np.nan, np.nan]))
)
```

dholzmueller:
Yes, that is what I would expect. Just wasn't sure if it's going to work with the nanmax/nanmin operations.

dholzmueller:
Just a thought, having quantile_range as a tuple instead of two separate values might make it harder for people to tune it in some hyperparameter optimization. I think it's probably not an important hyperparameter to tune, though.

rcap107:
> Just a thought, having quantile_range as a tuple instead of two separate values might make it harder for people to tune it in some hyperparameter optimization. I think it's probably not an important hyperparameter to tune, though.

This can be addressed in another PR

dholzmueller:
> This can be addressed in another PR

you mean, another one for the same release? because it's not as convenient to change the parametes after release, right?

rcap107:
> > This can be addressed in another PR
> 
> you mean, another one for the same release? because it's not as convenient to change the parametes after release, right?

It's not a very big deal, and it should not be a big change, up to you @Vincent-Maladiere 

(I'd rather have this PR in the release and open a new one later than risk not having it)

Vincent-Maladiere:
That's a good point, both sklearn's RobustScaler and MinMaxScaler have a tuple range as a hyperparameter, so I think this is fine.

dholzmueller:
okay, then let's leave it like it is.

GaelVaroquaux:
OK, I suggest that we merge, and worry about the NaN thing later: it can be addressed later

dholzmueller:
I think the NaN thing is not an issue, since Vincent tested it and it worked.

GaelVaroquaux:
Merged!

Thanks everyone

Issue #1286, opened 2025-04-07
Add a DropColumnIfConstant transformer


rcap107:
### Problem Description

It would be useful to have a transformer that automatically drops constant columns, i.e., columns that contain only one value. These columns do not bring any information and could therefore be dropped without issues. 

### Feature Description

The transformer could be similar to the current DropColumnIfNull transformer, which is activated with a flag in the TableVectorizer. It could be off by default, and turned on if needed.

rcap107:
A related feature might be adding a "DropColumnIfUnique", where the column is dropped if the number of unique values is equal (or very close to, maybe with a threshold?) to the number of rows in the dataframe. This should _definitely_ be off by default.

Themlaw:
is it still open or as it been solved ?

rcap107:
Hi @Themlaw, the issue has been addressed in #1313, thanks for checking

PR #1313, opened 2025-04-11
Adding DropUninformative

rcap107:
closes #1317 

Adding the `DropUninformative` transformer, which uses different tests to check whether if a column is considered to be "uninformative". 

For the moment, I have implemented
- Drop if too many nulls (I moved the logic from the transformer that is already there)
- Drop if constant
- Drop if all values are distinct

Missing values are considered as additional missing values as per @GaelVaroquaux's comment.

The idea adding new heuristics for "uninformative" columns here directly rather than having a bunch of different very small objects.

Still missing: docstring and examples, adding a deprecation warning for the drop if too many nulls transformer.


rcap107:
addressing #1286

CHANGES.rst
```diff
@@ -27,6 +27,11 @@ Lemaitre <glemaitre>`, :user:`Adrin Jalali <adrinjalali>`, :user:`Olivier Grisel
 <ogrisel>` and others participated through many discussions in defining the
 requirements and the public API.
 
+- Adding the ``DropUninformative`` transformer, which uses different heuristics
```
Vincent-Maladiere, line 30:
```suggestion
- Adding the :class:`DropUninformative` transformer, which uses different heuristics
```

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,137 @@
...
+class DropUninformative(SingleColumnTransformer):
+    """Drop column if it is found to be uninformative according to various criteria.
+
+    A column is considered to be "uninformative" if one or more of the following
+    issues are found:
+    - The fraction of missing values is larger than a certain fraction (by default,
+    all values must be null for the column to be dropped).
+    - The column includes only one unique value (the column is constant). Missing
+    values are considered a separate value.
+    - The number of unique values in the column is equal to the length of the column.
```
Vincent-Maladiere, line 20: I guess this category of non-informativeness should be handled with care because free-form text is informative but might have unique values.

We might need additional heuristics to separate non-informative IDs from free-form text. I'm thinking about mean text length, variance of text length, or looking for a structure –e.g. uuid4 are common types of ids that all look like `9268c073-1cb3-4817-a40d-d9296b0d5a8c`. You could also argue that numerical range IDs (from 0 to N) could also bring information if the data is not exchangeable.

There is also the somewhat nicher use case of mini-batching using partial fit, but we can ignore it for now.

WDYT?

jeromedockes: yes and there is no restriction on the types transformed by this column so for example a column of real numbers would also easily have all unique values.

this parameter is set to False by default, though... but I'm not sure I would ever want to set it to true

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,137 @@
...
+
+    Parameters
+    ----------
+    constant_column: bool, default=True
+        If True, drop the column if it contains only one unique value. Missing values
+        count as one additional distinct value.
+    column_is_id: bool, default=False
+        If True, drop the column if all values are distinct. Missing values count as
+        one additional distinct value.
+    threshold: float or None, default=1.0
```
jeromedockes, line 30: now that there are several criteria I think it should be called something like `null_fraction_threshold`

skrub/_drop_uninformative.py
rcap107: yes I totally agree the "column of uniques" should be kept off by default

the scenario I was considering was more along the lines of "the user has figured out the text columns and is handling them separately, anything else is an ID and can be dropped" 

maybe I could add a check that it's only for non-numeric columns, what do you think

jeromedockes: however if other people prefer to keep it I'm ok with it as long as it's off by default

jeromedockes:
> I can't replicate the test failures even using pixi

is it only on windows?

rcap107:
> > I can't replicate the test failures even using pixi
> 
> is it only on windows?

I've only seen it fail on windows, I'm not sure if it's because other runs are skipped by the failure

jeromedockes (commented):
really cool @rcap107 ! apart from adding to the reference index I only have nitpicks

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,137 @@
...
+        if self.null_fraction_threshold == 1.0:
+            return sbd.is_all_null(column)
+        # No nulls found, or no threshold
+        if self.null_count == 0 or self.null_fraction_threshold is None:
+            return False
+        return self.null_count / len(column) > self.null_fraction_threshold
+
+    def _drop_if_constant(self, column):
+        if self.constant_column:
+            if (sbd.n_unique(column) == 1) and (sum(sbd.is_null(column)) == 0):
```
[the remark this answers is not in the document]
rcap107: yes, I replaced it with `not sbd.has_nulls`

doc/reference/index.rst
```diff
@@ -85,6 +85,7 @@ Building a pipeline
    Cleaner
    SelectCols
    DropCols
+   DropUninformative
```
jeromedockes, line 88: we also need to import it in `__init__.py` and add it to `__all__`

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,134 @@
...
+
+__all__ = ["DropUninformative"]
+
+
+class DropUninformative(SingleColumnTransformer):
+    """Drop column if it is found to be uninformative according to various criteria.
+
+    A column is considered to be "uninformative" if one or more of the following
+    issues are found:
+    - The fraction of missing values is larger than a certain fraction (by default,
```
jeromedockes, line 16: I think we need a blank line + the list items continuations must be indented

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,134 @@
...
+        if self.null_fraction_threshold == 1.0:
+            return sbd.is_all_null(column)
+        # No nulls found, or no threshold
+        if self._null_count == 0 or self.null_fraction_threshold is None:
+            return False
+        return self._null_count / len(column) > self.null_fraction_threshold
+
+    def _drop_if_constant(self, column):
+        if self.drop_if_constant:
+            if (sbd.n_unique(column) == 1) and (not sbd.has_nulls(column)):
```
jeromedockes, line 74:
```suggestion
            if (sbd.n_unique(column) == 1) and (self._null_count == 0):
```

I guess we could avoid re-counting nulls like this?

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,134 @@
...
+                return True
+        return False
+
+    def _drop_if_id(self, column):
+        if self.drop_if_id and not sbd.is_numeric(column):
```
jeromedockes, line 79: not important here but as a general trick you can sometimes reduce the amount of indentation by returning early eg here we could do `if not self.drop_if_id: return False` and then the main body of the function would be one level less deep

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,134 @@
...
+            dropped depending on the threshold.
+        """
+        del y
+
+        self._check_params()
+
+        # Count nulls
+        self._null_count = sum(sbd.is_null(column))
+
+        self.drop_ = any(
```
jeromedockes, line 107: we can avoid computing the subsequent decisions if the previous one is sufficient for dropping. for example that might avoid needlessly counting unique values if we are above the null threshold. maybe something like

```
self.drop_ = False
if self._drop_if_too_many_nulls(column):
    self.drop_ = True
if self._drop_if_constant(column):
    ...
```

or `any(check(column) for check in [self._drop_if_too_many_nulls, self._drop_if_constant, ...])`

rcap107: thanks, it makes sense

jeromedockes (dismissed):
cool, thanks so much @rcap107 !

jeromedockes:
oops I approved th echanges too quick :)

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,172 @@
...
+    - The number of unique values in the column is equal to the length of the
+      column.
+
+    Parameters
+    ----------
+    drop_if_constant : bool, default=False
+        If True, drop the column if it contains only one unique value. Missing values
+        count as one additional distinct value.
+
+    drop_if_id : bool, default=False
```
Vincent-Maladiere, line 30: What about `drop_if_unique`? As we have no way to identify IDs, `drop_if_id` could be a little misleading? WDYT? Also, it could be nice to add an explainer saying that free-form text is likely to be dropped too, so to be used with caution

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,172 @@
...
+    By default, only null columns are dropped:
+
+    >>> du = DropUninformative()
+    >>> du.fit_transform(df["col1"])
+    []
+
+    It is also possible to drop constant columns, or specify a lower null fraction
+    threshold:
+
+    >>> df = pd.DataFrame({"col1": [1,2,None], "col2": ["const", "const", "const"]})
```
Vincent-Maladiere, line 53:
```suggestion
    >>> df = pd.DataFrame({"col1": [1, 2, None], "col2": ["const", "const", "const"]})
```

skrub/_drop_uninformative.py
```diff
@@ -0,0 +1,172 @@
...
+        if self.drop_null_fraction == 1.0:
+            return self._null_count == len(column)
+        # No nulls found, or no threshold
+        if self._null_count == 0 or self.drop_null_fraction is None:
+            return False
+        return self._null_count / len(column) > self.drop_null_fraction
+
+    def _drop_if_constant(self, column):
+        if self.drop_if_constant:
+            if (sbd.n_unique(column) == 1) and (not self._null_count > 0):
```
Vincent-Maladiere, line 107: Nitpick: simplify
```suggestion
            if (sbd.n_unique(column) == 1) and (self._null_count == 0):
```

Vincent-Maladiere: If we want, this can be simplified further as:
```python
return self.drop_if_constant and (sbd.n_unique(column) == 1) and (self._null_count == 0)
```
without hurting readability

Vincent-Maladiere (commented):
A few nitpicks and it's good to go :)

Vincent-Maladiere (approved):
LGTM, thank you @rcap107! :)

Issue #1297, opened 2025-04-10
Moving the user guide into examples?


jeromedockes:
The user guide is a bit shallow and a good part of it is redundant with some of the examples. Should we consider reorganizing the example gallery so that there is a part of it that is quite narrative, meant to be read like a user guide, and remove the current user guide?

Issue #1293, opened 2025-04-10
Improve the expressions examples


jeromedockes:
As discussed in the original PR, the examples about the skrub expressions still need some work

- They are too long
- There may still be some material in `/examples/FIXME/11b_expressions_explained_alternative_take_TODO_unify_both.py` that needs to be integrated into the other examples
- It might be useful, in addition to the longer, narrative examples which serve as a user guide, to have a small collection of very short, "howto" examples that focus on one specific task or aspect of the api

MarieSacksick:
Is this issue still a thing?

About each of the points:  
- They are too long

They don't look so long to me, so maybe they got shorter? Or is there one specific example that you have in mind that is still too long?

- There may still be some material in /examples/FIXME/11b_expressions_explained_alternative_take_TODO_unify_both.py that needs to be integrated into the other examples

That file doesn't exist anymore.
 
- It might be useful, in addition to the longer, narrative examples which serve as a user guide, to have a small collection of very short, "howto" examples that focus on one specific task or aspect of the api  

To me it's covered by all the dedicated and specific examples about data ops in the [user guide](https://skrub-data.org/stable/data_ops.html)

rcap107:
> * It might be useful, in addition to the longer, narrative examples which serve as a user guide, to have a small collection of very short, "howto" examples that focus on one specific task or aspect of the api
> 
> To me it's covered by all the dedicated and specific examples about data ops in the [user guide](https://skrub-data.org/stable/data_ops.html)

We have been talking about reworking the user guide at length, partly because a lot of the content that is already there is hard to parse. The idea is moving a lot of that content to the self-contained pages that are mentioned in the issue. Which subjects and what's the best way of doing that is still unclear.

PR #1314, opened 2025-04-11
Add skrub pipeline user guide

Vincent-Maladiere:
Address https://github.com/skrub-data/skrub/issues/1297, toward https://github.com/skrub-data/skrub/issues/1293


doc/skrub_pipeline.rst
```diff
@@ -0,0 +1,353 @@
...
+  :class:`~sklearn.model_selection.GridSearchCV`) becomes difficult in complex
+  pipelines.
+
+What is the difference with scikit-learn :class:`~sklearn.pipeline.Pipeline`?
+=============================================================================
+
+Scikit-learn pipelines represent a linear sequence of transformations on one
+table with a fixed number of rows.
+
+.. image:: ../../_static/sklearn_pipeline.svg
```
jeromedockes, line 39: maybe we need 1 less level of `../` now?

doc/skrub_pipeline.rst
```diff
@@ -0,0 +1,353 @@
...
+    :width: 500
+
+Skrub expressions, on the other hand, can manipulate any number of variables.
+The transformation they perform is not a linear sequence but any Directed
+Acyclic Graph of computations.
+
+.. image:: ../../_static/skrub_expressions.svg
+
+What is the difference with orchestrators like Apache Airflow?
```
jeromedockes, line 48: those 2 "what is the difference" sectiosn are very useful :+1:

examples/expressions/12_scikit_learn_estimators_in_expressions.py
```diff
@@ -1,292 +0,0 @@
...
-# %%
-# Note how ``apply`` works for supervised estimators: we pass the targets
-# (another skrub expression) as the argument for the parameter ``y``.
-#
-# If you unfold the dropdown to see the computation graph, you will see the
-# inner nodes that have been marked as ``X`` and ``y``: these nodes are highlighted
-# in blue and red, respectively. Furthermore, the ``data`` is an input variable,
-# so it is marked with a double box.
-#
-# To perform
```
jeromedockes, line 57: I think there may be a bit of explanation about mark_as_x etc to rescue out of this section and move elswhere but I'm not sure

examples/expressions/12_scikit_learn_estimators_in_expressions.py
```diff
@@ -1,292 +0,0 @@
...
-
-# %%
-
-# Note: using deferred or even defining a function is completely optional here,
-# because the computation it contains involves no control flow or assignments.
-# we only do it to simplify the computation graph by collapsing several
-# operations into a single function call.
-
-
-@skrub.deferred
```
jeromedockes, line 163: I agree this is redundant with the example 10. I wonder if there is another place we could illustrate the recommended practice of grouping steps together in `deferred` functions to simplify the graph avoid potential issues with pickling etc

jeromedockes (approved):
I think this is an improvement, thanks a lot @Vincent-Maladiere ! :rocket: I think once the conflicts and links are good we could merge and iterate in other prs

doc/skrub_pipeline.rst
```diff
@@ -0,0 +1,353 @@
+.. _skrub_pipeline:
+
+===================================================
+Skrub Pipeline: flexible machine learning pipelines
```
GaelVaroquaux, line 4: I hate to be that person, but that title is too generic: it does not make us guess what is different to other pipelines.

The cool thing about the skrub pipeline is that it applies to any wrangling tool.

Maybe "Cross-validate or tune any wranging pipeline"

Vincent-Maladiere: Agreed, was looking for something better. Let's use your suggestion

examples/expressions/12_scikit_learn_estimators_in_expressions.py
Vincent-Maladiere: That depends on the role we want the User Guide to play. I tried to keep it as light as possible so that readers understand the main expression concepts, without diving straight into all the details (the example showcases these details better IMO because they are built around complete use-cases). For instance, if I add `mark_as_X` in the User Guide, I should also add the `choose_from` mechanism, etc. Does that make sense?

doc/skrub_pipeline.rst
Vincent-Maladiere: What about "Cross-validate or tune any **data** pipeline", since I'm not sure most readers will be familiar with wrangling pipelines

Vincent-Maladiere:
The last TODO is to update the images, once we agree with the content

jeromedockes:
thanks a lot @Vincent-Maladiere  that is super useful!!

PR #1315, opened 2025-04-11
expressions cross_validate return DataFrame

gabrielapgomezji:
TO CONSIDER: No tests that covers the lines that I changed

### DIFF
--- a/skrub/_expressions/_estimator.py
+++ b/skrub/_expressions/_estimator.py
@@ -478,7 +478,12 @@ def cross_validate(expr_estimator, environment, **cv_params):
     >>> pred = X.skb.apply(log_reg, y=y)
     >>> search = pred.skb.get_randomized_search(random_state=0)
     >>> skrub.cross_validate(search, pred.skb.get_data())['test_score'] # doctest: +SKIP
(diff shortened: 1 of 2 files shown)


Vincent-Maladiere:
Hey @gabrielapgomezji and @jeromedockes, what context/issue does this PR address?

GaelVaroquaux:
> cross_validate can return a dataframe instead of a dictionary which has a more readable display

Yes, more readable and easier to work with

jeromedockes:
Hi @gabrielapgomezji it seems like the change to the documentation was pushed but not the one to `_estimator.py`, do you think that might be the case that maybe you have some changes locally that were not pushed to the PR?

gabrielapgomezji:
Hi, yes I have some local changes regarding the `cross_validate` but some tests failed because of the examples of the docstrings. I wanted to check it better at the end of the week.

jeromedockes (approved):
Thank you very much @gabrielapgomezji ! :rocket:

jeromedockes (approved):
Thank you very much, @gabrielapgomezji !! :rocket:

PR #1317, opened 2025-04-12
Added an option to drop cols with cardinality == 1 in `TableVectorizer`

### LINKED ISSUE #1286: Add a DropColumnIfConstant transformer

victoris93:
Closes #1286

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -428,6 +428,10 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
         this selection is disabled: no columns are dropped based on the number
         of null values they contain.
 
+    drop_cardinality_1 : bool, default=False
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
Hi @victoris93 ,

Thanks a lot for starting this. This is super useful!

I think that the present PR would probably benefit from merging the work started by @rcap107 :
https://github.com/skrub-data/skrub/pull/1313

Maybe hold on for a tiny bit until we merge the above. But, you can help by reviewing it :).

Cheers!

PR #1318, opened 2025-04-12
Fix a malformated docstring

GaelVaroquaux:
Line too long was creating a warning.

But also, the "See also" title does not take a colon


jeromedockes (approved):
thanks a lot!

Issue #1323, opened 2025-04-14
UX: picklability check is not compatible with the `.assign(col_name=lambda x: ...)` pattern in chained pandas calls.


ogrisel:
`skrub` expressions discourage the use of in-place pandas dataframe column transformations in favor of `.assign`. However, the latter is often used with `lambda` expression to make it possible to chain operations on pandas dataframes (e.g. `.rename(...).drop(...).assign(new_col=lambda df: df['renamed_col'] + 1)`).

Since pandas lacks support for deferred expression evaluation, their doc actually recommends the use of lambda expression as a workaround:

https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.assign.html#:~:text=Examples

However, skrub's picklability check conflicts with this scheme:

```python
# %%
import skrub
import pandas as pd

# %%
pandas_df = pd.DataFrame(
    {
        "Time (UTC)": pd.date_range("2023-01-01", "2023-01-03", freq="D"),
        "Temperature ℃": [17.0, 25.0, 18.5],
    },
)
raw_datasource = skrub.var("raw_datasource", pandas_df)
raw_datasource
# %%
raw_datasource.rename(
    columns={"Temperature ℃": "temp_c", "Time (UTC)": "time_utc"}
).assign(temp_f=lambda x: x.temp_c * 9 / 5 + 32)
```

```
PicklingError: The check to verify that the pipeline can be serialized failed.
```

<details>

```python
---------------------------------------------------------------------------
PicklingError                             Traceback (most recent call last)
File ~/code/skrub/skrub/_expressions/_expressions.py:223, in check_can_be_pickled(obj)
    222 try:
--> 223     dumped = pickle.dumps(obj)
    224     pickle.loads(dumped)

PicklingError: Can't pickle <function <lambda> at 0x1405e3920>: attribute lookup <lambda> on __main__ failed

The above exception was the direct cause of the following exception:

PicklingError                             Traceback (most recent call last)
File /Users/ogrisel/tmp/skrub_vars_sandbox.py:4
      1 # %%
      2 raw_datasource.rename(
      3     columns={"Temperature ℃": "temp_c", "Time (UTC)": "time_utc"}
----> 4 ).assign(temp_f=lambda x: x.temp_c * 9 / 5 + 32)

File ~/code/skrub/skrub/_expressions/_expressions.py:330, in _check_return_value.<locals>.check_call_return_value(*args, **kwargs)
    328 @functools.wraps(f)
[...]

PR #1324, opened 2025-04-14
use cloudpickle to serialize expressions held in an estimator + skip eager pickling checks

jeromedockes:
one possible fix for #1323

### DIFF
--- a/skrub/_expressions/_expressions.py
+++ b/skrub/_expressions/_expressions.py
@@ -22,7 +22,7 @@ from .._utils import PassThrough, short_repr
 from .._wrap_transformer import wrap_transformer
 from . import _utils
 from ._choosing import get_chosen_or_default
-from ._utils import FITTED_PREDICTOR_METHODS, NULL, _CloudPickle, attribute_error
+from ._utils import FITTED_PREDICTOR_METHODS, NULL, attribute_error
 
 __all__ = [
     "var",
@@ -420,8 +420,7 @@ class _ExprDoc:
         return f"""Skrub expression.\nDocstring of the preview:\n{doc}"""
 
 
-class Expr(_CloudPickle):
-    _cloudpickle_attributes = ["_skrub_impl"]
+class Expr:
     __hash__ = None
 
     __doc__ = _ExprDoc()
--- a/skrub/_expressions/_expressions.py
+++ b/skrub/_expressions/_expressions.py
@@ -420,7 +420,8 @@ class _ExprDoc:
         return f"""Skrub expression.\nDocstring of the preview:\n{doc}"""
 
 
-class Expr:
+class Expr(_CloudPickle):
+    _cloudpickle_attributes = ["_skrub_impl"]
     __hash__ = None
 
     __doc__ = _ExprDoc()
@@ -1134,7 +1135,7 @@ class GetItem(ExprImpl):
         return f"[{_get_preview(self.key)!r}]"
 
 
-class Call(_CloudPickle, ExprImpl):
+class Call(ExprImpl):
     _fields = [
         "func",
         "args",
@@ -1144,7 +1145,6 @@ class Call(_CloudPickle, ExprImpl):
         "defaults",
         "kwdefaults",
     ]
-    _cloudpickle_attributes = ["func"]
 
     def compute(self, e, mode, environment):
         func = e.func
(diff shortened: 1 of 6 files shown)


skrub/_expressions/_expressions.py
```diff
@@ -420,7 +420,8 @@ def __get__(self, instance, owner=None):
         return f"""Skrub expression.\nDocstring of the preview:\n{doc}"""
 
 
-class Expr:
+class Expr(_CloudPickle):
+    _cloudpickle_attributes = ["_skrub_impl"]
```
ogrisel, line 424: Is `_skrub_impl` guaranteed to not hold a reference to a large datastructure? Calling `cloudpickle.dumps` would be problematic otherwise (by allocating a large intermediate datastructure before serializing to disk/network).

jeromedockes: right, that's true. no it isn't and in fact it is quite likely to hold references to dataframes that users passed when initializing a variable


all in all I think we might better off removing the pickle-ability checks, and any attempt to force usage of cloudpickle, and just recommend in the documentation that users rely on cloudpickle if pickle cannot handle their pipeline (eg due to lambda expressions)

Vincent-Maladiere (approved):
Based on my understanding of the convo and the current limitations of the pickling mechanism, it LGTM –I haven't tested thoroughly, yet.

ogrisel:
I think this solution is problematic because it causes large memory copies and significant overhead when working with large preview data as discussed in https://github.com/skrub-data/skrub/issues/1323#issuecomment-2802213948 and the rest of the discussion.

I think the consensus with @jeremiedbb was rather to move the picklability check later (when calling `get_estimator`) which is a much more light weight object instead.

jeremiedbb:
@ogrisel I think you pinged the wrong person cause I have no idea what you're talking about 😄

ogrisel:
Sorry i wanted to ping @jeromedockes .

Vincent-Maladiere:
My understanding of this PR, while @jeromedockes is away:
- We don't check cloud-picklability of expressions anymore
- We only check cloud-picklability when calling `get_estimator` or get_{grid, randomized}_search`, after the expressions have been stripped from intermediate results (which can be voluminous)
- Pickling these objects with the regular `pickle` lib will result in calling cloudpickle in `__getstate__` and `__setstate__` for the stripped expr, which could contain lambda functions or functions not defined in a Python module (like in a notebook).

Do you agree with this understanding @ogrisel?

ogrisel:
Sounds good.

jeromedockes:
Yes the check now only happens when calling `get_estimator()` as we discussed. 

```
>>> import pickle
>>> import skrub

We can use lambda functions etc. No check happens during the construction of the
expression.

>>> e = skrub.var('x', 1).skb.apply_func(lambda x, f: f(x), lambda x: x * 2)
>>> e
<Call '<lambda>'>
Result:
―――――――
2

Pickling expressions is not encouraged because they may contain data, what you
want to pickle are the estimators you extract from an expression and fit. So no
special behavior related to pickling is done by the expression. For example
pickling an expression that contains a lambda will fail.

>>> pickle.dumps(e)
    ...
_pickle.PicklingError: Can't pickle <function <lambda> at 0x742a9d4209a0>: attribute lookup <lambda> on __main__ failed

The estimator (which contains an expression stripped of any data) is checked for
pickling when it is created and defines setstate to use cloudpickle when
serializing its `expr` attribute which references the (data-less) expression.
So pickling the estimator works even if it contains lambda functions.

>>> pipeline = e.skb.get_estimator()
>>> loaded = pickle.loads(pickle.dumps(pipeline))
>>> loaded.fit_transform({'x': 10})
20

Now we create an expression that really cannot be pickled, even with
cloudpickle:

>>> class NoPickle:
...     def __getstate__(self):
...         raise pickle.PicklingError('cannot pickle NoPickle')
...
...     def __call__(self, x):
...         return x * 2

No check when building the expression, it works as expected

>>> e = skrub.var('x', 1).skb.apply_func(lambda x, f: f(x), NoPickle())
>>> e
<Call '<lambda>'>
Result:
―――――――
2
>>> e.skb.eval({'x': 10})
20

However there is a check when we extract the estimator. What remains to do is a
way to disable that check as we discussed. However if that check fails it means
the estimator cannot be serialized even with cloudpickle, so it will not be very
useful: we cannot save it once fitted, we cannot use it with joblib, etc.

>>> pipeline = e.skb.get_estimator()
    ...
_pickle.PicklingError: cannot pickle NoPickle
```

PR #1328, opened 2025-04-16
Add subsampling to expressions

jeromedockes:
implementation of what we discussed this morning with @Vincent-Maladiere , @rcap107  and @glemaitre  : 

- expressions have a `.skb.subsample_previews()` method to indicate that a node should be subsampled when in a kind of "preview / dry-run" setting
- subsampling is always active when computing previews (those used for the repr of expressions, tab completion etc)
- when fitting an estimator, a grid search or cross-validation, subsampling is turned off by default. we can activate it by passing `subsampling=True`

example on the employee salaries:

```python

import skrub
import skrub.datasets


dataset = skrub.datasets.fetch_employee_salaries()

# the `subsample_previews` configures how the sampling is done when it takes place
# whether it takes place or not depends on the context:
#     - for preview the subsampling is always on
#     - for fitting, cross-val etc subsampling is only on when we ask for it explicitly
data = skrub.var("data", dataset.employee_salaries).skb.subsample_previews(n=200)

# the rest of the pipeline is the same as before
employees = data.drop(columns="current_annual_salary", errors="ignore").skb.mark_as_X()
salaries = data["current_annual_salary"].skb.mark_as_y()

vectorizer = skrub.TableVectorizer(
    high_cardinality=skrub.MinHashEncoder(n_components=8)
)
predictions = employees.skb.apply(vectorizer).skb.apply(
    HistGradientBoostingRegressor(
        learning_rate=skrub.choose_float(0.01, 0.8, log=True)
    ),
    y=salaries,
)
```

We can see the preview is computed on a subsample (200 rows):

```
>>> predictions
<Apply HistGradientBoostingRegressor>
Result (on a subsample):
――――――――――――――――――――――――
     current_annual_salary
0             70592.839294
1             96203.419352
2            105259.040121
3             51729.439016
4            103002.625230
..                     ...
195           94219.034011
196           64474.151164
197           62751.180976
198           78534.171710
199           93314.398036
[200 rows x 1 columns]
```

We can ask to keep the subsampling for fitting a grid or randomized search:

```
>>> quick_search = predictions.skb.get_randomized_search(
...     fitted=True, subsampling=True, cv=2, n_iter=2, random_state=0
... )

>>> quick_search.detailed_results_
   mean_test_score  ...  std_score_time
0         0.312805  ...        0.017096
1         0.293941  ...        0.000409
[2 rows x 7 columns]

>>> quick_search.detailed_results_["mean_fit_time"]
0    0.194568
1    0.179346
Name: mean_fit_time, dtype: float64
```

but by default it is off (note the higher computation time and better scores):

```
>>> full_search = predictions.skb.get_randomized_search(
...     fitted=True, cv=2, n_iter=2, random_state=0
... )

>>> full_search.detailed_results_
   mean_test_score  ...  std_score_time
0         0.885271  ...        0.000374
1         0.884144  ...        0.000446
[2 rows x 7 columns]

>>> full_search.detailed_results_["mean_fit_time"]
0    1.167464
1    1.149597
Name: mean_fit_time, dtype: float64
```

notes:

- the 2 subsampling modes available here are "head" or "random", and we provide `n` the number of rows to keep. That could be changed it is just to take advantage of the existing functions in `skrub._dataframe`
- for now it was added to example 10
- later, we could add something like `skrub.sample_csv` or similar that does the subsampling while loading the csv instead of reading everything then subsampling. it would be able to easily read whether it should apply the subsampling or not (depending on the mode etc.) from the private function `_subsampling.should_subsample()` added here


examples/expressions/10_expressions_intro.py
```diff
@@ -251,6 +258,25 @@
...
+# We can first run a small dry-run to check if our param search will run correctly:
+
+# %%
+quick_search = predictions.skb.get_randomized_search(
+    scoring="roc_auc",
+    n_iter=4,
+    n_jobs=4,
+    random_state=0,
+    fitted=True,
+    subsampling=True,
```
rcap107, line 271: Maybe add a comment like this to draw attention to the parameter? 
```suggestion
    subsampling=True, # force the randomized search to run only on the subsample
```

jeromedockes: thanks!

Vincent-Maladiere (approved):
Looks good! Thanks @jeromedockes

jeromedockes:
@GaelVaroquaux  any thoughts on this?

jeromedockes:
I agree @glemaitre  this is just to turn on and off temporary subsampling to speed-up the development and debugging. If subsampling is actually a part of the final pipeline like feature selection or any other transformation, that is a different problem -- it is just another transformation in the pipeline and probably needs no special support. it does change the number / order of rows but that is something skrub expressions naturally accomodate already. so the subsampling API here only focusses on simple, temporary subsampling

jeromedockes:
for the name `subsample_previews`, other ideas welcome. the advantage of `subsample_previews` is it says what we are doing (subsampling) and shows it doesn't systematically happen all the time (only for previews), but the drawback is it's a bit long. for the parameter, maybe `with_subsampling` instead of `subsampling`, it's a bit longer but maybe slightly more explicit.


otherwise I think this is review (or merging if there are no further comments)

Vincent-Maladiere (commented):
Some doc comments

GaelVaroquaux:
Is this feature demoed in one exemple?

With a quick glance (on my mobile phone, sorry), I wasn't sure.

For users to discover this feature, I feel that it is important that there is one exemple that demonstrates it. It might even be useful to have a dedicated example.

GaelVaroquaux:
OK, bit of review (which are going to come in disorder, sorry, I'm doing this review while task switching):

* Should we have a see also been "Expr.skb.preview()" and "Expr.skb.subsample_previews()" (both ways). I would like to make it as easy for users to discover new functions

GaelVaroquaux:
As I mentioned before, I'm still stuck on the name of the "subsampling" parameter to get_grid_search/get_randomized_search.

I stress that I believe it gives the wrong intuitions, because it gives the impression that the corresponding method does the subsampling.

We could have as names "undo_subsampling" or "keep_subsampling"

jeromedockes:
> I stress that I believe it gives the wrong intuitions, because it gives the impression that the corresponding method does the subsampling.

ok I don't mind changing to `keep_subsampling` but I'm not sure what you mean by the wrong impression. It is indeed during the call of `get_randomized_search` that subsampling will be applied or not (the randomized search will be fitted either on the full data or on a subsample)

jeromedockes:
To me it is more that it has been _configured_ above, and we decide when we launch the cross-val if we want to use it or not.

I guess you could also think about it as having subsampled and then undoing that but I find it a bit less natural -- how does one undo subsampling and recreate the data that has been dropped . I would also like users to easily realize `subsample_previews` does nothing by default when actually fitting their  randomized search, they still get the correct results on the full data. it just adds the possibility to turn on subsampling if they wish to do so

jeromedockes:
maybe it could be something like `activate_subsampling`, or `with_subsampling`?

GaelVaroquaux:
OK, I had a look at the example. I would rather have a separate example, as it is very easy to miss.

Also, it is overloading an example that has a lot in it. This example should be the first one that people read to have an idea of what the expressions do and what they are. It needs to be straight to the point, with not too much in it.

Small detail: sphinx-gallery does not pick up the call to "e.skb.subsample_preview" and thus does not link back to it in the example gallery of the method. Probably not easy to fix, but unfortunate

GaelVaroquaux:
> maybe it could be something like activate_subsampling, or with_subsampling?

But neither of these make it explicit that the subsampling has been defined above

jeromedockes:
> Small detail: sphinx-gallery does not pick up the call to "e.skb.subsample_preview" and thus does not link back to it in the example gallery of the method. Probably not easy to fix, but unfortunate

that seems to be the case for all the `.skb` methods. I guess to compensate we should explicitly insert those links in nearby text.

Note that not everyone likes those links in the sample code: I have seen people complain that they unintentionally click the link when trying to select text to copy-paste it

GaelVaroquaux:
> > Small detail: sphinx-gallery does not pick up the call to "e.skb.subsample_preview" and thus does not link back to it in the example gallery of the method. Probably not easy to fix, but unfortunate
> 
> that seems to be the case for all the `.skb` methods. I guess to compensate we should explicitly insert those links in nearby text.
> 
> Note that not everyone likes those links in the sample code: I have seen people complain that they unintentionally click the link when trying to select text to copy-paste it

I care more about the backlinks: looking at the docstrings and seeing examples where the function is demoed

doc/skrub_pipeline.rst
```diff
@@ -395,3 +395,99 @@ Finally, there are other situations where using :func:`deferred` can be helpful:
...
 - See :ref:`example_tuning_pipelines` for an example of hyper-parameter tuning using
   skrub pipelines.
+
+Subsampling
+~~~~~~~~~~~
+
+To speed-up the development of our pipeline, we can tell skrub to work on a
+subsample of our data.
+
+On any transformation that produces either a DataFrame, a Column (Series) or a
```
GaelVaroquaux, line 405: This is very abstract to me. I wouldn't be able to link it to code that I can write.

We should probably say less here, and more in an example.

doc/skrub_pipeline.rst
```diff
@@ -395,3 +395,72 @@ Finally, there are other situations where using :func:`deferred` can be helpful:
   expressions on dataframes.
 - See :ref:`example_tuning_pipelines` for an example of hyper-parameter tuning using
   skrub pipelines.
+
+Subsampling
+~~~~~~~~~~~
```
GaelVaroquaux, line 400: Not a request for change or action, just a comment: Super improved rewrite (@jeromedockes close this when you see it).

GaelVaroquaux, line 400:
```suggestion
Subsampling to develop pipelines with reduced computation
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

jeromedockes: thanks! I was just now working on further reducing the user guide part and adding an actual example in the gallery with the employee salaries, but do you think we should keep it the way it is instead? not sure if there is much more to say in an actual example (still slightly unsure about the respective roles of the user guide and the examples)

GaelVaroquaux:
> thanks! I was just now working on further reducing the user guide part and adding an actual example in the gallery with the employee salaries, but do you think we should keep it the way it is instead? not sure if there is much more to say in an actual example (still slightly unsure about the respective roles of the user guide and the examples)

I think that it would be idea to move what you have in an example and expand it a tiny amount, leaving the user-guide almost empty: only describing the purpose of subsampling and the things to have in mind.

Expanding it would enable to demo the `keep_subsampling` and the fact that cross-validation or random search can be done without subsampling.

skrub/_expressions/_skrub_namespace.py
```diff
@@ -644,6 +654,135 @@ def concat(self, others, axis=0):
...
+        3         0.380719       0.004827  1.398196
+
+        This example dataset is so small that the subsampling does not change
+        the fit computation time but we can tell the second search used the
+        full data from the higher scores. For datasets of a realistic size
+        using the subsampling allows us to do a "dry run" of the
+        cross-validation or model fitting much faster than when using the
+        full data.
+        """  # noqa : E501
+        return Expr(SubsamplePreviews(self._expr, n=n, how=how))
```
GaelVaroquaux, line 784: Super high level though / question:

would there be interest / would it be possible, to have SubsamplePreview rather be a subclass of Expr, rather than a form of composition, to make sorter stack traces?

skrub/_expressions/_skrub_namespace.py
```diff
@@ -719,12 +858,24 @@ def eval(self, environment=None):
...
         result
             The result of running the computation, ie of executing the
             pipeline's ``fit_transform`` on the provided data.
 
+        See Also
+        --------
+        Expr.skb.preview :
+            Access the preview of the result on the variables initial values,
+            with subsampling. Faster than ``eval`` but does not allow passing
+            new data and always applies subsampling.
```
GaelVaroquaux, line 877: Can you make it more clear on how `preview` differs from `subsample_preview` (it's not clear in my head, I must admit)?

skrub/_expressions/_skrub_namespace.py
jeromedockes: Expr is the wrapper that has almost nothing in its `__dict__` to catch almost all attribute access in `__getattr__`. the actual skrub logic is stashed away in its `_skrub_impl` attribute -- the `SubsamplePreviews` in this case

skrub/_expressions/_skrub_namespace.py
```diff
@@ -746,20 +897,83 @@ def eval(self, environment=None):
...
+        -------
+        preview result
+            The result of evaluating the expression on the data stored in its
+            variables.
+
+        See Also
+        --------
+        Expr.skb.subsample_previews :
+            Specify how to subsample an intermediate result when computing
+            previews.
```
GaelVaroquaux, line 937: OK, so what I seem to understand is that "subsample_previews" sets the subsample for the whole subsequent operations, unlike "preview"? Is that right?

I'd love to find a better name. I don't find that when reading code it's terribly clear for me what is happening.

skrub/_expressions/_skrub_namespace.py
GaelVaroquaux: But it still adds one level to the stack? (or maybe not)

skrub/_expressions/_subsampling.py
```diff
@@ -0,0 +1,139 @@
+import numpy as np
+
+from .. import _dataframe as sbd
+from . import _evaluation, _expressions
+
```
GaelVaroquaux, line 5: Can you add comments around here telling us (the reader of this code) what the following few symbols (the constant, the function, the objects) are for in the big-picture of the code organization?

I'm thinking about the next people exploring this code base

skrub/_expressions/_skrub_namespace.py
jeromedockes: `preview()` returns the actual computed value of the preview (for any node, whether there is subsampling or not)

`subsample_previews()` adds a node in the graph that performs subsampling (or not, when `keep_subsampling=False`).

I tried to make the distinction with the plural but I guess that's not clear. maybe it should be just `subsample()` -- I just wanted to hint in the name that by default subsampling only takes place for computing previews.

skrub/_expressions/_skrub_namespace.py
jeromedockes: in many cases no, because most code will access the `_skrub_impl` and use that directly. but in any case we kind of need those 2 classes: the wrapper `Expr` has virtually no functionality of its own, its role is  to have no attributes and capture everything in `__getattr__`. without it there could be clashes between skrub method names and the (dataframe) methods users want to call in the pipeline

GaelVaroquaux:
> maybe it should be just subsample()

I actually like "subsample". I would have gone for this: it's short and the fact that it lives in the skb namespace tells me that it is doing something specific to expressions.

> I just wanted to hint in the name that by default subsampling only takes place for computing previews.

If you really want to be explicit about that, we could have "subsample_for_previews".
I actually think that it is a bad it: the notion of "previews" might not be clear to users, and they can benefit from subsampling without mastering the concept of preview.

skrub/_expressions/_subsampling.py
```diff
@@ -0,0 +1,139 @@
...
+    subsampling: False
+    SkrubPipeline(expr=<Call 'load_data'>)
+    >>> e.skb.get_pipeline(keep_subsampling=True, fitted=True)
+    subsampling: True
+    SkrubPipeline(expr=<Call 'load_data'>)
+    """
+    return _expressions.Expr(ShouldSubsample())
+
+
+def _sample_numpy(a, n):
```
GaelVaroquaux, line 59: Shouldn't we have a way to control this sampling stream?

I would probably advise for using a dedicated random number generator (not seeded by default), and documenting somewhere how to control it. Some users will want to do that.

That comment applies to the whole subsampling code

jeromedockes: I think we want it to be seeded by default so that if you sample X and y separately you get the same indices and don't break their alignment. We can expose the seed parameter though, I just figured few people would use it

GaelVaroquaux:
> as a side note we need to decide what to do with the ken embeddings

Subsample them? Actually, this is not completely a joke. Do they give a good usecase for subsampling? I remember hating the corresponding examples because they took forever.

> as their download is failing roughly 1/4 ci builds lately

OK, if it's only a download issue, that points maybe to the need to rehost them elsewhere, no? Maybe we should open an issue on that?

examples/expressions/12_subsampling.py
```diff
@@ -0,0 +1,84 @@
...
+Subsampling for faster development
+==================================
+
+Here we show how to use :meth:`.skb.subsample_previews()
+<Expr.skb.subsample_previews>` to speed-up interactive creation of skrub
+expressions by subsampling the data when computing preview results.
+"""
+
+# %%
+from sklearn.ensemble import HistGradientBoostingRegressor
```
GaelVaroquaux, line 15: In narrative examples, I prefer doing the import of a symbol in the cell where I use it.

skrub/_expressions/_skrub_namespace.py
```diff
@@ -644,6 +654,135 @@ def concat(self, others, axis=0):
...
+            Number of rows to keep.
+
+        how : 'head' or 'random'
+            How subsampling should be done (when it takes place). If 'head',
+            the first ``n`` rows are kept. If 'random', ``n`` rows are sampled
+            randomly, without maintaining order and without replacement.
+
+        Returns
+        -------
+        DataFrame
```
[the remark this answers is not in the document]
jeromedockes: yes that's left over from the first version which only supported dataframes

examples/expressions/12_subsampling.py
```diff
@@ -0,0 +1,89 @@
...
+# :meth:`.skb.subsample() <Expr.skb.subsample>`.
+
+# %%
+data = full_data.skb.subsample(n=100)
+data
+
+# %%
+# The rest of the pipeline will now use only 100 points for its previews.
+#
+# .. note::
```
GaelVaroquaux, line 40:
```suggestion
# .. topic:: Subsampling only applies to previews by default
```
I tend to prefer using explicit title when possible, with titles being tiny summaries of the content.

examples/expressions/12_subsampling.py
```diff
@@ -0,0 +1,89 @@
...
+#    :meth:`.skb.preview() <Expr.skb.preview>`. For other methods such as
+#    :meth:`.skb.get_pipeline() <Expr.skb.get_pipeline>` or
+#    :meth:`.skb.cross_validate() <Expr.skb.cross_validate>`, *no subsampling is
+#    done by default*. We can explicitly ask for it with ``keep_subsampling=True``
+#    as we will see below.
+#
+# To continue our pipeline we now define X and y:
+
+# %%
+employees = data.drop(columns="current_annual_salary", errors="ignore").skb.mark_as_X()
```
GaelVaroquaux, line 53: That line is a bit long. It overflows in the website layout.

I guess we can't do much about this, because the code formatters like it this why.

examples/expressions/12_subsampling.py
```diff
@@ -0,0 +1,89 @@
...
+# And finally we apply a TableVectorizer then gradient boosting:
+
+# %%
+from sklearn.ensemble import HistGradientBoostingRegressor
+
+predictions = employees.skb.apply(skrub.TableVectorizer()).skb.apply(
+    HistGradientBoostingRegressor(), y=salaries
+)
+
+# %%
```
GaelVaroquaux, line 66:
```suggestion
# %%
# 
# All the lines above run very fast, including fitting the predictor above.
# 
```

examples/expressions/12_subsampling.py
```diff
@@ -0,0 +1,89 @@
...
+# a small subsample the scores will be very low but this might help us quickly
+# detect errors in our cross-validation scheme.
+
+# %%
+predictions.skb.cross_validate(keep_subsampling=True)
+
+# %%
+# By default, when we do not explicitly ask for ``keep_subsampling=True``, no
+# subsampling takes place. Here we run the cross-validation on the full data.
+# Note the longer ``fit_time`` and much better ``test_score``.
```
GaelVaroquaux, line 86:
```suggestion
# By default, when we do not explicitly ask for ``keep_subsampling=True``, no
# subsampling takes place.
# 
# Here we run the cross-validation **on the full data**.
# Note the longer ``fit_time`` and much better ``test_score``.
```

examples/expressions/12_subsampling.py
```diff
@@ -0,0 +1,89 @@
...
+full_data = skrub.var("data", dataset)
+full_data
+
+# %%
+# We are working with a dataset of over 9K rows. As we build up our pipeline,
+# we see previews of the intermediate results so we can check that it behaves
+# as we expect. However, if some estimators are slow, fitting them and
+# computing results on the whole data can slow us down.
+#
+# So we can tell skrub to subsample the data when computing the previews, with
```
GaelVaroquaux, line 30:
```suggestion
#
# Lightweight construction of the pipeline on a subsample
# -------------------------------------------------------------------------------------
# 
# We can tell skrub to subsample the data when computing the previews, with
```

examples/expressions/12_subsampling.py
```diff
@@ -0,0 +1,89 @@
...
+# We can also turn on subsampling for other methods of the expression, such as
+# :meth:`.skb.cross_validate() <Expr.skb.cross_validate>`. Here we run the
+# cross-validation on the small subsample of 100 rows we configured. With such
+# a small subsample the scores will be very low but this might help us quickly
+# detect errors in our cross-validation scheme.
+
+# %%
+predictions.skb.cross_validate(keep_subsampling=True)
+
+# %%
```
GaelVaroquaux, line 83:
```suggestion
# %%
# 
# Evaluating the pipeline on the full data
# --------------------------------------------------------
```

GaelVaroquaux (commented):
A few suggestions on the example.

These might well be my last comments on the PR. I think that I'm going to be ready to merge it

skrub/_expressions/_subsampling.py
jeromedockes: I added a comment for the constant and as we just discussed IRL I'll add high-level comments in the module containing the base class that apply to this class and other expression nodes

skrub/_expressions/_subsampling.py
jeromedockes: @GaelVaroquaux  I think this is the last outstanding comment. do you want to expose the seed as a parameter or keep it hard-coded to 0?

GaelVaroquaux: I don't want this seeded by default.

I would like a way for the user to globally control the seed / random number sequence for subsampling.

The way I would consider doing it is by having a dedicated rng, exposed as a global somewhere. But I'm open to other implementations of a solution.

jeromedockes: but if it's not seeded don't you think some users will shoot themselves in the foot by doing `X.skb.subsample(how='random')` and later `y.skb.subsample(how='random')` and ending up selecting different indices in X and y and breaking the correspondance between the 2?

GaelVaroquaux: Point taken.

OK, I think that a solution for now, without over-engineering, would be to expose a default "SUBSAMPLING_SEED" and use it.

Drawback is that it pollutes the skrub namespace. That would be an argument for giving up on my idea. As you wish

GaelVaroquaux: Skrub config is a good idea

GaelVaroquaux: Let's add this to an issue and move on

GaelVaroquaux:
LGTM. Merged! Thanks!!

PR #1336, opened 2025-04-22
fix missing return

jeromedockes:
There is a check that sees if a function called in an expression returns None as this probably indicates it modifies its input in-place instead of returning a new value. When it was decided it should emit a warning rather than raise an exception I forgot to add the return statement after removing the raise statement


jeromedockes:
thanks @Vincent-Maladiere

PR #1337, opened 2025-04-23
Small refactoring in expressions evaluation

jeromedockes:
this allows ExprImpl subclasses to define `eval` to override how they are evaluated. it simplifies things a bit by not needing to special case expressions that may not need all their children to be evaluated (eg an applied estimator does not need y for predict, if_else needs only one of the branches to be evaluated, ...)


rcap107 (approved):

jeromedockes:

Issue #1329, opened 2025-04-16
Set max_columns = 0 by default in patch_display and the expressions repr


jeromedockes:
I think the default for those tablereports that get triggered automatically should be the lightweight version without plots or correlations. when we want the full report we can either set those parameters or use `TableReport` directly

If we want a full tablereport for an expression's result, we can do `skrub.TableReport(my_expr.skb.eval())`. or we could also add a `.skb.table_report()` so that we can pass tablereport parameters to it. also, the tablereport config used in the expression's repr could be taken from patch_display if that is active -- but I would hold off on doing that because it's probably not worth the effort if we don't know how many users actually use the patch display. so my suggestion would be: just set those parameters to 0 for now

jeromedockes:
One thing we probably want to do is disable the message that says plotting has been skipped in that case. that seems too low-level to expose to users so we could

- have a private attribute `_extra_template_variables` containing additional variables to pass to the html template. skrub code could set no_max_columns_message = True in there and the message would not be shown
- instead of the `_extra_template_variables` attribute we could have a private subclass of the TableReport like LightReport that always sets that variable in the data passed to the template

I think I slightly prefer the first option to avoid introducing a new class and inheritance level

GaelVaroquaux:
> I think I slightly prefer the first option to avoid introducing a new class and inheritance level

Works for me

PR #1338, opened 2025-04-23
Removing distributions and associations tabs from TableReport in some cases

rcap107:
Addressing #1329 

After discussing with @jeromedockes, the objective of this PR is removing the "Distributions" and "Associations" tabs, as well as the associated computations, from the TableReport.

We also need to find a different place for the message informing the user that no plots are being made, and suppress the message when the TableReport is generated for an expression.


rcap107:
closed by #1340

PR #1339, opened 2025-04-24
better handling of integer columns in parallel coord plots

jeromedockes:
when rounding and converting to int we could end up with duplicate ticks which plotly does not like

before

screenshot_2025-04-24T15:19:43+02:00

now

screenshot_2025-04-24T15:20:10+02:00

(note the choice of ticks still could be improved in the case where we are with integers, in log scale, but a very small range as seen here -- it would be more visually pleasing if we also had a tick at 10, but that's less of an issue)


Vincent-Maladiere (approved):
A nice improvement!

jeromedockes:
the example is like this now:

screenshot_2025-04-24T17:03:48+02:00

jeromedockes:
@Vincent-Maladiere thanks for the review; I added a couple more changes to ensure the jitter doesn't modify the data range and that int columns have integer ticks -- if you don't mind having another look that would be very helpful

skrub/_expressions/_parallel_coord.py
```diff
@@ -90,14 +100,21 @@ def _prepare_obj_column(col):
     }
 
 
-def _prepare_numeric_column(col, log_scale):
+def _prepare_numeric_column(col, log_scale, is_int):
     vals = col.to_numpy()
     if log_scale:
         vals = np.log(vals)
     min_val, max_val = np.nanmin(vals), np.nanmax(vals)
     tickvals = np.linspace(min_val, max_val, 10).tolist()
```
glemaitre, line 108: Maybe we could delay the conversion to list and instead keep it as a NumPy array and make the conversion after the if/else block below.

skrub/_expressions/_parallel_coord.py
```diff
@@ -90,14 +100,21 @@ def _prepare_obj_column(col):
...
-    if pd.api.types.is_integer_dtype(col):
-        ticktext = [str(int(np.round(np.exp(v)))) for v in tickvals]
+    if pd.api.types.is_integer_dtype(col) or is_int:
+        if log_scale:
+            tickvals = np.exp(tickvals)
+        tickvals = np.unique(np.round(tickvals).astype("int"))
```
glemaitre, line 112: Just to be sure that it is not OS dependent (hopefully NumPy 2 does not do that anymore)

```suggestion
        tickvals = np.unique(np.round(tickvals).astype(np.int64))
```

skrub/_expressions/_parallel_coord.py
```diff
@@ -90,14 +100,21 @@ def _prepare_obj_column(col):
...
+        tickvals_label_space = tickvals.tolist()
+        if log_scale:
+            tickvals = np.log(tickvals)
```
glemaitre, line 115:
```suggestion
            # cannot make inplace log since `tickval` is of integral type
            tickvals = np.log(tickvals)
```

skrub/_expressions/_parallel_coord.py
```diff
@@ -90,14 +100,21 @@ def _prepare_obj_column(col):
...
+        tickvals = tickvals.tolist()
+        ticktext = list(map(str, tickvals_label_space))
     else:
         ticktext = list(
             map("{:.2g}".format, np.exp(tickvals) if log_scale else tickvals)
```
glemaitre, line 120: Instead of the `map` and because we call list, I would just do a list comprehension:

```python
ticktext = [f"{np.exp(val) if log_scale else val:.2g}" for val in tickvals]
```

glemaitre (commented):
It looks good. Just a couple of quick changes but feel free to adapt.

skrub/_expressions/_parallel_coord.py
jeromedockes: Here tickvals is an array with only 10 elements, so I would prefer not adding the mental charge of having to think about the implications of in-place computation and the use of `out=input_buffer` because the optimization is not worth it for 10 elements. what do you think?

jeromedockes:
Thanks for the review @glemaitre ! I think I addressed most comments, except I would prefer not to have to think of the implications of using numpy's `out` parameter for an array of size 10

glemaitre (approved):

PR #1341, opened 2025-04-24
Make the selectors module public

### LINKED ISSUE #1146: Making selectors public
In https://github.com/skrub-data/skrub/pull/1116, we discussed allowing users to use skrub selectors for the `cols` parameter. Selecting columns is very frequent in skrub transformers, and they might be useful for users. They are currently private, if we want to render them public, we need to:

- [ ] Document them. Their docstrings are already quite exhaustive but we need a dedicated example / reference in the doc
- [ ] Define for which parameters / transformers we allow selectors
- [ ] Support them in those cases (`Joiner`, `AggJoiner`, `MultiAggJoiner`, `SelectCols`, `DropCols`, `TableVectorizer`, etc)

jeromedockes:
closes #1146 

It is already exposed as a public attribute of skrub as they can be used with expressions so 

```python
from skrub import selectors as s
```

already works on the main branch

however the actual module is called `_selectors` so 

```python
import skrub.selectors
```

raises an importerror

this pr renames it to `selectors`

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,7 +30,7 @@ New features
 - The :mod:`selectors` module provides utilities for selecting columns to which
   a transformer should be applied in a flexible way. The module was created in
   :pr:`895` by :user:`Jérôme Dockès <jeromedockes>` and added to the public API
-  in :pr:`TODO` by :user:`Jérôme Dockès <jeromedockes>`.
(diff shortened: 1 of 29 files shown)


rcap107 (approved):

glemaitre (approved):
LGTM

glemaitre:
Merging this one. We could subsequently improve the documentation.

Issue #1342, opened 2025-04-24
Small feedback on Skrub Expressions


davidgasquez:
### Problem Description

Love the idea of `skrub` expressions. I've played with them a little bit and found two small things that I though about raising.

1. Pylance is not happy with `Expr`. It'll keep raising linter errors, even in the sample code, there is a `Operator "+" not supported for types "Expr" and "Expr"`.
 
```python
a = skrub.var("a", 10)
b = skrub.var("b", 6)
c = a + b
```

2. Cell output should be flexible and allow to show the original type. I like the TableReport that is automatically generated. That said, there are times where I'd love to just keep it in the original type (e.g: pandas df in the screenshot case).

![Image](https://github.com/user-attachments/assets/9f907a5f-6ffb-4da3-a20c-b3f70357c36a)

### Feature Description

1. Would be awesome to have better Pylance support.
2. Have a way to opt out of the TableReport output for Expr variables

--- 

Thanks for the work on `skrub`! It has replaced many in small in house scripts and been fun to play with.

jeromedockes:
Hi @davidgasquez , thank you very much for giving it a try and for the feedback :)

- 1 In order to work with inputs of any type the expressions are very "dynamic", so it will always be hard for static type checkers to analyze a program that manipulates skrub expressions. However there are probably some things we can do to improve the user experience. The first low-hanging fruit that comes to my mind is to explicitly annotate the functions that create expressions (such as `skrub.var`, `skrub.as_expr` etc.) with a `Any` return type. The checker will still not be able to detect genuine errors, but at least that should silence the spurious errors. For example if I annotate `skrub.var(...) -> Any` the pyright "Operator + not supported" error disappears for me for the provided example script. I'll open a PR for that tomorrow, happy to hear other ideas anyone may have.
- 2
  - With #1340 there will be no plots or pairwise column associations in the reports shown for expressions, which speeds them up considerably (not sure if the time it takes to generate the report was one of the reasons you want to opt out).
  - It is not as convenient but you can always retrieve the actual result of the evaluation by calling `eval`, and then either display it or manipulate it however you want. In the example you show you could replace `orders` in the last line with `orders.skb.eval()` and this will return the actual dataframe, and you would get the usual pandas display for the output of the cell.
[...]

PR #1343, opened 2025-04-25
Add stub file for `_expressions._expressions`

jeromedockes:
addresses part of #1342 

adding the pyi file tells static type checkers that many operators are defined for the Expr class even if they fail to figure it out from the module's text

the PR also adds a pixi task and github action to check if the pyi file needs to be updated


pyproject.toml
```diff
@@ -190,11 +190,15 @@ clean-doc = { cmd = "rm -rf _build/ && rm -rf auto_examples/ && rm -rf reference
 [tool.pixi.feature.lint.tasks]
 lint = { cmd = "pre-commit install && pre-commit run -v --all-files --show-diff-on-failure" }
 
+[tool.pixi.feature.check-pyi-diff.tasks]
+check = { cmd = "python build_tools/generate_expressions_stub.py | diff - skrub/_expressions/_expressions.pyi" }
```
rcap107, line 194: I don't understand this operation: from what I can see, this is first  running `generate_expressions_stub.py`, then piping the output from stdout to the `diff` command, and it's comparing it to `skrub/_expressions/_expressions.pyi`

Doesn't this compare a file to itself? I am missing where we're supposed to modify the code in case new functions get added

jeromedockes: the file is created by manually running the command and writing its output in the file. this compares the output of the command to the content of the file. so if something changed in skrub code that causes the output of the script to change (eg a function was added) this tells you "the automated output has changed but not the file, you probably need to update the file or to update the script (eg to explicitly ignore the added function)"

jeromedockes: it is not a precommit hook (it only runs on the CI) and it does not actually modify the file. it will fail and tell you that you need to manually re-run the script and write its output to the file (possibly after editing the script)

the reason this has to be checked and done manually is that the script may not be able to figure out the correct type hint for the added function

rcap107: got it, thanks for the clarification

jeromedockes: I'll add a comment in the script

glemaitre: It might be nice to add a comment to note that it only works in Unix system. Alternatively, we could have another Python script to compute the diff but I'm not sure it is worth.

.github/workflows/check_stub_files_diff.yaml
```diff
@@ -0,0 +1,24 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v4
+      - uses: prefix-dev/setup-pixi@v0.8.8
+        with:
+          pixi-version: v0.41.3
+          frozen: true
+
+      - name: Check diff
```
glemaitre, line 23:
```suggestion
      - name: Check stub file for `_expression.py` is up-to-date
```

glemaitre (approved):

jeromedockes:
Thanks @glemaitre . given that this does not touch the code but only stub file & CI, and that the longer it is in main the more chance we have on getting feedback on UX in IDEs I say we merge it and see what happens?

GaelVaroquaux (approved):
Approving and merging (as suggested by @jeromedockes : merging fast to iterate)

PR #1344, opened 2025-04-25
DISCUSSION change Expr.skb.cross_validate?

jeromedockes:
As discussed with @glemaitre , @rcap107 , @jovan-stojanovic and @Vincent-Maladiere  having 2 `cross_validate` and 2 `train_test_split` is confusing.

this removes one of each. Now we must do

```python
skrub.cross_validate(e.skb.get_estimator(), e.skb.get_data())
```

(and for train_test_split `e.skb.train_test_split()`)


GaelVaroquaux:
I must say that I find this quite disappointing: we're forcing much typing (and thus much errors for non advanced users) for a very common need.

I agree with the sentiment that the solution that we had was suboptimal (somewhat confusing API). I think that we need to find a new one. Maybe it was just a question of playing with the names.

glemaitre:
> I think that we need to find a new one. Maybe it was just a question of playing with the names.

To me it is more a problem of API. The fact that you can easily missed that you don't use the search CV was a compelling example that something is wrong.

In terms of naming, I'm wondering about something: if I'm not mistaken we usually refer to "cross-validate a model" rather than data. I'm wondering if don't miss a paradigm to mention that we apply a search-cv on the data or a nested search-cv.

jeromedockes:
> we're forcing much typing (and thus much errors for non advanced users) for a very common need.

Indeed that was the reason for having `Expr.skb.cross_validate` in the first place. Here are some other options we discussed, none of which were to everybody's liking:

1. keep things as they are, with both `skrub.cross_validate(e.skb.get_pipeline(), e.skb.get_data())` and the shorthand `e.skb.cross_validate()`
2. keep things as they are but rename `e.skb.cross_validate()` to `e.skb.cross_validate_default()`
3. remove `.skb.cross_validate` (this PR)
4. remove `skrub.cross_validate` and add an `estimator` argument to `.skb.cross_validate`. ATM it uses the expression's data by default but we can replace it by passing a new value. This would allow doing the same thing for the estimator. So `e.skb.cross_validate()` would be like `e.skb.cross_validate(e.skb.get_estimator(), e.skb.get_data())` and we could write for example `e.skb.cross_validate(estimator=e.skb.get_randomized_search(cv=10))`
5. add a `cross_validate` method on the generated pipeline / param search. We would write `e.skb.get_randomized_search(cv=10).cross_validate(e.skb.get_data())`. (note the randomized search does not have a reference to the data so the `get_data()` is not optional here)
6. remove `skrub.cross_validate` and keep `e.skb.cross_validate()` as it is (which makes nested cross-validation impossible)

To remind the context, the 2 issues that were raised about `.skb.cross_validate()` are:

- It is not obvious that we are cross-validating the default pipeline, without any hyperparameter search
- It means there are 2 different `cross_validate` functions (`skrub.cross_validate` and `e.skb.cross_validate` which is a shortcut for `skrub.cross_validate` when we don't need nested CV)

jeromedockes:
I'll move the removal of `skrub.train_test_split` to a different PR because that probably needs less discussion (`skrub.train_test_split` does not do anything that `e.skb.train_test_split` cannot do)

jeromedockes:
> I'm wondering if don't miss a paradigm to mention that we apply a search-cv on the data or a nested search-cv.

sounds interesting but I'm not sure I understand could you elaborate a bit?

jeromedockes:
another option (similar to 4. but perhaps looks a bit less weird) could be something like `skrub.cross_validate(expr=None, *, pipeline="default", environment="default")` where `expr` must be provided whenever `pipeline` or `environment` is left to "default" (and is unused when both `pipeline` and `environment` are provided). So the simple case would be `skrub.cross_validate(e)` instead of `e.skb.cross_validate()`. when looking at the function call it is still not obvious that the default pipeline is used but at least the word "default" is explicitly in the function signature


More generally I think having a word like "default" in the function name or signature (eg as in option 2.  above) may be the only way to satisfy both requirements "I shouldn't need to type the code that extracts the default pipeline" and "It should be obvious that I'm using the default pipeline"

Vincent-Maladiere:
I'm +1 for `skrub.cross_validate(expr=None, *, pipeline="default", environment="default")`

jeromedockes:
> another option (similar to 4. but perhaps looks a bit less weird) could be something like `skrub.cross_validate(expr=None, *, pipeline="expr.get_pipeline()", environment="expr.get_data()")` where `expr` must be provided whenever `pipeline` or `environment` is left to the default (and is unused when both `pipeline` and `environment` are provided). So the simple case would be `skrub.cross_validate(e)` instead of `e.skb.cross_validate()` and the nested cv would remain `skrub.cross_validate(pipeline=e.skb.get_randomized_search(), environment=e.skb.get_data())`. when looking at the function call it is still not obvious that the default pipeline is used but at least it is clear what is being used when looking at the signature.

@rcap107 , @Vincent-Maladiere and I are ok with this solution so I'll update the PR with it

PR #1348, opened 2025-04-28
remove skrub.train_test_split

jeromedockes:
as it is redundant with `Expr.skb.train_test_split`


Vincent-Maladiere (approved):
Nice simplification!!

jeromedockes:
thanks @Vincent-Maladiere . I don't think we need more than 1 review for this one

PR #1357, opened 2025-04-30
use doctests in skrub pipeline user guide

jeromedockes:
This adds the prompts recognized by doctest in the user guide section about expressions.
The main advantage is that we can check the code runs and produces the expected output with pytest so there is less chance that it becomes outdated.

also users are probably used to seeing such snippets (eg in the reference documentation, scikit-learn user guide, ...) so it might make it slightly easier to immediately recognize the code and the outputs

also adds a step in the tests CI workflow to run those doctests

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -36,6 +36,9 @@ jobs:
       - name: Run tests
         run: pixi run -e ${{ matrix.environment }} test -n auto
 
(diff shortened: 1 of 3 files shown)


rcap107 (approved):
Looks great, this will be very useful as we improve the user guide

jeromedockes:
thanks!

PR #1358, opened 2025-04-30
Adding `datetime_format` parameter to TableVectorizer and Cleaner for handling custom formats

rcap107:


jeromedockes (approved):
Thanks a lot @rcap107 ! super useful when we have weirdly formatted datetimes (which is all the time :D )

PR #1361, opened 2025-05-02
Add  a `default` parameter to choices

jeromedockes:
~Not sure if we want this~, but we discussed it once or twice with @Vincent-Maladiere : allowing to specify an explicit default value for the choices.

This PR adds a new parameter `default`. If provided, it is used in the default pipeline. It does not affect the parameter grid so for example it has no effect on the behavior of `.skb.get_randomized_search()` and `.skb.get_grid_search()`.

```
>>> import skrub

Normally the default is the middle of the range

>>> skrub.choose_float(0.0, 10.0).as_expr().skb.eval()
np.float64(5.0)

But we can provide our own default value instead

>>> e = skrub.choose_float(0.0, 10.0, name='c', default=2.0).as_expr()
>>> e.skb.eval()
2.0
>>> e.skb.get_pipeline().fit_transform({})
2.0

The param grid and hyperparameter search don't change

>>> print(e.skb.describe_param_grid())
- c: choose_float(0.0, 10.0, name='c', default=2.0)
```

### DIFF
--- a/skrub/_expressions/_choosing.py
+++ b/skrub/_expressions/_choosing.py
@@ -399,12 +399,12 @@ def choose_from(outcomes, *, name=None, default=NULL):
 
     Normally the default value is the first item:
 
-    >>> choose_from([1, 2, 3]).as_expr().skb.eval()
+    >>> choose_from([1, 2, 3]).default()
     1
 
     But we can override it with a different value:
 
-    >>> choose_from([1, 2, 3], default=2).as_expr().skb.eval()
+    >>> choose_from([1, 2, 3], default=2).default()
     2
     """
     if isinstance(outcomes, typing.Mapping):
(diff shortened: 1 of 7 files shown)


jeromedockes:
after discussion IRL we decided we do want this so I'll work on adding tests now

skrub/_expressions/_choosing.py
```diff
@@ -424,34 +438,93 @@ class Optional(Choice):
...
+            (value,),
+            {"name": self.name, "default": default},
+            defaults={"name": None, "default": OPTIONAL_VALUE},
         )
         return f"optional({args})"
 
 
-def optional(value, *, name=None):
-    """Construct a choice between a value and ``None``.
+def optional(value, *, name=None, default=OPTIONAL_VALUE):
```
jeromedockes, line 454: note: this is displayed as

screenshot_2025-05-06T15:01:09+02:00

jeromedockes:
this one is ready for review

GaelVaroquaux (approved):
LGTM

GaelVaroquaux:
Other reviewers: feel free to merge this one any time

PR #1364, opened 2025-05-02
Better error message when passing a function to `.skb.apply`

jeromedockes:
It is a likely mistake to use `.skb.apply()` when we really meant to use `.skb.apply_func()`. The current error message is something like `AttributeError: 'function' object has no attribute 'fit'`. This PR improves the error message by mentioning `apply_func`. It also checks for the case where we passed the class when we meant to pass an instance `TableVectorizer` instead of `TableVectorizer()`

### DIFF
--- a/skrub/_expressions/tests/test_errors.py
+++ b/skrub/_expressions/tests/test_errors.py
@@ -388,6 +388,13 @@ def test_apply_bad_type():
         skrub.X(0).skb.apply(...)
 
 
+def test_apply_bad_string():
+    with pytest.raises(
+        Exception, match=r".*should be `None`, the string 'passthrough' or"
+    ):
+        skrub.X(0).skb.apply("pass through")
+
+
 def test_method_call_failure():
     with pytest.raises(
         Exception,


skrub/_expressions/_expressions.py
```diff
@@ -628,7 +628,26 @@ def _check_wrap_params(cols, how, allow_reject, reason):
         raise ValueError(msg)
 
 
+def _check_estimator_type(estimator):
+    if inspect.isclass(estimator):
+        raise TypeError(
+            "Please provide an estimator instance to `apply`, rather than the class. "
```
GaelVaroquaux, line 634: Maybe it should say "a scikit-learn like estimator"?

GaelVaroquaux:
Sooooo useful!!! Thanks

Vincent-Maladiere:
That's a good catch. Shouldn't we have `apply_est/apply_func` rather than `apply/apply_func` then? WDYT?

jeromedockes:
> That's a good catch. Shouldn't we have apply_est/apply_func rather than apply/apply_func then? WDYT?

maybe... although it may not be ovious what "est" stands for ... we could do apply_estimator but that's making a very frequently used method much longer. I guess it depends a bit if we think apply_func will be used very frequently or not

jeromedockes:
In the examples we have so far we use `apply` a lot more than `apply_func` so it would seem ok to keep the short name for the estimators, but maybe in reality it will be more balanced

rcap107:
I checked

rcap107:
Out of curiosity I checked the documentation of pandas and polars to see if there is an equivalent to apply. Pandas has `apply`, but it also has one `applymap`. Polars does not have `apply`, but it has `map`, as well as a bunch of other functions that start with `map`.

I think it's fine to have `apply`/`apply_func` because I also think that people will use `apply` more often.

rcap107 (approved):
Very useful, thanks!

Issue #1282, opened 2025-04-05
Error in "Handling datetime features with the DatetimeEncoder" Example + Minor Corrections


semyonbok:
The example juxtaposes the performance of three regression models that use different configurations of `DatetimeEncoder`. These lines highlight that the one with spline feature pre-processing is better:

https://github.com/skrub-data/skrub/blob/00a6d709ddc20560e2dbb1dd2860b6ef66121d89/examples/03_datetime_encoder.py#L221-L227

Please notice that spline periodic encoding actually yields significantly worse RMSE. Its performance on the first two splits is preposterous, as visualized below. The DataViz also shows RMSE on entire test set, as well as on its last 96 entries: the spline pipeline keeps under-performing. 

![Image](https://github.com/user-attachments/assets/42846488-4af7-4815-9aad-9efa775badcc)

Also, a couple of minor corrections should be done:
* just "don't want to" is meant, and `resolution` is not explicitly set to "hour" in the preceding code example:
https://github.com/skrub-data/skrub/blob/00a6d709ddc20560e2dbb1dd2860b6ef66121d89/examples/03_datetime_encoder.py#L89-L92
* slight improvements to prediction plotting can be made:
![Image](https://github.com/user-attachments/assets/b3c815d8-bd94-40a9-b177-c8f90bd40957)

rcap107:
Closing this to avoid duplication, this issue is being tracked in #1629

PR #1366, opened 2025-05-02
Reworking example 3 on the DatetimeEncoder to include expressions and lagged features 

rcap107:
First version, needs editing cleanup and double checking the section on lagged features.

### DIFF
--- a/examples/03_datetime_encoder_expressions.py
+++ b/examples/03_datetime_encoder_expressions.py
@@ -1,0 +1,473 @@
+"""
+.. _example_datetime_encoder :
+
+===================================================
+Handling datetime features with the DatetimeEncoder
+===================================================
+
+In this example, we illustrate how to better integrate datetime features
+in machine learning models with the |DatetimeEncoder|.
+
+This encoder breaks down passed datetime features into relevant numerical
+features, such as the month, the day of the week, the hour of the day, etc.
+
+It is used by default in the |TableVectorizer|.
+
+
+.. |DatetimeEncoder| replace::
+    :class:`~skrub.DatetimeEncoder`
+
+.. |TableVectorizer| replace::
+    :class:`~skrub.TableVectorizer`
+
+.. |OneHotEncoder| replace::
+    :class:`~sklearn.preprocessing.OneHotEncoder`
+
+.. |TimeSeriesSplit| replace::
+    :class:`~sklearn.model_selection.TimeSeriesSplit`
+
+.. |ColumnTransformer| replace::
+    :class:`~sklearn.compose.ColumnTransformer`
+
+.. |make_column_transformer| replace::
+    :class:`~sklearn.compose.make_column_transformer`
+
+.. |RidgeCV| replace::
+    :class:`~sklearn.linear_model.RidgeCV`
+
+.. |SimpleImputer| replace::
+    :class:`~sklearn.impute.SimpleImputer`
+
+.. |StandardScaler| replace::
+    :class:`~sklearn.preprocessing.StandardScaler`
+
+.. |ToDatetime| replace::
+    :class:`~skrub.ToDatetime`
+
+.. |var| replace::
+    :meth:`skrub.var`
+
+.. |apply_func| replace::
+    :meth:`skrub.Expr.skb.apply_func`
+
+.. |train_test_split| replace::
+    :meth:`skrub.Expr.skb.train_test_split`
+
+.. _scikit-learn example:
+ `<https://scikit-learn.org/stable/auto_examples/applications/plot_time_series_lagged_features.html#generating-polars-engineered-lagged-featurehttps://scikit-learn.org/stable/auto_examples/applications/plot_time_series_lagged_features.html#generating-polars-engineered-lagged-featuress>`_
+
+"""
+
+# %%
+# A problem with relevant datetime features
+# -----------------------------------------
+#
+# We will use a dataset of bike sharing demand in 2011 and 2012.
+# In this setting, we want to predict the number of bike rentals, based
+# on the date, time and weather conditions.
+# In this example, we will use Polars dataframes instead of Pandas.
+
+from pprint import pprint
+
+import polars as pl
+
+import skrub
+from skrub import TableReport, datasets
+
+data = datasets.fetch_bike_sharing().bike_sharing
+data = pl.from_pandas(data)
+TableReport(data)
+
+###############################################################################
+# %% We can convert the dataframe's ``"date"`` column from string to datetime
+# using |ToDatetime|.
+
+from skrub import ToDatetime
+
+transformed_dates = ToDatetime().fit_transform(data["date"])
+
+print(
+    "original dtype:",
+    data["date"].dtype,
+    "\n\nconverted dtype:",
+    transformed_dates.dtype,
+)
+# %%
+###############################################################################
+# Encoding the features
+# .....................
+#
+# For the sake of the example, we will now encode the ``date`` column
+# with a |DatetimeEncoder| to show the operations that are executed by the
+# encoder. In practice, the ``DatetimeEncoder`` would be passed to a
+# |TableVectorizer| object, as the latter will handle parsing datetimes
+# and perform additional cleaning steps.
+#
+# During the instantiation of the |DatetimeEncoder|, we specify that we want
+# don't want to extract features with a resolution finer than hours. This is
+# because we don't want to extract minutes, seconds and lower units, as they
+# are unimportant here.
+
+from skrub import DatetimeEncoder
+
+# DatetimeEncoder has "hour" as default resolution
+date_enc = DatetimeEncoder().fit_transform(transformed_dates)
+
+print(transformed_dates, "\n\nHas been encoded as:\n\n", date_enc)
+# %%
+###############################################################################
+# We see that the encoder is working as expected: the column has
+# been replaced by features extracting the month, day, hour, and day of the
+# week. Additionally, the total seconds since Epoch are also added by default.
+
+###############################################################################
+# One-liner with the TableVectorizer
+# ..................................
+#
+# As mentioned earlier, the |TableVectorizer| makes use of the
+# |DatetimeEncoder| by default. Note that ``X["date"]`` is still
+# a string, but will be automatically transformed into a datetime in the
+# |TableVectorizer|.
+
+from skrub import TableVectorizer
+
+table_vec = TableVectorizer().fit(data)
+pprint(table_vec.get_feature_names_out())
+
+# %%
+###############################################################################
+# If we want to customize the |DatetimeEncoder| inside the |TableVectorizer|,
+# we can replace its default parameter with a new, custom instance.
+#
+# Here, for example, we want it to extract the day of the week:
+
+# use the ``datetime`` argument to use a custom DatetimeEncoder in the TableVectorizer
+table_vec_weekday = TableVectorizer(datetime=DatetimeEncoder(add_weekday=True)).fit(
+    data
+)
+pprint(table_vec_weekday.get_feature_names_out())
+
+###############################################################################
+# .. note:
+#     For more information on how to customize the |TableVectorizer|, see
+#     :ref:`sphx_glr_auto_examples_01_dirty_categories.py`.
+#
+# Inspecting the |TableVectorizer| further, we can check that the
+# |DatetimeEncoder| is used on the correct column(s).
+pprint(table_vec_weekday.transformers_)
+
+###############################################################################
+#
+# Feature engineering for linear models
+# ....................................................................
+#
+# The |DatetimeEncoder| can generate additional periodic features. These are
+# particularly useful for linear models. This is controlled by the
+# ``periodic encoding`` parameter which can be either  ``circular`` or ``spline``,
+# for trigonometric functions or B-Splines respectively. In this example, we use
+# ``spline``.
+
+table_vec_periodic = TableVectorizer(
+    datetime=DatetimeEncoder(add_weekday=True, periodic_encoding="spline")
+).fit(data)
+table_vec_periodic.transform(data)
+
+# Spline features for each part of the date (month in year, day in month, day in
+# week, day in year) are added.
+
+# %%
+###############################################################################
+# Prediction with datetime features
+# ---------------------------------
+#
+# In this section, we will use skrub Expressions to build a predictive pipeline
+# that pre-processes the data and performs hyperparameter optimization for various
+# of the steps involved in the preparation.
+# We will use a |RidgeCV| model as our learner. While it is not the most powerful
+# model, it will help us illustrate how feature engineering affects the prediction
+# performance.
+
+from sklearn.impute import SimpleImputer
+from sklearn.linear_model import RidgeCV
+
+# As we are working with timeseries, we need to use |TimeSeriesSplit| to perform
+# crossvalidation.
+from sklearn.model_selection import TimeSeriesSplit
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
+
+ts = TimeSeriesSplit()
+
+ts = TimeSeriesSplit(
+    n_splits=3,  # to keep the notebook fast enough on common laptops
+    gap=48,  # 2 days data gap between train and test
+    max_train_size=10000,  # keep train sets of comparable sizes
+    test_size=3000,  # for 2 or 3 digits of precision in scores
+)
+# We define a skrub |var| to start with.
+data_var = skrub.var("data", data)
+
+# We extract our input data (X) and the target column (y), and mark them as X and y.
+X = data_var[
+    "date", "holiday", "weathersit", "temp", "hum", "windspeed"
+].skb.mark_as_X()
+y = data_var["cnt"].skb.mark_as_y()
+X
+
+# %%
+# Now we define a simple default pipeline for the |RidgeCV| predictor, which includes
+# a |StandardScaler| for numerical and a |SimpleImputer|. We will not modify this
+# pipeline: we will instead focus on the pre-processing and feature engineering.
+default_ridge_pipeline = make_pipeline(StandardScaler(), SimpleImputer(), RidgeCV())
+
+# The "base" variant of the pipeline uses the default |TableVectorizer|, with
+# no hyperparameter search.
+vectorized = X.skb.apply(TableVectorizer())
+predictions_base = vectorized.skb.apply(default_ridge_pipeline, y=y)
+search_base = predictions_base.skb.get_grid_search(fitted=True, cv=ts)
+
+# %% We can observe the results directly using ``.detailed_results``. Unsurprisingly,
+# the results are not very good. In this case, performing a GridSearch is not
+# necessary, but we will do it for consistency with the next steps.
+search_base.detailed_results_
+
+# %%
+# Prediction with periodic encoders and hyperparameter optimization
+# -----------------------------------------------------------------
+#
+# For our second pipeline, we define a grid of hyperparameters using the skrub
+# ``choose_*`` functions. In this case, we use |choose_bool| and |choose_from|,
+# as we are not interested in tweaking numerical parameters.
+# The ``choose_*`` functions allow to easily prepare a grid of parameters to use
+# to perform hyperparameter optimization.
+# Here, we use them to turn on and off the ``weekday`` and ``total_seconds`` flag,
+# and to select the specific periodic encoder to use.
+datetime_encoder = DatetimeEncoder(
+    add_weekday=skrub.choose_bool(name="weekday"),
+    add_total_seconds=skrub.choose_bool(name="total_seconds"),
+    periodic_encoding=skrub.choose_from(
+        [
+            "spline",
+            "circular",
+            None,
+        ]
+    ),
+)
+
+vectorized = X.skb.apply(TableVectorizer(datetime=datetime_encoder))
+predictions_enc = vectorized.skb.apply(default_ridge_pipeline, y=y)
+
+# Note that, in general, randomized search should be used instead of grid search.
+# In this case, grid search is fine as we are interested in a grid of categorical
+# values.
+search_enc = predictions_enc.skb.get_grid_search(fitted=True, cv=ts)
+# %%
+search_enc.detailed_results_
+# %%
+# We can observe that the prediction results have improved a lot thanks to the
+# introduction of the periodic features, however they are still not very good.
+# We can also see that setting ``total_seconds`` to ``True`` seems to consistently
+# reduce the test score.
+#
+# Adding lagged features
+# ----------------------
+#
+# The prediction results are not very good, so in this section we will extend the
+# feature generation section to include lagged features.
+# This function is taken from a similar `scikit-learn example`_. As the lagged
+# features are based on the actual count, we go back to the original data to
+# perform feature engineering on that.
+
+
+# To introduce the lagged features, we use a |deferred| function that takes the
+# dataframe we are working with, then adds lagged features to it. Given a sample
+# in the dataset, lagged features add information relative to samples prior to it.
+# In this case, for each hourly sample we add information relative to the previous
+# three hours, the same hour on the prior day, as well as aggregate features
+# (mean, max, min) obtained by using a rolling mean of either 24 hours, or 7 days.
+@skrub.deferred
+def get_lagged_features(df):
+    lagged_df = df.select(
+        "cnt",
+        *[pl.col("cnt").shift(i).alias(f"lagged_count_{i}h") for i in [1, 2, 3]],
+        lagged_count_1d=pl.col("cnt").shift(24),
+        lagged_count_1d_1h=pl.col("cnt").shift(24 + 1),
+        lagged_count_7d=pl.col("cnt").shift(7 * 24),
+        lagged_count_7d_1h=pl.col("cnt").shift(7 * 24 + 1),
+        lagged_mean_24h=pl.col("cnt").shift(1).rolling_mean(24),
+        lagged_max_24h=pl.col("cnt").shift(1).rolling_max(24),
+        lagged_min_24h=pl.col("cnt").shift(1).rolling_min(24),
+        lagged_mean_7d=pl.col("cnt").shift(1).rolling_mean(7 * 24),
+        lagged_max_7d=pl.col("cnt").shift(1).rolling_max(7 * 24),
+        lagged_min_7d=pl.col("cnt").shift(1).rolling_min(7 * 24),
+    )
+    return lagged_df
+
+
+data_prep = data_var.drop("casual", "instant", "registered").skb.mark_as_X()
+data_lagged = data_prep.skb.apply_func(get_lagged_features).drop("cnt")
+X = data_prep.skb.concat([data_lagged], axis=1).drop("cnt")
+X
+# Notice that adding lagged samples will introduce null values for all the initial
+# samples, as there is no previous information for them.
+
+y = data_var["cnt"].skb.mark_as_y()
+# %% We use the datetime encoder defined above to observe the impact of the new
+# features on the predictions.
+
+vectorized = X.skb.apply(TableVectorizer(datetime=datetime_encoder))
+predictions_lagged = vectorized.skb.apply(default_ridge_pipeline, y=y)
+search_lagged = predictions_lagged.skb.get_grid_search(fitted=True, cv=ts)
+# %%
+search_lagged.detailed_results_
+
+# We can see that the lagged features improved the prediction performance by a
+# large margin. Periodic features bring some benefit.
+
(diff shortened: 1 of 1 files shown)


rcap107:
After discussing IRL about this example, we decided that it should be a stand-alone example about handling timeseries with skrub, and that we should work on a different (simpler) example to showcase the datetime encoder

GaelVaroquaux:
> it should be a stand-alone example about handling timeseries with skrub

"timeseries forecasting", to be precise

rcap107:
I am working on this example again and I was wondering if I should cut most of the content that's relevant specifically to the DatetimeEncoder and focus only on the forecasting part. 

The idea was having a simpler example that focuses only on the datetime encoder, so all the detail about it would be in the first example, while this would be more on how to combine the expressions with the encoder, and the lagged features.

Vincent-Maladiere:
> The idea was having a simpler example that focuses only on the datetime encoder, so all the detail about it would be in the first example, while this would be more on how to combine the expressions with the encoder, and the lagged features.

Looks like a good approach!

rcap107:
For the record, I added the KBin discretizer to have "another method" to play with the expressions, but I don't have strong feelings about keeping it or removing it

GaelVaroquaux:
> I am working on this example again and I was wondering if I should cut most of the content that's relevant specifically to the DatetimeEncoder and focus only on the forecasting part.

> The idea was having a simpler example that focuses only on the datetime encoder, so all the detail about it would be in the first example, while this would be more on how to combine the expressions with the encoder, and the lagged features.

Sounds reasonable to me!

GaelVaroquaux:
> I can't figure out where is the configuration that looks for the examples, so now that I moved the new example to the expression directory it doesn't get rendered 🤔

Did you find it? I believe that it is a regexp in the conf.py, in the sphinx-gallery configuration

rcap107:
> Did you find it? I believe that it is a regexp in the conf.py, in the sphinx-gallery configuration

In the end, I think I was looking at the wrong generated docs, now it's rendering the example properly 

It's why I removed the comment, next time I'll edit it instead

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+.. _scikit-learn example:
+ `<https://scikit-learn.org/stable/auto_examples/applications/plot_time_series_lagged_features.html#generating-polars-engineered-lagged-featurehttps://scikit-learn.org/stable/auto_examples/applications/plot_time_series_lagged_features.html#generating-polars-engineered-lagged-featuress>`_
+
+"""
+
+# %%
+# A problem with relevant datetime features
+# -----------------------------------------
+#
+# We will use a dataset of bike sharing demand in 2011 and 2012.
```
glemaitre:
```suggestion
# We use a dataset of bike sharing demand in 2011 and 2012.
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# In this setting, we want to predict the number of bike rentals, based
+# on the date, time and weather conditions.
+# In this example, we will use Polars dataframes instead of Pandas.
```
glemaitre:
```suggestion
# In this example, we use Polars dataframes instead of Pandas.
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+import polars as pl
+
+from skrub import TableReport, datasets
+
+data = datasets.fetch_bike_sharing().bike_sharing
+data = pl.from_pandas(data)
+TableReport(data)
```
glemaitre, line 98: I would add a small and succinct description of what I should look at or takeaway from the view on the table report. So for instance, since we are interested in the "time" part, we can mention to people to have particular attention to the "date" column and that if they check the "Stats" tab, they will see that it is a "String" column.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+# %%
+# Prediction with datetime features
+# ---------------------------------
+#
+# In this section, we will use skrub Expressions to build a predictive pipeline
```
glemaitre: For the remaining of the example, use the present instead of future (I always do the same mistake).

```suggestion
# In this section, we use skrub Expressions to build a predictive pipeline
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# that pre-processes the data and performs hyperparameter optimization for various
+# of the steps involved in the preparation.
+# We will use a |RidgeCV| model as our learner. While it is not the most powerful
+# model, it will help us illustrate how feature engineering affects the prediction
+# performance.
+
+
+from sklearn.impute import SimpleImputer
```
glemaitre: Here, you import the `SimpleImputer` but we did not justify why in the previous paragraph. So either, we mentioned already that the linear model does not handle missing value or we can postpone the import to when it is going to be used.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+from sklearn.preprocessing import StandardScaler
+
+ts = TimeSeriesSplit(
+    n_splits=3,  # to keep the notebook fast enough on common laptops
+    gap=48,  # 2 days (48 hours) data gap between train and test
+    max_train_size=10000,  # keep train sets of comparable sizes
+    test_size=3000,  # for 2 or 3 digits of precision in scores
+)
+
+import skrub
```
glemaitre: Those imports look weird at the end of the cell. Maybe on the top of where they are going to be used.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+from skrub import TableVectorizer
+
+###############################################################################
+# We define a skrub |var| to start with.
+
+
+data_var = skrub.var("data", data)
+
+
+# We extract our input data (``X``) and the target column (``y``), and mark them
```
glemaitre, line 143:
```suggestion
###############################################################################
# We extract our input data (``X``) and the target column (``y``), and mark them
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+from sklearn.impute import SimpleImputer
+from sklearn.linear_model import RidgeCV
+
+###############################################################################
+# Since we are working with timeseries, we need to use |TimeSeriesSplit| to perform
+# crossvalidation.
+from sklearn.model_selection import TimeSeriesSplit
+from sklearn.pipeline import make_pipeline
+from sklearn.preprocessing import StandardScaler
```
glemaitre: a bit the same remark for those import.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+# %%
+# We can use the |Cleaner| to clean the data and convert the date column to a
+# datetime column, as well as performing additional consistency and cleaning checks.
+
+
+from skrub import Cleaner
+
+X = X.skb.apply(Cleaner())
```
glemaitre: i might want to see a preview at this stage.

glemaitre: and then maybe we should use the subsampling feature?

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# Now we define a simple default pipeline for the |RidgeCV| predictor, which includes
+# a |StandardScaler| for numerical features and a |SimpleImputer| to handle
+# missing values. We will not modify this pipeline: we will instead focus on
+# pre-processing and feature engineering.
+
+
+default_ridge_pipeline = make_pipeline(StandardScaler(), SimpleImputer(), RidgeCV())
+
+
+# The "base" variant of the pipeline uses the default |TableVectorizer|, with
```
glemaitre, line 188:
```suggestion
###############################################################################
# The "base" variant of the pipeline uses the default |TableVectorizer|, with
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# no hyperparameter search.
+
+
+vectorized = X.skb.apply(TableVectorizer())
+predictions_base = vectorized.skb.apply(default_ridge_pipeline, y=y)
+search_base = predictions_base.skb.get_grid_search(fitted=True, cv=ts)
+
+# %%
+# We can observe the results directly using ``.detailed_results``. Unsurprisingly,
```
glemaitre:
```suggestion
# We can observe the results directly using ``.detailed_results_``. Unsurprisingly,
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# the results are not very good. In this case, performing a GridSearch is not
```
glemaitre:
```suggestion
# the results are not very good. In this case, performing a grid-search is not
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# necessary, but we will do it for consistency with the next steps.
```
glemaitre: I'm confused here. We only search for the `alpha` of the `RidgeCV` but this will happen if we get the estimator since it is embedded in the learner itself, right? From the output, I see a single line, so it would be mean that we don't have really have a grid-search using a `GridSearchCV`.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+)
+
+
+vectorized = X.skb.apply(datetime_encoder, cols=s.any_date()).skb.apply(
+    numeric_encoder, cols=s.numeric()
+)
+predictions_enc = vectorized.skb.apply(default_ridge_pipeline, y=y)
+
+
+# Note that, in general, randomized search should be used instead of grid search.
```
glemaitre, line 263:
```suggestion
###############################################################################
# Note that, in general, randomized search should be used instead of grid search.
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+        [
+            "spline",
+            "circular",
+            None,
+        ],
+        name="periodic_encoding",
+    ),
+)
+
+from sklearn.pipeline import make_pipeline
```
glemaitre: By the way, I think that we should always use `# %%`, it is a bit less verbose than the line of `#`. 

```suggestion
# %%
from sklearn.pipeline import make_pipeline
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+    ),
+    StandardScaler(),
+    SimpleImputer(),
+)
+
+
+vectorized = X.skb.apply(datetime_encoder, cols=s.any_date()).skb.apply(
+    numeric_encoder, cols=s.numeric()
+)
+predictions_enc = vectorized.skb.apply(default_ridge_pipeline, y=y)
```
glemaitre: I don't like `_enc` as a name :) I would prefer something descriptive.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+# Note that, in general, randomized search should be used instead of grid search.
+# In this case, grid search is fine as we are interested in a grid of categorical
+# values.
+
+
+search_enc = predictions_enc.skb.get_grid_search(fitted=True, cv=ts)
```
glemaitre: I'm wondering if we could have a way to drop the constant feature created by the `KBinsDiscretizer`, e.g. a `VarianceThreshold`

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+        lagged_max_24h=pl.col("cnt").shift(1).rolling_max(24),
+        lagged_min_24h=pl.col("cnt").shift(1).rolling_min(24),
+        lagged_mean_7d=pl.col("cnt").shift(1).rolling_mean(7 * 24),
+        lagged_max_7d=pl.col("cnt").shift(1).rolling_max(7 * 24),
+        lagged_min_7d=pl.col("cnt").shift(1).rolling_min(7 * 24),
+    )
+    return lagged_df
+
+
+# To add lagged features to the data, we go back to the original ``data_var``, and
```
glemaitre, line 334:
```suggestion
# %%
# To add lagged features to the data, we go back to the original ``data_var``, and
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# drop the columns that are not relevant. We will have to clean the data again.
+
+
+# %%
+data_prep = data_var.drop("casual", "instant", "registered").skb.mark_as_X()
+data_prep = data_prep.skb.apply(Cleaner())
+
+
+# We can use the |apply_func| method to apply the lagged feature function
```
glemaitre, line 343:
```suggestion
# %%
# We can use the |apply_func| method to apply the lagged feature function
```

examples/expressions/13_timeseries_forecasting.py
glemaitre: Or the `Cleaner`

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+# Notice that adding lagged samples will introduce null values for all the initial
+# samples, as there is no previous information for them.
+
+
+# %%
+y = data_var["cnt"].skb.mark_as_y()
+
+
+# %% We can use the datetime encoder defined above to observe the impact of the new
```
glemaitre:
```suggestion
# %%
# We can use the datetime encoder defined above to observe the impact of the new
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+# features on the predictions.
+
+
+vectorized = X.skb.apply(TableVectorizer(datetime=datetime_encoder))
+predictions_lagged = vectorized.skb.apply(default_ridge_pipeline, y=y)
+search_lagged = predictions_lagged.skb.get_grid_search(fitted=True, cv=ts)
+search_lagged.detailed_results_
+
+
+# We can see that the lagged features improved the prediction performance by a
```
glemaitre, line 373:
```suggestion
# %%
# We can see that the lagged features improved the prediction performance by a
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+
+# %% We can use the datetime encoder defined above to observe the impact of the new
+# features on the predictions.
+
+
+vectorized = X.skb.apply(TableVectorizer(datetime=datetime_encoder))
+predictions_lagged = vectorized.skb.apply(default_ridge_pipeline, y=y)
+search_lagged = predictions_lagged.skb.get_grid_search(fitted=True, cv=ts)
+search_lagged.detailed_results_
```
glemaitre: I might expect a parallel coordinate plot again.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+split = predictions_lagged.skb.train_test_split(
+    environment=predictions_lagged.skb.get_data(),
+    splitter=split_function,
+    cutoff="2012-01-01",
+)
+search_lagged = predictions_lagged.skb.get_grid_search(
+    cv=TimeSeriesSplit(), fitted=True
+).fit(split["train"])
+results_lagged = search_lagged.best_pipeline_.predict(split["test"])
```
glemaitre: A cleaner to remove the warning about constant feature might be nice.

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,488 @@
...
+
+ax.set_ylabel("Demand")
+
+ax.tick_params(axis="x", which="minor", labelsize=8)
+ax.tick_params(axis="x", which="major", pad=10)
+
+annotation_text = "Weekend"
+
+point1 = [datetime(2012, 11, 3, 14, 0, 0), 500]
+point2 = [datetime(2012, 11, 4, 14, 0, 0), 550]
```
glemaitre: I'm wondering if we could not have a better trade-off between complexity of the code and the niceness of the plot. Right now, it looks fantastic but the code is scary for non-matplotlib tweaker.

Could delegate a bit of the boilerplate to pandas `.plot` method?

examples/expressions/13_timeseries_forecasting.py
rcap107: this and a lot of other imports are not in the best places, I have been moving them where it makes more sense to import them

examples/expressions/13_timeseries_forecasting.py
rcap107: good idea, I'm adding both

examples/expressions/13_timeseries_forecasting.py
rcap107: I am not doing a grid search at all here, I was just using the same line I use in the rest of the example

I agree it does not make a lot of sense to do when there is only one configuration

examples/expressions/13_timeseries_forecasting.py
rcap107: agreed, I am switching to `# %%`

examples/expressions/13_timeseries_forecasting.py
rcap107: `_feat_eng`?

examples/expressions/13_timeseries_forecasting.py
rcap107: I dropped the KBins completely, I don't think it brings a lot of value here

rcap107:
Thanks a lot @glemaitre for all the comments, I have addressed most of them. I decided to remove the KBins discretizer because it was not bringing a lot of insight and instead it was making the code longer and more annoying to deal with. 

I'm trying to simplify the code for the plot.

rcap107:
the legend wasn't fixed, I am saving it outside the bounding box and sphinx is only saving whatever ends in the bounding box of the figure so legend and title are cut 

edit: I ended up placing the legend inside the axes so I avoided the issue even though I don't like the result

examples/expressions/13_timeseries_forecasting.py
glemaitre: Ah I meant a full English name: `predictions_with_rich_feature` or `predictions_with_lagged_feature` (we drop the `_with_` potentially.

rcap107: I changed to `_rich_features` and `_lagged_features`

rcap107:
ready for review

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
+"""
+.. _example_datetime_encoder :
+
+===============================================================
+Timeseries forecasting with DatetimeEncoder and lagged features
+===============================================================
+
+In this example, we illustrate how combine the datetime features generated by the
```
Vincent-Maladiere:
```suggestion
In this example, we illustrate how to combine the datetime features generated by the
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+|DatetimeEncoder| with the skrub expressions to forecast the the demand for bike
```
Vincent-Maladiere:
```suggestion
|DatetimeEncoder| with the skrub expressions to forecast the demand for bike
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+# -----------------------------------------
+#
+# We use a dataset of bike sharing demand in 2011 and 2012.
+# In this setting, we want to predict the number of bike rentals, based
+# on the date, time and weather conditions.
+# In this example, we use Polars dataframes instead of Pandas.
+
+
+import polars as pl
+
```
Vincent-Maladiere, line 86:
```suggestion
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+# We will use a |RidgeCV| model as our learner. While it is not the most powerful
+# model, it will help us illustrate how feature engineering affects the prediction
+# performance.
+
+
+from sklearn.linear_model import RidgeCV
+
+# %%
+# Since we are working with timeseries, we need to use |TimeSeriesSplit| to perform
+# crossvalidation.
```
Vincent-Maladiere:
```suggestion

# Since we are working with time series, we need to use |TimeSeriesSplit| to perform
# cross-validation.
from sklearn.linear_model import RidgeCV
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+
+# %%
+# As first step, we use the skrub |TableReport| to explore the data and get an idea
+# of the features we are working with.
+# In particular, we can see that the dataset contains a "date" column, and that
+# the "cnt" column is the target we want to predict. We can get a high level
+# overview of the features in the dataset by checking out the ``Stats`` tab in the
+# report. Among other things, we can see that the "date" column is a string column.
+# This is potentially problematic, and a good use case for the skrub |DatetimeEncoder|,
+# as we show in the rest of the example.
```
Vincent-Maladiere: I would bring this paragraph after the TableReport view WDYT?

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+# Prediction with datetime features
+# ---------------------------------
+#
+# In this section, we use the skrub |DatetimeEncoder|, paired with the skrub
+# Expressions to build a predictive pipeline
+# that pre-processes the data, performs feature engineering and executes
+# hyperparameter optimization to improve the prediction performance.
+# We will use a |RidgeCV| model as our learner. While it is not the most powerful
+# model, it will help us illustrate how feature engineering affects the prediction
+# performance.
```
Vincent-Maladiere: Repetition with L12? I don't think this is necessary

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+    max_train_size=10000,  # keep train sets of comparable sizes
+    test_size=3000,  # for 2 or 3 digits of precision in scores
+)
+
+
+# %%
+# We import ``skrub`` to access the skrub expressions and define a skrub |var|
+# to start with. We use |.skb.subsample| to reduce the size of the dataset used
+# in the preview to 100 rows, so that the notebook runs faster when developing.
+# When the models are trained, the expressions switch automatically to the full
```
Vincent-Maladiere:
```suggestion
# During training, subsampling is disabled, and the expressions use the full
```

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+# as X and y.
+X = data_var[
+    "date", "holiday", "weathersit", "temp", "hum", "windspeed"
+].skb.mark_as_X()
+y = data_var["cnt"].skb.mark_as_y()
+X
+
+
+# %%
+# We can use the |Cleaner| to clean the data and convert the date column to a
```
Vincent-Maladiere: "Clean the data" is too vague; could you list what this transformer is doing in this instance?

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+
+def split_function(X, y, cutoff=None):
+    mask = X["date"] < cutoff
+    X_train, X_test = X.filter(mask), X.filter(~mask)
+    y_train, y_test = y.filter(mask), y.filter(~mask)
+    return X_train, X_test, y_train, y_test
+
+
+# %%
+split_base = predictions_base.skb.train_test_split(
```
Vincent-Maladiere: Could you create a function or a for loop to accumulate results in a dict? This would be easier to read

examples/expressions/13_timeseries_forecasting.py
```diff
@@ -0,0 +1,527 @@
...
+    ),
+)
+
+ax.legend(
+    loc="upper right",
+    # bbox_to_anchor=(0.5, 1.06),
+    ncol=1,
+    borderaxespad=0.0,
+    frameon=True,
+)
```
Vincent-Maladiere, line 491: I have to admit, I'm not very convinced by these results. The DatetimeEncoder and Periodic Features fit is quite poor, and the Lagged one looks like a 1h shift of the ground truth, which is probably what the Ridge selected

Vincent-Maladiere (commented):
Hey Riccardo, I feel like this example is quite long and could be trimmed a little bit.

In particular, I think the datetime encoder could be made optional using choices, so that the focus is more "creating lags and optionally using the datetime encoder". We could nest some hyperparameters in the datetime encoder, but comparing the default to the tuned version isn't necessary IMO.

Also, using boosting trees might provide more compelling results if we're interested in having a good fit with the ground truth

rcap107:
> Hey Riccardo, I feel like this example is quite long and could be trimmed a little bit.
> 
> In particular, I think the datetime encoder could be made optional using choices, so that the focus is more "creating lags and optionally using the datetime encoder". We could nest some hyperparameters in the datetime encoder, but comparing the default to the tuned version isn't necessary IMO.
> 
> Also, using boosting trees might provide more compelling results if we're interested in having a good fit with the ground truth

The example was originally meant to show that periodic features are bringing a noticeable benefit, then we decided to add the lagged features to show that they are also useful. We decided to use ridge because it would make the difference more noticeable, while hgb would just work regardless of the periodic features (for the most part).

examples/expressions/13_timeseries_forecasting.py
rcap107: We have been struggling to get any decent result with this dataset from the beginning, and there are already two other issues about it #1282 #1394 

If you have better datasets to suggest, maybe we should switch to something else and give up on this.

Vincent-Maladiere:
> The example was originally meant to show that periodic features are bringing a noticeable benefit, then we decided to add the lagged features to show that they are also useful. We decided to use ridge because it would make the difference more noticeable, while hgb would just work regardless of the periodic features (for the most part).

I understand, but I'm still concerned with the length of the example. Having a noticeable effect for boosting trees, which are what most people use, would be nice, although I get the argument that our datasets might not give us a performance bump

examples/expressions/13_timeseries_forecasting.py
Vincent-Maladiere: Unfortunately, I don't. Perhaps we could include a disclaimer stating that this example introduces feature engineering rather than a silver bullet method for time series? Or we could simply focus only on the feature engineering part without displaying the downstream performance?

Vincent-Maladiere: What I mean is we don't need to motivate this kind of feature engineering by showing some performance, since it has been adopted by the community already

rcap107: Fine by me! What matters to me is having an example of feature engineering for time series somewhere in the docs, ideally to replace the current one

I can add a disclaimer

rcap107:
After discussion with @Vincent-Maladiere, we were thinking of removing this example and replacing it with a much simpler example that shows how to work on dates using the DatetimeEncoder. Then, we could link the masterclass by @ogrisel and @glemaitre for a far more advanced and nuanced example than what we can easily maintain in the documentation (of course, if the authors are fine with it)
 
I think this example is at the same time too long to be an example, and not in-depth enough to be particularly useful.

Thoughts?

GaelVaroquaux:
> I think this example is at the same time too long to be an example, and not in-depth enough to be particularly useful.

Fine with that.

I would think that we should also consider long term merging in stuff from the master-class into the skrub documentation. I think so for multiple reasons, but one of them is that it will give us good examples to look at in terms of improving the API (ie creating functionality in skrub that make these examples easier to follow)

rcap107:
>  I would think that we should also consider long term merging in stuff from the master-class into the skrub documentation. I think so for multiple reasons, but one of them is that it will give us good examples to look at in terms of improving the API (ie creating functionality in skrub that make these examples easier to follow)

Yes, we should definitely do that

rcap107:
Closing this

Vincent-Maladiere:
> I would think that we should also consider long term merging in stuff from the master-class into the skrub documentation

Agreed, and we should point to the masterclass content somewhere :)

PR #1368, opened 2025-05-04
DOC: front page vocabulary / priority

GaelVaroquaux:
1. Rename "Effortless pipelines" to "Effortless Tabular Learning"

2. Put the table report before the GapEncoder: it's a stronger seller

An iteration on the front page to change a bit the focus that we broadcast (this is just an iteration, we could be more, but I think that it is good to work with iteration)

### DIFF
--- a/doc/_templates/index.html
+++ b/doc/_templates/index.html
@@ -47,7 +47,7 @@ still full-width (unlike the article) #}
   <!-- Feature Section 1 -->
   <div class="row mb-0 row-padding-between-features">
     <div class="col-md-5 mb-5">
-      <h4 class="feature-title">Effortless Pipelines</h4>
+      <h4 class="feature-title">Effortless Tabular Learning</h4>
       <p class="feature-text">Create strong scikit-learn pipeline baselines effortlessly with
         <a class="reference internal"
         href="reference/generated/skrub.TableVectorizer.html#tablevectorizer"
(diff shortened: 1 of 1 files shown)


GaelVaroquaux:
I don't understand why the precommit code format is failing

GaelVaroquaux:
> Not a big deal, but the arrow doesn't point in the right direction on small screens with vertical folding. Should we remove it, or fix its layout?

I couldn't get it to work. If you have a better suggestion, I'm open

doc/_templates/index.html
```diff
@@ -111,14 +84,45 @@ <h4 class="feature-title">Interactive Data Exploration</h4>
         <a href="https://skrub-data.org/skrub-reports/">Try it on your dataset →</a>
       </p>
       <div class="row">
-        <p class="click-table-report-hint">Click anywhere on the table <i class="fa-solid fa-arrow-trend-down"></i> </p>
+        <p class="click-table-report-hint"><i class="fa-solid
+fa-caret-left"></i> Click anywhere on the table</p>
```
Vincent-Maladiere, line 88: This is working for me:

Screenshot 2025-05-12 at 12 21 05

```suggestion
fa-caret-left caret-icon"></i> Click anywhere on the table</p>
```

And then add:
```html
  <script>
    function updateCaretDirection() {
      const colLeft = document.querySelector('.col-md-5');
      const colRight = document.querySelector('.col-md-7');
      const icon = document.querySelector('.caret-icon');
  
      if (!colLeft || !colRight || !icon) return;
  
      const leftRect = colLeft.getBoundingClientRect();
      const rightRect = colRight.getBoundingClientRect();
  
      const isStacked = Math.abs(leftRect.top - rightRect.top) > 50; // threshold for vertical stack
  
      if (isStacked) {
        icon.classList.remove('fa-caret-left');
        icon.classList.add('fa-caret-down');
      } else {
        icon.classList.remove('fa-caret-down');
        icon.classList.add('fa-caret-left');
      }
    }
  
    window.addEventListener('resize', updateCaretDirection);
    window.addEventListener('DOMContentLoaded', updateCaretDirection);
  </script>
```

rcap107:
I was wondering, should we expand a bit the section on the datetime encoder? Maybe with this example: 

demo_periodic_features

With @jeromedockes we talked about adding it to the user guide, but maybe it could be useful to have in the the front page

doc/_templates/index.html
jeromedockes: In the css there seems to be a hard-coded cutoff with a media query at 720px width:

```
@media (min-width: 720px) {
    .col-md-5 {
        flex: 0 0 auto;
        width: 41.66666667%;
    }
}
```

so we can probably use the same with something like

```
         @media (min-width: 720px) {
             .left-or-down-pointing-arrow  {
                 transform: rotate(90deg);
             }
         }
```


(the 90 may need to be adjusted, and maybe it's the other way around depending on what the arrow looks like when not rotated I haven't checked)

jeromedockes:
> With @jeromedockes we talked about adding it to the user guide, but maybe it could be useful to have in the the front page

I would suggest to do it in a separate PR but indeed I like the plot shown by @rcap107 a lot: it's a plot I want to see to make sure I understood what the spline encoder is doing, but it's a sufficient amount of work that I'm glad to see it in the documentation rather than making it myself

doc/_templates/index.html
GaelVaroquaux: I much prefer the solution by @jeromedockes to javascript

GaelVaroquaux: OK, I've implemented a CSS fix

jeromedockes (approved):
LGTM, thanks @GaelVaroquaux ! :+1:

PR #1369, opened 2025-05-05
Add a datasets.toy_dataframe()?

### LINKED ISSUE #1189: Add a `datasets.toy_dataframe()`?
(or some better name)

a function that creates and returns a small dataframe with columns of various types that we can use in the docstring examples. the goal is to avoid the need of creating the data in the examples, and also not need to download an  actual dataset.

first step is make a list of docstrings that could benefit from such a function and seeing if we can come up with one that would suit a good proportion of docstrings

an alternative could be using one of the scikit-learn datasets that don't need network access, but I think IIRC they will not have many different types in their columns. maybe the titanic one?

note we have a similar function as a pytest fixture but the goal is different: cover many different weird combinations of int, float, missing values etc. whereas the new function would be for illustration not testing

Themlaw:
fix #1189 
added a toy dataframe generation to be used in the docstrings. Display various types of data that can be store in a dataframe. Such as integers, floats, strings, booleans, dates, and times.

### DIFF
--- a/skrub/datasets/_generating.py
+++ b/skrub/datasets/_generating.py
@@ -99,3 +99,22 @@ def toy_orders(split="train"):
     else:
         assert split == "all", split
     return Bunch(X=X, y=y, orders=X.assign(delayed=y), orders_=X, delayed=y)
+
+def toy_dataframe():
+    """
+    Create a toy dataframe with various data types. Such as strings, booleans, dates, and integers.
+    
+    Returns
+    -------
+    df : pd.DataFrame
+        A pandas DataFrame with 5 rows and 6 columns.
+    """
+    df = pd.DataFrame({
+        "ID": [1, 2, 3, 4, 5],
+        "name": ["Alice", "Bob", "Charlie", "David", "Eve"],
+        "product": ["pen", np.nan, "cup", "spoon", "fork"],
+        "is_active": [True, False, True, False, True],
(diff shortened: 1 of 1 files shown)


rcap107:
Hi @Themlaw, and thanks for working on this! I apologize for the slow response. 

I think it's a good start! Two main things:
- Given that we want to keep feature equivalence between polars and pandas, and that we might need either library in some examples, it would be good if you could modify the function so that it takes the dataframe engine (pandas or polars) as a parameter to produce the correct result.  
- Coverage is complaining, so you should add a generation test to `skrub/datasets/tests/test_generating.py` that asserts that the generated dataset is exactly what you're expecting. The `df_module` will help with that. 

We might also want to add more columns/different cases later, but this is already a good start.

rcap107:
Hey @Themlaw, it seems like there was a mess-up on our side: this feature had already been implemented during a sprint, and the issue was not marked as complete as it should have. 

Now the situation is that there is a function called `skrub.toy_orders` that is already used in a number of examples, so the current PR would be redundant. I'm really sorry about that :see_no_evil: 

However, the current implementation of `toy_orders` has some issues (it should be imported from ` skrub.datasets` rather than `skrub`, and it is not documented properly). Would you be interested in working on #1403 to address these problems?

Vincent-Maladiere:
Closing in favor of https://github.com/skrub-data/skrub/issues/1403

Issue #1362, opened 2025-05-02
Bad formatting in DatetimeEncoder documentation


rcap107:
### Describe the issue linked to the documentation

Formatting is broken in the DatetimeEncoder 

![Image](https://github.com/user-attachments/assets/30617b4e-c652-458f-91f3-e2b1d9a7a395)

We should also add more detail on the default (and unmodifiable) parameters that are used for the periodic encoders.

jeromedockes:
fixed by #1374

PR #1374, opened 2025-05-12
Fixing formatting and improving docstring 

rcap107:
Addresses #1362 and cleans a bit the docstring for the DatetimeEncoder

Note that sphinx is being sphinx so the generated docs for the ``DatetimeEncoder`` are not putting the summary where it's supposed to be (after the big Note), and I don't know how to fix that.


jeromedockes (approved):
thanks a lot @rcap107 !

PR #1376, opened 2025-05-12
Adding a demo of what periodic features added by the DatetimeEncoder look like

rcap107:
I added everything, but the index html is really messed up, and there may be conflicts with #1368 (I'll deal with them later)

### DIFF
--- a/doc/generate_datetime_example.py
+++ b/doc/generate_datetime_example.py
@@ -3,10 +3,12 @@ A script for generating a plot that displays how periodic features are encoded,
 to be added to the front page.
 """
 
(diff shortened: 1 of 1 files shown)


Vincent-Maladiere:
A few remarks after an IRL discussion with @jeromedockes:
- We should try to keep a single feature engineering section and avoid splitting it into text encoding and datetime encoding.
- We should either show a single image (gap encoding or datetime encoding) or show both, but align them horizontally or vertically so that the section doesn't look overcrowded with information.
- If we choose not to display the datetime encoding here, we can still show it in the user guide

PR #1378, opened 2025-05-13
Order in the parallel coordinate plot

### LINKED ISSUE #1312: Order in the parallel coordinate plot
Image

IMHO, in the parallel coordinate plot (from https://skrub-data.org/dev/reference/generated/skrub.ParamSearch.html#skrub.ParamSearch.plot_results), we should have all the parameters varied at one end of the plot (the left, IMHO) and all the values reported (score, fit time, score time) at the other end (the right, IMHO, because it is closer to the colorbar that displays score)

Vincent-Maladiere:
Fixes #1312

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,10 +41,6 @@ New features
 
 Changes
 -------
-
-- The score(s) columns of :class:`ParamSearch.results_` is now the right-most columns.
-  :pr:`1378` by :user:`Vincent Maladiere <Vincent-Maladiere>`.
-
 .. warning::
   The default high cardinality encoder for both :class:`TableVectorizer` and
   :meth:`tabular_learner` has been changed from :class:`GapEncoder` to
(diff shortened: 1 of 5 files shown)


CHANGES.rst
```diff
@@ -41,6 +41,10 @@ New features
 
 Changes
 -------
+
+- The score(s) columns of :class:`ParamSearch.results_` is now the right-most columns.
+  :pr:`1378` by :user:`Vincent Maladiere <Vincent-Maladiere>`.
+
```
jeromedockes, line 47:
```suggestion
```

the ParamSearch has never been released yet so this is not a change from users' point of view.

We will still want to mention all the recent prs related to expressions but it will be easier to create one summary changelog entry when preparing the release

jeromedockes (approved):

jeromedockes:
thank you very much @Vincent-Maladiere !

Issue #1380, opened 2025-05-14
Enable multi-scoring search in the parallel plot


Vincent-Maladiere:
Currently, the parallel plot hardcodes `mean_test_score` as a color value. This column in `search._get_cv_results_table()` get renamed when using multiple scoring in the search:

```python
import skrub
from skrub.datasets import fetch_employee_salaries
from skrub import TableVectorizer
from sklearn.linear_model import RidgeCV

data = fetch_employee_salaries()
X, y = skrub.X(data.X), skrub.y(data.y)

cols = skrub.choose_from([["gender"], ["employee_position_title"]], name="cols")
pred = X[cols].skb.apply(TableVectorizer()).skb.apply(RidgeCV(), y=y)
search = pred.skb.get_grid_search(
    fitted=True,
    scoring=["r2", "neg_root_mean_squared_error"],
    refit="r2",
)
search.results_
```
```
	cols	mean_test_r2	mean_test_neg_root_mean_squared_error
0	['employee_position_title']	0.6022776657776607	-18355.519056003614
1	['gender']	0.004850310775039612	-29025.44568535838
```

But it fails at the plotting of the parallel plot because mean_test_score is hardcoded.

```python
search.plot_results()
```
```
KeyError: 'mean_test_score'
```

PR #1383, opened 2025-05-14
Enable multi scoring for parallel plot

Vincent-Maladiere:
Fixes https://github.com/skrub-data/skrub/issues/1380

### DIFF
--- a/skrub/_expressions/_estimator.py
+++ b/skrub/_expressions/_estimator.py
@@ -655,7 +655,7 @@ class ParamSearch(_CloudPickleExpr, BaseEstimator):
         Cross-validation results containing parameters and scores in a dataframe.
         """
         try:
-            return self._get_cv_results_table()
+            return self._get_cv_results_table()[0]
         except NotFittedError:
             attribute_error(self, "results_")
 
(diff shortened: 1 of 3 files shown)


skrub/_expressions/_estimator.py
```diff
@@ -737,13 +737,14 @@ def _get_cv_results_table(self, return_metadata=False, detailed=False):
...
+        col_score = f"mean_test_{metric_names[0]}"
         table = table.sort_values(
-            f"mean_test_{metric_names[0]}",
+            col_score,
             ascending=False,
             ignore_index=True,
             kind="stable",
         )
-        return (table, metadata) if return_metadata else table
+        return (table, col_score, metadata) if return_metadata else (table, col_score)
```
jeromedockes, line 747: thanks for starting it @Vincent-Maladiere ! do you think it might be easier to just put the col_score in the metadata, eg `metadata["main_score_column"] = "mean_test_accuracy"` or something like that?

Vincent-Maladiere: Why not! How to handle the case when `return_metadata = False` though? 🤔

jeromedockes: that function is only called by parallel coord which passes `metadata=True` and `results_` and `detailed_results_` with metadata=False but those wouldn't use the col_score either so it shouldn't be a problem. I guess I am saying the sorting column name is metadata about the table in the same way that the list of log-scale column is and callers that need it should pass metadata=True and retrieve it from there.

as this is a private function it is also perfectly fine to remove the metadata=False option and always return the metadata, and callers that don't need it just discard it, as you prefer!

skrub/_expressions/tests/test_parallel_coord.py
```diff
@@ -82,3 +83,22 @@ def test_parallel_coord():
     assert dim["label"] == "score"
     dim = next(data)
     assert dim["label"] == "fit time"
+
+
+def test_multi_scoring():
```
jeromedockes, line 88:
```suggestion
def test_multi_scoring():
    pytest.importorskip("plotly")

```

jeromedockes (approved):
LGTM, thanks @Vincent-Maladiere !

PR #1385, opened 2025-05-15
set choice name in home page parallel coord

jeromedockes:
With the new order of columns, because plotly does not leave a margin on the left, the tick label gets cropped otherwise


GaelVaroquaux:
I believe that this is no longer needed, given https://github.com/skrub-data/skrub/pull/1386

Vincent-Maladiere:
Closing in favor of #1386

PR #1386, opened 2025-05-15
wrap parallel coordinate plot tick labels

### LINKED ISSUE #1367: Line return in "choose_from(...)" in parallel coordinate plots
Image

When the "choose_from(...)" label is too long, we just add line return at the spaces (as a word processing display would do) to make it multiline

jeromedockes:
closes #1367 

alternative to #1385 

needs testing on many pipelines and display combinations to see if it makes things better or worse on average

example on home page:

screenshot_2025-05-15T15:08:10+02:00

### DIFF
--- a/skrub/_expressions/_parallel_coord.py
+++ b/skrub/_expressions/_parallel_coord.py
@@ -70,8 +70,7 @@ def _add_jitter(column):
 
 
 def _wrap_label(column):
-    """ Add "<br>" to limit the width of a label to at most 20 chars
-    """
+    """Add "<br>" to limit the width of a label to at most 20 chars"""
     label = "<br>\n".join(textwrap.wrap(column["label"], width=20))
     return {**column, "label": label}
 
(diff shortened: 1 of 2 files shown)


skrub/_expressions/_parallel_coord.py
```diff
@@ -67,16 +69,21 @@ def _add_jitter(column):
     return column
 
 
+def _wrap_label(column):
```
GaelVaroquaux, line 72:
```suggestion
def _wrap_label(column):
    """ Add "<br>" to limit the width of a label to at most 20 chars
    """
```

GaelVaroquaux (approved):
This is great!

Thank you very much for doing this

skrub/_expressions/_parallel_coord.py
```diff
@@ -67,16 +69,23 @@ def _add_jitter(column):
...
+    """ Add "<br>" to limit the width of a label to at most 20 chars
+    """
```
GaelVaroquaux, line 74:
```suggestion
    """Add "<br>" to limit the width of a label to at most 20 chars"""
```

PR #1387, opened 2025-05-15
Change parallel coord order to put score last

jeromedockes:
This changes the order of metrics in the `ParamSearch.results_` and parallel coordinate plot to put the score on the right (next to the colorbar)

It also gets rid of a warning in `test_multi_scoring`


Vincent-Maladiere (approved):
Great

PR #1390, opened 2025-05-16
add some comments in _evaluation.py and _estimator.py

jeromedockes:


skrub/_expressions/_estimator.py
```diff
@@ -86,6 +101,7 @@ def __init__(self, expr):
         self.expr = expr
 
     def __skrub_to_Xy_pipeline__(self, environment):
+        """Convert to a scikit-learn compatible pipeline."""
```
GaelVaroquaux, line 104:
```suggestion
        """Convert to a fully scikit-learn compatible pipeline (fit takes X, y)."""
```

GaelVaroquaux (approved):
LGTM. Thanks a lot for the improvement!

PR #1392, opened 2025-05-16
WIP add notes on how to use only part of a pipeline in user guide

jeromedockes:
maybe this should be an example with a more realistic pipeline

### DIFF
--- a/doc/skrub_pipeline.rst
+++ b/doc/skrub_pipeline.rst
@@ -418,3 +418,67 @@ However, all of those methods have a ``keep_subsampling`` parameter that we can
 set to ``True`` to force using the subsampling when we call them.
 
 See more details in a :ref:`full example <example_subsampling>`.
+
+Arriving late and leaving early
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We can give a name to any node with :meth:`.skb.set_name() <Expr.skb.set_name>`.
+When this is done we can:
+
+- Bypass the computation of that node and override its result by passing it as a
+  key in the ``environment`` argument.
+- Truncate the pipeline after this node to obtain the intermediate result with
+  :meth:`SkrubPipeline.truncated_after`.
+
+Here is a toy example with 3 steps:
+
+>>> def load_data():
+...     print("load")
+...     return [1, 2, 3, 4]
+
+
+>>> def transform(x):
+...     print("transform")
+...     return [e * 10 for e in x]
+
+
+>>> def agg(x):
+...     print("predict")
+...     return max(x)
+
+
+>>> output = (
+...     skrub.deferred(load_data)()
+...     .skb.set_name("loaded")
+...     .skb.apply_func(transform)
+...     .skb.set_name("transformed")
+...     .skb.apply_func(agg)
+... )
+load
+transform
+predict
+>>> output
+<Call 'agg'>
+Result:
+―――――――
+40
+
+>>> pipeline = output.skb.get_pipeline(fitted=True)
+load
+transform
+predict
+
+Bypass the ``load_data`` step:
+
+>>> pipeline.transform({'loaded': [6, 5, 4]})
+transform
+predict
+60
+
+Stop after the ``transform`` step:
+
+>>> truncated = pipeline.truncated_after('transformed')
+>>> truncated.transform({})
+load
+transform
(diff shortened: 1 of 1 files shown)


doc/skrub_pipeline.rst
```diff
@@ -418,3 +418,68 @@ However, all of those methods have a ``keep_subsampling`` parameter that we can
 set to ``True`` to force using the subsampling when we call them.
 
 See more details in a :ref:`full example <example_subsampling>`.
+
+Arriving late and leaving early
```
rcap107, line 422: I really don't like this title, it's not telling me what this section is about at all
```suggestion
Partitioning the pipeline
```

rcap107: Or, 
```suggestion
Adding checkpoints to the pipeline
```

doc/skrub_pipeline.rst
```diff
@@ -418,3 +418,68 @@ However, all of those methods have a ``keep_subsampling`` parameter that we can
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+We can give a name to any node with :meth:`.skb.set_name() <Expr.skb.set_name>`.
+When this is done we can:
+
+- Bypass the computation of that node and override its result by passing it as a
+  key in the ``environment`` argument.
+- Truncate the pipeline after this node to obtain the intermediate result with
+  :meth:`SkrubPipeline.truncated_after`.
+
```
rcap107, line 432: I think it could be helpful to describe these as "checkpoints" in the evaluation, so that it is possible to start from/stop at a checkpoint. 

Another thing, isn't this equivalent to 
```python
loaded = url.skb.apply_func(load_data)
transformed = loaded.url.skb.apply_func(transform)
output = transformed.skb.apply_func(agg)
```

If that is the case, I think it would be useful to add an explanation of why it's not done in the "usual way". Even "it's also possible to assign names".

doc/skrub_pipeline.rst
```diff
@@ -418,3 +418,68 @@ However, all of those methods have a ``keep_subsampling`` parameter that we can
...
+...     return [item * 10 for item in x]
+
+
+>>> def agg(x):
+...     print("agg")
+...     return max(x)
+
+
+>>> url = skrub.var("url")
+>>> output = (
```
rcap107, line 451: Isn't this equivalent to 
```python
loaded = url.skb.apply_func(load_data)
transformed = loaded.url.skb.apply_func(transform)
output = transformed.skb.apply_func(agg)
```

If that is the case, I think it would be useful to add an explanation of why it's not done in the "usual way". Even "it's also possible to assign names".

jeromedockes: it is equivalent except that it also names the intermediate results using `set_name()` and we can later use those names to refer to the corresponding node in the fitted pipeline

Vincent-Maladiere (approved):
Looks good, nice effort to make this explainer short

doc/skrub_pipeline.rst
```diff
@@ -418,3 +418,80 @@ However, all of those methods have a ``keep_subsampling`` parameter that we can
...
+...     url.skb.apply_func(load_data)
+...     .skb.set_name("loaded")
+...     .skb.apply_func(transform)
+...     .skb.set_name("transformed")
+...     .skb.apply_func(agg)
+... )
+
+Above, we give a name to each intermediate result with ``.skb.set_name()`` so
```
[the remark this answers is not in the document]
jeromedockes: no with the current public api we can only find a node by name. note that indices is not really defined for a graph

doc/skrub_pipeline.rst
jeromedockes: maybe but 'checkpoint' could give the (false) impression that something is being saved and we can reload from that checkpoint or something, whereas here its just about providing a value for this result, or cutting out the rest of the pipeline after it, or finding the scikit-learn estimator that produced it

doc/skrub_pipeline.rst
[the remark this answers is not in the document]
Vincent-Maladiere: Does a node have a name by default?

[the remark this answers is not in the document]
jeromedockes: no (except variables)

PR #1396, opened 2025-05-21
Removing a redundant periodic encoding

rcap107:
Fixes #1376  (part of)

This is removing the "month of year" encoding, because it is doing the same thing as "day in year"

HOWEVER

The difference between the two is that the encoding for "month of year" is always added, while that for "day in year" is added only if `add_day_of_year` is set to True

Which behavior do we want to keep? Always adding the encoding, or adding it only if the flag is set?


rcap107:
I talked IRL with @GaelVaroquaux and the idea was to change the `add_day_of_year` parameter from just being a boolean flag to adding a string parameter (like "replace", or "both"). 

The behavior would be: 
- By default, add the month encoding
- If `add_day_of_year` is True, add the encoding for the day of the year (this keeps the redundancy)
- If `add_day_of_year` is `"replace"`, replace the encoding for month in year with that of day in year

Vincent-Maladiere:
Do we have examples/use-cases/ideas where having both values is beneficial?

GaelVaroquaux:
> Do we have examples/use-cases/ideas where having both values is beneficial?

It's easiest for a model (eg a tree) to select a given day of month if the day of month is represented.

rcap107:
At the moment, redundancy is present only if `add_day_of_year` is set to True. I would be more in favor of simply removing the periodic encoding for `day_of_year` and keep that for `month_of_year`. Then, `add_day_of_year` would simply add the number of the day and no other features.

rcap107:
ready for review

Vincent-Maladiere (approved):
LGTM!

PR #1404, opened 2025-05-22
Coalesce optional dependencies into dev

### LINKED ISSUE #1375: too many optional dependencies targets in toml
As discussed with @rcap107 , there are too many combinations of dependencies which makes it hard for contributors to get a working dev environment: `dev`, `doc`, `test`, `lint` etc.

We could revisit those to simplify it; my proposal would be that `dev` installs everything you need to run the python tests and build the documentation. the installation and contributing instructions can be updated accordingly

Vincent-Maladiere:
Fixes #1375

Let's wait for https://github.com/skrub-data/skrub/pull/1393 to be merged, as it changes the install and contributing sections.


pyproject.toml
```diff
@@ -128,7 +125,7 @@ ruff = "==0.4.8"
 pre-commit = "*"
 
 [tool.pixi.feature.optional.dependencies]
-pyarrow = "*"
+pyarrow = "<20.0.0"
```
rcap107, line 128: why is there a limit on the pyarrow version?

Vincent-Maladiere: There is an issue with the last version of pyarrow regarding loading parquet having unit32 as index. This is a temporary fix for us, I hope this will be fixed in a later version on pyarrow side.

Vincent-Maladiere: I'll see if we can workaround it in skrub in the meantime in this PR

Vincent-Maladiere: Ok, I simply loaded and rewrote the failing parquet file using polars, and it fixed our pyarrow issue, no need to change pyarrow version indeed :)

rcap107 (approved):
Looks good! Thanks a lot @Vincent-Maladiere

PR #1405, opened 2025-05-22
Fixing inconsistency in naming features for the StringEncoder

rcap107:
Features added by the StringEncoder started from 0 up to n_components-1, while other encoders have features in the range 1 to n_components.

This PR addresses the inconsistency.


skrub/_string_encoder.py
```diff
@@ -89,10 +89,15 @@ def get_feature_names_out(self):
 
         Returns
         -------
-        feature_names_out : list of str objects
+        feature_names_out : ndarray of str objects
```
Vincent-Maladiere, line 92: You're outputting a list here
```suggestion
        feature_names_out : list of str objects
```

skrub/_string_encoder.py
```diff
@@ -161,10 +166,11 @@ def fit_transform(self, X, y=None):
         self._is_fitted = True
         self.n_components_ = result.shape[1]
 
-        name = sbd.name(X)
-        if not name:
-            name = "tsvd"
-        self.all_outputs_ = [f"{name}_{idx}" for idx in range(self.n_components_)]
+        self._input_name = sbd.name(X)
+        if not self._input_name:
+            self._input_name = "tsvd"
```
Vincent-Maladiere, line 171: Simplifies:
```suggestion
        self._input_name = sbd.name(X) or "tsvd"
```

This default naming is a good idea, we could do the same in `TextEncoder`, `MinHashEncoder`, `GapEncoder`, WDYT?

CHANGES.rst
```diff
@@ -65,6 +65,9 @@ Changes
 
 - The :class:`SimpleCleaner` has been removed. use :class:`Cleaner` instead. :pr:`1370` by :user:`Riccardo Cappuzzo<rcap107>`.
 
+- Features generated by the :class:`StringEncoder` now are named with indices in range ``[1, n_components]``, rather than
```
Vincent-Maladiere, line 68:
```suggestion
- Features generated by the :class:`StringEncoder` are now named with indices in the range ``[1, n_components]``, rather than
```

Vincent-Maladiere (commented):
Thank you for this PR @rcap107!

skrub/_string_encoder.py
rcap107: yes we could do that, but maybe in a separate pr

skrub/_string_encoder.py
```diff
@@ -92,7 +92,12 @@ def get_feature_names_out(self):
         feature_names_out : list of str objects
             Transformed feature names.
         """
-        return list(self.all_outputs_)
+
+        num_digits = len(str(self.n_components - 1))
+        return [
+            f"{self._input_name}_{str(i+1).zfill(num_digits)}"
```
jeromedockes, line 98: for what it's worth I think the way @rcap107 had it in the StringEncoder (ie start counting from 0, like everywhere in python) made more sense

rcap107: yeah that's what made more sense to me, but I'd rather have one consistent format

GaelVaroquaux:
> I just realized that the periodic features in the DatetimeEncoder have the same problem (starting from 0)

> Should I rename those as well while I'm doing this?

Consistency is a good thing.

I do wonder however if we should start at 0 or 1. Python indexing starts at 0.

rcap107:
Talking with Franck offline, I think we should rename the features in the StringEncoder so they're in 1 to N, because if we're changing the defaults from Gap to StringEncoder then having the features in 0 to N-1 would break even more stuff

Vincent-Maladiere (commented):
Very nice, thanks for adding names and factorizing the utils function! 
- Could you add a test for the `get_feature_names_out()` method for the `TextEncoder`? 
- Could you add a default name for the gap encoder as well?

skrub/_string_encoder.py
```diff
@@ -70,7 +72,7 @@ class StringEncoder(SingleColumnTransformer):
     ... ], name='video comments')
 
     >>> enc.fit_transform(X) # doctest: +SKIP
-       video comments_0  video comments_1
+       video comments_1  video comments_1
```
rcap107, line 75: typo

Vincent-Maladiere:
skrub meeting conclusion: let's stay consistent with sklearn PCA and start indexing names at 0

rcap107:
ready for review

skrub/_text_encoder.py
```diff
@@ -260,7 +264,7 @@ def fit_transform(self, column, y=None):
 
         self.n_components_ = X_out.shape[1]
 
-        cols = self.get_feature_names_out()
+        cols = get_encoder_feature_names(self.input_name_, self.n_components_)
```
Vincent-Maladiere: Why not using
```python
cols = self.get_feature_names_out()
```
here?

Vincent-Maladiere (commented):
Hey @rcap107, thanks for moving this PR forward. I have a meta-suggestion:
- add a `get_feature_names_out` method to `SingleColumnTransformer`, and remove it for the TextEncoder, StringEncoder, MinHashEncoder, SplineEncoder, CircularEncoder (transformers like the Gap can overwrite it by defining their own `get_feature_names_out`)
- remove the `outputs_` attributes and call `get_feature_names_out()` instead. The utils `get_encoder_feature_names` should not be used inside individual transformers.

WDYT?

rcap107:
I did the refactoring, now it should be ready

skrub/_on_each_column.py
```diff
@@ -123,6 +123,24 @@ def __init_subclass__(subclass, **kwargs):
...
+
+        Each feature has format ``{input_name}_{n_component}`` where ``input_name``
+        is the name of the input column, or a default name for the encoder, and
+        ``n_component`` is the idx of the specific feature.
+
+        Returns
+        -------
+        list of str
+            The list of feature names.
+        """
```
Vincent-Maladiere, line 137: You probably need to check that the transformer has been fitted here

Vincent-Maladiere (commented):
One last comment!

skrub/_datetime_encoder.py
```diff
@@ -548,10 +548,8 @@ def fit_transform(self, X, y=None):
         self.is_fitted = True
         self.n_components_ = X_out.shape[1]
 
-        name = sbd.name(X)
-        self.all_outputs_ = [
-            f"{name}_spline_{idx}" for idx in range(self.n_components_)
-        ]
+        self.input_name_ = sbd.name(X) + "_spline"
```
Vincent-Maladiere, line 555: I just noticed that this would fail when the column X don't have a name (None)

skrub/_datetime_encoder.py
```diff
@@ -633,10 +631,8 @@ def fit_transform(self, X, y=None):
 
         self.n_components_ = 2
 
-        name = sbd.name(X)
-        self.all_outputs_ = [
-            f"{name}_circular_{idx}" for idx in range(self.n_components_)
-        ]
+        self.input_name_ = sbd.name(X) + "_circular"
```
Vincent-Maladiere, line 642: Ditto here

skrub/_datetime_encoder.py
```diff
@@ -548,10 +548,9 @@ def fit_transform(self, X, y=None):
         self.is_fitted = True
         self.n_components_ = X_out.shape[1]
 
-        name = sbd.name(X)
-        self.all_outputs_ = [
-            f"{name}_spline_{idx}" for idx in range(self.n_components_)
-        ]
+        name = sbd.name(X) if sbd.name(X) else ""
```
Vincent-Maladiere: Sorry to nitpick, but why not have a default name like you suggested below?

rcap107: both _CircularEncoder and _SplineEncoder are supposed to be used only internally by the DatetimeEncoder, so `sbd.name(X)` would always be hour/month/weekday etc

I could rename it as "periodic" but in practice it should never be None, which is why I didn't bother adding a check on it until you mentioned it

Vincent-Maladiere: You're right, I forgot this is not publicly exposed, and we create a name in DatetimeEncoder. You can revert it to:
```python
self.input_name_ = sbd.name(X) + "_spline"
```
sorry about that :)

Vincent-Maladiere (approved):
LGTM! (after reverting the sbd.name as said above)

skrub/_datetime_encoder.py
rcap107: at this point, wouldn't it be fine to keep as is, in case we end up exposing those classes? :thinking: maybe add a TODO to remember to set a default value if we do

Vincent-Maladiere: You can leave a TODO and revert it, I'd rather not have code that can't be reached

PR #1407, opened 2025-05-22
Add cleaner to narrative docs

GaelVaroquaux:
Add to the narrative docs, to make it easier to find.

The cleaning functionality of skrub is quite light but 1) it was recently praised on linkedin, 2) this PR makes it easier for people to find functionality

### DIFF
--- a/doc/cleaning.rst
+++ b/doc/cleaning.rst
@@ -6,6 +6,33 @@ Cleaning
 
 .. currentmodule:: skrub
 
+:class:`Cleaner`: sanitizing a dataframe
+=========================================
+
+:class:`Cleaner` sanitizes a dataframe, transforming it to a more
+consistent data representation with which it is easier to work. For
+instance, it detects null values represented as strings, parses dates,
+removes uninformative columns.
+
+To have reproducible transformations, it is implemented as a scikit-learn
+transformer:
+
+* to sanitize a given dataframe `df`::
+
+    >>> from skrub import Cleaner
+    >>> cleaner = Cleaner(drop_if_constant=True)
+    >>> clean_df = cleaner.fit_transform(df)
+
+* to apply the same exact operations to a new dataframe `df_test`
+  (new rows with the same columns)::
+
+   >>> clean_df_test = cleaner.transform(df_test)
+
+  Reusing the cleaner to transform new data ensures that if columns were
+  dropped on the first dataframe, they are dropped on the second.
+
+|
+
 :func:`deduplicate`: merging variants of the same entry
 =========================================================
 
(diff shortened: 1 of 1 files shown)


doc/cleaning.rst
```diff
@@ -6,6 +6,44 @@ Cleaning
 
 .. currentmodule:: skrub
 
+:class:`Cleaner`: sanitizing a dataframe
+=========================================
+
+:class:`Cleaner` sanitizes a dataframe, transforming it to a more
+consistent data representation with which it is easier to work. For
+instance, it detects null values represented as strings, parses dates,
+removes uninformative columns.
```
rcap107, line 15: Maybe mention that uninformative means "columns that contain only missing values, and constant columns"? 

I don't really like the "for instance" wording, I'd rather have something like "among other transformations..."

GaelVaroquaux: Happy to make a bit more explicit (but not too long, the number one problem with all we have is that it's too long, and people don't read it).

I'll push back on the "among other transformations": it makes longer with little increased precision

GaelVaroquaux: I like it! Thanks!!

doc/cleaning.rst
```diff
@@ -6,6 +6,44 @@ Cleaning
 
 .. currentmodule:: skrub
 
+:class:`Cleaner`: sanitizing a dataframe
+=========================================
+
+:class:`Cleaner` sanitizes a dataframe, transforming it to a more
+consistent data representation which is easier to work with: it detects
+null values represented as strings, parses dates, and removes
+uninformative columns (see docstring: :class:`Cleaner`).
```
Vincent-Maladiere, line 15: We display :class:`Cleaner` three times, should we remove the "see docstring" part?

doc/cleaning.rst
```diff
@@ -6,6 +6,44 @@ Cleaning
...
+    >>> cleaner = Cleaner(drop_if_constant=True)
+    >>> clean_df = cleaner.fit_transform(df)
+
+* to apply the same exact operations to a new dataframe `df_test`
+  (new rows with the same columns)::
+
+   >>> clean_df_test = cleaner.transform(df_test)
+
+  Reusing the cleaner to transform new data ensures that if columns were
+  dropped on the first dataframe, they are dropped on the second.
```
Vincent-Maladiere, line 43: Should add a link to our getting-started example, where we use the Cleaner in the  "# Generating an interactive report for a dataframe" section?

Vincent-Maladiere (approved):
Very helpful! Feel free to ignore my comments below

GaelVaroquaux:
> We display :class:Cleaner three times, should we remove the "see docstring" part?

I worry that people might not click on it. It's not a habit that beginners have.

> Should we add a link to our getting-started example, where we use the Cleaner in the "# Generating an interactive report for a dataframe" section?

IMHO, only if we do it by turning existing text into a link. I would not want to make this longer.

Issue #1408, opened 2025-05-22
Error of dataframe dispatch on non recognized types could be improved


GaelVaroquaux:
When giving an input that is not recognized by the dataframe dispatch mechanism, the error is not incredibly helpful:
```python
from skrub import datasets, TableReport
df = datasets.fetch_employee_salaries()
TableReport(df)
```

leads to:
```python
---------------------------------------------------------------------------
NotImplementedError                       Traceback (most recent call last)
      4 df = datasets.fetch_employee_salaries()
----> 5 TableReport(df)

File ~/dev/skrub/skrub/_reporting/_table_report.py:145, in TableReport.__init__(self, dataframe, n_rows, order_by, title, column_filters, verbose, max_plot_columns, max_association_columns)
    143 self.max_plot_columns = max_plot_columns
    144 self.max_association_columns = max_association_columns
--> 145 self.n_columns = sbd.shape(self.dataframe)[1]

File /usr/lib/python3.12/functools.py:909, in singledispatch.<locals>.wrapper(*args, **kw)
    905 if not args:
    906     raise TypeError(f'{funcname} requires at least '
    907                     '1 positional argument')
--> 909 return dispatch(args[0].__class__)(*args, **kw)

File ~/dev/skrub/skrub/_dataframe/_common.py:423, in shape(obj)
    421 @dispatch
    422 def shape(obj):
--> 423     raise NotImplementedError()

NotImplementedError: 
```

We could improve by adding a message to the NotImplementError, something like "Operation not supported on this object. Expecting a dataframe-like object."

PR #1416, opened 2025-05-26
Improve error message for non-dataframe input in the TableReport

Vincent-Maladiere:
Fixes https://github.com/skrub-data/skrub/issues/1408

Additionally, we could add this error message to all `sbd` functions. WDYT?

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -420,7 +420,10 @@ def _collect_polars_lazyframe(df):
 
 @dispatch
 def shape(obj):
-    raise NotImplementedError()
+    raise NotImplementedError(
+        "Operation not supported on this object. Expecting a Pandas or Polars"
+        f"dataframe, but got an object of type {type(obj)}."
+    )
 
 
 @shape.specialize("pandas")
--- a/skrub/_reporting/tests/test_table_report.py
+++ b/skrub/_reporting/tests/test_table_report.py
@@ -7,7 +7,7 @@ from pathlib import Path
 
 import pytest
 
-from skrub import TableReport, ToDatetime
+from skrub import TableReport, ToDatetime, datasets
 from skrub import _dataframe as sbd
 
 
@@ -268,3 +268,9 @@ def test_minimal_mode(pd_module):
     assert "data-test-associations-skipped" in html
     assert 'id="column-summaries-panel"' not in html
     assert 'id="column-associations-panel"' not in html
+
+
+def test_error_input_type():
+    df = datasets.fetch_employee_salaries()
+    with pytest.raises(NotImplementedError, match="not supported on this object"):
+        TableReport(df)


jeromedockes:
> Additionally, we could add this error message to all sbd functions. WDYT?

i think that would be good, and I would suggest replacing the NotImplementedError with a TypeError -- for user point of view the issue is not a missing implementation, it's that they passed an unsupported type

skrub/_dataframe/_common.py
```diff
@@ -116,10 +116,17 @@
 #
 
 
+def _raise(obj, kind="object"):
```
rcap107, line 120: maybe set "Dataframe" instead of "object" by default?

skrub/_dataframe/_common.py
```diff
@@ -193,7 +200,7 @@ def _is_column_polars(obj):
 
 @dispatch
 def to_list(col):
-    raise NotImplementedError()
+    raise _raise(col, type="Series")
```
rcap107: typo I think
```suggestion
    raise _raise(col, kind="Series")
```

skrub/_dataframe/_common.py
Vincent-Maladiere: no special opinion, obj appeared as slightly more generic but I can use dataframe as well

rcap107: if "obj" can only be series or dataframe, I'd lean towards being explicit about it and calling it dataframe

Vincent-Maladiere: It can also be a polars LazyFrame (and EagerFrame although we don't use it)

rcap107: maybe just "Frame" then? it's a bit weird to me to have Series in one case and an unspecified "object" in the other

rcap107:
I just realized there are more sbd functions spread around other files that should be updated, like in _datetime_encoder.py and _to_datetime.py

skrub/_dataframe/_common.py
Vincent-Maladiere: "Object" includes dataframe + series + any other type. In a function like the following, we need to stay generic in the error message:

```python
@dispatch
def is_pandas_extension_dtype(obj):
    raise _raise(obj)


@is_pandas_extension_dtype.specialize("pandas")
def _is_pandas_extension_dtype_pandas(obj):
    return pd.api.types.is_extension_array_dtype(obj)


@is_pandas_extension_dtype.specialize("polars")
def _is_pandas_extension_dtype_polars(obj):
    return False
```

It makes more sense to have the more generic type as the default IMO, but I don't have a strong opinion. These functions are not user-facing.

GaelVaroquaux:
> I think it's worth bringing them into the _common.py file for consistency and discoverability (currently, it's easy to forget them, as I just have)

As long as it does not lead to circular imports (which can be circumvented, but are a hint of bad code organization)

Vincent-Maladiere:
> As long as it does not lead to circular imports (which can be circumvented, but are a hint of bad code organization)

I reverted my message after further reading, as I now understand that these functions make sense within their respective contexts

skrub/_dataframe/_common.py
Vincent-Maladiere: WDYT?

rcap107: We can go with the more generic default, I just found it a bit weird to have "Series" and "object" as two possible values

Vincent-Maladiere: I'm not sure I understand your point: the argument "kind" can be "DataFrame", "Series", "LazyFrame" and "object".

rcap107: My understanding was that the message would be `... of type Series` for series and or `... of type Object` when kind was not "Series", so for all other objects (Dataframes, Lazyframes etc.) 

So that any time the kind was left unspecified, the default value would be "object", and I found it weird to have either "series" or "object", which is why I suggested to change it to "dataframe". 

In my last comment I thought there could only be one possible return value, so in that case I would have been in favor of having only a generic return value, but that's not the case so you can dismiss it

Vincent-Maladiere: ah got it, yes most of the time we have `kind="DataFrame"` or `kind="Series"`, and sometimes `kind="object"`

Vincent-Maladiere:
Technically, the coverage should be the same, because we're replacing an error with another one, but I can add a few tests to make codecov happier

rcap107:
> Technically, the coverage should be the same, because we're replacing an error with another one, but I can add a few tests to make codecov happier

makes sense, but why is it complaining then :thinking: just coverage things?

Vincent-Maladiere:
I'm not sure, Codecov has been quite buggy in the past

GaelVaroquaux:
> I'm not sure, Codecov has been quite buggy in the past

Yes. It's not incredibly reliable.

rcap107:
Thanks a lot @Vincent-Maladiere!

PR #1418, opened 2025-05-27
Add subsampling hint in TableReport used in expression

### LINKED ISSUE #1384: Add subsampling hint in TableReport used in expression
At the bottom of the first tab of the TableReport, there is some space (where it says "100 rows ✕ 9 columns. "):
Image

We should use this to add usability messages in general, when the TableReport is called from a given context.

Specifically, in the case of subsampled previews, we should write here "100 rows ✕ 9 columns (subsampled from xxxx rows)".

Vincent-Maladiere:
Fixes #1384

Need to add tests

Screenshot 2025-05-27 at 16 07 34

### DIFF
--- a/skrub/_expressions/_inspection.py
+++ b/skrub/_expressions/_inspection.py
@@ -20,6 +20,7 @@ from . import _utils
 from ._choosing import BaseNumericChoice, Choice
 from ._evaluation import choice_graph, clear_results, evaluate, graph, param_grid
 from ._expressions import Apply, Value, Var
+from ._subsampling import uses_subsampling
 
 
 def _get_jinja_env():
@@ -54,6 +55,10 @@ def node_report(expr, mode="preview", environment=None, **report_kwargs):
         report_kwargs.setdefault("verbose", False)  # Hide the progress bar
         report = TableReport(result, **report_kwargs)
         report._set_minimal_mode()
+        if uses_subsampling(expr):
+            # Evaluate the result without subsampling
+            full_result = expr.skb.eval(keep_subsampling=False)
+            report._display_subsample_info(sbd.shape(full_result)[0])
     else:
         try:
             report = result._repr_html_()
--- a/skrub/_reporting/_data/templates/dataframe-sample.html
+++ b/skrub/_reporting/_data/templates/dataframe-sample.html
@@ -85,7 +85,11 @@ data-spans__{{ i }}__{{ j }}
     <div>
         <strong>{{ summary.n_rows | format_number }}</strong> rows ✕
         <strong data-manager="ColumnFilterMatchCount"
-            data-test="n-columns-display">{{ summary.n_columns | format_number }}</strong> columns.
+            data-test="n-columns-display">{{ summary.n_columns | format_number }}</strong> columns
+        {% if 'original_n_samples' in summary %}
+         (subsampled from <strong>{{ summary.original_n_samples | format_number}}
+        </strong> rows)
+        {% endif %}
     </div>
     <div class="keyboard-hints">
         <div>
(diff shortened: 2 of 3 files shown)


skrub/_expressions/_inspection.py
```diff
@@ -54,6 +55,10 @@ def node_report(expr, mode="preview", environment=None, **report_kwargs):
         report_kwargs.setdefault("verbose", False)  # Hide the progress bar
         report = TableReport(result, **report_kwargs)
         report._set_minimal_mode()
+        if uses_subsampling(expr):
+            # Evaluate the result without subsampling
+            full_result = expr.skb.eval(keep_subsampling=False)
```
rcap107, line 60: `full_result` is needed to get the number of rows the dataframe started with? doesn't evaluating the full expression on all samples defeat the point of sampling it in the first place?

rcap107: I think it's still useful to have the "subsampled" hint if subsampling was done at any point, and just avoid adding the starting number of rows

Vincent-Maladiere:
Screenshot 2025-05-27 at 16 48 38

GaelVaroquaux:
Formulation: I would say "subsampled from more rows"

Vincent-Maladiere:
Screenshot 2025-06-02 at 15 24 09

skrub/_expressions/_inspection.py
```diff
@@ -54,6 +55,8 @@ def node_report(expr, mode="preview", environment=None, **report_kwargs):
         report_kwargs.setdefault("verbose", False)  # Hide the progress bar
         report = TableReport(result, **report_kwargs)
         report._set_minimal_mode()
+        if uses_subsampling(expr):
```
rcap107, line 58: is it possible to add a test for this? if it's necessary

rcap107 (approved):
Looks good to me, I think it makes it clear enough when subsampling is being used

Thanks @Vincent-Maladiere

skrub/_expressions/tests/test_inspection.py
```diff
@@ -58,6 +58,13 @@ def test_full_report():
     assert "This step did not run" in (out / "node_4.html").read_text("utf-8")
 
 
+@pytest.mark.skipif(not _inspection._has_graphviz(), reason="report requires graphviz")
+def test_preview_subsample():
+    X = datasets.fetch_employee_salaries().X
+    preview = skrub.X(X).skb.subsample(n=3)._repr_html_()
+    assert "subsample" in preview
```
rcap107, line 65: I'm wondering if this test could be expanded to make coverage happy

skrub/_expressions/_inspection.py
Vincent-Maladiere: this line should is during test, I think it's a bug on codecov side

jeromedockes:
thanks @Vincent-Maladiere ! a bit of css needs to be adjusted: normally the keyboard hints don't show up if there is not enough space for them and the shape does not wrap but as the text is now longer that needs to be tweaked

without subsampling no wrapping and either all are shown or hidden but they do not get truncated:

screenshot_2025-06-09T21:37:29+02:00

screenshot_2025-06-09T21:37:41+02:00


with subsampling

screenshot_2025-06-09T21:33:24+02:00

jeromedockes:
a simple thing you can do is adjust the "25ch" [here](https://github.com/skrub-data/skrub/blob/f03ba44edb55dfc75fea91b126254a56c3b8a7e8/skrub/_reporting/_data/templates/dataframe-sample.css#L162) to put something larger . in order to change it only when there is subsampling maybe you can introduce a new css class like .keyboard-hints-when-subsampling that overrides the flex-basis and in the template add that class only when the subsampling message is shown

Vincent-Maladiere:
Woops, I needed to take Jerome's answer into account. Will do in another PR

GaelVaroquaux:
> Woops, I needed to take Jerome's answer into account. Will do in another PR

Yes, please do that. Those "details" are important. Thanks!

Issue #1377, opened 2025-05-12
Adding a skrub global config


jeromedockes:
### Problem Description

Having something similar to pandas `options`, scikit-learn `config` etc. to tweak the behavior of skrub in one place. things like

- should the expressions display use tablereport or not
- should tablereport show plots or not, compute correlations or not
- maybe directories for skrub pipeline full reports, joblib cache, ...
- once subsampling in expressions is merged, enable or disable it globally

### Feature Description

something like 
```python
skrub.config['use_tablereport'] = False

with skrub.config_context(use_tablereport=False):
    ...
```

or similar

jeromedockes:
maybe the default config could actually be read from environment variables for users who want to set it in a more permanent way

Vincent-Maladiere:
> maybe the default config could actually be read from environment variables for users who want to set it in a more permanent way

Totally agree, env variables would help!

jeromedockes:
list of the options we want:

- [ ] using tablereport or not in expressions
- [ ] using tablereport or not for dataframe displays (replace patch_display)
- [ ] plot and association thresholds for tablereport
- [ ] random seed for subsampling in expressions (see [this comment](https://github.com/skrub-data/skrub/pull/1328#discussion_r2089422816))
- [ ] force or disable subsampling in all contexts
- [ ]

PR #1427, opened 2025-06-03
Add a skrub config manager

Vincent-Maladiere:
Fixes https://github.com/skrub-data/skrub/issues/1377

### DIFF
--- a/skrub/_config.py
+++ b/skrub/_config.py
@@ -10,7 +10,8 @@ from ._reporting import _patching
 _global_config = {
     "use_tablereport": os.environ.get("SKB_USE_TABLEREPORT", False),
     "use_tablereport_expr": os.environ.get("SKB_USE_TABLEREPORT_EXPR", True),
-    "tablereport_max_col": int(os.environ.get("SKB_TABLEREPORT_MAX_COL", 30)),
+    "max_plot_columns": int(os.environ.get("SKB_MAX_PLOT_COLUMNS", 30)),
+    "max_association_columns": int(os.environ.get("SKB_MAX_ASSOCIATION_COLUMNS", 30)),
     "subsampling_seed": int(os.environ.get("SKB_SUBSAMPLING_SEED", 0)),
     "enable_subsampling": os.environ.get("SKB_ENABLE_SUBSAMPLING", "default"),
 }
@@ -18,6 +19,11 @@ _threadlocal = threading.local()
 
 
 def _get_threadlocal_config():
+    """Return a thread-local copy of the global configuration.
+
+    This is used to ensure that each thread has its own configuration
+    without affecting the global configuration.
+    """
     if not hasattr(_threadlocal, "global_config"):
         _threadlocal.global_config = _global_config.copy()
     return _threadlocal.global_config
@@ -49,8 +55,8 @@ def get_config():
 def _apply_external_patches(config):
     if config["use_tablereport"]:
         _patching._patch_display(
-            max_plot_columns=config["tablereport_max_col"],
-            max_association_columns=config["tablereport_max_col"],
+            max_plot_columns=config["max_plot_columns"],
+            max_association_columns=config["max_plot_columns"],
         )
     else:
         # No-op if dispatch haven't been previously enabled
@@ -60,7 +66,8 @@ def _apply_external_patches(config):
 def set_config(
     use_tablereport=None,
     use_tablereport_expr=None,
-    tablereport_max_col=None,
+    max_plot_columns=None,
+    max_association_columns=None,
     subsampling_seed=None,
     enable_subsampling=None,
 ):
@@ -90,11 +97,18 @@ def set_config(
         This configuration can also be set with the ``SKB_USE_TABLEREPORT_EXPR``
         environment variable.
 
-    tablereport_max_col : int, default=None
-        Set both the ``max_plot_columns`` and ``max_association_columns`` argument
-        of :class:`~skrub.TableReport`. Default is 30.
+    max_plot_columns : int, default=None
+        Set the ``max_plot_columns`` argument of :class:`~skrub.TableReport`.
+        Default is 30.
 
-        This configuration can also be set with the ``SKB_TABLEREPORT_MAX_COL``
+        This configuration can also be set with the ``SKB_MAX_PLOT_COLUMNS``
+        environment variable.
+
+    max_association_columns : int, default=None
+        Set the ``max_association_columns`` argument of :class:`~skrub.TableReport`.
+        Default is 30.
+
+        This configuration can also be set with the ``SKB_MAX_ASSOCIATION_COLUMNS``
         environment variable.
 
     subsampling_seed : int, default=None
@@ -112,7 +126,7 @@ def set_config(
         - If ``"disable"``, subsampling is never used, so ``skb.subsample`` becomes a
           no-op.
         - If ``"force"``, subsampling is used in all expression evaluation modes
-          (preview, fit_transform, etc.).
+          (:func:`~skrub.Expr.skb.eval`, fit_transform, etc.).
 
         This configuration can also be set with the ``SKB_ENABLE_SUBSAMPLING``
         environment variable.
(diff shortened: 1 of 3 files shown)


skrub/_config.py
```diff
@@ -0,0 +1,258 @@
...
+from contextlib import contextmanager
+
+import numpy as np
+
+from ._reporting import _patching
+
+_global_config = {
+    "use_tablereport": os.environ.get("SKB_USE_TABLEREPORT", False),
+    "use_tablereport_expr": os.environ.get("SKB_USE_TABLEREPORT_EXPR", True),
+    "tablereport_threshold": int(os.environ.get("SKB_TABLEREPORT_THRESHOLD", 30)),
```
rcap107:
```suggestion
    "tablereport_col_threshold": int(os.environ.get("SKB_TABLEREPORT_COL_THRESHOLD", 30)),
```

skrub/_config.py
```diff
@@ -0,0 +1,258 @@
...
+    "subsampling_seed": int(os.environ.get("SKB_SUBSAMPLING_SEED", 0)),
+    "enable_subsampling": os.environ.get("SKB_ENABLE_SUBSAMPLING", "default"),
+}
+_threadlocal = threading.local()
+
+
+def _get_threadlocal_config():
```
rcap107, line 21: why do we need this function? (trying to understand the code better)

skrub/_config.py
```diff
@@ -0,0 +1,258 @@
...
+    tablereport_threshold=None,
+    subsampling_seed=None,
+    enable_subsampling=None,
+):
+    """Set global skrub configuration.
+
+    Parameters
+    ----------
+    use_tablereport : bool, default=None
+        The type of HTML representation used for the dataframes preview in skrub
```
rcap107, line 90: swapped docstring with the `_expr` parameter

skrub/_config.py
rcap107: or perhaps
```suggestion
    "tablereport_max_col": int(os.environ.get("SKB_TABLEREPORT_MAX_COL", 30)),
```

skrub/_config.py
```diff
@@ -0,0 +1,258 @@
...
+
+        - If ``True``, replace the default DataFrame HTML displays with
+          :class:`~skrub.TableReport`.
+        - If ``False``, the original Pandas or Polars dataframe HTML representation
+          will be used.
+
+        This configuration can also be set with the ``SKB_USE_TABLEREPORT_EXPR``
+        environment variable.
+
+    tablereport_threshold : int, default=None
```
rcap107: like above, it would also match the other parameter names

```suggestion
    tablereport_max_col : int, default=None
```

skrub/_config.py
```diff
@@ -0,0 +1,258 @@
...
+        environment variable.
+
+    subsampling_seed : int, default=None
+        Set the random seed of subsampling in skrub expressions
+        :func:`skrub.Expr.skb.subsample`, when ``how="random"`` is passed.
+
+        This configuration can also be set with the ``SKB_SUBSAMPLING_SEED`` environment
+        variable.
+
+    enable_subsampling : {'default', 'disable', 'force'}, default=None
```
rcap107, line 121: what's the use case for `force`?

skrub/_reporting/_table_report.py
```diff
@@ -139,8 +165,16 @@ def __init__(
         self.title = title
         self.column_filters = column_filters
         self.verbose = verbose
-        self.max_plot_columns = max_plot_columns
-        self.max_association_columns = max_association_columns
+        self.max_plot_columns = (
```
rcap107: I just realized there is no error checking anywhere for the type and value of `max_plot_columns` and `max_association_column`, I think we should add it

skrub/_reporting/_table_report.py
```diff
@@ -47,12 +48,37 @@ class TableReport:
...
+        change the default using :func:`set_config`:
+
+        >>> from skrub import set_config
+        >>> set_config(tablereport_threshold=30)
+
+        You can also enable this default more permanently via an environment variable:
+
+        .. code:: shell
+
+            export SKB_TABLEREPORT_THRESHOLD=30
```
rcap107: I'm wondering if the config variables for max_association and max_plot should be split, so that a user can turn off association columns but still have the plots for example

skrub/_reporting/_patching.py
```diff
@@ -131,3 +159,14 @@ def unpatch_display(pandas=True, polars=True):
         Directly create a report from a dataframe.
     """
     _change_display(_unpatch, _get_to_patch(pandas=pandas, polars=polars))
+    warnings.warn(
```
glemaitre, line 162: This looks not covered.

skrub/_reporting/tests/test_patch_display.py
```diff
@@ -8,88 +8,90 @@
 @pytest.mark.parametrize("repeat_patch", [1, 2])
 @pytest.mark.parametrize("repeat_unpatch", [1, 2])
 def test_patch_display(df_module, repeat_patch, repeat_unpatch, capsys):
-    df = df_module.make_dataframe(
-        dict(
-            a=[1, 2, 3, 4],
-            b=["one", "two", "three", "four"],
-            c=[11.1, 11.2, 11.3, 11.4],
+    with pytest.raises(DeprecationWarning):
```
glemaitre: `DeprecationWarning` will be hidden to the end-user (it is only for the developer). Since a user could have used  `patch_display`, a `FutureWarning` is more appropriate.

glemaitre: Also I think that you are going to match a single call to it. So it would be best to have this pattern:

```python
with pytest.raises(FutureWarning, match="..."):
    patch_display(...)
assert ...
with pytest.raises(FutureWarning, match="..."):
    unpatch_display(...)
```

skrub/tests/test_config.py
```diff
@@ -0,0 +1,124 @@
...
+    # Default: subsampling during preview mode
+    assert evaluate(expr.skb.subsample(n=3), mode="preview").shape[0] == 3
+
+    with config_context(enable_subsampling="disable"):
+        assert evaluate(expr.skb.subsample(n=3), mode="preview").shape[0] == X.shape[0]
+        with config_context(enable_subsampling="default"):
+            assert evaluate(expr.skb.subsample(n=3), mode="preview").shape[0] == 3
+
+
+def test_error():
```
glemaitre: You can refactor an parametrize the test:

```python
@pytest.mark.parametrize(
    "params, err_msg", [
         (dict(use_tablereport="hello"), "..."),
         ...
    ]
)
def test_error(params, err_msg):
    with pytest.raises(ValueError, match=err_msg):
        set_config(**params)
```

The parametrization allows you to know exactly which configuration of the test is potentially failing. It is kind of nice.

skrub/_config.py
glemaitre: It is hard for me without knowing about skrub `TableReport`, what threshold means. I would really expect to have columns in the name.

skrub/_reporting/tests/test_patch_display.py
Vincent-Maladiere: Right, thanks for pointing that out, I realised this somewhere else recently

skrub/_config.py
Vincent-Maladiere: Currently, in the expressions, you don't have a way to run `.skb.eval` (or `fit_predict`) with subsampling. The 'force' option would allow users to explicitly ask for it

skrub/tests/test_config.py
Vincent-Maladiere: heu right wtf Vincent

skrub/_config.py
Vincent-Maladiere: Let's go with `SKB_TABLEREPORT_MAX_COL`

skrub/_config.py
Vincent-Maladiere: Yes, so this function handles reading the thread-local config. The first time this config is read, it sets a copy of the global default `_global_config` to our thread-local variable `_threadlocal`, so that modifying the config only modifies `_threadlocal`, but not `_global_config`. It's a way to define different configurations for different threads. It's a direct adaptation from `sklearn._config.py`. Does that make more sense?

skrub/_reporting/_table_report.py
Vincent-Maladiere: I don't think we need this flexibility, but this can be up for discussion

skrub/_config.py
Vincent-Maladiere: WDYT?

skrub/_config.py
Vincent-Maladiere: I can add a comment to explain this

skrub/_config.py
rcap107: it makes sense, but what you explained should probably be added somewhere

skrub/_config.py
rcap107: it would be helpful, yes

skrub/_reporting/_table_report.py
rcap107: my reasoning for separating them is that I can see myself disabling the associations for all tables, while keeping the barplots instead

it's also a conversation we had with @jeromedockes around the time the relevant PRs were being merged

I think this is the kind of parameter that should be exposed to the user in a config, while it may not be as useful when calling TableReport on its own

Vincent-Maladiere: Ok, will add

PR #1440, opened 2025-06-13
Exposing conversion to Float32 in the Cleaner

### LINKED ISSUE #1439: Expose the conversion to float32 as a parameter for the Cleaner
As discussed on mattermost, the Cleaner is more likely to be used with the expressions as the TableVectorizer may be too rigid for some use cases. 

The Cleaner does a lot of the processing steps done by the TableVectorizer, but does not parse numbers or cast them to float32. This option should be exposed to the user.

rcap107:
Fixes #1439

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -182,8 +182,9 @@ class Cleaner(TransformerMixin, BaseEstimator):
     datetime_format : str, default=None
         The format to use when parsing dates. If None, the format is inferred.
 
-    convert_float32 : bool, default=False
-        If set to ``True``, convert numeric columns to ``np.float32`` dtype.
+    numerical_dtype : "float32" or None, default=None
+        If set to ``float32``, convert numeric columns to ``np.float32`` dtype. If
+        ``None``, numerical dtypes are not modified.
 
     n_jobs : int, default=None
         Number of jobs to run in parallel.
@@ -228,7 +229,7 @@ class Cleaner(TransformerMixin, BaseEstimator):
     - ``ToStr()``: convert columns to strings, unless they are numerical,
       categorical, or datetime.
 
-    If ``convert_float32`` is set to ``True``, the ``Cleaner`` will also convert
+    If ``numerical_dtype`` is set to ``float32``, the ``Cleaner`` will also convert
     numeric columns to ``np.float32`` dtype, ensuring a consistent representation
     of numbers and missing values. This can be useful if the ``Cleaner``
     is used as a preprocessing step in a skrub pipeline.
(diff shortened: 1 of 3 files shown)


rcap107:
I'm wondering if converting to float32 should be an option for the config manager #1427

skrub/_table_vectorizer.py
```diff
@@ -182,6 +182,9 @@ class Cleaner(TransformerMixin, BaseEstimator):
     datetime_format : str, default=None
         The format to use when parsing dates. If None, the format is inferred.
 
+    convert_float32 : bool, default=False
```
GaelVaroquaux: Here we are using a different API as in the TableVectorizer (and as in a few other objects in scikit-learn that do type conversions, I believe).

I understand why, but the differences are quite large (different parameter name, different type of accepted value), and it will make it hard for the user to memorize. Can we get closer to the canonical API?

GaelVaroquaux: eg we could have `numerical_dtype="float32"`

rcap107: here the option is either "convert to float32", or "keep as is", if we use `numerical_dtype="float32"`, what would be the other option? 

I am open to changing the name of the parameter, but given that in this case it's really an on-off switch I think a boolean flag makes the most sense

rcap107: do you have an example of the scikit-learn objects?

glemaitre: For `OneHotEncoder` and others, we usually have a `dtype` parameter (https://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.OneHotEncoder.html) to control the output dtype.

rcap107: something like `numerical_dtype=None` by default when numbers are not converted and `numerical_dtype="float32"` when the conversion must be done? 

tbh I'm still not convinced this is better than having boolean flag here

GaelVaroquaux:
> here the option is either "convert to float32", or "keep as is", if we use numerical_dtype="float32", what would be the other option?

"None", I would expect, to do nothing.

GaelVaroquaux:
> something like numerical_dtype=None by default when numbers are not converted and numerical_dtype="float32" when the conversion must be done?

Yes

> tbh I'm still not convinced this is better than having boolean flag here

It's about minimizing the specificities of each object: they should have a similar look and feel (like school uniforms :D )

rcap107:
ready for review

skrub/tests/test_table_vectorizer.py
```diff
@@ -405,6 +405,17 @@ def test_convert_float32():
     assert sbd.is_float(out["float"])
     assert sbd.is_integer(out["int"])
 
+    vectorizer = Cleaner(numerical_dtype="float32")
+    out = vectorizer.fit_transform(X)
+    assert sbd.is_float(out["float"])
```
GaelVaroquaux: Does this check for float, or float32?

rcap107: good catch, it only checks for float

rcap107: so I just realized that this test, like all other tests in table_vectorizer, is not testing for pandas, pandas-nullable-types and polars

I'll do a quick fix to get this test through, but we'll have to update all tests

GaelVaroquaux:
> I'll do a quick fix to get this test through, but we'll have to update all tests

Yes, let's just worry about the Float32 functionality in this PR and address the issue that you mention here in another PR.

jeromedockes:
just a small comment, I think in skrub we have been using "numeric" and not "numeric**AL**" everywhere so it would be nice to stay consistent

GaelVaroquaux (approved):
LGTM. Thanks!

PR #1442, opened 2025-06-15
[DISCUSSION] Adding `.skb.iter_eval`?

jeromedockes:
I wonder if it would be useful to allow evaluating an expression step by step, to inspect the results, debug, build reports, test etc. We could add a method `.skb.iter_eval()` that returns a generator which yields each sub-expression and its result as it evaluates them

```
>>> import skrub

>>> a = skrub.var('a')
>>> b = skrub.var('b')
>>> c = a * a + b

>>> for item in c.skb.iter_eval({'a': 10, 'b': 20}):
...     print(item)
(<Var 'a'>, 10)
(<BinOp: mul>, 100)
(<Var 'b'>, 20)
(<BinOp: add>, 120)
```

(note as the pr shows in terms of implementation there is almost nothing to do, as we already use a loop and not recursion to evaluate the expressions)

### DIFF
--- a/skrub/_expressions/_evaluation.py
+++ b/skrub/_expressions/_evaluation.py
@@ -60,12 +60,21 @@ class _Computation:
     # >>> skrub.as_expr(d).skb.eval()
 
     def __init__(self, target, generator):
+        self.target = target
         self.target_id = id(target)
         self.generator = generator
 
 
 class CircularReferenceError(ValueError):
     pass
+
+
+def _run_gen(g):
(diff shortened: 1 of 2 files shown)


glemaitre:
When @jeromedockes presented this feature, I thought about some discussions that we had in the past quite related (@GaelVaroquaux): being able to have a sort of probe in the graph for instance to get the feature importance. I'm wondering if this way of iterating would be useful to go in this direction.

PR #1445, opened 2025-06-17
Provide informative errors when expressions are passed as values to skrub variables

### LINKED ISSUE #1438: Better error message when an expression is passed to TableVectorizer or to other expressions
I was writing an example with the expressions, and when I tried to combine them with the TableVectorizer I hit an exception with an error message that wasn't very clear: 

```python
from skrub import StringEncoder, TextEncoder
enc = skrub.choose_from({"string": StringEncoder(), "text": TextEncoder()})

TableVectorizer(high_cardinality=enc) # Wrong

import pandas as pd
df = pd.DataFrame({"a": ["a", "b", "c"]})

TableVectorizer(high_cardinality=enc).fit_transform(df)
```

This gives as error message

```

TypeError: Cannot clone object 'choose_from({'string': StringEncoder(), 'text': TextEncoder()})' (type <class 'skrub._expressions._choosing.Choice'>): it does not seem to be a scikit-learn estimator as it does not implement a 'get_params' method.

```

Which makes sense since I'm passing an expression rather than an encoder. I think it may be confusing for users that are learning how to use expressions and the error message is not very clear about waht the error is, so we might want to detect whether the object that is passed is an expression and raise an error accordingly. 

Edit to add:

A similar problem happens if I try to define a skrub variable and use another expression as its value. The code should fail early rather than raising an error at the end when I'm crossvalidating the model

rcap107:
Fixes #1438

Error checking for the value in the variables is already done. I thought to implement both improvements in the same PR, but I'm not sure how to implement the error checking for the encoders. I was thinking about doing it in `_expressions/tests/test_evaluation.py` or somewhere in `_expressions`, but maybe I should add that to the tests for each encoder?

### DIFF
--- a/skrub/_expressions/_expressions.py
+++ b/skrub/_expressions/_expressions.py
@@ -797,7 +797,7 @@ def check_value_type(value):
 
     if needs_eval(value):
         raise TypeError(
-            "The `value` of a `skrub.var()` must not be a skrub"
+            "The `value` of a `skrub.var()` must not be a skrub "
             f"expression or skrub choice. Got: {type(value)}."
         )
 
--- a/skrub/_expressions/_expressions.py
+++ b/skrub/_expressions/_expressions.py
@@ -48,7 +48,7 @@ from .._reporting._utils import strip_xml_declaration
 from .._utils import PassThrough, short_repr
 from .._wrap_transformer import wrap_transformer
 from . import _utils
-from ._choosing import get_chosen_or_default
+from ._choosing import BaseChoice, get_chosen_or_default
 from ._utils import FITTED_PREDICTOR_METHODS, NULL, attribute_error
 
 __all__ = [
@@ -792,10 +792,10 @@ def check_name(name, is_var):
         )
 
 
-def check_value_type(value):
-    from ._evaluation import needs_eval
-
-    if needs_eval(value):
+def check_var_value(value):
+    """Checking that the value passed to a skrub variable is not an expression
+    or a choice."""
+    if isinstance(value, (BaseChoice, Expr)):
         raise TypeError(
             "The `value` of a `skrub.var()` must not be a skrub"
             f"expression or skrub choice. Got: {type(value)}."
@@ -927,7 +927,7 @@ def var(name, value=NULL):
     gallery.
     """
     check_name(name, is_var=True)
-    check_value_type(value)
+    check_var_value(value)
     return Expr(Var(name, value=value))
 
 
@@ -988,7 +988,7 @@ def X(value=NULL):
     >>> X.skb.is_X
     True
     """
-    check_value_type(value)
+    check_var_value(value)
     return Expr(Var("X", value=value)).skb.mark_as_X()
 
 
@@ -1050,7 +1050,7 @@ def y(value=NULL):
     >>> y.skb.is_y
     True
     """
-    check_value_type(value)
+    check_var_value(value)
     return Expr(Var("y", value=value)).skb.mark_as_y()
 
 
--- a/skrub/_expressions/_expressions.py
+++ b/skrub/_expressions/_expressions.py
@@ -792,8 +792,19 @@ def check_name(name, is_var):
         )
 
 
+def check_value_type(value):
+    from ._evaluation import needs_eval
+
+    if needs_eval(value):
+        raise TypeError(
+            "The `value` of a `skrub.var()` must not be a skrub"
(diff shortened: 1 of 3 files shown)


rcap107:
Discussing with @jeromedockes, we can't fix the error message for estimators defined outside skrub (which is the majority) without patching scikit-learn, so that won't be done in this PR

Given that, the PR is ready for review since the smaller check has been implemented

skrub/_expressions/_expressions.py
```diff
@@ -792,8 +792,19 @@ def check_name(name, is_var):
         )
 
 
+def check_value_type(value):
+    from ._evaluation import needs_eval
+
+    if needs_eval(value):
```
Vincent-Maladiere, line 800: I think this can be simplified to:
```python
    isinstance(value, (Expr, BaseChoice))
```
since we won't be able to detect expressions or choices nested in arbitrary objects.

skrub/_expressions/_expressions.py
```diff
@@ -792,8 +792,19 @@ def check_name(name, is_var):
         )
 
 
+def check_value_type(value):
```
Vincent-Maladiere: Maybe the link to `var` could be more explicit, like:
```python
    def check_var_value(value)
```

skrub/_expressions/_expressions.py
```diff
@@ -792,8 +792,19 @@ def check_name(name, is_var):
...
+    from ._evaluation import needs_eval
+
+    if needs_eval(value):
+        raise TypeError(
+            "The `value` of a `skrub.var()` must not be a skrub"
```
Vincent-Maladiere:
```suggestion
            "The `value` of a `skrub.var()` must not be a skrub "
```

Vincent-Maladiere (commented):
Hey Riccardo, here are some comments on this PR

skrub/_expressions/tests/test_evaluation.py
```diff
@@ -116,6 +116,14 @@ def test_needs_eval():
     )
 
 
+def test_skrub_encoder_error():
```
Vincent-Maladiere: Is it a TODO?

rcap107: whoops

it was meant to be a todo, but we decided to skip that so I need to remove it

Vincent-Maladiere (approved):
Other than that, LGTM! Waiting for @jeromedockes's review if he finds the time

skrub/_expressions/_expressions.py
jeromedockes: needs_eval will also (recursively) look inside builtin collections and scikit-learn estimators so I think using it as @rcap107 had it makes sense, although I agree that in most cases when that error happens it will be because someone passed an expression directly

skrub/_expressions/_expressions.py
jeromedockes: and maybe make it clear that it is private to this module `_check_var_value`

skrub/_expressions/_expressions.py
```diff
@@ -792,8 +792,19 @@ def check_name(name, is_var):
         )
 
 
+def check_var_value(value):
+    """Checking that the value passed to a skrub variable is not an expression
+    or a choice."""
+    if isinstance(value, (BaseChoice, Expr)):
+        raise TypeError(
+            "The `value` of a `skrub.var()` must not be a skrub "
+            f"expression or skrub choice. Got: {type(value)}."
```
jeromedockes: if you decide to go for `needs_eval` instead of just an `isinstance` check, you might want to show the object instead of just its type and reword to "must not contain a skrub expression [...]"

jeromedockes: you could even show the problematic expression as is done in `_evaluation._check_environment` but that's probably overkill

jeromedockes (approved):

skrub/_expressions/_expressions.py
GaelVaroquaux: Showing both object value and type would be useful

skrub/_expressions/_expressions.py
Vincent-Maladiere: ah right I forgot that we handle BaseEstimator, so
```python
needs_eval(Ridge(alpha=skrub.var("a", 1)))
```
is True, indeed.
Feel free to ignore my comment then.

PR #1451, opened 2025-06-18
Improve .expr.subsample docs

### LINKED ISSUE #1446: Improve UX of  `.skb.subsample` when X and y are separate
I have the following piece of code
```python
data = fetch_medical_charge()

X = skrub.X(data.X)
y = skrub.y(data.y)

X = X.skb.subsample(n=1000)
```

I did not set `y.skb.subsample` because I was expecting that it would also be subsampled (somehow), but that was not the case and computing the preview with a hgb failed because the size of X and y was different. 

```
RuntimeError: Evaluation of 'apply()' failed.
You can see the full traceback above. The error message was:
ValueError: Found input variables with inconsistent numbers of samples: [1000, 163065]
```

This isn't a problem if X and y are taken from the same dataset that is then sampled, but when that's not the case there is a problem. 

I know that it's possible to sample the first 1000 values, but given that there can also be random sampling we might want to add a seed to be able to align X and y when they start separate. 

In any case we need to explain in the docs that each variable must be sampled, if it's not based on a variable that's been sampled earlier.

rcap107:
Fixes #1446 

Todo: 
- Add an example in the docstring
- Add a cross-reference to the user guide
- Expand a bit on this in the user guide

### DIFF
--- a/doc/skrub_pipeline.rst
+++ b/doc/skrub_pipeline.rst
@@ -396,6 +396,8 @@ Finally, there are other situations where using :func:`deferred` can be helpful:
 - See :ref:`example_tuning_pipelines` for an example of hyper-parameter tuning using
   skrub pipelines.
 
+.. _user_guide_subsampling_ref:
+
 Subsampling to develop pipelines with reduced computation
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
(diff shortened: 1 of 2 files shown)


rcap107:
Talking with @Vincent-Maladiere, we thought of improving the error message by detecting when one of X and y is sampled, but the other isn't. This should be done in `.skb.apply` when an estimator that takes `y` is used. 

It'll be a different PR though

rcap107:
ready for review, since we're not going to detect subsampling in this PR

Vincent-Maladiere (approved):
LGTM!

PR #1456, opened 2025-06-19
Fixing typo in StringEncoder title

### LINKED ISSUE #1455: [DOC] Typo in String Encoder title in examples

First, thanks for the examples with text encoders. It was really useful for me.

But when I was looking at the examples I have a typo in the [String Encoder example](https://skrub-data.org/stable/auto_examples/02_text_with_string_encoders.html#sringencoder)
Basically it is written "SringEncoder"

Image

The problem comes from `examples/02_text_with_string_encoders.py` in [line 230](https://github.com/skrub-data/skrub/blob/90c9d61628f59d9ed4601659dfb3d6a6b9f287a4/examples/02_text_with_string_encoders.py#L230)

I can open a PR to adjust this; it would be really quick.

I'm just wondering if the `|StringEncoder|` replacer works on titles. 
If yes, I could use all the replacers on all the titles in advance, if not, I would just fix the typo.

vitorpohlenz:
Fixes #1455

Simple typo fix in [String Encoder Example](https://skrub-data.org/stable/auto_examples/02_text_with_string_encoders.html#sringencoder)

It was written "SringEncoder"
Now it is fixed:

image

### DIFF
--- a/examples/02_text_with_string_encoders.py
+++ b/examples/02_text_with_string_encoders.py
@@ -227,7 +227,7 @@ results.append(("TextEncoder", text_encoder_results))
 plot_box_results(results)
 
 # %%
-# SringEncoder
+# StringEncoder
 # ^^^^^^^^^^^^
 # |TextEncoder| embeddings are very strong, but they are also quite expensive to
 # use. A simpler, faster alternative for encoding strings is the |StringEncoder|,


vitorpohlenz:
Since those are minor modifications.
If someone has more improvements that I could add to the same PR, please let me know.

Vincent-Maladiere (approved):
Thanks!

PR #1458, opened 2025-06-19
Shortening docstring preamble and moving text to the notes

### LINKED ISSUE #1424: Documentation from API to User Guide

Reviewing the documentation of the functions, some of the descriptions are a bit long (for example `joiner`). It might be simpler to the user, when working with the functions, to have a brief description in the API and move the rest of the documentation to the User Guide.
I noticed that the following functions had a long description and would benefit for a simplification:
- ToCategorical
- ToDatetime
- StringEncoder
- TextEncoder
- TableVectorizer
- tabular_learner
- DropUninformative
- Deferred
- eval_mode
- skrub.Expr.skb.full_report
- skrub.Expr.skb.get_pipeline
- skrub.Expr.skb.get_grid_search
- skrub.Expr.skb.mark_as_X
- skrub.Expr.skb.mark_as_y
- Column_associations

Also, I noticed that some didn't have a description (I don't know if it is intentional, if it's missing or it shouldn't be like that):
- Filter
- NameFilter
- Selector

_No response_

rcap107:
As reported in #1424, a lot of docstrings have extremely long preambles that make it hard to understand what the function/object is doing. 

This PR addresses that issue by shortening a lot the head of the docstring: most of the info that is not necessary to understand how the function works is moved to either the examples or the Notes section. 

This is a first iteration to improve the current status of the docstrings: later we might want to do a second pass and move some of what ends in the notes to the User Guide.


skrub/_drop_uninformative.py
```diff
@@ -40,6 +31,18 @@ class DropUninformative(SingleColumnTransformer):
...
+
+    - The fraction of missing values is larger than a certain fraction (by default,
+      all values must be null for the column to be dropped).
+    - The column includes only one unique value (the column is constant). Missing
+      values are considered a separate value.
+    - The number of unique values in the column is equal to the length of the
+      column, i.e., all values are unique. This is only considered for non-numeric
+      columns. Missing values are considered a separate value. Note that this
+      may lead to dropping columns that contain free-flowing text.
+
```
Vincent-Maladiere, line 45: This should probably be put under the Notes section rather than the Examples section

Vincent-Maladiere (approved):
This looks better, thanks!

Vincent-Maladiere (approved):
Looks great, thank you very much @rcap107!

rcap107:
We can always iterate  later

Issue #1285, opened 2025-04-07
Improve API page: inspiration from scikit-learn


GaelVaroquaux:
The scikit-learn API page has a really interesting layout:
https://scikit-learn.org/dev/api/index.html

We should borrow from it

rcap107:
With the addition of the expressions, the list of functions is very crowded and would definitely benefit from grouping up related objects

PR #1460, opened 2025-06-20
Improve doc api reference

Vincent-Maladiere:
Fixes https://github.com/skrub-data/skrub/issues/1285


doc/api_reference.py
```diff
@@ -0,0 +1,282 @@
...
+    page of each module.
+description (required, `None` if not needed)
+    The additional description for the module to be placed under the module
+    docstring, before the sections start.
+sections (required)
+    A list of sections, each of which consists of:
+    - title (required, `None` if not needed): the section title, commonly it should
+      not be `None` except for the first section of a module,
+    - description (optional): the optional additional description for the section,
+    - autosummary (required): an autosummary block, assuming current module is the
```
rcap107, line 23: is the autosummary supposed to be written manually? from the name it doesn't sound like it

doc/api_reference.py
```diff
@@ -0,0 +1,282 @@
...
+                    "GapEncoder",
+                    "SimilarityEncoder",
+                    "ToCategorical",
+                    "DatetimeEncoder",
+                    "ToDatetime",
+                    "to_datetime",
+                ],
+            },
+            {
+                "title": "Generating an HTML report",
```
rcap107: I'd rather have a more explicit title, like "exploring the dataframe"

doc/api_reference.py
```diff
@@ -0,0 +1,282 @@
...
+                    "Selector",
+                    "string",
+                ],
+            }
+        ],
+    },
+    "expressions": {
+        "short_summary": "",
+        "description": (
+            "Generalization the scikit-learn pipeline. "
```
rcap107:
```suggestion
            "Generalizing the scikit-learn pipeline. "
```

doc/api_reference.py
```diff
@@ -0,0 +1,282 @@
+"""Configuration for the API reference documentation.
+
+CONFIGURING API_REFERENCE
```
rcap107, line 6: Is this the file that we need to look at if we want to modify/add/remove objects, or change the order of the section etc.? 

Could you please add that a comment somewhere here or in the other docs to explain how and where to do that?

rcap107:
Thanks a lot @Vincent-Maladiere, this is already looking so much better! 

I think that most of the infrastructure is there, and now it's mostly a matter of deciding the order and the titles of the sections.

For example, I would like to reorder the objects in the `skrub` section of the section navigation index so that related objects were closer to each other (so that all the Encoders are next to each other, all the Joiners are next to each other and so on). Is that doable?

doc/api_reference.py
Vincent-Maladiere: Maybe a slight misnomer, "autosummary" represents the list of the objects you want to pass to the autosummary module (ie all objects to bring in the API reference)

rcap107:
Discussed during the skrub meeting: 

- The current grouping of the objects is not very intuitive: objects should be grouped by functionality more than by namespace (all encoders together, all joiners together etc.). We might want to reflect the grouping that is used in the user guide. 
- For the moment, adding a search bar was more complex than expected, so it should be added at a later time

GaelVaroquaux:
Thanks a lot for this PR.

I had a very quick look.

On the rendered page, I don't see the "section" titles appearing. For instance, the api_renference.py file suggests that the first part, the "pipeline" part, has a title "building a pipeline", but I don't see it:
image

I think that it would really help if they appeared as dividers.
I do note that it is not the case in the scikit-learn docs, so maybe it is not feasible.

Vincent-Maladiere:
Thank you for your feedback. It's probably feasible, but I aimed to stick to the scikit-learn style. In this instance, if you click on "pipeline" on the left sidebar, you'll see that the title is indeed "Building a pipeline". So it all boils down to what we want. WDYT?

GaelVaroquaux:
> It's probably feasible, but I aimed to stick to the scikit-learn style. 

If we can improve on it, it is even better :)

> In this instance, if you click on "pipeline" on the left sidebar, you'll see that the title is indeed "Building a pipeline".

Indeed, and that's pretty cool!

> So it all boils down to what we want. WDYT?

That's true. I think that if we could have the section title separating a bit the big table it would be cool. If it's hard, then we move on and merge as is.

Vincent-Maladiere:
I can give it a try!

GaelVaroquaux:
> I can give it a try!

Thanks! And if you hit a snag, tell us, and we'll move on.

Vincent-Maladiere:
@GaelVaroquaux is this last version better?

GaelVaroquaux:
This is amazing!!

One tiny remark: can we add somewhere in the dev guide how to add symbols to the API reference, to make it easier for people to find the relevant files. It's now a bit hard to find this

Vincent-Maladiere:
>  can we add somewhere in the dev guide how to add symbols to the API reference

I added a small entry at [the bottom of the development section](https://output.circle-artifacts.com/output/job/fe962b7d-bb9f-4971-bc34-e2f649eccb6f/artifacts/0/doc/CONTRIBUTING.html#editing-the-api-reference-documentation)

CONTRIBUTING.rst
```diff
@@ -367,3 +367,11 @@ specific examples, you can use the following command with a regex pattern:
...
 
 This is especially helpful when you're only modifying or checking a few examples.
+
+
+Editing the API reference documentation
+---------------------------------------
+
+To add a new entry to the :ref:`API reference documentation<api_ref>` or change its
+content, head to ``doc/api_reference.py``. This data is then used by ``doc/conf.py``
+to render templates located at ``doc/reference/*.rst.template``.
```
GaelVaroquaux, line 377:
```suggestion
to render templates located at ``doc/reference/*.rst.template``.

|

Note that **all public functions and classes must be documented in the API
reference**, hence when adding a public function or class, a new entry must be
added, as detailed just above.
```

skrub/selectors/__init__.py
```diff
@@ -1,240 +1,3 @@
-"""
```
GaelVaroquaux: I was about to merge (I did a small change), and then I realized that this large section has disappeared.

I wanted to check that it is ok

Vincent-Maladiere: Yes, I'd say 80% of it has been pushed to the User guide, and the remaining 20% was mostly internal dev stuff. I brought these 20% back to this file, with a link to the user guide for the public-facing part of the selector API.

GaelVaroquaux:
Dostring failures :|

Vincent-Maladiere:
darn! fixed :)

Issue #1410, opened 2025-05-23
Specify desired float precision for TableReport


ancri:
### Problem Description

Loving skrub so far! By default, TableReport shows too many float digits for my use cases, making its output less readable. In polars, there's the neat `float_precision` argument available in various places which allows me to control this. I haven't seen an equivalent in skrub. Is this something which could be added?

### Feature Description

```python
TableReport(df, float_precision=2)
```

and perhaps:
```python
path_display(float_precision=2)
```

georgescutelnicu:
Hey, id like to work on this but i want to check if my approach sounds good first.

The plan is to add a `float_precision` arg (None by default) to the `TableReport` class and create a func that rounds all float values accordingly. Then id modify the html and html_snippet methods to use this func in the return statement (`self._round_floats(self._summary) if self.float_precision is not None else self._summary`).

Does this sound like the right direction?

Vincent-Maladiere:
Hey @georgescutelnicu, thank you for volunteering for this issue.

In my opinion, we should only use the `float_precision` kwarg for displaying the value of the table, not for computing statistics or plotting. So, this value should be passed to the `summarize_dataframe` function via `_summary_kwargs`, and then passed to `_sample_table.make_table`. Precision casting should then be applied inside `{_PandasTable, _PolarsTable, }.add_table_body` on `sub_df`, which is a copy of the original dataframe.

Vincent-Maladiere:
Alternatively, we could make that change in the HTML generation directly, but I'd rather avoid bringing additional logic there if possible

rcap107:
> In my opinion, we should only use the `float_precision` kwarg for displaying the value of the table, not for computing statistics or plotting. So, this value should be passed to the `summarize_dataframe` function via `_summary_kwargs`, and then passed to `_sample_table.make_table`. Precision casting should then be applied inside `{_PandasTable, _PolarsTable, }.add_table_body` on `sub_df`, which is a copy of the original dataframe.

I also prefer this approach. One thing to consider is that both the Pandas and Polars versions must be implemented in each of `_PandasTable` and `_PolarsTable`.

georgescutelnicu:
Sounds good to me. Ill come up with a pr for this, if not tonight then definitely tomorrow. Thanks for the clarification and guidance.

jeromedockes:
IMO the number of shown digits is a display / formatting choice so it should be handled where the formatting happens (by choosing the appropriate format string, in the template) rather than by actually rounding data

[...]

PR #1463, opened 2025-06-20
TableReport float precision

georgescutelnicu:
Related to #1410

### DIFF
--- a/skrub/_reporting/_sample_table.py
+++ b/skrub/_reporting/_sample_table.py
@@ -95,7 +95,7 @@ def make_table(df, max_top_slice_size=5, max_bottom_slice_size=5):
 
 @make_table.specialize("pandas", argument_type="DataFrame")
 def _make_table_pandas(df, max_top_slice_size=5, max_bottom_slice_size=5, float_precision=None):
-     return _PandasTable(df, max_top_slice_size, max_bottom_slice_size, float_precision).table_data
+    return _PandasTable(df, max_top_slice_size, max_bottom_slice_size, float_precision).table_data
 
 
 @make_table.specialize("polars", argument_type="DataFrame")
--- a/skrub/_reporting/_sample_table.py
+++ b/skrub/_reporting/_sample_table.py
@@ -121,7 +121,7 @@ def _pick_slice_sizes(df, max_top_size, max_bottom_size):
 class _PolarsTable:
     """Helper for _make_table_polars."""
 
-    def __init__(self, df, max_top_slice_size, max_bottom_slice_size, float_precision=None):
+    def __init__(self, df, max_top_slice_size, max_bottom_slice_size, float_precision):
         self.df = df
         self.max_top_slice_size = max_top_slice_size
         self.max_bottom_slice_size = max_bottom_slice_size
@@ -342,7 +342,7 @@ class _PandasTable:
     stop_i, stop_j are one past the last i, j coords (df.shape[0], df.shape[1])
     """
 
-    def __init__(self, df, max_top_slice_size, max_bottom_slice_size, float_precision=None):
+    def __init__(self, df, max_top_slice_size, max_bottom_slice_size, float_precision):
         self.df = df
         self.max_top_slice_size = max_top_slice_size
         self.max_bottom_slice_size = max_bottom_slice_size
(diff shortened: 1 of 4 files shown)


skrub/_reporting/_sample_table.py
```diff
@@ -63,6 +63,7 @@
   rowspan, colspan, scope: the values for the html attributes with the same names
 """
 import pandas as pd
+import polars as pl
```
rcap107: polars is not a required dependency, so it may not be installed (this might also be the reason tests fail) 

```suggestion
try:
    import polars as pl
except ImportError:
    pass
```

skrub/_reporting/_sample_table.py
```diff
@@ -170,6 +172,12 @@ def add_df_slices(self):
             )
 
     def add_table_body(self, sub_df, part_name, start_i):
+        if self.float_precision is not None:
+            sub_df = sub_df.with_columns(
+                pl.col(col).round(self.float_precision)
```
rcap107: If you import selectors, this. can be simplified as
```python
sub_df.with_columns(cs.float().round(self.float_precision))
```

skrub/_reporting/_table_report.py
```diff
@@ -127,13 +129,15 @@ def __init__(
...
         max_association_columns=30,
+        float_precision=None,
     ):
         n_rows = max(1, n_rows)
         self._summary_kwargs = {
             "order_by": order_by,
             "max_top_slice_size": -(n_rows // -2),
             "max_bottom_slice_size": n_rows // 2,
             "verbose": verbose,
+            "float_precision": float_precision,
```
rcap107, line 195: there should be a check on the value of `float_precision` somewhere, because pandas and polars have different behaviors for negative values

rcap107 (commented):
Thanks for the PR! I added a few comments, but aside from that I think you're on the right track 👍 

You'll also have to update the CHANGES.rst document to explain the change, and tests are failing because polars is not installed for some configurations

skrub/_reporting/_table_report.py
georgescutelnicu: i added a check for negative `float_precision` in `_make_table_polar`, please take a look and lmk if it matches what u expected. Also im aware of the related warning

rcap107:
> Edit2: I guess we can also get rid of the rounding applied directly to the sub_df and instead round each value `v` (if its a float) when we iterate on every row of the sub_df. Im also aware of the linting issues.

This would avoid the inconsistency between Polars and Pandas and it would keep the formatting where it's supposed to be used (the templating code), so it's a better solution overall. @jeromedockes suggested the same thing in the issue 

There should still be a check for `float_precision` that raises an exception when the number is < 0, and the relative test

georgescutelnicu:
> > Edit2: I guess we can also get rid of the rounding applied directly to the sub_df and instead round each value `v` (if its a float) when we iterate on every row of the sub_df. Im also aware of the linting issues.
> 
> This would avoid the inconsistency between Polars and Pandas and it would keep the formatting where it's supposed to be used (the templating code), so it's a better solution overall. @jeromedockes suggested the same thing in the issue
> 
> There should still be a check for `float_precision` that raises an exception when the number is < 0, and the relative test

Just to clarify, should we move the rounding to every value `v` when iterating through the `sub_df` rows, or do we want to skip any rounding upfront and just pass the `float_precision` to the html template to handle the formatting there? 

For the latter, we'd still need to use rounding to handle negative `float_precision` values in order to replicate pandas behaviour i guess.

rcap107:
Hey @georgescutelnicu, I overlooked a different solution that would simplify this PR a lot, and would make it more convenient to set up. 

A recent PR implemented a global config for skrub, which could be used here to set the precision everywhere without having to deal with the internals of the TableReport. 

You might have to merge upstream/main on your branch, but once you do you'll have the `skrub/_config.py` file to modify. All there is to do is add a "floating_precision" parameter to the config, which means updating the `_global_config` dictionary, `set_config` and `config_context`. 

Then, you can go in `skrub/_reporting/utils.py` and update what's in `format_number` to something like this:
```python
        var = get_config()["float_precision"]
        return f"{number:#.{var}g}"
```

And finally update the template `skrub/_reporting/_data/template/dataframe-sample.html` by setting 
```html
{{ cell["value"]}}
```
to 

```html
{{ cell["value"] | format_number}}
```

By doing this, the user would be able to set the desired number of digits through `set_config` and the `config_context`. It also sidesteps the issue of having different behaviors with pandas and polars.

Unfortunately, it also means that most of what you've done so far will have to be rewritten, I'm really sorry about that 🙈

rcap107:
Closing this in favor of #1470

PR #1467, opened 2025-06-24
correct missing line and point (.)

emilienbattel09:
Addition of a line and missing point.

### DIFF
--- a/skrub/_expressions/_skrub_namespace.py
+++ b/skrub/_expressions/_skrub_namespace.py
@@ -2076,7 +2076,8 @@ class SkrubNamespace:
     def set_name(self, name):
         """Give a name to this expression.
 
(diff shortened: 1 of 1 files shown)


rcap107:
Hey @emilienbattel09, thanks for the second PR! 

This one is stuck because the pre-commit checks are failing. In order to fix that you will have to install `pre-commit` and run it on the files you are uploading. 
You can find some info on this in the [skrub dev guide](https://skrub-data.org/stable/CONTRIBUTING.html#formatting-and-pre-commit-checks). 

Here it's not a big deal, but it will become more important for larger PRs where code formatting helps a lot with reviewing and maintenance.

rcap107 (approved):
Thanks @emilienbattel09 !

Issue #1267, opened 2025-03-31
Remove distribution tiles when distributions aren't computed


Vincent-Maladiere:
Follow up on https://github.com/skrub-data/skrub/pull/1255#issuecomment-2766633436. In the `TableReport`, when the number of columns is greater than some threshold, we don't compute or display marginal distributions.

Removing these tiles entirely could improve the TableReport further since tiles stats are already presented in the 'stats' panel in a more compact format.

<img width="788" alt="Image" src="https://github.com/user-attachments/assets/52dcce51-c88f-43cb-b97f-b10d6a965e85" />

rcap107:
This was addressed (partially) in #1340 for the expressions. From what I remember, the hang up was deciding where to display the message about having too many columns. By keeping the tabs, we can show it in there, without the tabs, we need to figure out where to put it.

Vincent-Maladiere:
What I was suggesting was keeping the "Distributions" tab and removing the individual tiles. The blue message can stay.

Vincent-Maladiere:
Like this:

<img width="829" alt="Image" src="https://github.com/user-attachments/assets/1b2c2ccf-bc7b-4157-b525-99243ffe463d" />

PR #1469, opened 2025-06-25
Remove empty tiles when no column distribution is plotted

Vincent-Maladiere:
Fixes https://github.com/skrub-data/skrub/issues/1267


rcap107:
I just realized there is no example of this behavior anywhere in the documentation and I'm wondering if it should be added

Vincent-Maladiere:
It's a good point, maybe we should have a small example fully dedicated to the TableReport?

rcap107:
Yes, it would be useful, docstrings can't really show these functions properly

It could also be a good sprint PR

PR #1470, opened 2025-06-26
add global float_precision config

georgescutelnicu:
Related to #1410 

This is a follow up to #1463 pr. The earlier approach handled `float_precision` directly in `TableReport`. Im opening this new pr to keep the history clean and focused on the new approach.

In the test i used the target `y = fetch_employee_salaries().y` instead of `X` because we needed a float column to check `float_precision` against.


skrub/_config.py
```diff
@@ -184,6 +193,13 @@ def set_config(
             )
         local_config["enable_subsampling"] = enable_subsampling
 
+    if float_precision is not None:
+        if not isinstance(float_precision, numbers.Integral) or float_precision <= 0:
```
Vincent-Maladiere, line 197: A bit of a nitpick: should we allow `float_precision = 0`? This would display integer values like `1.24 -> 1`

skrub/_reporting/_utils.py
```diff
@@ -87,7 +87,10 @@ def format_number(number):
     if isinstance(number, numbers.Integral):
         return f"{number:,}"
     if isinstance(number, numbers.Real):
-        return f"{number:#.3g}"
+        from skrub._config import get_config
```
Vincent-Maladiere, line 91: If there is no specific reason to place this line here, it should be placed at the top of the file

skrub/_reporting/_utils.py
```diff
@@ -87,7 +87,10 @@ def format_number(number):
...
+
+        var = get_config()["float_precision"]
+        return f"{number:#.{var}g}"
```
Vincent-Maladiere, line 94: Pandas display doesn't force decimal representation, so I think we should be consistent with that. WDYT @rcap107?
```suggestion
        return f"{number:.{var}g}"
```

Reproducer:
```python
import pandas as pd
x = 123
print(f"{x:.3g}")
pd.set_option("display.precision", 3)
pd.DataFrame(dict(a=[x]))
```

Vincent-Maladiere (commented):
Hey @georgescutelnicu, the PR is already in good shape! Some feedback

skrub/_reporting/_utils.py
rcap107: importing at the top of the file causes a circular import because of the `dispatch`

Vincent-Maladiere: ok, then add a comment please

skrub/_reporting/_utils.py
rcap107: works for me

skrub/_config.py
georgescutelnicu: isn't setting it to 0 giving the same output as setting it to 1 since the formatting uses significant digits rather than fixed decimal places? 

From python docs (string — Common string operations): `'g'` General format. For a given precision `p >= 1`, this rounds the number to `p` significant digits and then formats the result in either fixed-point format or in scientific notation, depending on its magnitude. A precision of `0` is treated as equivalent to a precision of `1`.

skrub/_reporting/_utils.py
jeromedockes: In your example x and the series are integer @Vincent-Maladiere but if you make it a float it does print the decimal point

```
>>> import pandas as pd
... x = 123
... print(f"{x:.3g}")
... pd.set_option("display.precision", 3)
... pd.Series(x, dtype=float)
123
0    123.0
dtype: float64
```

I would keep it as it helps distinguish floats from ints

Vincent-Maladiere: You could argue that casting integers to floats doesn't help distinguish floats from ints either. So both options are valid, and I don't have a strong pref.

jeromedockes: but this formatting is only applied to floats, not ints so ints would not be cast to floats and their representation would not be modified in any case

Vincent-Maladiere: ah yes, indeed

georgescutelnicu: allright, i rushed that change a bit. i will revert to the original formatting

skrub/_config.py
Vincent-Maladiere: Yes, my understanding was slightly off, but the question remains. If 0 is equivalent to 1, should we keep it?

georgescutelnicu: i thought it might confuse users into thinking `0` behaved differently, but since its treated the same as `1`, we could keep it allowing for flexibility i guess, since it doesnt seem to cause any harm

Vincent-Maladiere: Ok, I don't have a strong preference

Issue #1429, opened 2025-06-05
Expose `OnEachColumn` and `OnSubFrame`


Vincent-Maladiere:
This [Stack Overflow issue](https://stackoverflow.com/questions/37685412/avoid-scaling-binary-columns-in-sci-kit-learn-standsardscaler) sparked a discussion among the skrub developers about the easiest solution using skrub.

In addition to the skrub expressions, one solution that came to mind is to use `skrub._on_each_column.OnEachColumn`:

```python
import numpy as np
import pandas as pd
from sklearn.preprocessing import StandardScaler
import skrub  # your friend here
from skrub import selectors as s
from skrub._on_each_column import OnEachColumn

X = np.hstack((
    np.random.random((1000, 2)),
    np.random.randint(2, size=(1000, 2)))
)
X = pd.DataFrame(X, columns=list("abcd"))
y = np.random.random(1000)

# Create a column selector with arbitrary logic.
not_binary = s.filter(lambda col: col.nunique() > 2)
preprocessor = OnEachColumn(StandardScaler(), cols=not_binary)
print(preprocessor.fit_transform(X))

# Bring it into a sklearn pipeline
from sklearn.pipeline import make_pipeline
from sklearn.linear_model import Ridge

pipe = make_pipeline(preprocessor, Ridge()).fit(X, y)
```

1. Should we publicly expose `OnEachColumn` and `OnSubFrame`? Their main use case would be to apply a transformer using selectors
2. Should we rename these classes by highlighting how they relate to `skb.apply`? `Apply`, `ApplyCols`?

GaelVaroquaux:
As you know, I'm in favor of exposing at, with name "ApplyOnCols"

rcap107:
skrub meeting suggestion: `ApplyToCols` or `ApplyToColumns`

Vincent-Maladiere:
We can now answer the StackOverflow post :)

PR #1478, opened 2025-06-27
Expose `ApplyToCols` and `ApplyToFrame`

Vincent-Maladiere:
Fixes https://github.com/skrub-data/skrub/issues/1429

### DIFF
--- a/examples/10_apply_on_cols.py
+++ b/examples/10_apply_on_cols.py
@@ -21,10 +21,12 @@ X, y = data.X, data.y
 X
 
 # %%
-# Our goal is now to apply a :class:`~skrub.StringEncoder` to two columns of our choosing: ``division`` and ``employee_position_title``.
+# Our goal is now to apply a :class:`~skrub.StringEncoder` to two columns of our
+# choosing: ``division`` and ``employee_position_title``.
 #
 # We can achieve this using :class:`~skrub.ApplyToCols`, whose job is to apply a
-# transformer to multiple columns independently, and let unmatched columns through without changes.
+# transformer to multiple columns independently, and let unmatched columns through
+# without changes.
 # This can be seen as a handy drop-in replacement of the
 # :class:`~sklearn.compose.ColumnTransformer`.
 #
@@ -42,19 +44,14 @@ Xt = apply_string_encoder.fit_transform(X)
 Xt
 
 # %%
-# Let's now imagine that our two previous columns ``"division"``,
-# ``"employee_position_title"`` are highly correlated, and we want to reduce 
-# the number of components of their vector representation jointly. 
+# In addition to the :class:`~skrub.ApplyToCols` class, the
+# :class:`~skrub.ApplyToFrame` class is useful for transformers that work on multiple
+# columns at once, such as the :class:`~sklearn.decomposition.PCA` which reduces the
+# number of components.
 #
-# This time, we can use :class:`~skrub.ApplyToFrame`, which applies a transformer to
-# a subset of columns jointly as a single dataframe, rather than separately like :class:`~skrub.ApplyToCols`.
-#
-# We apply a :class:`~sklearn.decomposition.PCA` via
-# :class:`~skrub.ApplyToFrame` jointly to the ``60`` previous vector columns, which can be identified easily because they are all prefixed with ``"lsa"``.
-#
-# To select them without hardcoding their names, we introduce skrub selectors,
-# which allow for flexible matching pattern and composable logic. See
-# :ref:`selectors<selectors>` for further details.
+# To select columns without hardcoding their names, we introduce
+# :ref:`selectors<selectors>`, which allow for flexible matching pattern and composable
+# logic.
 #
 # The regex selector below will match all columns prefixed with ``"lsa"``, and pass them
 # to :class:`~skrub.ApplyToFrame` which will assemble these columns into a dataframe and
@@ -97,7 +94,9 @@ pipeline.fit_transform(Xt)
 # Suppose we want to apply an :class:`~sklearn.preprocessing.OrdinalEncoder` on
 # categorical columns with low cardinality (e.g., fewer than ``40`` unique values).
 #
-# We define a column filter using skrub selectors with a lambda function.
+# We define a column filter using skrub selectors with a lambda function. Note that
+# the same effect can be obtained directly by using
+# :func:`~srkub.selectors.cardinality_below`.
 from sklearn.preprocessing import OrdinalEncoder
 
 low_cardinality = s.filter(lambda col: col.nunique() < 40)
--- a/examples/10_apply_on_cols.py
+++ b/examples/10_apply_on_cols.py
@@ -1,0 +1,140 @@
+"""
+Hands-On with Column Selection and Transformers
+===============================================
+
+In previous examples, we saw how skrub provides powerful abstractions like
+:class:`~skrub.TableVectorizer` and :func:`~skrub.tabular_learner` to create pipelines.
+
+In this new example, we show how to gain more flexibility with pipelines by selecting
+dataframe columns using arbitrary logic, and applying transformers to them.
+"""
+
+# %%
+# We begin with loading a dataset with heterogeneous datatypes, and replacing Pandas's
+# display with the TableReport display via :func:`skrub.set_config`.
+import skrub
+from skrub.datasets import fetch_employee_salaries
+
+skrub.set_config(use_tablereport=True)
+data = fetch_employee_salaries()
+X, y = data.X, data.y
+X
+
+# %%
+# Our goal is now to apply a :class:`~skrub.StringEncoder` to two specific columns.
+#
+# We can achieve this using :class:`~skrub.ApplyToCols`, whose job is to apply a
+# transformer to multiple columns independently, and to simply pass through unmatched
+# columns. This can be seen as a handy drop-in replacement of the
+# :class:`~sklearn.compose.ColumnTransformer`.
+#
+# Since we selected two columns and set the number of components to ``30`` each,
+# :class:`~skrub.ApplyToCols` will create ``2*30`` embedding columns in the dataframe
+# ``Xt``, which we prefix with ``lsa_``.
+from skrub import ApplyToCols, StringEncoder
+
+apply_string_encoder = ApplyToCols(
+    StringEncoder(n_components=30),
+    cols=["division", "employee_position_title"],
+    rename_columns="lsa_{}",
+)
+Xt = apply_string_encoder.fit_transform(X)
+Xt
+
+# %%
+# Let's now imagine that our two previous columns ``"division"``,
+# ``"employee_position_title"`` are highly correlated, and we want to reduce their
+# vector representations jointly.
+#
+# This time, we can use :class:`~skrub.ApplyToFrame`, which applies a transformer to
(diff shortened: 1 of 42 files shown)


rcap107:
This will be a fun one to review

Vincent-Maladiere:
ahahah indeed. I apologised in advance

jeromedockes:
> ahahah indeed. I apologised in advance

:sweat_smile: thanks for plowing through

skrub/_apply_to_each_column.py
```diff
@@ -11,7 +11,7 @@
 from . import _utils, selectors
 from ._join_utils import pick_column_names
 
-__all__ = ["OnEachColumn", "SingleColumnTransformer", "RejectColumn"]
+__all__ = ["ApplyToCols", "SingleColumnTransformer", "RejectColumn"]
```
jeromedockes: total nitpick: maybe I would name that file `_apply_to_cols.py`?

GaelVaroquaux:
Is this ready to review / merge?

rcap107:
I'll review this tomorrow, I haven't had the time yet

GaelVaroquaux:
> I'll review this tomorrow, I haven't had the time yet

Thanks Riccardo!!!

skrub/_apply_to_frame.py
```diff
@@ -5,14 +5,14 @@
...
+__all__ = ["ApplyToFrame"]
 
 
-class OnSubFrame(TransformerMixin, BaseEstimator):
+class ApplyToFrame(TransformerMixin, BaseEstimator):
     """Apply a transformer to part of a dataframe.
 
     A subset of the dataframe is selected and passed to the transformer (as a
-    single input). This is different from ``OnEachColumn`` which fits a
+    single input). This is different from ``ApplyToCols`` which fits a
```
rcap107:
```suggestion
    single input). This is different from ``ApplyToCols``, which fits a
```

skrub/_apply_to_frame.py
```diff
@@ -78,8 +78,8 @@ class OnSubFrame(TransformerMixin, BaseEstimator):
     2  0.0   0.0  100.0     0.0
     3  0.0   0.0    0.0  1000.0
     >>> from sklearn.decomposition import PCA
-    >>> from skrub._on_subframe import OnSubFrame
-    >>> OnSubFrame(PCA(n_components=2)).fit_transform(df).round(2)
+    >>> from skrub._apply_to_frame import ApplyToFrame
```
rcap107:
```suggestion
    >>> from skrub import ApplyToFrame
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
+"""
+Hands-On with Column Selection and Transformers
+===============================================
+
+In previous examples, we saw how skrub provides powerful abstractions like
+:class:`~skrub.TableVectorizer` and :func:`~skrub.tabular_learner` to create pipelines.
+
+In this new example, we show how to gain more flexibility with pipelines by selecting
```
rcap107:
```suggestion
In this new example, we show how to create more flexible pipelines by selecting
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+dataframe columns using arbitrary logic, and applying transformers to them.
```
rcap107:
```suggestion
and transforming dataframe columns using arbitrary logic.
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+import skrub
+from skrub.datasets import fetch_employee_salaries
+
+skrub.set_config(use_tablereport=True)
+data = fetch_employee_salaries()
+X, y = data.X, data.y
+X
+
+# %%
+# Our goal is now to apply a :class:`~skrub.StringEncoder` to two specific columns.
```
rcap107:
```suggestion
# Our goal is now to apply a :class:`~skrub.StringEncoder` to two columns of our choosing: ``division`` and ``employee_position_title``.
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+#
+# We can achieve this using :class:`~skrub.ApplyToCols`, whose job is to apply a
+# transformer to multiple columns independently, and to simply pass through unmatched
```
rcap107:
```suggestion
# transformer to multiple columns independently, and let unmatched columns through without changes.
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+# columns. This can be seen as a handy drop-in replacement of the
```
rcap107:
```suggestion
# This can be seen as a handy drop-in replacement of the
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+apply_string_encoder = ApplyToCols(
+    StringEncoder(n_components=30),
+    cols=["division", "employee_position_title"],
+    rename_columns="lsa_{}",
+)
+Xt = apply_string_encoder.fit_transform(X)
+Xt
+
+# %%
+# Let's now imagine that our two previous columns ``"division"``,
```
rcap107: I'm not convinced by this phrasing: either they are highly correlated (which they do not seem to be, from the associations) and so there is a reason for this operation, or they aren't. If they aren't, I think adding an imaginary reason would just add to confusion: it's an example, I think that showing what is happening is more useful

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+# ``"employee_position_title"`` are highly correlated, and we want to reduce their
+# vector representations jointly.
+#
+# This time, we can use :class:`~skrub.ApplyToFrame`, which applies a transformer to
+# columns jointly, rather than separately like :class:`~skrub.ApplyToCols`.
```
rcap107:
```suggestion
# a subset of columns jointly as a single dataframe, rather than separately like :class:`~skrub.ApplyToCols`.
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+    cols=["division", "employee_position_title"],
+    rename_columns="lsa_{}",
+)
+Xt = apply_string_encoder.fit_transform(X)
+Xt
+
+# %%
+# Let's now imagine that our two previous columns ``"division"``,
+# ``"employee_position_title"`` are highly correlated, and we want to reduce their
+# vector representations jointly.
```
rcap107:
```suggestion
# the number of components of their vector representation jointly. 
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+    StringEncoder(n_components=30),
+    cols=["division", "employee_position_title"],
+    rename_columns="lsa_{}",
+)
+Xt = apply_string_encoder.fit_transform(X)
+Xt
+
+# %%
+# Let's now imagine that our two previous columns ``"division"``,
+# ``"employee_position_title"`` are highly correlated, and we want to reduce their
```
rcap107:
```suggestion
# ``"employee_position_title"`` are highly correlated, and we want to reduce 
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+# vector representations jointly.
+#
+# This time, we can use :class:`~skrub.ApplyToFrame`, which applies a transformer to
+# columns jointly, rather than separately like :class:`~skrub.ApplyToCols`.
+#
```
rcap107: I think it would be useful to be more explicit, something like


This is useful for transformers that work on multiple columns at once, such as the PCA that we want to use to reduce the number of components.

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+# We apply a :class:`~sklearn.decomposition.PCA` via
+# :class:`~skrub.ApplyToFrame` to the ``60`` previous vector columns jointly. These
```
rcap107:
```suggestion
# :class:`~skrub.ApplyToFrame` jointly to the ``60`` previous vector columns, which can be identified easily because they are all prefixed with ``"lsa"``.
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+# columns are easy to identify because they are all prefixed with ``"lsa"``.
```
rcap107:
```suggestion
```

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,140 @@
...
+
+# %%
+# Let's run through one more example to showcase the expressiveness of the selectors.
+# Suppose we want to apply an :class:`~sklearn.preprocessing.OrdinalEncoder` on
+# categorical columns with low cardinality (e.g., fewer than ``40`` unique values).
+#
+# We define a column filter using skrub selectors with a lambda function.
+from sklearn.preprocessing import OrdinalEncoder
+
+low_cardinality = s.filter(lambda col: col.nunique() < 40)
```
rcap107, line 102: Given that selectors include `.cardinality_below`, I think there should either be a mention to the function ("this can also be done with .cardinality_below"), or a different example should be used

examples/10_apply_on_cols.py
Vincent-Maladiere: Ok, I'll simplify the narrative

examples/10_apply_on_cols.py
Vincent-Maladiere: True, I can mention it

GaelVaroquaux: Should we address this?

(I also want terribly to merge, because this is very useful. If I merge too fast, maybe address in a new PR)

examples/10_apply_on_cols.py
```diff
@@ -0,0 +1,138 @@
...
+# using a logical operator. This resulting selector matches string
+# columns with cardinality below ``40``.
+#
+# We can also define the opposite selector ``high_cardinality`` using the negation
+# operator ``~`` and apply a :class:`skrub.StringEncoder` to vectorize those
+# columns.
+from sklearn.ensemble import HistGradientBoostingRegressor
+
+high_cardinality = ~low_cardinality
+make_pipeline(
```
GaelVaroquaux: I find it a bit ugly that we are creating a pipeline, fitting it, but not storing it to a model.

I can see people copy-pasting this, and not understanding that they need to actually assign the return of this to a variable.

```suggestion
model = make_pipeline(
```

Vincent-Maladiere: Makes sense, updated!

GaelVaroquaux (approved):
Merging!

Thanks a lot. This is a big deall

examples/10_apply_on_cols.py
GaelVaroquaux: Darn, I think that we forgot to do this change. I'm doing it in main...

examples/10_apply_on_cols.py
Vincent-Maladiere: yes it has been adressed in the text section above, L97:
```
# We define a column filter using skrub selectors with a lambda function. Note that
# the same effect can be obtained directly by using
# :func:`~srkub.selectors.cardinality_below`.
```

Issue #1486, opened 2025-07-05
cross-validation on an expression pipeline chaining more than 2 polars join raises `DuplicateError`


ogrisel:
### Describe the bug

The following code builds a set of features by iteratively joining a dataframes wrapped in variables.

The 3 joins are defined within a for loop.

Everything runs fine when running evaluating this pipeline on the full training set.

However, when running this same pipeline in via skrub's cross-validation, it triggers the following crash.

**Note**: the crash disappears if we change the pipeline to only chain 2 joins instead of 3.

### Steps/Code to Reproduce

```python
# %%
import skrub
from datetime import datetime
import polars as pl

common_time_range = pl.datetime_range(
    start=datetime(2023, 1, 12, 0, 0),
    end=datetime(2023, 1, 24, 0, 0),
    time_zone="UTC",
    interval="1d",
).alias("time")

time = skrub.var(
    "time",
    pl.DataFrame().with_columns([common_time_range]),
)

data_variables = {}
var_specs = [("a", 0), ("b", -2), ("c", 1)]
# var_specs = [("a", 0), ("b", -2)]  # with only two variable, the problems does not occur


for name, offset in var_specs:
    data_variables[name] = skrub.var(
        name,
        pl.DataFrame().with_columns(
            [
                pl.datetime_range(
                    start=datetime(2023, 1, 3 + offset, 0, 0),
                    end=datetime(2023, 1, 29 + offset, 0, 0),
                    time_zone="UTC",
                    interval="1d",
                ).alias("time"),
                pl.arange(27).alias("value"),
            ]
        ),
    )

# %%
features = time
for name, data in data_variables.items():
    features = features.join(
        data.rename(lambda x: x if x == "time" else x + "_" + name),
        on="time",
    )
features = features.drop("time").skb.mark_as_X()
# %%
target_df = skrub.var(
    "target",
    pl.DataFrame().with_columns(
        [
            common_time_range,
            pl.arange(12, 25).alias("target"),
        ]
    ),
)
target = target_df["target"].skb.mark_as_y()

# %%
from sklearn.linear_model import Ridge


predictions = features.skb.apply(Ridge(), y=target)
predictions.skb.eval()  # no problem when evaluating the pipeline outside of a cross-validation


# %%
predictions.skb.cross_validate(cv=5, scoring=["r2", "neg_mean_absolute_error"])

```

### Expected Results

[...]

PR #1488, opened 2025-07-06
Dont use preview in eval

jeromedockes:
After #1486  I think it is better to always re-run the full evaluation in `.skb.eval()` and not reuse the previews, which can be accessed through `.skb.preview()` anyway. the assumption that the result will be the same when there is no subsampling and the data is the same is incorrect


Vincent-Maladiere:
I can continue this one, as we'd like to have it for releasing 0.6.0

jeromedockes:
>  I can continue this one, as we'd like to have it for releasing 0.6.0

thanks, I think in terms of code there is nothing more to it but I would maybe go over the examples to see if preview should be put forward more or used in some places where eval was used, now that eval() always retriggers a full computation

Vincent-Maladiere:
I think we should rename `.preview()` to `.eval_with_cache()` or something more self-explanatory/explicit. Currently, the behavior of `.preview()` compared to `.eval()` is not obvious. WDYT?

jeromedockes:
I have the impression that preview() communicates a bit better the intended functionality, which is to get a quick preview on the result on preview data. I think the caching part is more of an implementation detail

rcap107:
I also think that `preview` conveys the use case better. In the docs we present `eval` as the "re-run everything possibly on new data", and I don't think that `with_cache` gives the idea of giving a "preview" :thinking:

GaelVaroquaux:
+1  for preview 
 
 
 
 
 
 
 
 
 rcap107  left a comment  (skrub-data/skrub#1488) 
 
 I also think that  preview  conveys the use case better. In the docs we present  eval  as the "re-run everything possibly on new data", and I don't think that  with_cache  gives the idea of giving a "preview" 🤔

Vincent-Maladiere:
I'm not very convinced, but ok.

PR #1495, opened 2025-07-09
Add dynamic placement of arrow in string encoders benchmark plot

ArturoAmorQ:
Similar to the work done in https://github.com/scikit-learn/scikit-learn/pull/31062, this PR changes the hard-coded position of the arrow that indicates best models in the [Various string encoders example](https://skrub-data.org/stable/auto_examples/02_text_with_string_encoders.html) to a dynamic placing in proportion of `xlim` and `ylim`.

It also takes the opportunity to set the `random_state` of the `StringEncoder` for reproducibility.

### DIFF
--- a/examples/02_text_with_string_encoders.py
+++ b/examples/02_text_with_string_encoders.py
@@ -237,7 +237,7 @@ plot_box_results(results)
 # case, 30.
 from skrub import StringEncoder
 
-string_encoder = StringEncoder(ngram_range=(3, 4), analyzer="char_wb")
+string_encoder = StringEncoder(ngram_range=(3, 4), analyzer="char_wb", random_state=0)
 
 string_encoder_pipe = make_pipeline(
     TableVectorizer(high_cardinality=string_encoder),
--- a/examples/02_text_with_string_encoders.py
+++ b/examples/02_text_with_string_encoders.py
@@ -301,13 +301,23 @@ def plot_performance_tradeoff(results):
         ax.set_ylabel("ROC AUC")
         ax.set_title("Prediction performance / training time trade-off")
 
+    log_xlim = np.log10(ax.get_xlim())
+    x_log_range = log_xlim[1] - log_xlim[0]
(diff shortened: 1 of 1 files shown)


examples/02_text_with_string_encoders.py
```diff
@@ -301,13 +301,23 @@ def plot_performance_tradeoff(results):
...
 
+    log_xlim = np.log10(ax.get_xlim())
+    x_log_range = log_xlim[1] - log_xlim[0]
+    y_range = ax.get_ylim()[1] - ax.get_ylim()[0]
+    arrow_tip_x_log = log_xlim[0] + 0.05 * x_log_range
+    arrow_tip_y = ax.get_ylim()[1] - 0.05 * y_range
+    arrow_start_x_log = arrow_tip_x_log + 0.04 * x_log_range
+    arrow_start_y = arrow_tip_y - 0.2 * y_range
+    arrow_tip_x = np.power(10, arrow_tip_x_log)
+    arrow_start_x = np.power(10, arrow_start_x_log)
```
GaelVaroquaux: Ouch.

This adds a lot of chunk of code to an example for cosmetic reasons.

I don't like that. More code like this is frightening for users. Examples should strive to have less code, not more code

Vincent-Maladiere: Maybe we could turn this cell into a publicly exposed plot function? These pareto-front plots are quite popular IMO

ArturoAmorQ: > Maybe we could turn this cell into a publicly exposed plot function?

Not sure if that falls in the skrub scope, but yes. It sounds like something needed in the market. For the moment I managed to simplify the code and now it effectively adds one line only, which I think is worth the results.

**This PR:**
sphx_glr_02_text_with_string_encoders_006_this_PR(1)

**Main:**
sphx_glr_02_text_with_string_encoders_006_main(1)

rcap107:
thanks a lot @ArturoAmorQ !

Issue #1454, opened 2025-06-19
Showing values for bar plots in TableReport


alamehm:
### Problem Description

Hello,
It would be great if we can see the value of a horizontal bar , or a vertical bar in the TableReport plot.
Currently the situation is the following: 
- For vertical bars, we cannot know the exact value of a bar (X and Y), especially when we have a single tick label in the axes, with large values .
- For horizontal bars, we can see the percentage of each bar, but not the value as well. 

**Suggestion:** 
- Showing more info about each bar `on_hover` , similar to what is used in other libraries like `plotly`.
- This way the layout of TableReport will remain the same, but the info will be shown `on_hover` only

Below is an example of the current TableReport plots.

![Image](https://github.com/user-attachments/assets/7247b065-746e-41d8-8f37-0de8f118150d)

### Feature Description

- Adding an `on_hover` event to the plots
- This may open the doors to simplify the implementation of "tailored" plots in TableReport

### Alternative Solutions

Using off-the-shelf solutions may solve it ex.`plotly`, but this will add a new dependency for `skrub` , and it is not straight forward to execute javascript inside templates (currently in skrub, the plots exist inside template tags)

rcap107:
Hey @alamehm, thanks a lot for the suggestion! 

I agree that the current version of the plots in the TableReport could be improved (having only one tick in some cases should be fixed, for example). 

I think that having the exact value of a bar could be quite useful, and straightforward to implement. It could also be a kwarg or a configuration parameter for convenience. 

I am not keen on adding a dependency on plotly for that, and in general I'm afraid that replacing the current plots with a plotly version would likely introduce a lot of issues. 

Tailored plots are a cool idea that I remember from the workshop, and while there is some refactoring to be done, it shouldn't be too difficult. 

So, I like the suggestion but I am not convinced that implementing it with plotly would be the best way of doing it  🤔

alamehm:
@rcap107, I agree with you about the issues of adding a new dependency.

lionelkusch:
I try to open an Pull Request for plotting the value on the histogram.

PR #1496, opened 2025-07-09
Add a text for displaying values on histogram.

lionelkusch:
For the histogram, I add text for displaying the percentage of each bar, the same as for values_count. I try to reuse the idea of display value for displaying on the histogram.

This pull request was made for fixing the issue #1454.

### DIFF
--- a/skrub/_reporting/_plotting.py
+++ b/skrub/_reporting/_plotting.py
@@ -199,6 +199,17 @@ def _robust_hist(values, ax, color):
     n_high_outliers = (high < values).sum()
     n, bins, patches = ax.hist(inliers)
     n_out = n_low_outliers + n_high_outliers
+
+    # Display percentage on the bar of the histrogram
+    threshold_display = np.max(n) / 2
+    for x_, value in zip(bins, n):
+        percentage = value / np.sum(n)
+        percentage_string = _utils.format_percent(percentage)
+        if value > threshold_display:
+            ax.text(x_, value - threshold_display, percentage_string, rotation='vertical')
+        else:
+            ax.text(x_, value + threshold_display * 0.1, percentage_string, rotation='vertical')
+
     if not n_out:
         return 0, 0
     width = bins[1] - bins[0]


Vincent-Maladiere:
Hey @lionelkusch thank you for this PR! https://github.com/skrub-data/skrub/issues/1454 is kinda tricky and we're not sure how to move forward.

Your suggestion is interesting, but my understanding is that we'd rather have the height of each vertical bar rather than the percentile of their distribution. Additionally, we'd also like to have the total count for the categorical distribution (displayed horizontally). We need to see whether adding these figures would take too much space to be useful or not.

We can use this PR to explore different options.

lionelkusch:
This PR is quite limited to add information to the histogram is the same spirit as for the value_counted.
I pushed further in PR #1506 . For adding the height of the bar. 
Tell me what final form you want.

lionelkusch:
> > Addressing your remarks:

> > >   The problem when you combine the two in the same graphic is that the text will be longer than the columns. This won't look nice and it will be difficult to automatically add the text to the graphic.

> >  If the text crossing the bar edge doesn't look good, we can choose to always have the text outside of the bar, increasing the figure size slightly if necessary.

> > >  Adding the number of elements creates some variability and creates a difficult problem for the position of the text relative to the bar.

> > I think this can be addressed by controlling the number of digits in the format using custom rules, like full representation when the entries are smaller than 10,000 and scientific notation otherwise. We would have an upper bound on the size of the text.

> > -----------------------------------
> > What do the others think?

I agree on all points here and I would also prefer having both numbers visible at the same time. 

> Adding the number of elements creates some variability and creates a difficult problem for the position of the text relative to the bar.

How about adding the % to the right (the top) of the plot, aligned with each bar? With a fixed format (like 3 significant digits) the space should remain mostly constant, while the actual value is still represented in the same way as it is now. If the % sign takes too much space, it could be added above the column of numbers as an annotation.

_Originally posted by @rcap107 in https://github.com/skrub-data/skrub/issues/1506#issuecomment-3089379783_

I put back the comments of the PR #1506 about the graphic representation.

rcap107:
I quite like the current version, thanks a lot @lionelkusch! 

I just realized is that we might want to remove the % numbers for numerical features, because bins for numerical values are defined somewhat arbitrarily, so the % numbers for them are not really meaningful.

rcap107:
Hey @lionelkusch, do you think you finish this PR? As we discussed, the current version is good, the only change would be removing the % from the numerical columns 

I hope there won't be too many conflicts to deal with

lionelkusch:
@rcap107  I am sorry, I forgot to do the modification.
I did it. I think that the PR is read.

rcap107:
thanks a lot! I'm wondering if the plots would look better without the right and top spines, like the current version :thinking: 

Screenshot_20250924_131132

Screenshot_20250924_131208

lionelkusch:
I propose you this solution:
image

GaelVaroquaux:
Is there a place where I can look at a full HTML version, to interact a bit with it?

rcap107:
> Is there a place where I can look at a full HTML version, to interact a bit with it?

I was comparing the table report in the homepage of the generated docs with that in the live version

rcap107:
> I propose you this solution: image

I think it's ok, I was worried that the tick might be read by someone as a "minus" or "negative", but it doesn't really make sense for percentages

lionelkusch:
I removed the tick line; I think it's not needed it for the readability of the figure.

GaelVaroquaux:
It's quite OK, actually:

image

My concern is: information overload. The more we put, the harder it is for a user to focus and to get the message. It brings us to the gold rule of communication: less in more.

But you've managed to keep it really light.

An ideal option, would have been to have a "hover": information that appears when the mouse goes over the relevant graph. But it's hard to do with our current toolkit.

I'm +1 for merge on this. Congratulations!!

rcap107:
Hey @lionelkusch, I keep on losing track of this issue :see_no_evil:  my bad

We can merge after removing the barplot labels for the numerical features, like here
image

The bins are decided arbitrarily, so the size of each bin is not particularly informative.

(I thought I had already mentioned this here, but I guess I forgot)

lionelkusch:
Are you sure that you want to remove the label?
I ask the question because it seems to me that this was the starting point of the PR, as described in the issue #1454.

rcap107:
> Are you sure that you want to remove the label? I ask the question because it seems to me that this was the starting point of the PR, as described in the issue #1454.

#1454 mentioned both categorical and numerical features, but given that the distribution of numerical features is generated with "best effort" bins, the actual value of the bins isn't very meaningful, so I think it can be removed.

For categorical features, each category is well delimited so it makes sense to have both the number of occurrences and the %, so the current version is good and should remain as you wrote it.

lionelkusch:
done

rcap107 (approved):
Looks good to me, thanks a lot @lionelkusch. It was a bit convoluted (my fault for that), but the result is a nice addition to the TableReport.

MA-inria:
Thank you all for merging this PR

PR #1498, opened 2025-07-09
feat: Add cardinality information in table report

### LINKED ISSUE #1414: Highlight high cardinality columns in the TableReport
It may be useful to highlight high cardinality columns in some way in the table report (different color? bold font? not sure). By default high cardinality is 40 in the TableVectorizer, so that may be used as the threshold.

It could also be a parameter to have in the skrub configuration #1377, and it could be set by the user as a TableReport parameter if they have specific requirements and/or they know what their value for "high cardinality" should be.

rouk1:
This PR introduces a information on column cardinality in the table report.
As a bonus `cardinality_threshold` as been added as a configuration integer.

UI preview:
Screenshot 2025-07-09 at 15 20 46

Closes #1414.

### DIFF
--- a/skrub/_config.py
+++ b/skrub/_config.py
@@ -15,6 +15,7 @@ _global_config = {
     "subsampling_seed": int(os.environ.get("SKB_SUBSAMPLING_SEED", 0)),
     "enable_subsampling": os.environ.get("SKB_ENABLE_SUBSAMPLING", "default"),
     "float_precision": int(os.environ.get("SKB_FLOAT_PRECISION", 3)),
+    "cardinality_threshold": int(os.environ.get("SKB_CARDINALITY_THRESHOLD", 40)),
 }
 _threadlocal = threading.local()
 
@@ -72,6 +73,7 @@ def set_config(
     subsampling_seed=None,
     enable_subsampling=None,
     float_precision=None,
+    cardinality_threshold=None,
 ):
     """Set global skrub configuration.
 
@@ -140,6 +142,14 @@ def set_config(
         This configuration can also be set with the ``SKB_FLOAT_PRECISION``
         environment variable.
 
+    cardinality_threshold: int,  default=40
+        Control the threshold value used to warn user if they have
+        high cardinality columns in there dataset. It's also used as the default
+        value of ``TableVectorizer`` `cardinality_threshold` parameter.
+
+        This configuration can also be set with the ``SKB_CARDINALITY_THRESHOLD``
+        environment variable.
+
     See Also
     --------
     get_config : Retrieve current values for global configuration.
@@ -200,6 +210,16 @@ def set_config(
             )
         local_config["float_precision"] = float_precision
 
+    if cardinality_threshold is not None:
+        if (
+            not isinstance(cardinality_threshold, numbers.Integral)
+            or cardinality_threshold < 0
+        ):
+            raise ValueError(
+                "'cardinality_threshold' must be a positive"
+                "integer, got {cardinality_threshold!r}"
+            )
+
     _apply_external_patches(local_config)
 
 
@@ -213,6 +233,7 @@ def config_context(
     subsampling_seed=None,
     enable_subsampling=None,
     float_precision=None,
+    cardinality_threshold=None,
 ):
     """Context manager for global skrub configuration.
 
@@ -281,6 +302,14 @@ def config_context(
         This configuration can also be set with the ``SKB_FLOAT_PRECISION``
         environment variable.
 
+    cardinality_threshold: int,  default=40
+        Control the threshold value used to warn user if they have
+        high cardinality columns in there dataset. It's also used as the default
+        value of ``TableVectorizer`` `cardinality_threshold` parameter.
+
+        This configuration can also be set with the ``SKB_CARDINALITY_THRESHOLD``
+        environment variable.
+
     Yields
     ------
     None.
@@ -305,6 +334,7 @@ def config_context(
         subsampling_seed=subsampling_seed,
         enable_subsampling=enable_subsampling,
         float_precision=float_precision,
+        cardinality_threshold=cardinality_threshold,
     )
 
     try:
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -8,8 +8,8 @@ from sklearn.pipeline import make_pipeline
 from sklearn.preprocessing import OneHotEncoder
 from sklearn.utils.validation import check_is_fitted
 
+from . import _config, _utils
 from . import _dataframe as sbd
-from . import _utils
 from . import selectors as s
 from ._check_input import CheckInputDataFrame
 from ._clean_categories import CleanCategories
@@ -730,7 +730,7 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
     def __init__(
         self,
         *,
-        cardinality_threshold=40,
+        cardinality_threshold=_config.get_config()["cardinality_threshold"],
         low_cardinality=LOW_CARDINALITY_TRANSFORMER,
         high_cardinality=HIGH_CARDINALITY_TRANSFORMER,
         numeric=NUMERIC_TRANSFORMER,
--- a/skrub/_reporting/_data/templates/column-summary.html
+++ b/skrub/_reporting/_data/templates/column-summary.html
@@ -78,6 +78,13 @@
                 <dt>(constant) String length</dt>
                 <dd>{{ column.constant_string_length }}</dd>
                 {% endif %}
+
+                <dt>Cardinality</dt>
+                {% set cardinality_class = "warning" if column.is_high_cardinality else "ok" %}
+                <dd class=" {{ cardinality_class }}">
+                    {{ "high" if column.is_high_cardinality else "low" }}
+                    ({{ ">" if column.is_high_cardinality else "<" }} {{ config["cardinality_threshold"] }})
+                </dd>
             </dl>
         </div>
 
@@ -88,7 +95,7 @@
             <div class="copybutton-grid">
                 <div class="box">
                     <pre id="{{ val_id }}"
-                         data-copy-text="{{ column.constant_value.__repr__() }}">{{ column.constant_value }}</pre>
+                        data-copy-text="{{ column.constant_value.__repr__() }}">{{ column.constant_value }}</pre>
                     {{ buttons.copybutton(val_id) }}
                 </div>
             </div>
--- a/skrub/_reporting/_html.py
+++ b/skrub/_reporting/_html.py
@@ -8,6 +8,7 @@ import secrets
 import jinja2
 import pandas as pd
 
+from skrub import _config
 from skrub import _dataframe as sbd
 from skrub import selectors as s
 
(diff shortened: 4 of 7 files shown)


CHANGES.rst
```diff
@@ -44,6 +44,10 @@ New features
   and :func:`unpatch_display` are deprecated and will be removed in the next release
   of skrub. :pr:`1427` by :user:`Vincent Maladiere <Vincent-Maladiere>`.
 
+- :func:`get_config`, :func:`set_config` and :func:`config_context` now have a new
+  parameter `cardinality_threshold` that controls the threshold value used to warn user if they have
+  high cardinality columns in there dataset. :pr:`1498` by :user:`rouk1 <rouk1>`.
```
```suggestion
  high cardinality columns in their dataset. :pr:`1498` by :user:`rouk1 <rouk1>`.
```

jeromedockes:
thanks a lot, this seems helpful to quickly get an idea of what kind of column we are looking at. As the information is redundant with the number of unique values just above, I wonder if instead of adding a new row we could change the display of the number of unique values to make it more salient eg change its color or size depending on the cardinality, or display a little icon next to it? the benefit would be to reduce the height of the card and the amount of text it contains. WDYT?

GaelVaroquaux:
I agree with Jérôme. I would really like not to make the report bigger, add a line to it, in particular since it is partly redundant with what we have.

Screen real estate, and amount of information displayed is a crucial aspect of UI/UX

rouk1:
I've made some adjustments according to your feedbacks.
There is no more new line in the column summary. Instead if high cardinality is detected a warning button appears. Clicking or hovering the button opens a tooltip with a label that may be challenged.

(I had to made a small adjustment on tooltip width politic, I use a new CSS [function](https://developer.mozilla.org/en-US/docs/Web/CSS/calc-size), which if not supported will just fallback to autosizing).

Screenshot 2025-07-10 at 14 21 29
Screenshot 2025-07-10 at 14 21 09


Let me know what you think of it.

GaelVaroquaux:
I looooooove the screenshots.

I haven't been able to play with it (didn't quickly find a generated version), but what I see in the screenshot really looks nice.

More generally (and that's for a different PR), I'd love more of these things, these little tooltips so useful

Vincent-Maladiere:
@GaelVaroquaux you can look at the table in the home page: https://output.circle-artifacts.com/output/job/97eb2af4-0345-4017-be12-1718bada1c10/artifacts/0/doc/index.html

GaelVaroquaux:
> @GaelVaroquaux you can look at the table in the home page: https://output.circle-artifacts.com/output/job/97eb2af4-0345-4017-be12-1718bada1c10/artifacts/0/doc/index.html

Good point. And it confirms my assessment: this feature is really cool.

+1 for merge on my side.

@rouk1 : while you're at it, could you add a tooltip to the "danger sign" in the association column that would say "Some columns are highly redundant", if you have time (tooltip should display only if the danger sign displays)

Vincent-Maladiere:
- Should the high cardinality be marked as "danger"? Or is it more an information?
- You have a slight typo, "This column **h**as a high cardinality (> 40)."

rouk1:
> @rouk1 : while you're at it, could you add a tooltip to the "danger sign" in the association column that would say "Some columns are highly redundant", if you have time (tooltip should display only if the danger sign displays)

There is already a tooltip on the association tab. I think it has been implemented another way. I suppose because it sounds strange to add a button in a tab...
Screenshot 2025-07-10 at 15 25 22

@Vincent-Maladiere typo has been fixed and button now contains a _i_ instead of a warning symbol.

GaelVaroquaux:
> Should the high cardinality be marked as "danger"? Or is it more an information?

I like the way that it is currently done. Not too invasive, but still catches the attention

jeromedockes (approved):
Thanks a lot @rouk1  this is really neat!

rcap107:
Before we merge I want to point out that the PR is also making a change in the TableVectorizer that hasn't been documented

GaelVaroquaux:
> Before we merge I want to point out that the PR is also making a change in the TableVectorizer that hasn't been documented

As in adding an argument?

rcap107:
The init has changed from
`cardinality_threshold=40`
to
`cardinality_threshold=_config.get_config()["cardinality_threshold"]`

and I think that should be mentioned somewhere. The docstring at least should mention that the threshold can be set in the global config.

rouk1:
I've added a line in the docstring that mentions the global config. Please correct my english if needed. 🇫🇷 🥖 🧀

rcap107:
> I've added a line in the docstring that mentions the global config. Please correct my english if needed. 🇫🇷 🥖 🧀

works for me, but not for docstring :see_no_evil:

skrub/_config.py
```diff
@@ -200,6 +210,16 @@ def set_config(
...
         local_config["float_precision"] = float_precision
 
+    if cardinality_threshold is not None:
+        if (
+            not isinstance(cardinality_threshold, numbers.Integral)
+            or cardinality_threshold < 0
+        ):
+            raise ValueError(
+                "'cardinality_threshold' must be a positive"
+                "integer, got {cardinality_threshold!r}"
```
GaelVaroquaux: We have a missing "f" string, no?
```suggestion
                f"integer, got {cardinality_threshold!r}"
```

rouk1: Fixed thx !

rouk1:
> > I've added a line in the docstring that mentions the global config. Please correct my english if needed. 🇫🇷 🥖 🧀
> 
> works for me, but not for docstring 🙈

Something was off in my ruff local conf. Should be good now 🤞🏻

skrub/_table_vectorizer.py
```diff
@@ -730,7 +731,7 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
     def __init__(
         self,
         *,
-        cardinality_threshold=40,
+        cardinality_threshold=_config.get_config()["cardinality_threshold"],
```
GaelVaroquaux: OK, I think that we have a bug here.

This does not do what you probably think that it does: the function call "_config.get_config..." is evaluated when the code is imported, not when the function is called. As a consequence the default cannot be changed via set_config.

The right pattern is to set the value to "None", and in the function body, check that it is none and if so call get_config. Does this make sense?

rouk1: You're right ! 
I changed that for an explicit comparison with None (maybe one wants to pass 0 ?).
I updated the docstring accordingly, let me know if it is understandable.

GaelVaroquaux (approved):
LGTM. Thanks!

GaelVaroquaux:
Hum, I'm confused. We have strange failing tests. But they seem legit to me.

@rouk1 maybe a failed merge / rebase with main?

skrub/_table_vectorizer.py
```diff
@@ -742,7 +745,11 @@ def __init__(
         datetime_format=None,
         n_jobs=None,
     ):
-        self.cardinality_threshold = cardinality_threshold
+        self.cardinality_threshold = (
+            cardinality_threshold
+            if cardinality_threshold is not None
+            else _config.get_config()["cardinality_threshold"]
+        )
```
GaelVaroquaux: Actually, this logic must go in fit, and it must create a variable "self.cardinality_threshold_", else the object thinks that its default has been changed.

But bigger picture: I don't like the idea of a data-processing pipeline default being changed by a global. It's very brittle IMHO

GaelVaroquaux:
> @rouk1 maybe a failed merge / rebase with main?

Ok, it's something more fundamental, see:
https://github.com/skrub-data/skrub/pull/1498/files#r2200592242

But thinking of it, I think that the TableVectorizer should not depend on a global / config.

Sorry, I'm coming in late and should have spotted this before. But globals modifying the behavior of data processing is a bad pattern IMHO, that will lead to subtle bugs in user pipelines

rouk1:
> > @rouk1 maybe a failed merge / rebase with main?
> 
> Ok, it's something more fundamental, see: https://github.com/skrub-data/skrub/pull/1498/files#r2200592242
> 
> But thinking of it, I think that the TableVectorizer should not depend on a global / config.
> 
> Sorry, I'm coming in late and should have spotted this before. But globals modifying the behavior of data processing is a bad pattern IMHO, that will lead to subtle bugs in user pipelines

Should I just revert the changes in `TableVectorizer` and put back the default value to a literal ? (Source issues was indicating that threshold should be the same as in `TableVectorizer` so I tried to factorize this. But this is probably out of the scope of this PR.

Question stays should this threshold in the table report be a config element ? Do you prefer a simple kwarg in the constructor ?

I'll fix tests during refactor once we agree on what should be done.

ps: marking this PR draft as it's moving to much for a ready to review PR ^^

GaelVaroquaux:
> Should I just revert the changes in TableVectorizer and put back the default value to a literal ? (Source issues was indicating that threshold should be the same as in TableVectorizer so I tried to factorize this. But this is probably out of the scope of this PR.

I would say so.

> Question stays should this threshold in the table report be a config element ? Do you prefer a simple kwarg in the constructor ?

Which-ever people prefer. I think that displays being modified by globals is fine. So, if there was a discussion at some point pushing for the config thing, I will consider that people put in more thoughts than me and think that we should go forward with the discussion

Vincent-Maladiere:
I think it's a good idea to put this kwarg in TableReport init function, hopefully it would make the config options easier to understand :)

GaelVaroquaux:
> I think it's a good idea to put this kwarg in TableReport init function, hopefully it would make the config options easier to understand :)

I'm not sure that I parse your sentence, @Vincent-Maladiere . Can you explain what you mean?

Vincent-Maladiere:
Sorry if that wasn't clear; if `set_config` options like `high_cardinality_threshold` can also be found in the `TableReport`, my hope is that users will become familiar with this option first when playing directly with the `TableReport`. Later, it will become easier for them to use `set_config` since they've understood what its options stand for. Does that make sense?

rouk1:
I reverted modification on the `TableVectorizer` class, and CI is green !

Please decide if:
- I remove all modification on the global config and pass the threshold as a kwarg of `TableReport`'s constructor
- We merge this as it is right now (with a global config threshold value)

rouk1:
> Sorry if that wasn't clear; if `set_config` options like `high_cardinality_threshold` can also be found in the `TableReport`, my hope is that users will become familiar with this option first when playing directly with the `TableReport`. Later, it will become easier for them to use `set_config` since they've understood what its options stand for. Does that make sense?

Oh so there is a third option:

1. I remove all modification on the global config and pass the threshold as a kwarg of TableReport's constructor
2. We merge this as it is right now (with a global config threshold value)
3. I add a kwarg to `TableReport`'s contructor that is none by default and fallbacks to the global config.

GaelVaroquaux:
I'm +1 with the current state of the PR. I would merge it if it were me

Vincent-Maladiere:
Let's merge! Thank you for your patience and this very cool contribution @rouk1 !

rouk1:
Undrafted then !

Ping me i you want me to implement the kwarg !

GaelVaroquaux:
Great. Thanks a lot, this is super useful.

I particularly appreciate the UI "touch"

GaelVaroquaux:
Merged, by the way :)

PR #1505, opened 2025-07-09
DOC: Create an example that gives a use case.

MarieSacksick:
Use case, and checking how/if skrub can answer it.


Vincent-Maladiere:
Hey @MarieSacksick! Are you planning to continue working on this? :)

MarieSacksick:
Yes, but most likely next week. I have some progress from yesterday that I can add already.

MarieSacksick:
Hello @Vincent-Maladiere, it's a quite simple usecase, but here I have what I want. I just can't find how to actually do the prediction :sweat_smile:

examples/99_use_case.py
```diff
@@ -0,0 +1,139 @@
...
+# it's how I found it in https://skrub-data.org/dev/reference/generated/skrub.Expr.skb.apply.html#skrub.Expr.skb.apply,
+# but it seems very heavy to me.
+# Why do I have to regive X, it's supposed to have been defined before?
+# Apparently not, because it's buggy.
+# full_data_ops.skb.get_pipeline().fit({"X": X_dict})
+
+# %%
+# Other test
+trained_pipeline = full_data_ops.skb.get_pipeline().fit({"X": X_dict, "y": y_np})
+# %%
```
rcap107:
```suggestion
#%%
from sklearn.ensemble import HistGradientBoostingClassifier
hgb = HistGradientBoostingClassifier()
predictions = full_data_ops.skb.apply(hgb, y=y)
predictions
```
This adds the classifier at the end of the pipeline (the vectorizer needs to be fitted, but it isn't classifying anything)

Here you can also `prediction.skb.cross_validate()`

rcap107: However the actual learner hasn't been fitted yet

examples/99_use_case.py
```diff
@@ -0,0 +1,139 @@
...
+import pickle
+
+saved_model = pickle.dumps(trained_pipeline)
```
rcap107: So you fit it here by passing `fitted=True` to the `get_pipeline` function
```suggestion
import pickle

saved_model = pickle.dumps(predictions.skb.get_pipeline(fitted=True))
```

This will fit on the data you pass as the preview, if you want to pass new data here you can do it using the dictionary.

examples/99_use_case.py
```diff
@@ -0,0 +1,139 @@
...
+    "cc_emails": [generate_email() for _ in range(random.randint(0, 5))],
+}
+
+loaded_model = pickle.loads(saved_model)
+# here I'm quite puzzled, because none of two options work.
+# while it's what is described in
+# https://skrub-data.org/dev/auto_examples/expressions/10_expressions_intro.html
+loaded_model.predict({"X": X_input})
+loaded_model.skb.predict({"X": X_input})
+# %%
```
rcap107: Now you can actually predict 
```suggestion
loaded_model.predict({"X": X_input})
# %%
```

rcap107:
Hi @MarieSacksick, I added a few comments that should help. I think the main thing that is not clear from the docs, is that you're not suppose to apply `.fit()` directly to the data op, but do it either from a parameter search (with `fitted=True` to get the fitted best model), or by fitting directly the learner with `.skb.get_pipeline(fitted=True)`. Once you have a fitted learner, you can save it an apply it to new data like in the example. 

I see now that these patterns are never shown clearly in the documentation, which is a big problem. Thanks a lot for spotting the issues.

MarieSacksick:
Thanks @rcap107 for your feedbacks, it seems that it's all running now :)!  

When I run the last cell several times in a row, I have a weird behavior though:  
image
  

The prediction, instead of being only [1] seems to be concatenated with old ones.

MarieSacksick:
If you think that this example is interesting to keep in the documentation, I can add some text to make it more appropriate for real docs.

rcap107:
> Thanks @rcap107 for your feedbacks, it seems that it's all running now :)!
> 
> When I run the last cell several times in a row, I have a weird behavior though: image
> 
> The prediction, instead of being only [1] seems to be concatenated with old ones.

I think that what is happening is that you defined X_input as a single dictionary, rather than a list of dicts, which means that `pd.DataFrame` expands the `cc_emails` field rather than putting it into a list like it was in the definition of X_dicts

If you put X_input in a list it starts working and returns only a single value as intended

examples/99_use_case.py
```diff
@@ -0,0 +1,124 @@
...
+# in my microservice:
+X_input = {
+    "id": generate_id(),
+    "sender": generate_email(),
+    "title": generate_text(max_str_length=10, min_str_length=2),
+    "content": generate_text(max_str_length=100, min_str_length=10),
+    "date": generate_datetime(),
+    "cc_emails": [generate_email() for _ in range(random.randint(0, 5))],
+}
```
rcap107:
```suggestion
X_input = [{
    "id": generate_id(),
    "sender": generate_email(),
    "title": generate_text(max_str_length=10, min_str_length=2),
    "content": generate_text(max_str_length=100, min_str_length=10),
    "date": generate_datetime(),
    "cc_emails": [generate_email() for _ in range(random.randint(0, 5))],
}]
```

rcap107:
> If you think that this example is interesting to keep in the documentation, I can add some text to make it more appropriate for real docs.

Yes, I think it would be very useful if you can expand a bit the narrative :+1:

examples/99_use_case.py
MarieSacksick: thanks, it fixes my trouble! It breaks a bit my demo that it can be used for both stream and batch scoring, but it's fine.

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
+"""
+Usecase: developing locally, and avoiding to repeat code in production
+=======================================================================
+
+We will imagine a use case here and try to use skrub to answer it.
+For now, we start with a simple "flat" use case.
+Next, we can imagine that the data scientist has to predict fraud in retail. The target
+is at the basket level, while most information is at the items level.
+
```
rcap107: I think this part is referring to a different example? 
```suggestion
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+"""
+
+# %%
+###############################################################################
+# A typical use case
+# ------------------
+#
+# As a data scientist, I'm given a project where I have to predict if an email is fishy.
+# I am developing and testing my models locally, whether in a notebook or in a script.
```
rcap107:
```suggestion
# I am developing and testing my models locally: in a notebook or a python script.
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+# Once I'm happy, I want to deploy my model.
+# In this use case, the model is used in a microservice, and the model has to return the
+# score for each row. The microservice receives a payload in a json.
```
rcap107: This is not very clear to me: what is the score? Could you expand a bit on what the microservice is doing for someone who isn't familiar with them?

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+# To avoid having to recode the pipeline once the model is validated into the
+# microservice, which is both error-prone and troublesome, I would like to have the
+# skrub pipeline ready, and just have to load it.
+# It is also helpful, because it forces me to know before hand what kind of data I will
+# or can have at the entrance of the microservice, and avoid to build a model based on
+# information that is not accessible yet, in this part of the product pipeline.
```
rcap107:
```suggestion
# Working in this way is also helpful, because it forces me to know beforehand what kind 
# of data I will have at the entrance of the microservice, and avoids building a model based on
# information that is not accessible yet in this part of the product pipeline.
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+# ------------------
+#
+# As a data scientist, I'm given a project where I have to predict if an email is fishy.
+# I am developing and testing my models locally, whether in a notebook or in a script.
+# Once I'm happy, I want to deploy my model.
+# In this use case, the model is used in a microservice, and the model has to return the
+# score for each row. The microservice receives a payload in a json.
+# To avoid having to recode the pipeline once the model is validated into the
+# microservice, which is both error-prone and troublesome, I would like to have the
+# skrub pipeline ready, and just have to load it.
```
rcap107:
```suggestion
# To avoid having to recode the pipeline once the model is validated into the
# microservice, which is both error-prone and troublesome, I would like to be
# able to load an object that executes the same operations as the pipeline: 
# the skrub learner can help with this. 
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+# It is also helpful, because it forces me to know before hand what kind of data I will
+# or can have at the entrance of the microservice, and avoid to build a model based on
+# information that is not accessible yet, in this part of the product pipeline.
+# For instance, in my use case, I want to detect a spam before it reaches the receiver
+# mailbox. Therefore, I cannot use a feature which would use it the receiver opens the
+# email.
```
rcap107:
```suggestion
# For instance, in my use case, I want to detect a spam email before it reaches the receiver
# mailbox. Therefore, I cannot use a feature which is available only when the receiver opens 
# the email.
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+# The variable X for now is a list of dicts. It's not something that an estimator can
+# handle directly.
+# Let's convert it to a pandas DataFrame.
+import pandas as pd
+
+df = X.skb.apply_func(pd.DataFrame)
+
+# %%
+# For this example, we will use a strong baseline, with a tabular learner.
+learner = skrub.tabular_learner("classification")
```
rcap107:
```suggestion
# For this example, we will use a strong baseline, with a tabular pipeline.
model = skrub.tabular_pipeline("classification")
```

the `tabular_learner` has been renamed `tabular_pipeline` in a recent PR, exactly because it could be confused with the DataOps learner

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+
+# We can now apply the learner to the data.
+# The pipeline is fitted when applying the learner to the data.
+predictions = df.skb.apply(learner, y=y)
```
rcap107:
```suggestion
# We can now apply the predictive model to the data.
# The DataOps plan is ready after applying the model to the data.
predictions = df.skb.apply(learner, y=y)
```

the plan isn't fitted yet at this point

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+
+predictions.skb.draw_graph()
+
+# %%
+# To end the explorative work, I can save the model to a file somewhere.
```
rcap107:
```suggestion
# To end the explorative work, I can save the fitted learner to a file somewhere.
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+import joblib
+
+with open("model.pkl", "wb") as f:
+    joblib.dump(predictions.skb.get_pipeline(fitted=True), f, protocol=5)
```
rcap107: Is the `protocol` parameter needed? If it is, please explain why, otherwise it should be removed

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+df = X.skb.apply_func(pd.DataFrame)
+
+# %%
+# For this example, we will use a strong baseline, with a tabular learner.
+learner = skrub.tabular_learner("classification")
+
+# We can now apply the learner to the data.
+# The pipeline is fitted when applying the learner to the data.
+predictions = df.skb.apply(learner, y=y)
+
```
rcap107:
```suggestion
# We can then explore the full plan: 
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+predictions.skb.draw_graph()
+
+# %%
+# To end the explorative work, I can save the model to a file somewhere.
+import joblib
+
+with open("model.pkl", "wb") as f:
```
rcap107: I think it's better if this is made explicit
```suggestion
with open("learner.pkl", "wb") as f:
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+    "sender": generate_email(),
+    "title": generate_text(max_str_length=10, min_str_length=2),
+    "content": generate_text(max_str_length=100, min_str_length=10),
+    "date": generate_datetime(),
+    "cc_emails": [generate_email() for _ in range(random.randint(0, 5))],
+}
+
+# I just have to load the model and use it to predict the score for this input.
+with open("model.pkl", "rb") as f:
+    loaded_model = joblib.load(f)
```
rcap107:
```suggestion
# I just have to load the learner and use it to predict the score for this input.
with open("learner.pkl", "rb") as f:
    loaded_learner = joblib.load(f)
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+
+prediction = loaded_model.predict({"X": [X_input]})
```
rcap107:
```suggestion
prediction = loaded_learner.predict({"X": [X_input]})
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+    "title": generate_text(max_str_length=10, min_str_length=2),
+    "content": generate_text(max_str_length=100, min_str_length=10),
+    "date": generate_datetime(),
+    "cc_emails": [generate_email() for _ in range(random.randint(0, 5))],
+}
+
+# I just have to load the model and use it to predict the score for this input.
+with open("model.pkl", "rb") as f:
+    loaded_model = joblib.load(f)
+
```
rcap107:
```suggestion
# ``X_input`` must be passed as a list so that it can be parsed correctly as a dataframe by Pandas. 
```

examples/10_use_case.py
```diff
@@ -0,0 +1,154 @@
...
+prediction = loaded_model.predict({"X": [X_input]})
+prediction
+
+# %%
+###############################################################################
+# Conclusion
+# ----------
+#
+# Thanks to skrub pipeline, I have the insurance that all the transformations and
```
rcap107:
```suggestion
# Thanks to the skrub DataOps and learner, I have the insurance that all the transformations and
```

examples/10_use_case.py
MarieSacksick: By "next", I mean in another iteration, once (if?) we want to complete this example to show more of skrub's features. It may not make sense now, if you want to keep examples short and to the point.

examples/10_use_case.py
MarieSacksick: the score is the prediction.  
Sure, I'll be more precise :) !

examples/10_use_case.py
rcap107: It's better if we keep the example short, and add new examples later

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+# To do this, they want to use a machine learning model, contained into a microservice.
+# The microservice receives a payload in a json, and returns a score between 0 and 1,
+# depending on how likely it is that the email is a spam.
+#
+# To avoid having to recode the pipeline once the model is validated into the
+# microservice, which is both error-prone and troublesome, I would like to be
+# able to load an object that executes the same operations as the pipeline:
+# the skrub learner can help with this.
+# Working in this way is also helpful,
+# because it forces me to know beforehand what kind of data I will have at the entrance
```
rcap107:
```suggestion
# because it forces me to know beforehand what kind of data I have at the entrance
```
sticking to present tense rather than future

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+# cannot use a feature which is available only when the receiver opens the email.
+
+# %%
+###############################################################################
+# Let's create the data set
+# ------------------
+#
+# Given my use case, the data set is not a dataframe, but a list of dict.
+# We are going to generate a fully random data set. We will not have a look at the
+# quality of the prediction, we want to focus on the pipeline construction.
```
rcap107:
```suggestion
# In this use case, the emails to be tested when the model is put in production
# are not contained in a dataframe, but in a json. As a result, our training data
# should also be contained in a list of dictionaries
# We are going to generate a fully random data set. We will not have a look at the
# quality of the prediction, since we want to focus on the pipeline construction.
```

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+
```
rcap107:
```suggestion
# Generating the training data
# --------------------
# In this section, we define a few functions that help us with generating the 
# training data in dictionary form. 
```

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+    random_length = random.randint(min_str_length, max_str_length)
+    random_text = "".join(
+        random.choice(string.ascii_letters + string.digits + string.punctuation)
+        for _ in range(random_length)
+    )
+    return random_text
+
+
+# %%
+n_samples = 1000
```
rcap107:
```suggestion
# We generate 1000 training samples and store them in a list of dictionaries: 

n_samples = 1000
```

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+# variable.
+import skrub
+
+X = skrub.X(X)
+y = skrub.y(y)
+
+# %%
+# The variable X for now is a list of dicts. It's not something that an estimator can
+# handle directly.
+# Let's convert it to a pandas DataFrame.
```
rcap107:
```suggestion
# Let's convert it to a pandas DataFrame using :meth:`.skb.apply_func() <DataOp.skb.apply_func>`
```
this should render properly

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+    }
+    for _ in range(n_samples)
+]
+
+
+# generate array of 1 and 0 to represent the target variable
+y = np.random.binomial(n=1, p=0.9, size=n_samples)
+
+# %%
+# Let's start our skrub DataPlan by indicating what are the features and the target
```
rcap107:
```suggestion
# Building the DataOps plan
# ---------------------------
# Let's start our DataOps plan by indicating what are the features and the target
```
in the end, we decided to go with "DataOps plan", rather than having the "Data plan" as a separate concept

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+# %%
+# The variable X for now is a list of dicts. It's not something that an estimator can
+# handle directly.
+# Let's convert it to a pandas DataFrame.
+import pandas as pd
+
+df = X.skb.apply_func(pd.DataFrame)
+
+# %%
+# For this example, we will use a strong baseline, with a tabular pipeline.
```
rcap107:
```suggestion
# For this example, we will use a strong baseline, with skrub's :func:`tabular_pipeline() <tabular_pipeline>`.
```

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+
+# We can now apply the predictive model to the data.
+# The DataOps plan is ready after applying the model to the data.
+predictions = df.skb.apply(tab_pipeline, y=y)
+
+# We can then explore the full plan:
+predictions.skb.draw_graph()
+
+# %%
+# To end the explorative work, I can save the fitted learner to a file somewhere.
```
rcap107:
```suggestion
# To end the explorative work, we need to build the learner, fit it, and save it to a file. 
# Passing ``fitted=True`` to the :meth:`.skb.make_learner() <DataOp.skb.make_learner>`
# function makes it so that the learner is fitted on the data that has been passed to the 
# variables of the DataOps plan. 
```

examples/10_use_case.py
```diff
@@ -0,0 +1,157 @@
...
+prediction
+
+# %%
+###############################################################################
+# Conclusion
+# ----------
+#
+# Thanks to the skrub DataOps and learner, I have the insurance that all the
+# transformations and preprocessing done when developing the models are similar to the
+# ones done in production.
```
rcap107:
```suggestion
# Thanks to the skrub DataOps and learner, I am assured that all the 
# transformations and preprocessing done during model development
#  are the exactly the same that are done in production.
```

examples/11_use_case.py
```diff
@@ -0,0 +1,165 @@
...
+import skrub
+
+X = skrub.X(X)
+y = skrub.y(y)
+
+# %%
+# The variable X for now is a list of dicts. It's not something that an estimator can
+# handle directly.
+# Let's convert it to a pandas DataFrame using
+# :meth:`.skb.apply_func() <DataOp.skb.apply_func>`
```
Vincent-Maladiere:
```suggestion
# Let's convert it to a pandas DataFrame using :func:`~skrub.DataOp.skb.apply_func`.
```

examples/11_use_case.py
```diff
@@ -0,0 +1,165 @@
...
+import pandas as pd
+
+df = X.skb.apply_func(pd.DataFrame)
+
+# %%
+# For this example, we will use a strong baseline, with skrub's
+# :func:`tabular_pipeline() <tabular_pipeline>`.
```
Vincent-Maladiere:
```suggestion
# :func:`~skrub.tabular_pipeline()`.
```

examples/11_use_case.py
```diff
@@ -0,0 +1,165 @@
...
+# %%
+# To end the explorative work, we need to build the learner, fit it, and save it to a
+# file.
+# Passing ``fitted=True`` to the :meth:`.skb.make_learner() <DataOp.skb.make_learner>`
+# function makes it so that the learner is fitted on the data that has been passed to
+# the variables of the DataOps plan.
+import joblib
+
+with open("learner.pkl", "wb") as f:
+    joblib.dump(predictions.skb.make_learner(fitted=True), f)
```
Vincent-Maladiere: To make this more readable?
```suggestion
    learner = predictions.skb.make_learner(fitted=True)
    joblib.dump(learner, f)
```

examples/11_use_case.py
```diff
@@ -0,0 +1,165 @@
...
+# The DataOps plan is ready after applying the model to the data.
+predictions = df.skb.apply(tab_pipeline, y=y)
+
+# We can then explore the full plan:
+predictions.skb.draw_graph()
+
+# %%
+# To end the explorative work, we need to build the learner, fit it, and save it to a
+# file.
+# Passing ``fitted=True`` to the :meth:`.skb.make_learner() <DataOp.skb.make_learner>`
```
Vincent-Maladiere:
```suggestion
# Passing ``fitted=True`` to the :func:`~skrub.DataOp.skb.make_learner`
```

examples/11_use_case.py
```diff
@@ -0,0 +1,170 @@
...
+# the skrub learner can help with this.
+# Working in this way is also helpful,
+# because it forces us to know beforehand what kind of data we have at the entrance
+# of the microservice, and avoids building a model based on information that is not
+# accessible yet in this part of the product pipeline. For instance, in this use case,
+# we want to detect a spam email before it reaches the receiver mailbox. Therefore, we
+# cannot use a feature which is available only when the receiver opens the email.
+#
+# We will not have a look at the
+# quality of the prediction, since this example focuses on the pipeline construction.
```
Vincent-Maladiere: Slight rewording:

```suggestion
# As a team of data scientists, we are tasked with a project to predict whether an email is potentially 
# malicious (i.e., spam or phishing). We develop and test our models locally, either in a Jupyter
# notebook or within a Python script. Once we are satisfied with the model's performance, we move
# on to deploying it.
#
# In this use case, every time the email provider receives a new email, they want to verify whether
# it is spam before displaying it in the recipient’s inbox. To achieve this, they plan to integrate
# a machine learning model within a microservice. This microservice will accept an email’s data as
# a JSON payload and return a score between 0 and 1, indicating the likelihood that the email is spam.
#
# To avoid rewriting the entire data pipeline when moving from model validation to production
# deployment, which is both error-prone and inefficient, we prefer to load an object that
# encapsulates the same processing pipeline used during model development. This is where the
# :class:`~skrub.SkrubLearner` can help.
#
# Adopting this workflow also has the benefit of forcing us to clearly define the type of data
# that will be available at the input of the microservice. It helps ensure we build models that
# rely only on information accessible at this specific point in the product pipeline. For
# example, since we want to detect spam before the email reaches the recipient’s inbox,
# we cannot use features that are only available after the recipient opens the email.
#
# Since this example is focused on the pipeline construction itself, we won't look at our model performance.
```

MarieSacksick:
@Vincent-Maladiere, @rcap107, question about where to put this example; for now it's at the root of the examples, but it looks like you have a [subdirectory for data ops](https://skrub-data.org/dev/auto_examples/data_ops/index.html), shouldn't it be there instead?

Vincent-Maladiere:
Yes, I think it should be placed there!

rcap107 (approved):
Looks good to me! Thanks a lot @MarieSacksick

examples/data_ops/13_use_case.py
```diff
@@ -0,0 +1,171 @@
...
+# To end the explorative work, we need to build the learner, fit it, and save it to a
+# file.
+# Passing ``fitted=True`` to the :func:`~skrub.DataOp.skb.make_learner`
+# function makes it so that the learner is fitted on the data that has been passed to
+# the variables of the DataOps plan.
+import joblib
+
+with open("learner.pkl", "wb") as f:
+    learner = predictions.skb.make_learner(fitted=True)
+    joblib.dump(learner, f)
```
rcap107, line 140: Just wondering, is saving on disk here going to cause issues in the CI? it doesn't seem to be the case, but in other examples we usually save in bytestrings :thinking:

GaelVaroquaux:
> Just wondering, is saving on disk here going to cause issues in the CI? it doesn't seem to be the case, but in other examples we usually save in bytestrings 🤔

I think that in examples it should be fine.
In tests, you want to be careful, as you might be running on read-only disks.

Vincent-Maladiere (approved):
Very nice, thank you @MarieSacksick!

PR #1506, opened 2025-07-09
Add a check box in the distribution of TableReport to switch between a percentage of numbers in the plot.

### LINKED ISSUE #1454: Showing values for bar plots in TableReport

lionelkusch:
This PR is based on the PR #1496.

This PR solves the desire requirement behaviour of the issue #1454 .

The principle of this feature is to use JavaScript to send a message which changes the style of the html division of the figure between 'none' and 'block' (default value). This means that there are 2 figures created for each column: one with a percentage label and one with normal numbers.

WARNING: The percentage is fixed in the size of the display number, which is not the case for the number. This can create some not rendering well for large numbers.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -117,6 +117,9 @@ Changes
 - A new parameter ``float_precision`` has been added to the global config to control the number of significant digits
   displayed for floating-point values in reports. :pr:`1470` by :user:`George S <georgescutelnicu>`.
 
+- The display of :class:`TableReport` has the new capability to switch between display percentage and the value of
+  the histogram and the value_counts plot. :pr:`1506` by :user:`Lionel Kusch<lionelkusch>`.
+
 Bugfixes
 --------
 - Fixed a bug that caused the :class:`StringEncoder` and :class:`TextEncoder` to raise an exception if the
--- a/skrub/_reporting/_data/templates/column-summaries.html
+++ b/skrub/_reporting/_data/templates/column-summaries.html
@@ -18,7 +18,7 @@
 	    </div>
 	    <button type="button" data-test="select-all-columns" data-manager="SelectAllVisibleColumns">Select all</button>
 	    <button type="button" data-test="deselect-all-columns" data-manager="DeselectAllColumns">Deselect all</button>
-	    <input type="checkbox" data-test="percentage-display" data-manager="PercentageDisplay">%</input>
+	    <input type="checkbox" data-test="percentage-display" data-manager="PercentageDisplay" checked >%</input>
 	</div>
     </div>
 
--- a/skrub/_reporting/_data/templates/column-summaries.html
+++ b/skrub/_reporting/_data/templates/column-summaries.html
@@ -18,6 +18,7 @@
 	    </div>
 	    <button type="button" data-test="select-all-columns" data-manager="SelectAllVisibleColumns">Select all</button>
 	    <button type="button" data-test="deselect-all-columns" data-manager="DeselectAllColumns">Deselect all</button>
+	    <input type="checkbox" data-test="percentage-display" data-manager="PercentageDisplay">%</input>
 	</div>
     </div>
 
--- a/skrub/_reporting/_data/templates/column-summary.html
+++ b/skrub/_reporting/_data/templates/column-summary.html
@@ -98,7 +98,10 @@
         {% for plot_name in column.plot_names %}
         <div>
             <div class="margin-t-m" data-svg-needs-adjust-viewbox data-test="switch-display" data-manager="SwitchFigures" style="display: block;">
-                {{ column[plot_name] | safe }}
+                {{ column[plot_name][1] | safe }}
+            </div>
+            <div class="margin-t-m" data-svg-needs-adjust-viewbox data-test="switch-display" data-manager="SwitchFigures" style="display: none;">
+                {{ column[plot_name][0] | safe }}
             </div>
             {% if plot_name == "value_counts_plot" %}
             <details data-test="frequent-values-details">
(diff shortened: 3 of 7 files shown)


jeromedockes:
thanks a lot @lionelkusch this is neat! However I wonder if it is worth doubling to computation time used to produce a report as this is something we've been meaning to keep fast for interactive use in a notebook. and there is the added complexity to the html and javascript 🤔  What do others think?

GaelVaroquaux:
I have the same question as @jeromedockes (sorry @lionelkusch , we really appreciate your involvement)

lionelkusch:
In the short time of the sprint, I want it to propose something functional. The creation of two independent figures was the simplest solution for me.
However, if you want to include this new feature, I can provide a new version more optimise, which generates two versions of the same figure with different labels. This will increase a few milliseconds the time of computation.

Vincent-Maladiere:
What about a simpler "abs value (percentile)" kind of display, for each bar?

Vincent-Maladiere:
@lionelkusch do you plan to continue working on this PR?

lionelkusch:
Yes, I can take care of that if you agree on what you want displayed.

Vincent-Maladiere:
I think you can go ahead with my comment above

lionelkusch:
There are 2 displays one for the percentage and one with the full value.
The "abs value (percentile)" is for which display?

Additionally, I don't fully understand what you mean by "abs value (percentile)".
Can you give an example or the mathematical formula?

Vincent-Maladiere:
You can combine both. The height of each column corresponds to the total number of elements in the bin, and the text is the number of elements in the bin, with the percentile representation between parentheses. This applies to both numerical and categorical columns.

lionelkusch:
The problem when you combine the two in the same graphic is that the text will be longer than the columns. This won't look nice and it will be difficult to automatically add the text to the graphic.
This is linked to my warning that for the moment, the percentile is fixed in the number of characters, which simplifies the position of the text. Adding the number of elements creates some variability and creates a difficult problem for the position of the text relative to the bar.

In my opinion, it's better to stay with 2 representations: one with percentile and one with absolute numbers and switch between them.

lionelkusch:
This is an optimised version of the generation of the figures without generating twice the figures.

I released that it is not possible to duplicate or copy a figure of matplolib.
I decided to add the label at the end of the creation of the figure.
The addition of the label at the end allows me to add the label, serialise the figure and remove them just after getting the figure without them and restart.

I changed a bit of the label using the _utils for the format of the number and adjusted the position.
Tell me if the result is good for you.

Vincent-Maladiere:
Thanks for continuing this work. Overall, I'm still not convinced by the double plot approach, because:
- The % box is hard to notice, so most people won't use it. Maybe we could mitigate this by adding the button to each plot? Anyway, I'm not fond of adding complexity to the UI.
- This complexifies the logic significantly.
- Performance-wise, I noted a 10% increase in compute time, which is not very significant. However, the disk footprint increases by 50%, which I still think is a lot for this feature, although the disk space itself is not big in absolute terms.

Addressing your remarks:

> The problem when you combine the two in the same graphic is that the text will be longer than the columns. This won't look nice and it will be difficult to automatically add the text to the graphic.

If the text crossing the bar edge doesn't look good, we can choose to always have the text outside of the bar, increasing the figure size slightly if necessary.

> Adding the number of elements creates some variability and creates a difficult problem for the position of the text relative to the bar.

I think this can be addressed by controlling the number of digits in the format using custom rules, like full representation when the entries are smaller than 10,000 and scientific notation otherwise. We would have an upper bound on the size of the text.

---

What do the others think?

lionelkusch:
I agree with you that this increase the complexity for the manage of the text and that there is an increase in memory for the generation of the second figure.

rcap107:
> ...
> What do the others think?

I agree on all points here and I would also prefer having both numbers visible at the same time. 

> Adding the number of elements creates some variability and creates a difficult problem for the position of the text relative to the bar.

How about adding the % to the right (the top) of the plot, aligned with each bar? With a fixed format (like 3 significant digits) the space should remain mostly constant, while the actual value is still represented in the same way as it is now. If the % sign takes too much space, it could be added above the column of numbers as an annotation.

lionelkusch:
By your remarks, I see that the switching views is not interesting.
I propose a static version where the percentage and the value are represented at the same time in PR #1496.
I let you look at it for continuing the conversation.
I am sorry not to continue into this PR but I prefer to keep this PR focused only on the dynamic approach and the PR #1496 for static approaches. I hope you can understand me.

rcap107:
> By your remarks, I see that the switching views is not interesting. I propose a static version where the percentage and the value are represented at the same time in PR #1496. I let you look at it for continuing the conversation. I am sorry not to continue into this PR but I prefer to keep this PR focused only on the dynamic approach and the PR #1496 for static approaches. I hope you can understand me.

It makes sense, and thanks for working on both versions. We might revisit this version in the future, but I think that for the time being the static version would already be quite useful.

jeromedockes:
in the end the alternative PR was merged so we can clsoe this

Issue #1502, opened 2025-07-09
Code block formatting is broken in the Install from source page


rcap107:
### Describe the issue linked to the documentation

See screenshot 

<img width="928" height="413" alt="Image" src="https://github.com/user-attachments/assets/69d70944-c036-4003-be83-0a19dc8a9f1e" />

PR #1507, opened 2025-07-09
fix code block formatting broken in the Install from source page rst format

GaelVaroquaux:
Should fix https://github.com/skrub-data/skrub/issues/1502


rcap107 (approved):

rcap107:
looks like it does, thanks a lot @GaelVaroquaux

PR #1514, opened 2025-07-11
add a variable to catch the pipeline

GaelVaroquaux:
A missing change from https://github.com/skrub-data/skrub/pull/1478

### DIFF
--- a/examples/10_apply_on_cols.py
+++ b/examples/10_apply_on_cols.py
@@ -70,7 +70,7 @@ Xt
 # a :class:`~sklearn.pipeline.Pipeline`.
 from sklearn.pipeline import make_pipeline
 
-make_pipeline(
+model = make_pipeline(
     apply_string_encoder,
     apply_pca,
 ).fit_transform(X)


examples/10_apply_on_cols.py
```diff
@@ -70,7 +70,7 @@
 # a :class:`~sklearn.pipeline.Pipeline`.
 from sklearn.pipeline import make_pipeline
 
-make_pipeline(
+model = make_pipeline(
```
Vincent-Maladiere, line 73: I'm afraid you've made a typo 😬
Here, `model` is the transformed dataframe, not the pipeline

GaelVaroquaux:
> -make_pipeline(
> +model = make_pipeline(

> I'm afraid you've made a typo 😬
> Here, model is the transformed dataframe, not the pipeline

Right. I read the example wrong.

I think that we need to break this line in two line: one that creates the model, the other that transforms the dataframe.

Do you think that you can do it, you'll be better than me (obviously, given the above)?

Vincent-Maladiere:
No worries, I can do it!

PR #1524, opened 2025-07-17
ENH: adding train/test splits to the employee salaries dataset

rcap107:
This is similar to what is already available for the credit card fraud dataset, but in this case I set the default return value to "all" to keep the original behavior. Employee salaries is used all over the library as an example, so changing the default means updating it everywhere


rcap107:
does this PR even need the changelog

GaelVaroquaux:
> does this PR even need the changelog

Nope

PR #1525, opened 2025-07-18
Various fixes for config

### LINKED ISSUE #1523: minor issues with skrub config
- how do I set `max_plot_columns` to always plot? I would expect setting it to None to mean "no maximum" but it still uses the default 30
- similarly (less important) there does not seem to be a way to set subsampling seed to None. maybe not a big deal since this is risky due to the possibility of subsampling several items separately
- BUG: setting `enable_subsampling` to "force" will also subsample even during `predict`; we never want subsampling in that case (`y_score` would have a different shape than `y_true`)

rcap107:
Fixes #1523


rcap107:
Looks like I can't approve my own PR :upside_down_face:

rcap107:
Looks good to me, except for coverage being annoying

Vincent-Maladiere:
This should fix codecov issues

skrub/_data_ops/_subsampling.py
```diff
@@ -13,14 +13,15 @@
 
 def _should_subsample(mode, environment):
     enable_subsampling = get_config()["enable_subsampling"]
+    if "fit" not in mode and mode != "preview":
+        # We never want to subsample outside of fit, fit_transform and preview.
```
GaelVaroquaux, line 17: Is this behavior documented somewhere?

GaelVaroquaux: It must be

GaelVaroquaux (approved):
LGTM

PR #1527, opened 2025-07-18
Fixing a bug in two examples introduced by refactoring

rcap107:
Docs don't render because I changed a name and forgot to update them


GaelVaroquaux (approved):
LGTM. Thanks!

PR #1529, opened 2025-07-20
DOC - Reworking and extending DataOps userguide

rcap107:

PR #1531, opened 2025-07-21
Improve choices example

Vincent-Maladiere:
This PR suggests a lot of rewording and adds links to references of the choices example. To be useful, this example should be significantly shorter, and most parts detailing the features of the data ops or choices should go to the user guides, in https://github.com/skrub-data/skrub/pull/1529.

### DIFF
--- a/examples/data_ops/11_choices.py
+++ b/examples/data_ops/11_choices.py
@@ -3,7 +3,7 @@
 
 .. _example_tuning_pipelines:
 
-Tuning DataOps
+Hyperparameter tuning with DataOps
 ==============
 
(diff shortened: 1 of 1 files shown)


examples/data_ops/11_choices.py
```diff
@@ -2,39 +2,44 @@
 .. currentmodule:: skrub
 
 .. _example_tuning_pipelines:
+
 Tuning DataOps
```
rcap107: I think we can be more explicit here
```suggestion
Hyperparameter tuning with DataOps
```

rcap107 (approved):
Looks much better, thanks a lot @Vincent-Maladiere

PR #1534, opened 2025-07-22
FIX Renaming more instances of "expression" to "data_op" in the codebase

rcap107:
I found some more instances of expressions/pipelines that have slipped by in the main refactoring


rcap107:
I did not touch the examples because we're still working on those in separate PRs

rcap107:
Talking with @jeromedockes, we were thinking of renaming the `full_report` to `evaluation_report`, what do you think @Vincent-Maladiere ?

Vincent-Maladiere:
I don't have a strong opinion. I think the concept of evaluation in ML is quite different from the data ops evaluation, so it might be confusing.

Vincent-Maladiere (approved):

rcap107:
Let's stick to `full_report` for the time being

PR #1538, opened 2025-07-23
better error message when apply() fails

jeromedockes:
when a node fails the line where it was defined is added to the error message and the error is stored in the node to display n the report.

during a previous refactoring a change was made that caused this to only be the case for nodes that define compute(), and thus not Apply nodes, this PR fixes that


skrub/_data_ops/_evaluation.py
```diff
@@ -268,29 +268,12 @@ def handle_data_op(self, data_op):
 
     def _eval_data_op(self, data_op):
         impl = data_op._skrub_impl
-        if hasattr(impl, "eval"):
-            return (yield from impl.eval(mode=self.mode, environment=self.environment))
-        return (yield from super().handle_data_op(data_op))
-
-    def handle_choice(self, choice):
```
jeromedockes, line 275: git does not produce a great diff for this change, but basically it is moving the try/except that was around _skrub_impl.compute() to put it in _eval_data_op. the handle_choice, handle_choice_match functions are unchanged

PR #1543, opened 2025-07-23
Fix - `get_feature_names_out` fails in two cases

### LINKED ISSUE #1540: Discussion: is get_features_name_out something missing or on purpose?
Hello :) !

I am trying to access the features names outside at the end of a pipeline where I use a GapEncoder and a DropCol, because I need them to plot feature importance with feature names beside. 

My question is: is it meant to be implemented at some point? Should I use a workaround instead of get_features_name_out?

code snippets:  
```
from skrub import GapEncoder

datasets = fetch_employee_salaries()
X, y = datasets.X[:500], datasets.y[:500]

preprocessing = make_column_transformer(
    (GapEncoder(n_components=100), "division"),
)

preprocessing.fit(X)
preprocessing.get_feature_names_out()
```

and for dropcols:  
```
from skrub import DropCols
from sklearn.linear_model import LinearRegression

datasets = fetch_employee_salaries()
X, y = datasets.X[:500], datasets.y[:500]

model = make_pipeline(DropCols(['gender', 'department', 'department_name', 'division',
       'assignment_category', 'employee_position_title', 'date_first_hired']), LinearRegression())

model.fit(X, y)
model.get_feature_names_out()
```

They both output an error.  
Thank you :)!

rcap107:
Fixes #1540


rcap107:
The tests are failing with the `min-optional-deps` configuration because the column transformer in min-deps configuration does not support polars dataframes: in `all_columns`, it is expecting a pandas index, while polars gives a list.

How should I deal with this ? I can remove the test with the column transformer (I added it mostly to check for the behavior in #1540), but maybe there is a way to skip the test for the given configuration. What's the best choice here?

CHANGES.rst
```diff
@@ -9,6 +9,16 @@ Release history
 Ongoing development
 ===================
 
+Bugfixes
+--------
+
+- ``get_feature_names_out`` now works correctly when used by :class:`GapEncoder`,
+  :DropCols:, :class:`SelectCols:` from within a scikit-learn ``Pipeline``. In
```
Vincent-Maladiere:
```suggestion
  :class:`DropCols`, :class:`SelectCols:` from within a scikit-learn ``Pipeline``. In
```

skrub/tests/test_select_cols.py
```diff
@@ -64,10 +66,19 @@ def test_drop(df_module):
     assert Drop().fit(col).transform(col) == []
 
 
-def test_get_feature_names_out(df_module):
-    df = df_module.example_dataframe
+def test_get_feature_names_out(df):
+    # df = df_module.example_dataframe
```
Vincent-Maladiere:
```suggestion
```

rcap107:
> Hey @rcap107! After testing on different CI versions with pixi, it looks that the error you have is specific to Polars on column transformers, on a prior version of sklearn. We can go two ways:
> 
> * We force conversion from polars to Pandas for old sklearn versions
> * or we ignore it and suggest people to upgrade to latest sklearn and polars if they aim at using column transformer
> 
> WDYT?

Personally I'd rather go with the second, but since it's on sklearn side I wonder if someone that is stuck on that version just can't upgrade it :thinking: 

So I think the first choice is overall better, but we need to document it properly

rcap107:
> > Hey @rcap107! After testing on different CI versions with pixi, it looks that the error you have is specific to Polars on column transformers, on a prior version of sklearn. We can go two ways:
> > 
> > * We force conversion from polars to Pandas for old sklearn versions
> > * or we ignore it and suggest people to upgrade to latest sklearn and polars if they aim at using column transformer
> > 
> > WDYT?
> 
> Personally I'd rather go with the second, but since it's on sklearn side I wonder if someone that is stuck on that version just can't upgrade it 🤔
> 
> So I think the first choice is overall better, but we need to document it properly

I'm trying to implement this, but it's not clear to me where I should force the conversion to Polars, given that the call that fails is inside the code of the column transformer, and not on our side

Unless we convert by default (again, where?) polars dataframes to pandas if the sklearn version is too low.

Vincent-Maladiere:
You're right, the issue is on sklearn side. One way to do it would be to patch the `fit_transform` method of the ColumnTransformer for sklearn < 1.?.0. in `skrub/sklearn_compat.py`. Or we could ignore it and skip fail the test in this situation.

rcap107:
I had to patch the fit_transform and transform methods of the ColumnTransformer so that they convert `X` to pandas if it is a polars dataframe or series

skrub/_sklearn_compat.py
```diff
@@ -189,10 +189,56 @@ def _is_fitted(estimator, attributes=None, all_or_any=all):
...
         return len(fitted_attrs) > 0
 
+    try:
+        import polars as pl
+        from sklearn.compose._column_transformer import ColumnTransformer
+
+        ColumnTransformer.original_fit_transform = ColumnTransformer.fit_transform
+        ColumnTransformer.original_transform = ColumnTransformer.transform
+
+        def _patched_fit_transform(self, X, y=None):
```
Vincent-Maladiere: We could simplify the logic slightly by using a decorator, and use it like:

```python
ColumnTransformer.fit_transform = wrap_to_pandas(ColumnTransformer.fit_transform)
ColumnTransformer.transform = wrap_to_pandas(ColumnTransformer.transform)
```

skrub/_sklearn_compat.py
```diff
@@ -189,10 +189,56 @@ def _is_fitted(estimator, attributes=None, all_or_any=all):
...
+            """Patched version of fit_transform that handles polars DataFrames"""
+            original_type = None
+            if isinstance(X, (pl.DataFrame, pl.Series)):
+                original_type = type(X)
+                X = X.to_pandas()
```
Vincent-Maladiere: what happens if y is not None?

Vincent-Maladiere (commented):
Hey @rcap107, some feedback

skrub/_sklearn_compat.py
rcap107: y is forwarded unchanged to the ColumnTransformer in the fit_transform, transform does not take y at all

rcap107:
as it turns out, the bug with DropCols is also affecting any situation where ApplyToCols is used, both explicitly, or from within the dataops evaluation, so it's a more serious issue than I had first understood

skrub/_sklearn_compat.py
Vincent-Maladiere: Would it be an issue if y was a pandas DataFrame/series?

Vincent-Maladiere:
> the bug with DropCols is also affecting any situation where ApplyToCols is used

I'm not sure to follow, could you show an example?

rcap107:
> > the bug with DropCols is also affecting any situation where ApplyToCols is used
> 
> I'm not sure to follow, could you show an example?

this code fails on main, but it works with the fix in this PR

```python
#%%
import skrub
from skrub import DropCols
import polars as pl
import skrub.selectors as s
from skrub._apply_to_cols import ApplyToCols

df = pl.DataFrame({
    "a": [1, 2, 3],
    "b": [4, 5, 6],})

# %%
app = ApplyToCols(DropCols(s.glob("")),)
app.fit_transform(df)
# %%
v = skrub.var("df", df)
v.skb.apply(app)

```

rcap107:
The reason this issue is stuck is because the ColumnTransformer does not work with polars on scikit-learn < 1.5, so I tried to patch it to fix the problem. However, this was a bigger issue than anticipated.

I realized that the test that was failing was something that I added to reproduce the error in #1540, so the problem has been there for much longer. 

I think we should ignore that particular problem in this PR, and tackle separately. Here, we can address the specific problems with #1540.

rcap107:
In short, I'll remove the extra test that is failing and all the sklearn compat code, and continue with the rest so we can merge the fix

skrub/tests/test_select_cols.py
```diff
@@ -64,10 +66,18 @@ def test_drop(df_module):
     assert Drop().fit(col).transform(col) == []
 
 
-def test_get_feature_names_out(df_module):
-    df = df_module.example_dataframe
+def test_get_feature_names_out(df):
```
glemaitre, line 69:
```suggestion
def test_get_feature_names_out(df):
    """Check the behaviour of `get_feature_names_out`.

    Non-regression test for:
    https://github.com/skrub-data/skrub/issues/1256
    https://github.com/skrub-data/skrub/issues/1540
    """
```

glemaitre (approved):
It looks a similar fix than in https://github.com/skrub-data/skrub/issues/1256 so LGTM.

PR #1546, opened 2025-07-24
Deprecate ken embeddings

### LINKED ISSUE #1516: Deprecate Ken Embeddings
Should we remove the Ken Embeddings example and deprecate them entirely? They are not usable in practice, and we wouldn’t recommend that users try them. Plus, downloading them breaks the CI at times. WDYT?

Vincent-Maladiere:
Fixes #1516


rcap107:
the end of an era

rcap107 (approved):

PR #1552, opened 2025-07-25
ApplyToFrame supports Sparse output

### LINKED ISSUE #1513: skb.apply does not support some scikit-learn estimators
### Describe the bug

The `skb.apply` and thus expressions (data plan, data ops) do not allow to plug-in standard scikit learn vectorizers such as `CountVectorizer`, `HashingVectorizer`, `TfidfVectorizer` etc. because they output a sparse matrix while expressions expect a dataframe. 

There is currently no way to use `set_output(transform="pandas")` on these transformers and I don't think it will be possible anytime soon in scikit-learn (https://github.com/scikit-learn/scikit-learn/discussions/22377).

### Steps/Code to Reproduce

```
import skrub
df = skrub.toy_orders().X
x = skrub.var("x", df)

# Works:
x.skb.apply(skrub.MinHashEncoder(), cols=["product"])

# Works (numeric transformer)
from sklearn.preprocessing import StandardScaler
x.skb.apply(StandardScaler(), cols=["quantity"])

# Works (on dataframe)
from sklearn.feature_extraction.text import HashingVectorizer
enc = HashingVectorizer()
enc.fit_transform(df["product"])

# Doesn't work (inside expressions)
x.skb.apply(HashingVectorizer(), cols=["product"])
```

### Expected Results

Support for `CountVectorizer`, `HashingVectorizer`, `TfidfTransformer` so we can use them in expressions.

### Actual Results

```
TypeError: HashingVectorizer.fit_transform returned a result of type csr_matrix, but a pandas DataFrame was expected. If HashingVectorizer is a custom transformer class, please make sure that the output is a pandas container when the input is a pandas container. One way of enabling a transformer to output pandas DataFrames is inheriting from the sklearn.base.TransformerMixin class and defining the 'get_feature_names_out' method. See https://scikit-learn.org/stable/auto_examples/miscellaneous/plot_set_output.html for details.
```

### Versions

```shell
Latest version (0.6.dev0)
```

Vincent-Maladiere:
Fixes https://github.com/skrub-data/skrub/issues/1513

### DIFF
--- a/skrub/_apply_to_frame.py
+++ b/skrub/_apply_to_frame.py
@@ -1,3 +1,6 @@
+import warnings
+
+from scipy import sparse as sp
 from sklearn.base import BaseEstimator, TransformerMixin, clone
 from sklearn.utils.validation import check_is_fitted
 
@@ -173,6 +176,11 @@ class ApplyToFrame(TransformerMixin, BaseEstimator):
             The transformed data.
         """
         self.all_inputs_ = sbd.column_names(X)
+
+        # for sklearn
+        self.feature_names_in_ = self.all_inputs_
+        self.n_features_in_ = len(self.all_inputs_)
+
         self._columns = selectors.make_selector(self.cols).expand(X)
         to_transform = selectors.select(X, self._columns)
         if self.keep_original:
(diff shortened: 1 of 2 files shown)


skrub/_apply_to_frame.py
```diff
@@ -184,6 +192,18 @@ def fit_transform(self, X, y=None):
...
+            if sp.issparse(transformed):
+                warnings.warn(
+                    f"The output of {type(self.transformer_).__name__!r} is a sparse"
+                    " array or sparse matrix."
+                )
+                if sbd.column_names(to_transform) != sbd.column_names(X):
+                    raise ValueError(
+                        "When a transformer outputs a sparse array or sparse matrix,"
+                        " all columns of the input dataframe must beselected. Got"
+                        f" cols={self.cols!r} instead of skrub.selectors.all()."
```
vitorpohlenz, line 204: Hi Vincent, I have a naive question/suggestion about this `raise`, would it be a good idea to put a brief observation in the docstring of `cols` arg to use all columns for sparse objects (despite that it is the default behaviour) ?

jeromedockes:
it is more complicated than that because due to set_output scikit-learn''s TransformerMixin will also raise an error when trying to convert a sparse matrix to a dataframe. as it is small and has conflicts I think it is easier to open a new PR if we want to revisit this topic

Issue #1566, opened 2025-08-22
META-issue: deprecating features and bumping min requirements for 0.7.0


rcap107:
I'm opening this issue to track a few deprecations and breaking changes that we need to consider. 

Minimum requirements
- We need to bump scikit-learn requirements to 1.4 (currently it's 1.2)
- Python 3.9 stops being supported in October 2025, so we need to bump the minimum requirements to 3.10

Deprecations
- [x] KEN embeddings  #1546 
- [x] `DropIfTooManyNulls`
- [x] `tabular_learner` 
- [x] `patch_display` and `unpatch_display`
- [x] `specific_transformers` parameter in `TableVectorizer`
- [x] Values for `.skb.apply(how=...)` #1628 

These are most of the changes that I have found so far.

rcap107:
Adding #1628

rcap107:
Completed.
`patch_display` and `unpatch_display` were un-deprecated, and `specific_transformers` will be kept for the time being

PR #1567, opened 2025-08-22
MAINT - Removing some deprecated features

### LINKED ISSUE #900: Failing fetch ken embeddings test

rcap107:
Addressing part of #1566


rcap107:
Ready for review, to be merged in 0.7.0

skrub/_reporting/_patching.py
```diff
@@ -1,10 +1,7 @@
 import importlib
-import warnings
```
jeromedockes, line 2: Thanks @rcap107 ! Actually some discussion started about using the config to patch the display:

https://github.com/skrub-data/skrub/issues/1523#issuecomment-3460873988

I'm not so sure this should be handled with the config. So maybe we could keep the rest of the PR but hold off on removing patch_display until that has been settled? (or maybe even un-deprecate it 🤔 )

rcap107: We can do that, and deal with patch_display later

skrub/conftest.py
```diff
@@ -242,3 +244,20 @@ def use_fit_transform(request):
 xfail_with_download_error = pytest.mark.xfail(
     raises=OSError, match="Can't download the file '.*' from urls.*"
 )
+
+BASE_CONFIG = get_config()
+
+
+@pytest.fixture(autouse=True)
+def _reset_config_to_base():
```
jeromedockes:
```suggestion
def reset_config_to_base():
```

nitpick: as it's meant for other modules to (auto)use

skrub/conftest.py
```diff
@@ -242,3 +244,20 @@ def use_fit_transform(request):
...
+    """Autouse fixture that resets config to base_config before each test.
+
+    This ensures that tests run in isolation and don't affect each other's
+    configuration state, even when running in parallel or in different orders.
+    This prevents race conditions where one test's config changes could leak
```
jeromedockes: the config is stored in thread local data so I'm not sure 'in parallel' or 'race conditions' are really the issue here, but rather that as you say above we want to make sure each test cleans up after itself does not alter the program's global state, possibly affecting subsequent tests

jeromedockes (approved):
Thanks @rcap107 !

PR #1578, opened 2025-08-27
DOC - adding a guide on how to write examples 

rcap107:


doc/tutorial_example.rst
```diff
@@ -0,0 +1,215 @@
+.. _tutorial_write_example:
+
+How to write an example for the skrub gallery
```
jovan-stojanovic: Long for a section name, maybe just "Writing an example"?

doc/tutorial_example.rst
```diff
@@ -0,0 +1,215 @@
...
+- If your example uses the same objects multiple times, you can define a replacement at the top
+  of the docstring using the ``.. replace::`` directive, and then use the replacement
+  instead of the full directive.
+- You can also add references to other sections of the documentation using the
+  ``:ref:`label``` directive, where ``label`` is the label of the section you want to reference.
+
+For example, if your example uses the ``TableVectorizer`` class multiple times, you can
+define a replacement at the top of the docstring. Then, you might want to add a
+reference to the user guide section about the ``TableVectorizer`` class. This can be
+done as follows:
```
jovan-stojanovic: Maybe it's a good place to use a reference to the TableVectorizer function in the documentation itself :)

jovan-stojanovic (commented):
Super cool PR.

A general comment: 
in the "Development" section of the website, there are subsections including the one you want to add and "Contributing to skrub". "Contributing to skrub" is then also a section in the drop-down of "More". We might want to add an issue to clear this up.

doc/tutorial_example.rst
rcap107: this is a fully separate page so I think that the more expressive title makes sense here

doc/tutorial_example.rst
rcap107: good idea

jovan-stojanovic (approved):

PR #1572, opened 2025-08-25
MAINT - Bumping minimum python and sklearn requirements

rcap107:
As discussed in #1566, we need to increase the minimum requirements: python to 3.10 and sklearn to 1.4.2

PR #1580, opened 2025-08-28
Adding robots.txt and sitemap to the doc website

rcap107:
With this, search engines should not index the dev pages 

I updated the pixi.lock file because I needed to add the sphinx_sitemap extension


pyproject.toml
```diff
@@ -137,6 +138,7 @@ platforms = ["linux-64", "osx-arm64", "win-64"]
 pytorch = { version = ">=2.4.1,<3" }
 transformers = "*"
 sentence-transformers = ">=3.0,<5"
+pyarrow = "<21"
```
jovan-stojanovic: I see this is related to #1572, is it necessary to be included in this PR?

rcap107: yes, #1572  is not going to be merged soon and without the pin tests break

GaelVaroquaux (approved):
LGTM merging

PR #1581, opened 2025-08-29
DOC - Cleaning pass on all documents in the doc

rcap107:
The main change is in the 00_getting_started example (the introduction to skrub), which I rewrote almost entirely.

Everything else is a collection of small fixes, typos, wrong formatting and the like


GaelVaroquaux (approved):
LGTM, merging. Thanks!

PR #1588, opened 2025-09-06
Update our development status

GaelVaroquaux:
I noticed that we still broadcast a beta status: https://pypi.org/project/skrub/

I think that we should change this. Here is a PR doing it.


MarieSacksick:
wouhou!

PR #1594, opened 2025-09-08
FIX - .skb.concat fails when it is passed a dataframe with non-strings as column names

### LINKED ISSUE #1593: Trouble in skrub._dataframe.concat()
### Describe the bug

Sorry for the blurry title, I don't really know what's not going on correctly.  
I know it's a private module, so it may change and not be complete, but my use-case doesn't seem so weird? (cf reproducible piece of code below).  
Is it because the column name by default of `to_frame()` is not a string?

### Steps/Code to Reproduce

```python
import pandas as pd
import numpy as np

X = pd.DataFrame(
            np.random.rand(100, 5), columns=[f"Feature number {i}" for i in range(5)]
        )

y = pd.Series(np.ones(100))
y = y.to_frame()

df = sbd.concat(X, y, axis=1)
```

### Expected Results

No error is thrown, and the concatenation goes through.  

With this, it's ok:  
```python
y_bis = np.ones(100)
y_bis = pd.DataFrame(y_bis, columns=["Target"])

df = sbd.concat(X, y_bis, axis=1)
```

### Actual Results

```pytb
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/skrub/_dataframe/_common.py", line 385, in _concat_pandas
    _check_same_type(dataframes)
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/skrub/_dataframe/_common.py", line 373, in _check_same_type
    raise TypeError(msg)
TypeError: Mixing types is not allowed, got dataframes at position [0], series at position [1].
>>> y = y.to_frame()
>>> df = sbd.concat(X, y, axis=1)
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/skrub/_dataframe/_common.py", line 392, in _concat_pandas
    dataframes = _join_utils.make_column_names_unique(*dataframes)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/skrub/_join_utils.py", line 189, in make_column_names_unique
    new_names = pick_column_names(sbd.column_names(df), forbidden_names=used)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/skrub/_join_utils.py", line 153, in pick_column_names
    new_name = _get_new_name(name, forbidden_names)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/skrub/_join_utils.py", line 161, in _get_new_name
    tags = re.findall(tag_pattern, suggested_name)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/marie/anaconda3/envs/skore_test/lib/python3.12/re/__init__.py", line 217, in findall
    return _compile(pattern, flags).findall(string)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: cannot use a string pattern on a bytes-like object
```

### Versions

```shell
using skrub 0.6.0

System:
    python: 3.12.8 | packaged by Anaconda, Inc. | (main, Dec 11 2024, 16:31:09) [GCC 11.2.0]
executable: /home/marie/anaconda3/envs/skore_test/bin/python
   machine: Linux-6.8.0-65-generic-x86_64-with-glibc2.35

Python dependencies:
      sklearn: 1.6.1
          pip: 24.2
   setuptools: 75.1.0
        numpy: 2.2.0
        scipy: 1.14.1
       Cython: None
       pandas: 2.2.3
   matplotlib: 3.9.3
       joblib: 1.4.2
threadpoolctl: 3.5.0

Built with OpenMP: True

threadpoolctl info:
       user_api: blas
   internal_api: openblas
    num_threads: 12
         prefix: libscipy_openblas
       filepath: /home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/numpy.libs/libscipy_openblas64_-6bb31eeb.so
        version: 0.3.28
threading_layer: pthreads
   architecture: Haswell

       user_api: blas
   internal_api: openblas
    num_threads: 12
         prefix: libscipy_openblas
       filepath: /home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/scipy.libs/libscipy_openblas-c128ec02.so
        version: 0.3.27.dev
threading_layer: pthreads
   architecture: Haswell

       user_api: openmp
   internal_api: openmp
    num_threads: 12
         prefix: libgomp
       filepath: /home/marie/anaconda3/envs/skore_test/lib/python3.12/site-packages/scikit_learn.libs/libgomp-a34b3233.so.1.0.0
        version: None
```

rcap107:
Fixes #1593 for data ops concat

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -24,6 +24,8 @@ Bugfixes
   when applied to categorical columns. :pr:`1570` by :user:`Riccardo Cappuzzo<rcap107>`.
 - Fixed the display of DataOp objects in google colab cell outputs (no output
   was displayed). :pr:`1590` by :user:`Jérôme Dockès <jeromedockes>`.
+- Fixed an error that occurred when using ``.skb.concat`` with a pandas dataframe
+  with column names that aren't strings. :pr:`1594` by :user:`Riccardo Cappuzzo<rcap107>`.
 
 Release 0.6.1
 ===================
--- a/skrub/_data_ops/_data_ops.py
+++ b/skrub/_data_ops/_data_ops.py
@@ -1699,7 +1699,15 @@ class Concat(DataOpImpl):
                 f"Invalid axis value {e.axis!r} for concat. Expected one of 0 or 1."
             )
 
-        result = sbd.concat(e.first, *e.others, axis=e.axis)
+        if e.axis == 1:
+            first = _check_column_names(e.first)
+            others = list(map(_check_column_names, e.others))
+        else:
+            # No need to sanitize column names if concatenating vertically.
+            first = e.first
+            others = e.others
+
+        result = sbd.concat(first, *others, axis=e.axis)
 
         if e.axis == 1:
             if mode == "preview" or "fit" in mode:
--- a/skrub/_data_ops/tests/test_data_ops.py
+++ b/skrub/_data_ops/tests/test_data_ops.py
@@ -337,13 +337,11 @@ def test_concat_vertical_duplicate_cols():
     assert out_1.shape[0] == out_2.shape[0] == 4
 
 
-def test_concat_vertical_non_str_colname():
-    X = pd.DataFrame(
-        np.random.rand(100, 5), columns=[f"Feature number {i}" for i in range(5)]
-    )
+def test_concat_horizontal_non_str_colname():
+    X = pd.DataFrame({"a": [0, 1], "b": [2, 3]})
     X_op = skrub.X(X)
 
-    y = pd.Series(np.ones(100))
+    y = pd.Series([1, 1])
     y_no_name = (
         y.to_frame()
     )  # note that the name is left empty so pandas will auto-name it 0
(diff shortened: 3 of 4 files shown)


jeromedockes:
Thanks @rcap107!

- I think converting to strings on the fly with a warning might be better than raising. Note we are already possibly changing some column names in concat anyway to avoid duplicates. Also it is what we do in the `Cleaner` and `TableVectorizer` and in `.skb.apply()` [here](https://github.com/skrub-data/skrub/blob/4b4c16414c4669f8dc73e79987aa684159546d19/skrub/_data_ops/_data_ops.py#L1244).
- I think it simplifies things to enforce invariants like the "column names should be strings" closer to the boundary of skrub because
  - when we choose to raise, for users the exception is raised closer to their code making it easier to understand the traceback
  - when we choose to "fix" the input, it is better to do it early on so that all of the processing is working with the same, sanitized input
  - skrub internals can make that assumption without needing to check it every time

At the moment `Cleaner` (and thus `TableVectorizer` and `tabular_pipeline`), and `.skb.apply()` do this column name check & conversion near the start of their processing. `.skb.concat` already has similar checks to enforce that all its arguments are dataframes, so we could add the check there. (The `skb.apply` check also covers `.skb.select` and `.skb.drop` because they are implemented by applying `SelectCols` and `DropCols`.)

I therefore suggest adding the check to those already performed by Concat [here](https://github.com/skrub-data/skrub/blob/4b4c16414c4669f8dc73e79987aa684159546d19/skrub/_data_ops/_data_ops.py#L1672), something like:

```diff
diff --git a/skrub/_data_ops/_data_ops.py b/skrub/_data_ops/_data_ops.py
index a2cb85a1..b08b73f4 100644
--- a/skrub/_data_ops/_data_ops.py
+++ b/skrub/_data_ops/_data_ops.py
@@ -1699,7 +1699,15 @@ class Concat(DataOpImpl):
                 f"Invalid axis value {e.axis!r} for concat. Expected one of 0 or 1."
             )

-        result = sbd.concat(e.first, *e.others, axis=e.axis)
+        if e.axis == 1:
+            first = _check_column_names(e.first)
+            others = map(_check_column_names, e.others)
+        else:
+            # No need to sanitize column names if concatenating vertically.
+            first = e.first
+            others = e.others
+
+        result = sbd.concat(first, *others, axis=e.axis)

         if e.axis == 1:
             if mode == "preview" or "fit" in mode:
```

jeromedockes (approved):
thank you @rcap107 !

PR #1596, opened 2025-09-11
store & report dataop evaluation timing

jeromedockes:
This adds to the dataop's report the time it took to evaluate each node

ATM it is shown here in the `full_report`:

image

zoom:

image

I also took the opportunity to do a bit of refactoring & slightly improving comments in the run() function in which the timing is added

### DIFF
--- a/skrub/_data_ops/_estimator.py
+++ b/skrub/_data_ops/_estimator.py
@@ -141,8 +141,8 @@ class SkrubLearner(_CloudPickleDataOp, BaseEstimator):
         full_report_kwargs : dict
             See :meth:`DataOp.skb.full_report`
 
-        Result
-        ------
+        Returns
+        -------
         dict
             The result of ``DataOp.skb.full_report``: a dict containing
             ``'result'``, ``'error'`` and ``'report_path'``.


skrub/_reporting/_data/templates/data_ops/node.html
```diff
@@ -70,12 +70,14 @@ <h3>This step was used by:</h3>
...
-                <code class="node-creation-stack">
-{{ node_creation_stack_description }}
-                </code>
-            </pre>
+                    <pre><code class="node-creation-stack">{{ node_creation_stack_description }}</code></pre>
+                </div>
+                {% if eval_duration is not none %}
+                <div>
+                    <h3>Computation time</h3>
+                    {{ "{:.2g}".format(eval_duration) }} s
```
rcap107: I think minutes should also be added, I could imagine some nodes taking longer than 60 seconds to fit on larger data

rcap107:
This looks really good, thanks a lot @jeromedockes 

Maybe this should also be added to the param searches? though in that case it would just be the sum of all runs.

skrub/_reporting/_data/templates/data_ops/node.html
jeromedockes: ok I improved a bit the formatting:

image

jeromedockes:
> Maybe this should also be added to the param searches? though in that case it would just be the sum of all runs.

Maybe, yes. ATM we don't have those kind of detailed node-by-node reporting for the param searches, I guess what is interesting in param search is more aggregated information such as the mean fit time of the whole pipeline as shown in the `results_` or parallel coord plot, otherwise it would be a bit unwieldly. and then you might want to dive in and do a full report of the `best_learner_`

in any case I would suggest exploring that in another issue; now at least the information is available in the node

rcap107:
> in any case I would suggest exploring that in another issue; now at least the information is available in the node

Yes better do that

rcap107:
Can you add a mention of the evaluation time in the docstring for the `full_report`? Aside from that, the PR is good. 

We should have a proper section on the functions of the full report in the user guide as well (in a separate PR).

PR #1599, opened 2025-09-11
DOC - Add link to pixi in CONTRIBUTING.rst

star1327p:
Add link to pixi in CONTRIBUTING.rst
https://pixi.sh/latest/


rcap107:
Good addition, thanks @star1327p! Do you mind adding the same link in the section on building the documentation? I can imagine people missing the first link if they're only working on the docs.

I'm also wondering if we should have a full section on how to use pixi in this document 🤔 that'll be for a different PR though

Issue #1477, opened 2025-06-27
Let `TableReport` accept `skrub` expressions as input


ogrisel:
### Problem Description

At the moment, the following code snippet:

```python
import polars as pl
import skrub

skrub.TableReport(skrub.var("df", pl.DataFrame({"a": range(3)})))
```

raises `NotImplementedError` while the usual:

```python
skrub.TableReport(pl.DataFrame({"a": range(3)}))
```

works as expected which can be confusing.

### Feature Description

I agree that any skrub expression wrapping a dataframe naturally embeds the `TableReport` view as part of their HTML representation for Jupyter notebook when they end up in the output of a notebook cell.

However, I suspect that some users might want to use `TableReport` programmatically outside of any notebook and the above `NotImplementedError` exception can be confusing.

rcap107:
Writing
```python
skrub.TableReport(skrub.var("df", pl.DataFrame({"a": range(3)})).skb.eval())
```
returns the proper report. 

We could either change the error message to suggest the `.skb.eval()`, or maybe add a parameter to the TableReport that forces the evaluation if it is passed an expression. 

I'm leaning towards having a first PR that adds the error message, and later a PR to add full support

Vincent-Maladiere:
> We could either change the error message to suggest the .skb.eval()

I'm +1 for a clear error message

GaelVaroquaux:
>     We could either change the error message to suggest the .skb.eval()

> I'm +1 for a clear error message

I agree that I would prefer an error message than having the TableReport transparently accepting expressions.

The reason is that I think that it is important that users understand that they are manipulating different type of objects.

The counter argument could be that we should make it as simple as possible to replace normal df code by expression. I'm happy being convinced of this if other people (eg @ogrisel) think that this is the way to go.

ogrisel:
Improving the error message would already be helpful.

jeromedockes:
With the recently-merged #1416  the error is now

```
TypeError: Operation not supported on this object. Expecting a Pandas or Polars object, but got an object of type <class 'skrub._expressions._expressions.Expr'>.
```

which is already much better than NotImplementedError. Maybe the check introduced by #1416 should special-case expressions with a tailored message so that we get a good message whenever an expression reaches a "dispatched" function, which might happen in a few other cases

jeromedockes:
[...]

PR #1607, opened 2025-09-12
Adding a special case to _common so that Data Ops get a more informative message

rcap107:
I added a check to the dispatch module so that if it gets passed a Data Op a more specific error message is shown. 
This should address #1477 

With this, 

```python
import skrub
import pandas as pd

df = pd.DataFrame({"value": [12,3]})

v = skrub.var("data", df)

skrub.TableReport(v)
```
gets the message

```
TypeError: Operation not supported on DataOps. Expecting a Pandas or Polars object, but got a DataOp. You may want to use `.skb.eval()` to evaluate the DataOp and obtain a dataframe.
```


skrub/_dataframe/_common.py
```diff
@@ -119,6 +119,14 @@
 
 
 def _raise(obj, kind="object"):
+    from .._data_ops._data_ops import DataOp
+
+    if isinstance(obj, DataOp):
+        raise TypeError(
+            "Operation not supported on DataOps. Expecting a Pandas or Polars "
```
GaelVaroquaux:
```suggestion
            "This operation is not supported on DataOps. It Expects a Pandas or Polars "
```

skrub/_dataframe/_common.py
```diff
@@ -119,6 +119,14 @@
...
+            f"{kind}, but got a DataOp. You may want to use `.skb.eval()` to "
+            "evaluate the DataOp and obtain a dataframe."
```
GaelVaroquaux:
```suggestion
            f"{kind}, but got a DataOp. You may want to use `.skb.eval()` to "
            "evaluate the DataOp into a dataframe and run your operation on it."
```

skrub/_dataframe/tests/test_common.py
```diff
@@ -40,6 +41,12 @@ def test_not_implemented():
         params = [None] * n_params
         with pytest.raises(TypeError):
             func(*params)
+        dop = [skrub.var("a")] * n_params
+        with pytest.raises(
+            TypeError,
+            match=r"Operation not supported on DataOps.*Expecting a Pandas*",
```
GaelVaroquaux: Need to change this :)

skrub/_dataframe/_common.py
```diff
@@ -119,6 +119,18 @@
...
+    if isinstance(obj, DataOp):
+        raise TypeError(
+            "Expected a Pandas or Polars DataFrame, but got a skrub DataOp. "
+            "A function that expects an actual value cannot be applied directly"
+            " to a DataOp; you may want to (i) use op.skb.eval() or op.skb.preview()"
+            "to evaluate the dataop and turn it into an actual value or "
+            "(ii) use op.skb.apply_func or op.skb.apply to schedule the operation"
+            " for later execution (when the dataop is evaluated) "
+            "rather than computing it immediately."
+        )
```
jeromedockes:
```suggestion
        raise TypeError(
            """Expected a Pandas or Polars DataFrame, but got a skrub DataOp.
A function that expects an actual value cannot be applied directly to a DataOp;
you may want to (i) use op.skb.eval() or op.skb.preview() to evaluate the
dataop and turn it into an actual value or (ii) use op.skb.apply_func() or
op.skb.apply() to schedule the operation for later execution (when the dataop is
evaluated) rather than computing it immediately.
 """
        )
```

thanks @rcap107 . I would suggest actually wrapping the explanation with hard line breaks as shown in this comment, because some environments that display error messages in the browser do not wrap line but make the output scrollable instead, so most of the message might be hidden to some users if it is just one really long line.

as we discussed later we can make the error message much shorter by including a link to a short section in the doc dedicated to explaining this issue and the fix

jeromedockes: not related to your PR, but the name `_raise` is way too vague for a function that is used in many places as this one is. also the code that calls it is written `raise _raise(...)` but actually the function itself raises an exception so it should just be `_raise()` (or rather something clearer like `_raise_dispatch_unregistered_type()`

but that is for another PR

jeromedockes (approved):
thank you @rcap107 !

PR #1610, opened 2025-09-14
avoid extra predict in dataops fit

jeromedockes:
Currently, calling 'fit()' on a learner calls 'predict()' on the final estimator (if any) on the train set. getting this prediction is only useful for previews, not for fitting the learner.

with this PR, this extra predict() does not happen anymore.


examples/data_ops/11_data_ops_intro.py
```diff
@@ -1,5 +1,4 @@
 """
-
```
jeromedockes, line 2: the whitespace changes to examples are just to trigger their build in CI

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -268,11 +268,7 @@ def apply(
         <Apply DummyClassifier>
         Result:
         ―――――――
-           delayed
-        0    False
-        1    False
-        2    False
-        3    False
+        array([False, False, False, False])
```
jeromedockes, line 271: we now see directly the output of predict() not transformed into a dataframe

skrub/_data_ops/_data_ops.py
```diff
@@ -769,7 +770,7 @@ def _check(reason):
     if how == "full_frame":
         _check("`how` is 'full_frame'")
         return estimator
-    if not hasattr(estimator, "transform"):
+    if hasattr(estimator, "predict") or not hasattr(estimator, "transform"):
         _check("`estimator` is a predictor (not a transformer)")
```
rcap107, line 774: I'm seeing here that _check is then calling `_check_wrap_params`, but I don't understand how the fact it's a predictor or a transformer affects that part

jeromedockes: if it's a predictor the estimator should never be wrapped in applytocols or applytoframe: it is always applied directly to the whole data. so in this case we check that cols is s.all() and how is "full_frame" or "auto", then return the original, unmodified estimator

PR #1615, opened 2025-09-17
MAINT - adding explicit linkcheck command

rcap107:
The linkcheck function is useful for testing links that may be broken, but calling make linkcheck runs all the examples which takes forever for no good reason, so I added a noplot version to both the makefile and pyproject


jeromedockes (approved):
super useful, thanks!

there seem to be a couple of broken links indeed :/

should this be mentioned in the developer docs to help people know it's there?

rcap107:
> super useful, thanks!
> 
> there seem to be a couple of broken links indeed :/
> 
> should this be mentioned in the developer docs to help people know it's there?

I'll add a note on that and merge

PR #1617, opened 2025-09-17
use dataop name in Call __repr__ if available

jeromedockes:
tiny tweak to the `__repr__` of Call nodes: when they have a name, always use it as <Call "the name">

in this (nonsensical) example

```
>>> import skrub

>>> add = skrub.deferred(lambda x:lambda y:x+y)
>>> a, b = skrub.var('a'), skrub.var('b')
>>> add(a).skb.set_name('add a')(b)
<Call 'add a'>
```

in the main branch the output would be `<Call 'Call'>` which is not as nice


rcap107 (approved):
Looks goofd to me, thanks @jeromedockes

Issue #1441, opened 2025-06-13
TableVectorizer tests only check pandas dataframes


rcap107:
Tests in `test_table_vectorizer.py` are all using example pandas only dataframes, rather than using `df_module` and testing pandas, pandas nullable types, and polars

We should update all the tests to address this

rcap107:
from the skrub meeting: this is low priority and something to do when we have spare time

shivanshutiwari35:
Hi @rcap107, since this issue was discussed to be a low priority. If it is alright, I can take it up. With this, I will also be able to get more familiar with the codebase. 

Let me know if it's okay for me to proceed, or if there is any other instruction for me.

rcap107:
Hi @shivanshutiwari35, thanks for offering your help! Yes, you're more than welcome to work on this issue. 

You might want to work on this a few tests at a time (at least at the beginning), there are a lot of tests to go through and it might be easier to review and go over a few at a time rather than covering everything.

shivanshutiwari35:
Thanks @rcap107 ,  I’ll start with a small batch of tests and open a PR for review. Based on the feedback, I’ll continue iterating in parts to keep it manageable.

Vincent-Maladiere:
Hey @shivanshutiwari35, thank you for your interest! If you haven't already, I recommend that you read the [contributing documentation](https://skrub-data.org/dev/CONTRIBUTING.html) to get familiar with our workflow.

shivanshutiwari35:
Thanks, @rcap107, I’ll go through the contributing guide thoroughly and get started.

rcap107:
Closed by #1618

PR #1618, opened 2025-09-18
1441 table vect tests

gabrielapgomezji:
Modifying tests from table vectorizer to add df_module.
Changed a part, but there are others that might need a bit more change as test_handle_unknown_category

### DIFF
--- a/skrub/tests/test_table_vectorizer.py
+++ b/skrub/tests/test_table_vectorizer.py
@@ -410,15 +410,6 @@ def test_convert_float32(df_module):
     Test that the TableVectorizer converts float64 to float32
     when using the default parameters.
     """
-    # data = {
-    #     "int": [15, 56, 63, 12, 44],
-    #     "float": [5.2, 2.4, 6.2, 10.45, 9.0],
-    #     "str1": ["public", "private", "private", "private", "public"],
-    #     "str2": ["officer", "manager", "lawyer", "chef", "teacher"],
-    #     "cat1": ["yes", "yes", "no", "yes", "no"],
-    #     "cat2": ["20K+", "40K+", "60K+", "30K+", "50K+"],
-    # }
-    # X = df_module.make_dataframe(data)
     X = _get_clean_dataframe()
     vectorizer = TableVectorizer()
     out = vectorizer.fit_transform(X)
--- a/skrub/tests/test_table_vectorizer.py
+++ b/skrub/tests/test_table_vectorizer.py
@@ -193,8 +193,16 @@ def _get_missing_values_dataframe(categorical_dtype="object"):
     )
 
 
-def test_get_preprocessors():
-    X = _get_clean_dataframe()
+def test_get_preprocessors(df_module):
+    data = {
+        "int": [15, 56, 63, 12, 44],
+        "float": [5.2, 2.4, 6.2, 10.45, 9.0],
+        "str1": ["public", "private", "private", "private", "public"],
+        "str2": ["officer", "manager", "lawyer", "chef", "teacher"],
+        "cat1": ["yes", "yes", "no", "yes", "no"],
+        "cat2": ["20K+", "40K+", "60K+", "30K+", "50K+"],
+    }
+    X = df_module.make_dataframe(data)
     steps = _get_preprocessors(
         cols=X.columns,
         drop_null_fraction=1.0,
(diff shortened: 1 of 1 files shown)


rcap107:
Addressing #1441

skrub/tests/test_table_vectorizer.py
```diff
@@ -193,8 +193,16 @@ def _get_missing_values_dataframe(categorical_dtype="object"):
     )
 
 
-def test_get_preprocessors():
-    X = _get_clean_dataframe()
+def test_get_preprocessors(df_module):
```
gabrielapgomezji, line 212: For the moment I am changing the dataframe inside the test but later I will update the _get_clean_dataframe function

skrub/tests/test_table_vectorizer.py
```diff
@@ -577,16 +636,18 @@ def test_handle_unknown_category():
...
 
     # +2 for binary columns which get one category dropped
-    n_zeroes = X["str2"].nunique() + X["cat2"].nunique() + 2
+    n_zeroes = sbd.n_unique(X["str2"]) + sbd.n_unique(X["cat2"]) + 2
+    
+    colnames = sbd.column_names(X_trans_unknown)[2:n_zeroes]
     assert_array_equal(
-        X_trans_unknown.iloc[0, 2:n_zeroes],
-        np.zeros_like(X_trans_unknown.iloc[0, 2:n_zeroes]),
+        sbd.slice(X_trans_unknown[colnames], 0, 1).to_numpy().squeeze(),
```
gabrielapgomezji, line 629: Not sure if this is the best way of doing it

rcap107: I suggested this solution, but I am not sure either it's the cleanest way of doing it

rcap107:
This PR was complicated by the fact that some of the tests are checking for pandas-specific behaviors and conditions. 

Some tests must be skipped with min-optional-reqs because polars support was added in scikit-learn 1.4, so in some specific cases tests could not pass when run using polars. 

The main changes revolved around modifying the functions `_get_clean_dataframe` etc so that they'd use `df_module`, rather than returning directly a pandas dataframe. As a result, fixtures had to defined in a different way (see `X_tuple_fixture`). 

Some other changes had to be made to get around how polars and pandas have different indexing and dtype representations. 

At this point, most of the tests (those that are not being skipped, at least) fully support both pandas and polars. However, there are some tests that may need some refinement. 

Given that this isn't a user-facing issue, and in the interest of avoiding conflicts, I think we should merge this and leave further improvements to future PRs.

skrub/tests/test_table_vectorizer.py
```diff
@@ -307,121 +326,147 @@ def passthrough_vectorizer():
     )
 
 
-@pytest.mark.parametrize("X, dict_expected_types", X_tuples)
-def test_auto_cast(X, dict_expected_types):
+def test_auto_cast(X_tuples_fixture):
```
GaelVaroquaux: I would have found it markedly more readable to use "@pytest.mark.parametrize" (https://docs.pytest.org/en/stable/how-to/parametrize.html#parametrize-basics) instead of the fixture.

Is there a reason that you did not consider this option?

skrub/tests/test_table_vectorizer.py
```diff
@@ -495,79 +540,87 @@ def test_get_feature_names_out():
...
+@pytest.fixture
+def fit_transform_inputs(df_module):
+    """Generate test inputs for fit_transform equivalence test."""
+    return [
+        _get_clean_dataframe(df_module),
+        _get_dirty_dataframe(df_module, categorical_dtype="object"),
+        _get_dirty_dataframe(df_module, categorical_dtype="category"),
+        _get_mixed_types_dataframe(df_module),
+        _get_mixed_types_array(),
+    ]
```
GaelVaroquaux: Same thing, I would much rather prefer a "@pytest.mark.parametrize", which is more explicit, than the fixture

rcap107: Is this what you'd expect with parametrize? I find this far less readable than the fixture to be honest.

```python
@pytest.mark.parametrize(
    "X",
    [
        pytest.param(_get_clean_dataframe, id="clean_dataframe"),
        pytest.param(lambda df_module: _get_dirty_dataframe(df_module, categorical_dtype="object"), id="dirty_dataframe_object"),
        pytest.param(lambda df_module: _get_dirty_dataframe(df_module, categorical_dtype="category"), id="dirty_dataframe_category"),
        pytest.param(_get_mixed_types_dataframe, id="mixed_types_dataframe"),
        pytest.param(lambda _: _get_mixed_types_array(), id="mixed_types_array"),
    ],
)
```
Unless there is a way of accessing df_module without the lambda functions, I think that having `parametrize` makes harder to understand

skrub/tests/test_table_vectorizer.py
rcap107: I don't think `parametrize` plays well with the fact we need to pass `df_module` to the functions

skrub/tests/test_table_vectorizer.py
GaelVaroquaux: Yes, this is what I was expecting.

I find this more readable because it introduce less conceptual things to understand and more locality in the code: the values are defined explicitly close where they are used. The fixture defines things very implicitly and far.

GaelVaroquaux (approved):
LGTM

Thanks a lot @gabrielapgomezji and @rcap107 !!

PR #1623, opened 2025-09-19
show transformed data not dataop itself in fit report

jeromedockes:
The `learner.report(env, mode="fit")` currently shows for each node the dataop itself, rather than the intermediate result (eg as fit_transform would)

for example

```python
from sklearn.dummy import DummyRegressor
import skrub
from skrub import datasets

data = skrub.var("data")
X = data.skb.drop("current_annual_salary").skb.mark_as_X()
y = data["current_annual_salary"].skb.mark_as_y()
hgb = X.skb.apply(skrub.TableVectorizer()).skb.apply(
    HistGradientBoostingRegressor(), y=y
)
dummy = X.skb.apply(DummyRegressor(), y=y)
learner = (
    skrub.choose_from({"hgb": hgb, "dummy": dummy}, name="model")
    .as_data_op()
    .skb.make_learner()
)

df = datasets.fetch_employee_salaries().employee_salaries
learner.report(environment={"data": df}, mode="fit")
learner.report(environment={"data": df}, mode="fit_transform")
```

# For the first report ("fit") in the main branch the (wrong) displayed output is the display of the intermediate data op itself:

image

# In the fixed report it is the actual output of this step:

image

# (note the output of the final supervised estimator in "fit" mode is the estimator itself -- but not the dataop)

image


rcap107 (approved):
LGTM, thanks @jeromedockes

PR #1627, opened 2025-09-22
DOC - Improving getting started example

rcap107:
- Removed section on datasets
- Added replace directives
- Split exploration and cleaning sections

### DIFF
--- a/examples/00_getting_started.py
+++ b/examples/00_getting_started.py
@@ -43,8 +43,11 @@ from skrub import TableReport
 TableReport(employees_df)
 
 # %%
-#
 # You can use the interactive display above to explore the dataset visually.
+#
+# It is also possible to tell skrub to replace the default pandas & polars
+# displays with |TableReport| by modifying the global config with
+# |set_config|.
 #
 # .. note::
 #
@@ -73,24 +76,6 @@ employees_df = Cleaner().fit_transform(employees_df)
 TableReport(employees_df)
 
 # %%
-# It is also possible to tell skrub to replace the default pandas & polars
-# displays with |TableReport| by modifying the global config with
-# |set_config|.
-
-from skrub import set_config
-
-set_config(use_table_report=True)
-
-employees_df
-
-# %%
-# This setting can easily be reverted:
-
-set_config(use_table_report=False)
-
-employees_df
-
-# %%
 # Easily building a strong baseline for tabular machine learning
 # --------------------------------------------------------------
 #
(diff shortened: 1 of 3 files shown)


examples/00_getting_started.py
```diff
@@ -80,11 +55,27 @@
 #
 #    .. _example reports: https://skrub-data.org/skrub-reports/examples/
 #    .. _demo: https://skrub-data.org/skrub-reports/
+#
+# From the Report above, we can see that there are datetime columns, so we use the
+# |Cleaner| to parse them.
+
+# %%
+# Sanitizing data with the |Cleaner|
```
jeromedockes, line 68: I find the interleaving a bit weird, we are describing the tablereport, then we start on the cleaner, then go back to the config for making the tablereport the default display, and later go back to the cleaner. I would group tablereport and then cleaner. 

Also I think we have too many tables displayed which makes the thing very noisy and adds a lot of scrolling. I think we can just mention the config (with a link) without a demo of displaying the same tablereport we already saw once more, plus the default pandas output

examples/00_getting_started.py
```diff
@@ -130,11 +121,11 @@
 # Assembling data
 # ---------------
 #
-# ``skrub`` allows imperfect assembly of data, such as joining dataframes
-# on columns that contain typos. ``skrub``'s joiners have ``fit`` and
+# Skrub allows imperfect assembly of data, such as joining dataframes
```
jeromedockes: I think we can cut the section on Joiner or reduce it and move it after the encoders

examples/00_getting_started.py
rcap107: I reordered the sections a bit, I wonder if we should keep the TableReport after the Cleaner to show that it does something :thinking:

examples/00_getting_started.py
```diff
@@ -2,75 +2,53 @@
...
+.. |TableVectorizer| replace:: :class:`~skrub.TableVectorizer`
+.. |Joiner| replace:: :class:`~skrub.Joiner`
+.. |SquashingScaler| replace:: :class:`~skrub.SquashingScaler`
+.. |DatetimeEncoder| replace:: :class:`~skrub.DatetimeEncoder`
+.. |ApplyToCols| replace:: :class:`~skrub.ApplyToCols`
+.. |StringEncoder| replace:: :class:`~skrub.StringEncoder`
+.. |TextEncoder| replace:: :class:`~skrub.TextEncoder`
 """
 
 # %%
```
glemaitre:
```suggestion
```

examples/00_getting_started.py
```diff
@@ -2,75 +2,53 @@
...
+from skrub import TableReport
 
-employees_df = Cleaner().fit_transform(employees_df)
 TableReport(employees_df)
 
 # %%
-#
 # You can use the interactive display above to explore the dataset visually.
 #
+# It is also possible to tell skrub to replace the default pandas & polars
```
glemaitre:
```suggestion
# It is also possible to tell skrub to replace the default pandas and polars
```

examples/00_getting_started.py
```diff
@@ -80,31 +58,29 @@
 #
 #    .. _example reports: https://skrub-data.org/skrub-reports/examples/
 #    .. _demo: https://skrub-data.org/skrub-reports/
+#
+# From the Report above, we can see that there are datetime columns, so we use the
+# |Cleaner| to parse them.
```
glemaitre:
```suggestion
# From the report above, we see that there are columns with date and time stored
# as `object` dtype (cf. "Stats" tab of the report). We therefore use the |Cleaner| to
# parse them and obtain a `datetime` dtype column. We see in the next section that this
# transformer does additional cleaning.
```

examples/00_getting_started.py
```diff
@@ -80,31 +58,29 @@
...
+# dataframe by parsing nulls and dates, and by dropping "uninformative" columns
+# (e.g., columns with too many nulls or that are constant).
+#
 
-set_config(use_table_report=False)
+from skrub import Cleaner
 
-employees_df
+employees_df = Cleaner().fit_transform(employees_df)
+TableReport(employees_df)
```
glemaitre, line 78: I feel that we need a sentence to conclude this section regarding the processing specifically for this transformation.

examples/00_getting_started.py
glemaitre: Maybe, we could simplify this text here because it is redundant with the next section. I would propose a broader statement than focusing on the datetime. For instance, we could state that after reading a table, it is common that the dtype are not inferred properly or contains unnecessary columns and that we use the `Cleaner` for that purpose.

And in the next section, we point out specifically to the datetime column.

examples/00_getting_started.py
```diff
@@ -119,82 +95,34 @@
```
glemaitre, line 103: Could we show the model repr:

```python
model = tabular_pipeline("regressor")
model

# %%
results = cross_validate(model, employees_df, salaries)
results["test_score"]
```

examples/00_getting_started.py
```diff
@@ -243,22 +171,40 @@
...
+# ``transform`` methods, storing information about the data across calls.
+#
+# The |Joiner| allows fuzzy-joining multiple tables, where each row of
+# a main table will be augmented with values from the best match in the auxiliary table.
+# You can control how distant fuzzy-matches are allowed to be with the
+# ``max_dist`` parameter.
+#
+# Skrub also allows you to aggregate multiple tables according to various strategies.
+# You
+# can see other ways to join multiple tables in :ref:`user_guide_joining_dataframes`.
```
glemaitre:
```suggestion
# You can see other ways to join multiple tables in
# :ref:`user_guide_joining_dataframes`.
```

Issue #1614, opened 2025-09-17
DOC - User guide section on dealing with sparse outputs


rcap107:
Even though skrub's tagline is "machine learning with _dataframes_", users may need to deal with sparse data depending on the transformers they use: see #1513 

It's still possible to use DataOps in this situation by setting `how="full_frame"` in `.skb.apply`, however this is not well documented in the docstrings and in the user guide. 

One side effect that should be mentioned in the user guide when covering this is that a script like the following  

```python
from sklearn.decomposition import PCA
from skrub.datasets import toy_orders
from sklearn.feature_extraction.text import HashingVectorizer
X = toy_order().X
import skrub
X = skrub.X(X)
X.skb.apply(HashingVectorizer(), how="full_frame")
X_csr = X.skb.apply(HashingVectorizer(), how="full_frame")
X_pca = X_csr.skb.apply(PCA())
```

the problem with this is that then concatenating X_pca with a dataframe fails
```python
X_pca.skb.concat([X])
```
and it becomes necessary to do `X_pca.skb.apply_func(pd.DataFrame)` to wrap the result back into something that can be concatenated.

The user guide needs to be updated accordingly (afterwards, we can link to that part of the guide in error messages). 

We might also want to change from "full_frame" to a different name ("no_wrap"?).

jeromedockes:
> We might also want to change from "full_frame" to a different name ("no_wrap"?).

and we can make the other options more similar to the corresponding class names:

```
X.skb.apply(transformer, how='cols') -> wrap transformer in ApplyToCols
X.skb.apply(transformer, how='frame') -> wrap transformer in ApplyToFrame
X.skb.apply(transformer, how='no_wrap') -> do not wrap transformer, apply directly to X
X.skb.apply(transformer, how='auto') -> decide by inspecting transformer (the default)
```

so the behavior would be the same but wrt to current names the renaming would be

```
columnwise -> cols
subframe -> frame
full_frame -> no_wrap
auto -> auto
```

WDYT?

rcap107:
> > We might also want to change from "full_frame" to a different name ("no_wrap"?).
> 
> and we can make the other options more similar to the corresponding class names:
> 
> ```
> X.skb.apply(transformer, how='cols') -> wrap transformer in ApplyToCols
> X.skb.apply(transformer, how='frame') -> wrap transformer in ApplyToFrame
> X.skb.apply(transformer, how='no_wrap') -> do not wrap transformer, apply directly to X
[...]

PR #1628, opened 2025-09-22
rename values for .skb.apply(how=...)

jeromedockes:
As discussed in [1614](https://github.com/skrub-data/skrub/issues/1614#issuecomment-3304425737) , this changes the values for the `how` parameter of `.skb.apply()` to match the names of the corresponding transformers. The options are now

```
'auto' -> let skrub decide, the default
'cols' -> ApplyToCols
'frame' -> ApplyToFrame
'no_wrap' -> no wrapping
```

the behavior is unchanged, only the values change. The old values 'auto', 'columnwise', 'sub_frame', 'full_frame' are kept with a FutureWarning

### DIFF
--- a/skrub/_data_ops/tests/test_data_ops.py
+++ b/skrub/_data_ops/tests/test_data_ops.py
@@ -12,6 +12,7 @@ from sklearn.linear_model import LogisticRegression
 from sklearn.model_selection import train_test_split
 
 import skrub
+from skrub import ApplyToCols
 from skrub import selectors as s
 from skrub._data_ops import _data_ops
 from skrub._utils import PassThrough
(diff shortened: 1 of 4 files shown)


skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -329,45 +330,64 @@ class A(_data_ops.DataOpImpl):
...
 
+def test_apply_invalid_how():
+    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+    X = skrub.var("X", df)
+    t = PassThrough()
+    for how in ["auto", "cols", "frame", "no_wrap"]:
+        assert list(X.skb.apply(t, how=how).skb.eval().columns) == ["a", "b"]
+    with pytest.raises(RuntimeError, match="`how` must be one of"):
+        X.skb.apply(t, how="bad value")
+    with pytest.warns(FutureWarning, match="'columnwise' has been renamed to 'cols'"):
```
rcap107, line 387:
```suggestion
    # TODO: remove when old names are dropped in 0.7.0
    with pytest.warns(FutureWarning, match="'columnwise' has been renamed to 'cols'"):
```

PR #1630, opened 2025-09-23
rename splitter -> split_func in DataOp.skb.train_test_split

jeromedockes:
"splitter" makes it sound like it expects a [cv splitter](https://scikit-learn.org/stable/glossary.html#term-cross-validation-splitter) , but what is expected here is a function with the same signature as scikit-learn's `train_test_split`

It might get even more confusing if we later add a method to iterate over the splits obtained from an actual splitter (as opposed to a single split)

the suggestion here is to rename the parameter to `split_func` and deprecate `splitter`


rcap107 (approved):
LGTM, thanks @jeromedockes

PR #1632, opened 2025-09-24
DOC - Adding a summary page on all the dataops features

rcap107:


rcap107:
From IRL conversation: this is mostly fine, but it should be summarized more, and links should be put into see also blocks.

rcap107:
Closing this as the docs have evolved a lot and we are thinking of overhauling the way data ops are presented anyway

PR #1634, opened 2025-09-25
MAINT refactoring get_feature_names_out to define it in encoders

rcap107:
Remove `get_feature_names_out` from the `SingleColumnTransformer` class and add it to the specialized encoders since they might specialized depending on their own parameters.


rcap107:
This is the first part of an issue on refactoring the `get_feature_names_out` function. Here I moved it out of the SingleColumnTransformer, later we might want to create a class that abstracts it. We can't use the SingleColumnTransformer for that because not all the encoders that inherit from this use the same `get_feature_names_out`

glemaitre (approved):
Just pushed a small fixed to make sure that `_SplineEncoder` follows the scikit-learn API (not overwrite attributes) and make sure that we cover the line that was not covered before.

PR #1642, opened 2025-10-02
discussion/wip allow passing kwargs in .skb.apply()

jeromedockes:
This allows passing extra named arguments to the methods of estimators passed to `.skb.apply()`. Still hesitating slightly about whether this is needed.

Also, we should agree on the exact interface: a single dict like `kwargs={"fit": {"sample_weight": sw}}`, several like `fit_kwargs={"sample_weight": sw}` (what is currently in this PR), other variations...

### DIFF
--- a/skrub/_apply_to_cols.py
+++ b/skrub/_apply_to_cols.py
@@ -73,8 +73,11 @@ class SingleColumnTransformer(BaseEstimator):
 
     __single_column_transformer__ = True
 
-    def fit(self, column, y=None):
+    def fit(self, column, y=None, **kwargs):
         """Fit the transformer.
+
+        This default implementation simply calls ``fit_transform()`` and
+        returns ``self``.
 
         Subclasses should implement ``fit_transform`` and ``transform``.
 
@@ -87,12 +90,15 @@ class SingleColumnTransformer(BaseEstimator):
         y : column or dataframe
             Prediction targets.
 
+        kwargs :
+            Extra named arguments are passed to ``self.fit_transform()``.
+
         Returns
         -------
         self
             The fitted transformer.
         """
-        self.fit_transform(column, y=y)
+        self.fit_transform(column, y=y, **kwargs)
         return self
 
     def _check_single_column(self, column, function_name):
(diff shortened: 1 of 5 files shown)


rcap107:
I feel like I don't particularly like any of the solutions we discussed with @jeromedockes, but at the same time not having this feature would be too limiting, since people use other libraries that need these kwargs (and scikit-learn has some estimators that need them too).

I think this version with fit_kwargs and predict_kwargs is the better solution.

jeromedockes:
example passing sample weights (computed from X) to a ridge regressor:

```python
import polars as pl

import skrub
import skrub.datasets
import skrub.selectors as s

dataset = pl.from_pandas(skrub.datasets.fetch_employee_salaries().employee_salaries)
# skrub.TableReport(dataset).open()
data = skrub.var("data")
y_col = "current_annual_salary"
X = data.skb.select(s.all() - y_col).skb.mark_as_X()
y = data[y_col].skb.mark_as_y()


def compute_weights(X):
    return (-(((X["year_first_hired"] - 2025) / 20) ** 2)).exp().rename("weight")


weight = X.skb.apply_func(compute_weights)
pred = X.skb.apply(skrub.TableVectorizer()).skb.apply(
    Ridge(), y=y, fit_kwargs={"sample_weight": weight}
)
```
graph

rcap107 (commented):
> example passing sample weights (computed from X) to a ridge regressor:
> 
> ```python
> import polars as pl
> from sklearn.linear_model import Ridge
> 
> import skrub
> import skrub.datasets
> import skrub.selectors as s
> 
> dataset = pl.from_pandas(skrub.datasets.fetch_employee_salaries().employee_salaries)
> # skrub.TableReport(dataset).open()
> data = skrub.var("data")
> y_col = "current_annual_salary"
> X = data.skb.select(s.all() - y_col).skb.mark_as_X()
> y = data[y_col].skb.mark_as_y()
> 
> 
> def compute_weights(X):
>     return (-(((X["year_first_hired"] - 2025) / 20) ** 2)).exp().rename("weight")
> 
> 
> weight = X.skb.apply_func(compute_weights)
> pred = X.skb.apply(skrub.TableVectorizer()).skb.apply(
>     Ridge(), y=y, fit_kwargs={"sample_weight": weight}
> )
> ```
> graph
> 
>  

Could you add an example like this to the docstring? Maybe we could also add a mention to xgboost since it's likely what a lot of people will use this for.

Aside from that, I think it's good

PR #1645, opened 2025-10-02
Fix pyarrow missing module when using ToDatetime with polars dataframe

MarieSacksick:
Bugfixing the pyarrow missing import.  
tested in local.  
Will try to do another PR to add a CI about it.

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -277,6 +277,7 @@ def _to_pandas_polars_column(obj):
     try:
         return obj.to_pandas()
     except ImportError:
+        # pyarrow is needed for polars .to_pandas() and may not be installed
         return pd.Series(to_numpy(obj))
 
 


skrub/_dataframe/_common.py
```diff
@@ -265,11 +265,21 @@ def _to_pandas_pandas(obj):
...
 
 
-@to_pandas.specialize("polars")
+@to_pandas.specialize("polars", argument_type="DataFrame")
 def _to_pandas_polars(obj):
     return obj.to_pandas()
 
 
+@to_pandas.specialize("polars", argument_type="Column")
+def _to_pandas_polars(obj):
```
jeromedockes: it works like this but we adopted the convention to give unique names to each of the functions we define in this module so I suggest this change
```suggestion
def _to_pandas_polars_column(obj):
```

jeromedockes (commented):
thanks a lot @MarieSacksick !

apart from the small nitpick LGTM :)

skrub/_dataframe/_common.py
```diff
@@ -265,11 +265,21 @@ def _to_pandas_pandas(obj):
...
     return obj.to_pandas()
 
 
+@to_pandas.specialize("polars", argument_type="Column")
+def _to_pandas_polars_column(obj):
+    """Convert a polars Series to a pandas Series, through numpy when pyarrow is not
+    installed."""
+    try:
+        return obj.to_pandas()
+    except ImportError:
```
jeromedockes, line 279:
```suggestion
    except ImportError:
        # pyarrow is needed for polars .to_pandas() and may not be installed
```

jeromedockes (approved):

jeromedockes:
Thank you very much @MarieSacksick !! the coverage missed line is expected and will be tackled in #1648 

LGTM

Issue #1655, opened 2025-10-02
Bug during the tests when downloading the data


MarieSacksick:
### Describe the bug

In a PR, I received this error: https://github.com/skrub-data/skrub/actions/runs/18197615376/job/51807943917?pr=1654  

```
FAILED skrub/tests/test_config.py::test_use_table_report_data_ops - OSError: [Errno 66] Directory not empty: '/Users/runner/skrub_data/employee_salaries/tmpetbgh3p8/employee_salaries' -> '/Users/runner/skrub_data/employee_salaries/employee_salaries'
```

According to @jeromedockes, the error comes from here: https://github.com/skrub-data/skrub/blob/dd0b01065bbbdee3276711566f952f35a5c63abe/skrub/datasets/_utils.py#L267

### Steps/Code to Reproduce

That's the problem: we can't always reproduce, because it's due to the parallelization of tests and the fact that they try to download the data simultaneously.  

### Expected Results

The tests shouldn't be affected by that.  

### Actual Results

The error above.  

### Versions

```shell
N/A
```

MarieSacksick:
To me there are two elements to act upon: make the _extract_archive safe to //, and also, try to avoid as much as possible to download datasets in tests.

MarieSacksick:
Also having this issue https://github.com/skrub-data/skrub/actions/runs/18198643176/job/51811568325?pr=1656:  
```
FAILED skrub/_reporting/tests/test_table_report.py::test_error_input_type -   
OSError: [Errno 39] Directory not empty:   
'/home/runner/skrub_data/employee_salaries/tmpl5m634bk/employee_salaries' ->   
 '/home/runner/skrub_data/employee_salaries/employee_salaries'
```

rcap107:
This issue should have been fixed

PR #1654, opened 2025-10-02
Feat: Add title to report

MarieSacksick:
Closes https://github.com/skrub-data/skrub/issues/1649

PR #1656, opened 2025-10-02
chore: alleviate datasets downloads in test

MarieSacksick:
To start the issue https://github.com/skrub-data/skrub/issues/1655, and to unlock the PR https://github.com/skrub-data/skrub/pull/1654

### DIFF
--- a/skrub/tests/test_config.py
+++ b/skrub/tests/test_config.py
@@ -44,10 +44,11 @@ def test_config_context():
 
 
 def test_use_table_report_data_ops(simple_df):
+    X = skrub.X(simple_df)
     with config_context(use_table_report_data_ops=True):
-        assert _use_table_report(simple_df)
+        assert _use_table_report(X)
         with config_context(use_table_report_data_ops=False):
-            assert not _use_table_report(simple_df)
+            assert not _use_table_report(X)
 
 
 def test_use_table_report(simple_df):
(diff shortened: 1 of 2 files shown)


skrub/tests/test_config.py
```diff
@@ -1,16 +1,31 @@
...
 
 
+@pytest.fixture
+def simple_df():
+    return pd.DataFrame(
+        {
+            "A": [1, 2, 3, 4, 5],
+            "B": ["a", "b", "a", "b", "c"],
+        }
+    )
```
rcap107: Eventually we would like to remove pandas as a dependency, so we should try to use df_module wherever we can 

```suggestion
def simple_df(df_module):
    return df_module.make_dataframe(
        {
            "A": [1, 2, 3, 4, 5],
            "B": ["a", "b", "a", "b", "c"],
        }
    )
```

skrub/tests/test_config.py
```diff
@@ -1,16 +1,31 @@
...
+
+
+@pytest.fixture
+def simple_series():
+    return pd.Series([1, 2, 3, 4, 5], name="A")
+
```
rcap107, line 27: Same here

```suggestion

@pytest.fixture
def simple_series(df_module):
    return df_module.make_column(name="A", values=[1, 2, 3, 4, 5])

```

rcap107:
hah, as it turns out tests are breaking because of the new polars-without-pyarrow conf :laughing:

PR #1657, opened 2025-10-02
Implementation for pruning full reports of the skrub-data folder 

dierickxsimon:
I can`t find any issues related to this PR, but this problem came up during the pydata2025 sprints. It`s a fix to ensure that the skrub-data folder does not get full of unused reports when creating a `full_report`. 

This fix ensure that when creating a `full_report` with `output_dir=None` older reports in the skrub_data folder are deleted.

### DIFF
--- a/skrub/_data_ops/_skrub_namespace.py
+++ b/skrub/_data_ops/_skrub_namespace.py
@@ -1309,6 +1309,11 @@ class SkrubNamespace:
         was defined, the time it took to run, and more) and a display of the
         intermediate result (or error).
 
+        By default the report is stored in the skrub data folder.
+        TODO complete the documentation!! write a note for the rest.
+
+        note: reports older then one week are deleted in the skrub data folder
+
         Parameters
         ----------
         environment : dict or None (default=None)
--- a/skrub/_data_ops/_utils.py
+++ b/skrub/_data_ops/_utils.py
@@ -1,5 +1,6 @@
 import enum
 import os
+import shutil
 import time
 import traceback
 
@@ -69,18 +70,14 @@ def prune_folder(path: str):
     if not os.path.exists(path):
         return
 
-    files = sorted(
-        (
-            (f, os.path.getmtime(os.path.join(path, f)))
-            for f in os.listdir(path)
-            if os.path.isfile(os.path.join(path, f))
-        ),
-        key=lambda f: f[1],
-        reverse=True,
+    folders = (
+        (f, os.path.getmtime(os.path.join(path, f)))
+        for f in os.listdir(path)
+        if os.path.isdir(os.path.join(path, f))
     )
-    for f in files:
+    for dir in folders:
         try:
-            if f[1] < time_threshold and f[0].startswith("full_data_op_report_"):
-                os.remove(os.path.join(path, f[0]))
+            if dir[1] < time_threshold and dir[0].startswith("full_data_op_report_"):
+                shutil.rmtree(os.path.join(path, dir[0]))
         except Exception:
             pass
--- a/skrub/_data_ops/_utils.py
+++ b/skrub/_data_ops/_utils.py
@@ -1,4 +1,6 @@
 import enum
+import os
+import time
 import traceback
 
 from joblib.externals import cloudpickle
@@ -60,3 +62,25 @@ def format_exception(e):
 def format_exception_only(e):
     """compatibility for python < 3.10"""
     return traceback.format_exception_only(type(e), e)
+
+
+def prune_folder(path: str):
+    time_threshold = time.time() - 7 * 24 * 3600  # 7 days ago
+    if not os.path.exists(path):
+        return
+
+    files = sorted(
+        (
+            (f, os.path.getmtime(os.path.join(path, f)))
+            for f in os.listdir(path)
+            if os.path.isfile(os.path.join(path, f))
+        ),
+        key=lambda f: f[1],
+        reverse=True,
+    )
+    for f in files:
+        try:
+            if f[1] < time_threshold and f[0].startswith("full_data_op_report_"):
+                os.remove(os.path.join(path, f[0]))
+        except Exception:
+            pass
--- a/skrub/_data_ops/tests/test_utils.py
+++ b/skrub/_data_ops/tests/test_utils.py
@@ -5,17 +5,14 @@ import time
 from skrub._data_ops import _utils
 
 
-def test_prune_folder_with_standard_name():
+def test_prune_folder_with_standard_name_dirs():
     eight_days_ago = time.time() - 8 * 24 * 3600
     with tempfile.TemporaryDirectory() as tmpdir:
         for i in range(3):
-            with open(os.path.join(tmpdir, f"full_data_op_report_{i}.txt"), "w") as f:
-                f.write("This is a test file.\n")
+            dirname = os.path.join(tmpdir, f"full_data_op_report_{i}")
+            os.mkdir(dirname)
             if i < 2:
-                os.utime(
-                    os.path.join(tmpdir, f"full_data_op_report_{i}.txt"),
-                    (eight_days_ago, eight_days_ago),
-                )
+                os.utime(dirname, (eight_days_ago, eight_days_ago))
 
         assert len(os.listdir(tmpdir)) == 3
 
@@ -25,15 +22,12 @@ def test_prune_folder_with_standard_name():
         assert len(remaining_items) == 1
 
 
-def test_prune_folder_with_nonstandard_name():
+def test_prune_folder_with_nonstandard_name_dirs():
     eight_days_ago = time.time() - 8 * 24 * 3600
     with tempfile.TemporaryDirectory() as tmpdir:
-        with open(os.path.join(tmpdir, "other_report.txt"), "w") as f:
-            f.write("This is a test file.\n")
-            os.utime(
-                os.path.join(tmpdir, "other_report.txt"),
-                (eight_days_ago, eight_days_ago),
-            )
+        dirname = os.path.join(tmpdir, "other_report")
+        os.mkdir(dirname)
+        os.utime(dirname, (eight_days_ago, eight_days_ago))
 
         assert len(os.listdir(tmpdir)) == 1
 
@@ -41,3 +35,4 @@ def test_prune_folder_with_nonstandard_name():
 
         remaining_items = os.listdir(tmpdir)
         assert len(remaining_items) == 1
+        assert remaining_items[0] == "other_report"
--- a/skrub/_data_ops/tests/test_utils.py
+++ b/skrub/_data_ops/tests/test_utils.py
@@ -1,0 +1,43 @@
+import os
+import tempfile
+import time
+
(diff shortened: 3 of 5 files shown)


dierickxsimon:
I' m running in some issues with the ci/cd, but it's a bit unclear for me what exactly the problem is.

rcap107:
It looks like the CI issue is related to #1658, so now that that PR is merged it should be fixed

skrub/_data_ops/_utils.py
```diff
@@ -60,3 +63,21 @@ def format_exception(e):
...
     return traceback.format_exception_only(type(e), e)
+
+
+def prune_folder(path: str):
+    if not os.path.exists(path):
+        return
+    time_threshold = time.time() - 7 * 24 * 3600  # 7 days ago
+
+    folders = (
+        (f, os.path.getmtime(os.path.join(path, f)))
```
dierickxsimon: All tests are passing except the codecov (not enough test coverage). In the test cov report these  lines (line 73-77) are not covered. Maybe because assignment of the variable is only done on line 73.

skrub/_data_ops/_utils.py
```diff
@@ -60,3 +63,21 @@ def format_exception(e):
...
+        for f in os.listdir(path)
+        if os.path.isdir(os.path.join(path, f))
+    )
+    for dir in folders:
+        try:
+            if dir[1] < time_threshold and dir[0].startswith("full_data_op_report_"):
+                shutil.rmtree(os.path.join(path, dir[0]))
+        except Exception:
```
dierickxsimon: This line isn’t covered by the tests: I added the try-except block to make sure it fails silently if the file happens to still be open. I’m not sure if this is really needed (and if so, whether the exception block should be tested). Thanks in advance!

GaelVaroquaux: Unrelated comment: it's much much preferable to capture a specific Exception rather than the base class. So this code should capture the specific one that is trigger by the problem.

Also, I would always add a small comment on why we capture the exception.

skrub/_data_ops/_utils.py
```diff
@@ -60,3 +63,21 @@ def format_exception(e):
...
+    time_threshold = time.time() - 7 * 24 * 3600  # 7 days ago
+
+    folders = (
+        (f, os.path.getmtime(os.path.join(path, f)))
+        for f in os.listdir(path)
+        if os.path.isdir(os.path.join(path, f))
+    )
+    for dir in folders:
+        try:
+            if dir[1] < time_threshold and dir[0].startswith("full_data_op_report_"):
```
rcap107: A couple of changes: I moved the name condition to the first if, and I unpacked `dir` into `dir_path` and `dir_time`. `dir` is a python keyword so it should not be used as a variable name. 

```suggestion

    folders = (
        (f, os.path.getmtime(os.path.join(path, f)))
        for f in os.listdir(path)
        if os.path.isdir(os.path.join(path, f)) 
        and f.startswith("full_data_op_report_"))
    )
    for dir_path, dir_time in folders:
        try:
            if dir_time < time_threshold :
```

skrub/_data_ops/_utils.py
rcap107: I think we should catch the exception as a `PermissionError`, and the exception should be tested. However, this should be a Windows specific error, so the test should be skipped when the platform isn't Windows with something like

```python
import sys
import pytest

@pytest.mark.skipif(sys.platform != 'win32', reason="Test is Windows-specific")
def test_windows_specific_functionality():
    # test code here
```

If you don't have a Windows machine to test this one, it will have to be tested on CI which is going to be annoying unfortunately.

skrub/_data_ops/_utils.py
rcap107: This is probably going to break the pre-commit checks

skrub/_data_ops/_utils.py
jeromedockes: I agree with the general rule but here I think catching a broad exception type makes sense: we are performing an accessory background task (cleanup of old files created by other programs), so we don't want to crash the program if it fails. it might fail for other reasons beside the one mentioned here: another thread may have removed it already between the listdir() and the rmtree() calls, we may not have permission if it was created by another user, etc. . In those cases failing the current report generation would not help, so I suggest we emit a warning (so that if they see the warning often the user might want to report it or manually delete the directory if it is large) and proceed with the main task without raising an error

jeromedockes: to test on other platforms than windows (and even on windows actually), use [unittest.mock](https://docs.python.org/3/library/unittest.mock.html#) and the pytest [monkeypatch](https://docs.pytest.org/en/stable/how-to/monkeypatch.html) fixture to mock rmtree so it raises an exception (set side_effect = OSError() on the mock)

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1307,7 +1307,13 @@ def full_report(
...
         intermediate computation, some information (the line of code where it
         was defined, the time it took to run, and more) and a display of the
-        intermediate result (or error).
+        intermediate result (or error). By default, the report is stored in
+        a timestamped subdirectory of the skrub data folder.
+
+        .. note::
+
+            reports starting with ``full_data_op_report_`` that are stored
+            in the skrub data folder are automatically deleted after 7 days.
```
jeromedockes: maybe we can tweak the wording to make it clearer that each execution of the report checks for old files and removes them? so that wary users don't wonder if we are trying to schedule something or leave some daemon cleanup process laying around

dierickxsimon: sound good to me. I will change it!

jeromedockes:
> sound good to me. I will change it!

and maybe we could add something along the lines of "for reports that you want to keep, provide an output directory" so users know what to do if they don't want it removed? Thanks!

skrub/_data_ops/_utils.py
GaelVaroquaux: I'm fine with @jeromedockes point: we can not worry about being specific and catch a broad exception

dierickxsimon: Okay then i will just catch a broad exception and write a test for it, I willl work on it today or tomorrow!

skrub/_data_ops/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+from skrub._data_ops import _utils
+
+
+def test_prune_folder_with_standard_name_dirs():
+    eight_days_ago = time.time() - 8 * 24 * 3600
+    with tempfile.TemporaryDirectory() as tmpdir:
+        for i in range(3):
+            dirname = os.path.join(tmpdir, f"full_data_op_report_{i}")
+            os.mkdir(dirname)
+            if i < 2:
```
rcap107:
```suggestion
            os.mkdir(dirname)
            # setting the access time of the first two files 
            # to 8 days ago, so that they should be pruned
            if i < 2:
```

skrub/_data_ops/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+        remaining_items = os.listdir(tmpdir)
+        assert len(remaining_items) == 1
+
+
+def test_prune_folder_with_nonstandard_name_dirs():
+    eight_days_ago = time.time() - 8 * 24 * 3600
+    with tempfile.TemporaryDirectory() as tmpdir:
+        dirname = os.path.join(tmpdir, "other_report")
+        os.mkdir(dirname)
+        os.utime(dirname, (eight_days_ago, eight_days_ago))
```
rcap107:
```suggestion
        os.mkdir(dirname)
        # setting access time to 8 days ago
        os.utime(dirname, (eight_days_ago, eight_days_ago))
```

skrub/_data_ops/tests/test_utils.py
```diff
@@ -0,0 +1,61 @@
...
+
+        assert len(os.listdir(tmpdir)) == 1
+
+        _utils.prune_folder(tmpdir)
+
+        remaining_items = os.listdir(tmpdir)
+        assert len(remaining_items) == 1
```
rcap107:
```suggestion
        # the report has a non-default name, 
        # so it should not be pruned
        assert len(remaining_items) == 1
```

skrub/_data_ops/_utils.py
```diff
@@ -62,3 +66,25 @@ def format_exception(e):
...
     """compatibility for python < 3.10"""
     return traceback.format_exception_only(type(e), e)
+
+
+def prune_folder(path: str):
+    if not os.path.exists(path):
+        return
+    time_threshold = time.time() - 7 * 24 * 3600  # 7 days ago
+
+    folders = (
```
jeromedockes: nitpick: the rest of skrub uses the (newer) pathlib module for those kind of filesystem manipulation rather than os, so for consistency it would be great to use it here, especially since the path you get from `_get_output_dir` is already a pathlib.Path 

for example you can get the mtime with [stat](https://docs.python.org/3/library/pathlib.html#pathlib.Path.stat), the Path object representing the directory has iterdir() and glob() methods, you can join with the `/` operator or `joinpath`, etc.

also maybe we could name `folders` `directories` instead? since we use the term 'directory' rather than 'folder' for other variables (same for the function name prune_folder)

finally why do you prefer 2 separate loops for finding the directories and then removing them?

skrub/_data_ops/_utils.py
```diff
@@ -62,3 +66,25 @@ def format_exception(e):
...
+        (f, os.path.getmtime(os.path.join(path, f)))
+        for f in os.listdir(path)
+        if os.path.isdir(os.path.join(path, f)) and f.startswith("full_data_op_report_")
+    )
+    for dir_path, dir_time in folders:
+        try:
+            if dir_time < time_threshold:
```
jeromedockes: i guess the test could be outside the try block

skrub/_data_ops/_utils.py
```diff
@@ -62,3 +66,25 @@ def format_exception(e):
...
+                shutil.rmtree(os.path.join(path, dir_path))
```
jeromedockes: I wonder if we should have an even stricter check of the name (eg check with a regex there is a timestamp) just to be extra sure we don't erase something we shouldn't ... but I guess "full_data_op_report_" is already very specific 🤔

skrub/_data_ops/_utils.py
dierickxsimon: totally agree with the variable naming — I’ll change it, and I’ll switch to using the pathlib module as well. As for the two separate loops: I initially used them for debugging purposes so I could easily print all the folders that were selected. But maybe its better to and even more readable to put it all in one loop. I will change it!

dierickxsimon:
@jeromedockes  I updated all code based on your comments. I also cleaned up the pytests to make them a bit more readable and to remove the code duplication.

jeromedockes:
> @jeromedockes I updated all code based on your comments. I also cleaned up the pytests to make them a bit more readable and to remove the code duplication.

awesome, thank you for this feature @dierickxsimon ! this is an important addition to make the reports convenient to use :)

PR #1658, opened 2025-10-02
avoid error when 2 threads fetch the same dataset at the same time

jeromedockes:
solves half of #1655  (the concurrency issue in the fetcher); the other half (too many useless downloads in tests) is tackled in another pr #1656


rcap107 (approved):
thanks @jeromedockes, this bug was quite annoying

PR #1659, opened 2025-10-06
FIX - Fixing a condition in conftest for polars w/out pyarrow

rcap107:
#1648 added a slight mistake in how the mark is checked. This PR fixes that and renames the marker so the name is more explicit.


rcap107:
> oupsy, you're correct.

Happens! Not a big deal

MarieSacksick:
Once you merge with main, there is an additional place where you'll have to change the variable name due to an addition, here: https://github.com/skrub-data/skrub/pull/1656/files#diff-9278359efc8c91f314bcf3ae2a7b867c67d653bb233ce367995dd874a1c3db15R112.

jeromedockes:
thanks @rcap107  and @MarieSacksick  :)

Issue #1294, opened 2025-04-10
Create dedicated doc sections to explain some of the expressions error messages


jeromedockes:
Some of the errors can be difficult to understand at first: those related to the fact that expressions are lazy. We discussed adding sections in the doc and showing the url in the error message

rcap107:
Related issues: 
- #1571 
- #1773

rcap107:
Another error I've encountered a few times and that I always forget is this:

```sh
"dot" with args ['-Tsvg', '/var/folders/s5/sh5p03613f99c34ty0s6p3tc0000gn/T/tmpgqs7a9fj/tmpyr57lrb_'] returned code: 1

stdout, stderr:
 b''
b'Format: "svg" not recognized. No formats found.\nPerhaps "dot -c" needs to be run (with installer\'s privileges) to register the plugins?\n'
```

Which is accompanied by this not very helpful exception:

```python
File ~/work/skrub/skrub/_data_ops/_inspection.py:340, in _check_graphviz()
    338 if _has_graphviz():
    339     return
--> [340](https://file+.vscode-resource.vscode-cdn.net/Users/rcap/work/skrub/~/work/skrub/skrub/_data_ops/_inspection.py:340) raise ImportError("Please install pydot and graphviz to draw data_op graphs.")

ImportError: Please install pydot and graphviz to draw data_op graphs.
```

The solution is simple (`dot -c` in a terminal in whatever environment is being used), but it's not very obvious, so it might warrant being added to the "common errors" list.

PR #1661, opened 2025-10-06
WIP 🚧 allow setting dataop choices with optuna

jeromedockes:
setting skrub.DataOp choices with Optuna

Apparently @rcap107 was asked a few times about combining skrub dataops and optuna, which is a great idea.

This is very rough first draft of what this could look like. Still missing some functionality, tests & docs etc.

## search estimator interface

The DataOp now has a `.skb.make_optuna_search()` method, which can be used to search for the best choice outcomes with optuna.

For example:

```python
import skrub
from sklearn.linear_model import LogisticRegression
from sklearn.ensemble import RandomForestClassifier
from sklearn.dummy import DummyClassifier

X_a, y_a = make_classification(random_state=0)
X, y = skrub.X(X_a), skrub.y(y_a)
logistic = LogisticRegression(C=skrub.choose_float(0.1, 10.0, log=True, name="C"))
rf = RandomForestClassifier(
    n_estimators=skrub.choose_int(3, 30, name="n estimators"), random_state=0
)
classifier = skrub.choose_from(
    {"logistic": logistic, "random forest": rf, "dummy": DummyClassifier()},
    name="classifier",
)
pred = X.skb.apply(classifier, y=y)

search = pred.skb.make_optuna_search(fitted=True, random_state=0)  # <--- 🎯 optuna --------------

print("\nparam grid:")
print(pred.skb.describe_param_grid())

print("best params:")
print(search.best_learner_.describe_params())
```

prints

```
[... optuna output]
[I 2025-10-06 23:33:43,972] Trial 9 finished with value: 0.8399999999999999 and parameters: {'2:classifier': '0:logistic', '0:C': 0.10037026315054523}. Best is trial 6 with value: 0.9099999999999999.

param grid:
- classifier: 'logistic'
  C: choose_float(0.1, 10.0, log=True, name='C')
- classifier: 'random forest'
  n estimators: choose_int(3, 30, name='n estimators')
- classifier: 'dummy'

best params:
{'n estimators': 7, 'classifier': 'random forest'}
```

We can pass our own optuna study to control how it is created
For example above change `make_optuna_search()` to:

```python
import optuna

study = optuna.create_study(
    storage="sqlite:///db.sqlite3", direction="maximize", study_name="classifier"
)

search = pred.skb.make_optuna_search(fitted=True, n_trials=200, study=study)
```

and we can monitor the search progress / explore results with `❯ optuna-dashboard sqlite:///db.sqlite3`

## low-level interface

For more control over the use of optuna, the SkrubLearner.collect_params() method can be used to obtain a set of parameters from an optuna trial. this is useful for more advanced optuna features (multi-objective, ask-and-tell interface, ...). Here is a simple example revisited from the optuna tutorial:

```python
import optuna
import skrub

x = skrub.choose_float(-10, 10, name="x").as_data_op()
err = (x - 2) ** 2

def objective(trial):
    learner = err.skb.make_learner()
    params = learner.collect_params(trial)
    learner.set_params(**params)
    return learner.fit_transform({})


study = optuna.create_study(direction="minimize")
study.optimize(objective, n_trials=100)
print(study.best_params)
```

prints something similar to

```
[...]
[I 2025-10-06 23:41:17,610] Trial 99 finished with value: 1.210419581181497 and parameters: {'0:x': 0.8998092978117398}. Best is trial 61 with value: 0.0005685118895572634.
{'0:x': 2.0238434873614843}
```

### DIFF
--- a/skrub/_data_ops/_estimator.py
+++ b/skrub/_data_ops/_estimator.py
@@ -1,22 +1,28 @@
 # Scikit-learn-ish interface to the skrub DataOps
+import sys
 
+import numpy as np
 import pandas as pd
+from joblib import effective_n_jobs
 from sklearn import model_selection
 from sklearn.base import BaseEstimator, TransformerMixin, clone
 from sklearn.exceptions import NotFittedError
+from sklearn.utils import check_random_state
 from sklearn.utils.validation import check_is_fitted
 
 from .. import _join_utils
+from .._utils import short_repr
 from ._choosing import BaseNumericChoice, get_default
 from ._data_ops import Apply, check_subsampled_X_y_shape
 from ._evaluation import (
     choice_graph,
-    chosen_or_default_outcomes,
+    eval_choices,
     evaluate,
     find_first_apply,
     find_node_by_name,
     find_X,
     find_y,
+    get_choice_default,
     get_params,
     param_grid,
     set_params,
@@ -87,6 +93,42 @@ class _CloudPickleDataOp(_CloudPickle):
     """
 
     _cloudpickle_attributes = ["data_op"]
+
+
+def _is_optuna_trial(obj):
+    try:
+        trial_type = sys.modules["optuna"].Trial
+    except (KeyError, AttributeError):
+        return False
+    return isinstance(obj, trial_type)
+
+
+def _optuna_trial_suggestions(trial):
+    def policy(choice_id, display_name, choice):
+        name = f"{choice_id}:{display_name}"
+        if isinstance(choice, BaseNumericChoice):
+            if choice.to_int:
+                func = trial.suggest_int
+                default_step = 1
+            else:
+                func = trial.suggest_float
+                default_step = None
+            return func(
+                name,
+                choice.low,
+                choice.high,
+                log=choice.log,
+                step=getattr(choice, "step", default_step),
+            )
+        if choice.outcome_names is None:
+            outcome_names = list(map(short_repr, choice.outcomes))
+        else:
+            outcome_names = choice.outcome_names
+        outcome_names = [f"{i}:{n}" for i, n in enumerate(outcome_names)]
+        result = trial.suggest_categorical(name, outcome_names)
+        return int(result.split(":", 1)[0])
+
+    return policy
 
 
 class SkrubLearner(_CloudPickleDataOp, BaseEstimator):
@@ -178,10 +220,32 @@ class SkrubLearner(_CloudPickleDataOp, BaseEstimator):
     def set_params(self, **params):
         if "data_op" in params:
             self.data_op = params.pop("data_op")
-        set_params(
-            self.data_op, {int(k.lstrip("data_op__")): v for k, v in params.items()}
-        )
+
+        def to_id(key):
+            if key.startswith("data_op__"):
+                return int(key.lstrip("data_op__"))
+            return int(key.split(":", 1)[0])
+
+        def to_idx(val):
+            if isinstance(val, str):
+                return int(val.split(":", 1)[0])
+            return val
+
+        set_params(self.data_op, {to_id(k): to_idx(v) for k, v in params.items()})
         return self
+
+    def collect_params(self, policy):
+        if _is_optuna_trial(policy):
+            policy = _optuna_trial_suggestions(policy)
+        elif isinstance(policy, str) and policy == "default":
+            policy = get_choice_default
+        else:
+            raise ValueError(
+                "Policy should be a optuna.Trial instance or the string 'default', got"
+                " object of type {type(obj).__name__!r}: {obj!r}"
+            )
+        choices = eval_choices(self.data_op, policy)
+        return {f"data_op__{k}": v for k, v in choices.items()}
 
     def get_param_grid(self):
         grid = param_grid(self.data_op)
@@ -402,9 +466,7 @@ class SkrubLearner(_CloudPickleDataOp, BaseEstimator):
         parameters (outcomes of `choose_*` objects contained in the
         DataOp).
         """
-        return describe_params(
-            chosen_or_default_outcomes(self.data_op), choice_graph(self.data_op)
-        )
+        return describe_params(eval_choices(self.data_op), choice_graph(self.data_op))
 
 
 # Xy_pipeline because it is an actual scikit-learn pippeline rather than
@@ -894,3 +956,88 @@ class _XyParamSearch(_XyPipelineMixin, ParamSearch):
 
     def _call_predictor_method(self, name, X, y=None):
         return getattr(self.best_learner_, name)(self._get_env(X, y))
+
+
+class OptunaSearch(_CloudPickleDataOp, BaseEstimator):
+    def __init__(
+        self,
+        data_op,
+        study=None,
+        cv=None,
+        random_state=None,
+        n_jobs=None,
+        n_trials=10,
+        timeout=None,
+        callbacks=None,
+        catch=(),
+        refit=True,
+    ):
+        self.data_op = data_op
+        self.study = study
+        self.cv = cv
+        self.random_state = random_state
+        self.n_jobs = n_jobs
+        self.n_trials = n_trials
+        self.timeout = timeout
+        self.callbacks = callbacks
+        self.catch = catch
+        self.refit = refit
+
+    def fit(self, environment):
+        import optuna
+        import optuna.samplers
+
+        if self.study is None:
+            random_state = check_random_state(self.random_state)
+            seed = random_state.randint(np.iinfo("int32").max)
+            # The default sampler, we instantiate it to set the seed
+            sampler = optuna.samplers.TPESampler(seed=seed)
+            self.study_ = optuna.create_study(direction="maximize", sampler=sampler)
+        else:
+            self.study_ = self.study
+
+        def objective(trial):
+            learner = self.data_op.skb.make_learner()
+            params = learner.collect_params(trial)
+            learner.set_params(**params)
+            cv_results = learner.data_op.skb.cross_validate(environment, cv=self.cv)
+            return cv_results["test_score"].mean()
+
+        n_jobs = effective_n_jobs(self.n_jobs)
+        self.study_.optimize(
+            objective,
+            n_jobs=n_jobs,
+            n_trials=self.n_trials,
+            timeout=self.timeout,
+            callbacks=self.callbacks,
+            catch=self.catch,
+        )
+        self.best_params_ = self.study_.best_params
+        if not self.refit:
+            return self
+        best_learner = self.data_op.skb.make_learner()
+        best_learner.set_params(**self.study_.best_params)
+        best_learner.fit(environment)
+        self.best_learner_ = best_learner
+        return self
+
+    def __getattr__(self, name):
+        if name not in supported_modes(self.data_op):
+            attribute_error(self, name)
+
+        def f(*args, **kwargs):
+            return self._call_predictor_method(name, *args, **kwargs)
+
+        f.__name__ = name
+        return f
+
+    def _call_predictor_method(self, name, environment):
+        check_is_fitted(self, "study_")
+        if not hasattr(self, "best_learner_"):
+            raise AttributeError(
+                "This parameter search was initialized with `refit=False`. "
+                f"{name} is available only after refitting on the best parameters. "
+                "Please pass another value to `refit` or fit a learner manually "
+                "using the `best_params_` or `study_` attributes."
+            )
+        return getattr(self.best_learner_, name)(environment)
--- a/skrub/_data_ops/_evaluation.py
+++ b/skrub/_data_ops/_evaluation.py
@@ -950,39 +950,52 @@ def set_params(data_op, params):
             target.chosen_outcome = v
 
 
-class _ChosenOrDefaultOutcomes(_DataOpTraversal):
-    """Helper for `chosen_or_default_outcomes`."""
+def get_choice_default(choice_id, display_name, choice):
+    if isinstance(choice, _choosing.Choice):
+        return choice.chosen_outcome_idx or 0
+    else:
+        return choice.chosen_outcome_or_default()
 
-    def run(self, data_op):
+
+class _ChoiceEvaluator(_DataOpTraversal):
+    """Helper for `eval_choices`."""
+
+    def run(self, data_op, policy):
+        graph = choice_graph(data_op)
+        data_op_choices = graph["choices"]
+        self.display_names = graph["choice_display_names"]
+        self.choice_ids = {id(c): k for k, c in data_op_choices.items()}
+        self.Xy_choices = graph["Xy_choices"]
+        self.policy = policy
         self.chosen = {}
         self.results = {}
         _ = super().run(data_op)
         return self.chosen
 
     def handle_choice(self, choice):
-        if id(choice) in self.results:
-            return self.results[id(choice)]
-        if not isinstance(choice, _choosing.Choice):
-            # We have a NumericChoice, the outcome is simply a number
-            outcome = choice.chosen_outcome_or_default()
-            self.chosen[id(choice)] = outcome
-            self.results[id(choice)] = outcome
-            return outcome
-        # We have a Choice and need to visit the chosen outcome (it may contain
-        # further choices).
-        idx = choice.chosen_outcome_idx or 0
-        self.chosen[id(choice)] = idx
-        outcome = choice.outcomes[idx]
-        result = yield outcome
-        self.results[id(choice)] = result
-        return result
+        c_id = self.choice_ids[id(choice)]
+        if c_id in self.results:
+            return self.results[c_id]
+        display_name = self.display_names[c_id]
+        if c_id in self.Xy_choices:
+            # Clamp to default if ancestor of X or y
+            param = get_choice_default(c_id, display_name, choice)
+        else:
+            param = self.policy(c_id, display_name, choice)
+        if isinstance(choice, _choosing.Choice):
+            value = yield choice.outcomes[param]
+        else:
+            value = param
+        self.chosen[c_id] = param
+        self.results[c_id] = value
+        return value
 
     def handle_choice_match(self, choice_match):
         outcome = yield choice_match.choice
         return (yield choice_match.outcome_mapping[outcome])
 
 
-def chosen_or_default_outcomes(data_op):
+def eval_choices(data_op, policy=get_choice_default):
     """Get the selected or default outcomes for choices in the DataOp.
 
     Return a mapping from the choice's ID (0, 1, ... -- see `choice_graph`) to
@@ -1000,7 +1013,7 @@ def chosen_or_default_outcomes(data_op):
     >>> from sklearn.linear_model import Ridge
     >>> from sklearn.dummy import DummyRegressor
     >>> from skrub import choose_from, choose_float
-    >>> from skrub._data_ops._evaluation import chosen_or_default_outcomes, choices
+    >>> from skrub._data_ops._evaluation import eval_choices, choices
 
     >>> e = choose_from(
     ...     [DummyRegressor(), Ridge(alpha=choose_float(0.1, 1.0, name="alpha"))],
@@ -1016,18 +1029,15 @@ def chosen_or_default_outcomes(data_op):
 
     In the default configuration, the 'regressor' chooses the DummyRegressor. So the
     'alpha' choice is not used. It does not appear in the
-    `chosen_or_default_outcomes` result:
+    `eval_choices` result:
 
-    >>> pprint(chosen_or_default_outcomes(e))
+    >>> pprint(eval_choices(e))
     {1: 0}
 
     Here we only see that choice 'regressor' (ID 1) in chooses its first outcome
     (index 0), and the choice 'alpha' (ID 0) does not appear.
     """  # noqa: E501
-    data_op_choices = choices(data_op)
-    short_ids = {id(c): k for k, c in data_op_choices.items()}
-    outcomes = _ChosenOrDefaultOutcomes().run(data_op)
-    return {short_ids[k]: v for k, v in outcomes.items()}
+    return _ChoiceEvaluator().run(data_op, policy=policy)
 
 
 class _Found(Exception):
--- a/skrub/_data_ops/_skrub_namespace.py
+++ b/skrub/_data_ops/_skrub_namespace.py
@@ -19,7 +19,13 @@ from ._data_ops import (
     check_name,
     deferred,
 )
-from ._estimator import ParamSearch, SkrubLearner, cross_validate, train_test_split
+from ._estimator import (
+    OptunaSearch,
+    ParamSearch,
+    SkrubLearner,
+    cross_validate,
+    train_test_split,
+)
 from ._evaluation import (
     choices,
     clone,
@@ -1286,14 +1292,12 @@ class SkrubNamespace:
         ... )
         >>> pred = X.skb.apply(selector, y=y).skb.apply(classifier, y=y)
         >>> print(pred.skb.describe_defaults())
-        {'k': 9, 'classifier': 'logistic', 'C': 1.0...}
+        {'k': 9, 'C': 1.0..., 'classifier': 'logistic'}
         """
-        from ._evaluation import choice_graph, chosen_or_default_outcomes
+        from ._evaluation import choice_graph, eval_choices
         from ._inspection import describe_params
 
-        return describe_params(
-            chosen_or_default_outcomes(self._data_op), choice_graph(self._data_op)
-        )
+        return describe_params(eval_choices(self._data_op), choice_graph(self._data_op))
 
     def full_report(
         self,
@@ -1788,6 +1792,40 @@ class SkrubNamespace:
 
         search = ParamSearch(
             self.clone(), model_selection.RandomizedSearchCV(None, None, **kwargs)
+        )
+        if not fitted:
+            return search
+        return search.fit(
+            env_with_subsampling(self._data_op, self.get_data(), keep_subsampling)
+        )
+
+    def make_optuna_search(
+        self,
+        *,
+        study=None,
+        cv=None,
+        random_state=None,
+        n_jobs=None,
+        n_trials=10,
+        timeout=None,
+        callbacks=None,
+        catch=(),
+        refit=True,
+        fitted=False,
+        keep_subsampling=False,
+    ):
+        _check_keep_subsampling(fitted, keep_subsampling)
+        search = OptunaSearch(
+            self.clone(),
+            study=study,
+            cv=cv,
+            random_state=random_state,
+            n_jobs=n_jobs,
+            n_trials=n_trials,
+            timeout=timeout,
+            callbacks=callbacks,
+            catch=catch,
+            refit=refit,
         )
         if not fitted:
             return search


rcap107:
This is really cool! Thanks a lot @jeromedockes!

GaelVaroquaux:
Thanks a lot for the PR!!

> The DataOp now has a .skb.make_optuna_search()

From an API perspective I would much prefer if we could avoid making a function for each hyper-parameter search method. It will scale very poorly as we add methods. I do realize that we might need different functions, as there will be different signatures, but can we find commonalities?

rcap107:
> Thanks a lot for the PR!!
> The DataOp now has a .skb.make_optuna_search()
> From an API perspective I would much prefer if we could avoid making a function for each hyper-parameter search method. It will scale very poorly as we add methods. I do realize that we might need different functions, as there will be different signatures, but can we find commonalities?

You mean something like this? 

```python
predictions.skb.make_param_search(kind="random", ...)
```

Is it going to be necessary? Are there so many other param search methods we can add?

GaelVaroquaux:
> You mean something like

> predictions.skb.make_param_search(kind="random", ...)

Yes

jeromedockes:
The main parameters which differ between those search strategies are:

- `study`, `timeout`: only for optuna. (also `n_iter` is called `n_trials` in optuna, and optuna has a few additional less important parameters)
- `n_iter`/`n_trials`, `random_state`: only for optuna and randomized_search, not for grid_search

all but `random_state` are important/often-used parameters.

In the return value, the optuna search will have a `study_` attribute which the scikit-learn ones don't. We may also later want to add capabilities like resuming the search which only apply to optuna.

I guess it's a tradeoff between having several functions or having a few "param a is ignored unless param b = x" in the docstring and different return types. Scikit-learn chose to have a different class for each search strategy. Regarding the addition of a new search strategy, if we go for multiple methods the cost is to create a new method (which is annoying and clutters the namespace), if we go for a single method, it depends on how many new parameters the new strategy needs to introduce (docstring gets more complex as the ratio shared parameters / all parameters decreases).

As strategies to add I'm guessing you have in mind the HalvingRandomSearchCV and HalvingGridSearchCV, once those stop being experimental in scikit-learn? And there are also other packages like hyperopt, not sure if we will want to support more or if optuna is enough (I would say just optuna should be good for the foreseeable future).

jeromedockes:
In any case I think I'll split this PR. I'll remove the `make_optuna_search()` to keep only the possibility to pass a trial in `.skb.make_learner(choose=optuna_trial)` (the name "choose" is up for debate). This is the main addition which by itself enables the use of optuna with the data ops, and a minimal addition to the skrub API. The `make_search` method is boilerplate around that which I can add in another PR once we settle on the API question.

GaelVaroquaux:
Thanks @jeromedockes !

After discussion with @rcap107 , here are my thoughts:

This PR caters to optuna power user. Many people are not optuna users, but would like to flip a switch to use optuna.

Looking at the example https://output.circle-artifacts.com/output/job/8860ed56-510b-47b7-b260-b71da3da42a1/artifacts/0/doc/auto_examples/data_ops/13_optuna_choices.html#sphx-glr-auto-examples-data-ops-13-optuna-choices-py with @rcap107 it seems that we could do a: `make_iter_search` that could use either scikit-learn's RandomSearch, or optuna (or more in the long run).

The internals of `make_iter_search` would really be wrapping code like the one that you wrote in the example:

```python
import optuna


def objective(trial):
    learner = pred.skb.make_learner(choose=trial)
    cv_results = skrub.cross_validate(learner, environment=env, cv=cv)
    return cv_results["test_score"].mean()


study = optuna.create_study(direction="maximize")
study.optimize(objective, n_trials=16)

best_learner = pred.skb.make_learner(choose=study.best_trial)
```

We would then modify the existing example to show first at the top how to easily use optuna, just flipping a switch from code using RandomSearch, and then continue with your example for advanced users.

We'll also need to do a bit of impedance matching, converting "n_iter" which is scikit-learn's convention to "n_trials", and exposing the optuna solver.

jeromedockes:
thanks @GaelVaroquaux yes that is the plan, to have the scikit-learn estimator interface for most users, and the more bare-bones direct optuna use for optuna users. As I mentioned I opted for splitting the PR so that I could make progress on the core functionality while we decided on the `make_<estimator>()` specific API. Note the benefit of optuna is not only the choice of samplers but also features that make it user-friendly (like the visualizations or resuming searches), some of which will probably be more easily available through the "optuna user" interface.

jeromedockes:
> we could do a: make_iter_search that could use either scikit-learn's RandomSearch, or optuna (or more in the long run).

a couple of questions:

- can we list the parameters of `make_iter_search` here (or in another issue)? I'm guessing only `n_iter`, `random_state`, `n_jobs`, refit`, and the CV params (cv and scoring)?
-  what other methods do we plan to add in the long run?
- is `make_grid_search` also replaced by `make_iter_search`?

> We'll also need to do a bit of impedance matching, converting "n_iter" which is scikit-learn's convention to "n_trials", and exposing the optuna solver.

I suspect most of the impedance matching work would revolve around the "scoring" and "refit" parameters, and what happens when there are multiple scorers

GaelVaroquaux:
>   • can we list the parameters of make_iter_search here (or in another issue)? I'm guessing only n_iter, random_state, n_jobs, refit`, and the CV params (cv and scoring)?

Let's start with a short list and consider adding stuff later. It's much easier to add stuff than change / rename

>   • what other methods do we plan to add in the long run?

I don't know; it's love to support scikit-learn's successive halfing

>   • is make_grid_search also replaced by make_iter_search?

Not for now, I would say: it differs strongly in that one cannot control the number of iterations / trials. We can always try to fold it back in later if we want.

> I suspect most of the impedance matching work would revolve around the "scoring" and "refit" parameters, and what happens when there are multiple scorers

Yes indeed. Let's start by supporting only a single scorers, we can add support for multiple scorers later.

rcap107:
How are we going to deprecate `make_randomized_search`? 

We could deprecate it with the next minor release (if we're having one), but maybe we should deprecate it directly in 0.7.0 when we add this PR

GaelVaroquaux:
> How are we going to deprecate make_randomized_search?

> We could deprecate it with the next minor release (if we're having one), but maybe we should deprecate it directly in 0.7.0 when we add this PR

What do you think is best?

rcap107:
> How are we going to deprecate make_randomized_search? We could deprecate it with the next minor release (if we're having one), but maybe we should deprecate it directly in 0.7.0 when we add this PR
> What do you think is best?

I think it's better if we deprecate in 0.7 and remove it in 0.8

GaelVaroquaux:
> I think it's better if we deprecate in 0.7 and remove it in 0.8

Great!

jeromedockes:
we are getting ahead of ourselves 😅 the new function needs to be implemented before deprecating anything

jeromedockes:
actually the name may not need to change: `make_iter_search` is confusing as all methods iterate over a bunch of stuff, that's not the important part. the difference between the 2 functions we are talking about is one does a full grid while the other makes some kind of random sampling (with several possible techniques), so `make_randomized_search` makes sense and avoids breaking backwards compatibility

jeromedockes:
(not related to the name but the optuna option can also be added to the `make_grid_search`; optuna also has a grid search implementation)

rcap107:
> actually the name may not need to change: `make_iter_search` is confusing as all methods iterate over a bunch of stuff, that's not the important part. the difference between the 2 functions we are talking about is one does a full grid while the other makes some kind of random sampling (with several possible techniques), so `make_randomized_search` makes sense and avoids breaking backwards compatibility

The reason for `make_iter_search` was because these searches have a number of iterations as a budget, which I still think makes sense, though I don't have strong feelings on the specific name.

rcap107:
> (not related to the name but the optuna option can also be added to the `make_grid_search`; optuna also has a grid search implementation)

Do we want to add that support though? I don't think people use optuna for grid search

GaelVaroquaux:
I don't have a strong opinion on names, and I'm perfectly fine recycling and extending the make_random_search function for now. It's a pragmatic way forward

jeromedockes:
> The reason for `make_iter_search` was because these searches have a number of iterations as a budget, which I still think makes sense, though I don't have strong feelings on the specific name. 

"n_iter" is not a very good name in the first place ("iter" is too generic, for example "n_trials" is better) so I don't think it should drive the rest of the API. again the difference is full grid vs sampling, but both involve iteration. actually the grid one really does iterate over a list of param combinations, whereas the random one samples them on the fly -- so "make_iter_search" better describes the grid search than the random search 🙃

also optuna allows setting a timeout rather than a number of trials, which I think should be exposed because it is very useful -- often we have a time budget (e.g. a 15-min coffee break or a fixed budget for fair comparison between different models/methods) rather than a specific number of trials we need

jeromedockes:
> Do we want to add that support though? I don't think people use optuna for grid search

you still may want it for the benefits of optuna like visualization and having progress information (see #1242 ). the sampling strategies are only one reason to use optuna and perhaps not even the most important one for most users. but the grid search one can be added later

GaelVaroquaux:
> "n_iter" is not a very good name in the first place ("iter" is too generic, for example "n_trials" is better)

That was by design, to be applicable to many settings.


> also optuna allows setting a timeout rather than a number of trials, which I think should be exposed because it is very useful -- often we have a time budget (e.g. a 15-min coffee break) rather than a specific number of trials we need

Yes, that's pretty cool, and I do think that we should expose it long term. I'm completely open to merging before we expose it and adding it later

jeromedockes:
> That was by design, to be applicable to many settings.

I understand, and for the parameter name it's not so bad because the context makes it less ambiguous, but for a function name out of context it's less appropriate IMO

jeromedockes:
> the sampling strategies are only one reason to use optuna 

put differently, I don't think the main disadvantage of the RandomizedSearchCV is the random sampling, which works pretty well, but rather the user experience that once you launch it it's a black box (you don't know how far along it is, when it will end, what is the best score and if it still improves) and if you interrupt it you lose all the already-completed work.

rcap107:
> > the sampling strategies are only one reason to use optuna
> 
> put differently, I don't think the main disadvantage of the RandomizedSearchCV is the random sampling, which works pretty well, but rather the user experience that once you launch it it's a black box (you don't know how far along it is, when it will end, what is the best score and if it still improves) and if you interrupt it you lose all the already-completed work.

I can say that being able to understand and track the training results is something that came up in various flavors in questions at Pydata (mostly related to mlflow, but the gist is the same)

examples/data_ops/1131_optuna_choices.py
```diff
@@ -0,0 +1,134 @@
...
+print(pred.skb.describe_param_grid())
+
+# %%
+# Load data for the example
+
+# %%
+from sklearn.datasets import fetch_california_housing
+from sklearn.model_selection import KFold
+
+env = {}
```
rcap107: I don't think we have other examples of this way of defining the environment elsewhere, so maybe it's better to add a short sentence explaining it

doc/modules/data_ops/validation/hyperparameter_tuning.rst
```diff
@@ -30,21 +30,26 @@ for ``alpha``.
...
-   uses default values for each of the choices. For numeric choices it is the
-   middle of the range, and for :func:`choose_from` it is the first option we
-   give it.
+   When we do :meth:`.skb.make_learner() <DataOp.skb.make_learner>`, the
+   pipeline we obtain does not perform any hyperparameter tuning. The pipeline
+   we obtain by default uses default values for each of the choices. For numeric
+   choices it is the middle of the range (unless an explicit default has been
+   set when creating the choice), and for :func:`choose_from` it is the first
+   option we give it. We can also obtain random choices, or choices suggested by
+   an optuna :class:`trial <optuna.trial.Trial>`, by passing the ``choose``
```
rcap107:
```suggestion
   an Optuna :class:`trial <optuna.trial.Trial>`, by passing the ``choose``
```

examples/data_ops/1131_optuna_choices.py
```diff
@@ -0,0 +1,134 @@
...
+#
+# Here we return a single score (R²), but multi-objective
+# optimization is also possible. Please refer to the Optuna documentation for
+# more information.
+
+# %%
+import optuna
+
+
+def objective(trial):
```
rcap107, line 157: This function is using a lot of parameters that are defined outside the scope and are not passed as arguments. It works here because they're defined in the outer scope, but is there a way of passing these arguments to the function? 

I think this detail may be missed by readers and I am not sure if it's possible to define an `objective` somewhere and use it elsewhere (out of the scope).

doc/modules/data_ops/validation/hyperparameter_tuning.rst
```diff
@@ -105,43 +110,60 @@ The default choices for an DataOp, those that get used when calling
...
 
->>> search.best_learner_.describe_params() # doctest: +SKIP
-{'α': 0.054...}
+>>> search.best_learner_.describe_params()  # doctest: +SKIP
+{'α': 0.000479...}
 
 It is also possible to use :meth:`ParamSearch.plot_results` to visualize the results
 of the search using a parallel coordinates plot.
 
+This could also be done with optuna, either by passing ``backend='optuna'`` to
```
rcap107:
```suggestion
This could also be done with Optuna, either by passing ``backend='optuna'`` to
```

doc/modules/data_ops/validation/hyperparameter_tuning.rst
```diff
@@ -105,43 +110,60 @@ The default choices for an DataOp, those that get used when calling
...
+
+
+Rather than fitting a randomized or grid search to find the best combination, it
+is also possible to obtain an iterator over different parameter combinations, to
+inspect their outputs or to have manual control over the model selection, using
+:meth:`.skb.iter_learners_grid() <DataOp.skb.iter_learners_grid>` or
+:meth:`.skb.iter_learners_randomized() <DataOp.skb.iter_learners_randomized>` (
+which yield the candidate pipelines that are explored by the grid and randomized
+search respectively), or with the ``choose`` parameter of
+:meth:`.skb.make_learner() <DataOp.skb.make_learner>`.
```
rcap107, line 162: Slight rewording
```suggestion
Rather than fitting a randomized or grid search to find the best combination, it
is also possible to obtain an iterator over different parameter combinations to
inspect their outputs or to have manual control over the model selection. This can
be done with :meth:`.skb.iter_learners_grid() <DataOp.skb.iter_learners_grid>` or
:meth:`.skb.iter_learners_randomized() <DataOp.skb.iter_learners_randomized>` (
which yield the candidate pipelines that are explored by the grid and randomized
search respectively), or with the ``choose`` parameter of
:meth:`.skb.make_learner() <DataOp.skb.make_learner>`.
```

examples/data_ops/1131_optuna_choices.py
```diff
@@ -0,0 +1,164 @@
+"""
+.. currentmodule:: skrub
+.. _example_optuna_choices:
+
+Tuning DataOps with Optuna
+==========================
+
+This example shows how to use `Optuna
+<https://optuna.readthedocs.io/en/stable/>`_ to tune the hyperparameters a
```
rcap107: "in a", "with a"?

examples/data_ops/1131_optuna_choices.py
```diff
@@ -0,0 +1,164 @@
...
+skrub :class:`DataOp`. As seen in the previous example, skrub DataOps can contain
+"choices", objects created with :func:`choose_from`, :func:`choose_int`,
+:func:`choose_float`, etc. and we can use hyperparameter search techniques to
+pick the best outcome for each choice. Performing this search with Optuna
+allows us to benefit from its many features such as state-of-the-art search
```
rcap107:
```suggestion
allows us to benefit from its many features, such as state-of-the-art search
```

examples/data_ops/1131_optuna_choices.py
```diff
@@ -0,0 +1,164 @@
...
+
+# %%
+# This allows us to use Optuna's reporting capabilities provided in
+# `optuna.visualization
+# <https://optuna.readthedocs.io/en/stable/reference/visualization/>`_ or
+# `optuna-dashboard
+# <https://optuna-dashboard.readthedocs.io/en/latest/getting-started.html>`_.
+
+# %%
+import optuna
```
rcap107, line 113: I just noticed that there was no need to import optuna before, setting `backend="optuna"` was enough to do the trick.

I wonder if this should be highlighted in the text.

skrub/_data_ops/_estimator.py
```diff
@@ -180,9 +180,18 @@ def get_params(self, deep=True):
     def set_params(self, **params):
```
rcap107, line 180: I'm trying to understand this function. 

Here, `set_params` is calling itself recursively and at each call it checks whether a `data_op` is in the parameters. 

What I don't understand is the dictionary generation: isn't the result of the dictionary calling `to_id` and `to_idx` a mapping between integer values? I am guessing this is the step that is picking the specific configuration for the specific hyperparameter, but I am not sure that's the correct interpretation.

As for the diff, if I'm understanding it correctly, it's due to the fact that the parameters used by optuna have a naming scheme that's different from that used in the data ops. 

Is this right?

skrub/_data_ops/_estimator.py
```diff
@@ -47,7 +46,8 @@
     "multimetric_",
 ]
 _SEARCH_FITTED_ATTRIBUTES = _SKLEARN_SEARCH_FITTED_ATTRIBUTES_TO_COPY + [
-    "best_learner_"
+    "best_learner_",
+    "refit_",
```
rcap107, line 50: why is this needed?

skrub/_data_ops/_estimator.py
rcap107: Now that I'm looking at the _ChoiceEvaluator I can follow better what is happening here

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -288,10 +290,138 @@ def predict(self, X):
...
+        all_cf, all_ci = set(), set()
+        for _ in range(200):
+            random = dop.skb.make_learner(choose=choose)
+            cf, ci = random.fit_transform({})
+            all_cf.add(cf)
+            all_ci.add(ci)
+        assert all_cf == {"a", "b", "c"}
+        assert all_ci == set(range(10, 16))
+    for choose in [check_random_state(1), "random (1 ) "]:
+        seeded = dop.skb.make_learner(choose=choose)
```
rcap107, line 322:
```suggestion
    # testing numpy RandomState and random state with seed 1
    # also testing random spacing in the seed number
    for choose in [check_random_state(1), "random (1 ) "]:
        seeded = dop.skb.make_learner(choose=choose)
```

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -288,10 +290,138 @@ def predict(self, X):
...
+
+def test_make_learner_choices_before_X():
+    # check that choices that are children of X (same for y) are forced to take
+    # their default value
+    msg = "The following choices are used in the construction of X or y.*choice_clamped"
+    with pytest.warns(UserWarning, match=msg):
+        X = (
+            skrub.as_data_op(skrub.choose_from(("a", "b", "c"), name="choice_clamped"))
+            + "x"
+        ).skb.mark_as_X()
```
rcap107, line 338: Running this line in a terminal gives me this message: 
```python
/Users/rcap/work/skrub/skrub/_data_ops/_evaluation.py:853: UserWarning: The following choices are used in the construction of X or y, so their value cannot be tuned because they are needed outside of the cross-validation loop. They will be clamped to their default value: [choose_from(('a', 'b', 'c'), name='choice_clamped')]
```

What I am understanding from the test is that the choice is clamped to the default value of the `choose_from`, which should be `a` because it's the first. From the message, this does not come true at all. Is it a bug?

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -288,10 +290,138 @@ def predict(self, X):
...
+        return {k: v for k, v in learner.get_params().items() if k != "data_op"}
+
+    dop = skrub.as_data_op(
+        (skrub.choose_from(("a", "b", "c")), skrub.choose_int(10, 15, default=11))
+    )
+
+    sampler = optuna.samplers.TPESampler(seed=0)
+    study = optuna.create_study(sampler=sampler)
+
+    learner = dop.skb.make_learner(choose=study.ask())
```
rcap107, line 363: How were the values of `a` and `b` chosen? Is it a result of setting seed=0? Could you add a comment to explain this?

skrub/_data_ops/tests/test_estimators.py
```diff
@@ -329,11 +344,121 @@ def test_multimetric():
...
+        )
+        search = data_op.skb.make_randomized_search(
+            n_iter=5, cv=2, random_state=0, backend=randomized_search_backend
+        ).fit(data)
+    assert search.best_score_ > 0.6
+    assert search.results_.shape[0] == 5
+    assert search.results_["mean_test_score"].isna().any()
+    # the search should pick logistic regression
+    assert search.best_learner_.get_params()["data_op__1"] == 0
+
```
rcap107: making this a bit more explicit
```suggestion
# setting error_score="raise" to make sure the exception
# from BadClassifier is raised and caught
```

skrub/_data_ops/_optuna.py
```diff
@@ -0,0 +1,401 @@
...
+        pass
+    return scorer
+
+
+class OptunaParamSearch(_BaseParamSearch):
+    """Learner that evaluates a skrub DataOp with hyperparameter tuning.
+
+    This class is not meant to be instantiated manually, ``OptunaParamSearch``
+    objects are created by calling :meth:`DataOp.skb.make_randomized_search()`
+    on a :class:`DataOp`.
```
rcap107:
```suggestion
    on a :class:`DataOp`, after setting ``backend="optuna"``.
```

examples/data_ops/1131_optuna_choices.py
jeromedockes: why not, not sure what to say or where though

examples/data_ops/1131_optuna_choices.py
jeromedockes: it is very common to use global variables like `skrub` or `pred` here so I think it should be ok

> I am not sure if it's possible to define an objective somewhere and use it elsewhere

yes it's not a problem (python like most languages uses lexical scope)

skrub/_data_ops/tests/test_data_ops.py
jeromedockes: I'm not sure I understand, what is a bug? when the choice is before X and y it should not be tuned. so we issue a warning and freeze it to the default. this is what we assert below: you can see that the first letter is always "a", whereas the one after "x" is allowed to vary and take the possible choice outcomes a, b, c

skrub/_data_ops/tests/test_data_ops.py
jeromedockes: they are chosen by the optuna trial study.ask() . yes they should always be the same as the sampler is seeded, I'll add a comment

skrub/_data_ops/_estimator.py
jeromedockes: the refit_ attribute is needed to sort the results table to show the best runs first (refit_ can be a metric name)

skrub/_data_ops/_estimator.py
jeromedockes: yes this is basically parsing the param names and values to set them on the data op. in scikit-learn set params for nested objects they are chained with `__` so we remove the `data_op__` part. for optuna the names and values we pass to optuna are chosen so that the optuna logs, plots etc will be informative so we remove the part that was just added for the sake of optuna display. I'll add comments to explain this

jeromedockes (commented):
thanks for the review @rcap107 :)

skrub/_data_ops/tests/test_data_ops.py
rcap107: Yes, I explained myself poorly. I understand what the code is doing, but I think that the warning message is not clear: I would rather see the default value of the choose_from, rather than `choose_from(...)`. The current error message does not tell me (at least, explicitly) what it means to "clamp to the default value".

jeromedockes: ok I understand now. I'll improve the warning message in another PR as it's a bit orthogonal to this one

jeromedockes: although here the important thing is not really the default values, but the choices themselves (their names or, if there aren't any, their repr) so that the user can easily find them in their code and figure out why they are above X in the graph and if / how to fix that. the message is that we should either remove the choice or rearrange the pipeline or position of mark_as_X() so that it does not get clamped anymore, regardless of the value it gets clamped to

rcap107: > ok I understand now. I'll improve the warning message in another PR as it's a bit orthogonal to this one

Makes sense to have it in a separate PR 

> although here the important thing is not really the default values, but the choices themselves (their names or, if there aren't any, their repr) so that the user can easily find them in their code and figure out why they are above X in the graph and if / how to fix that. the message is that we should either remove the choice or rearrange the pipeline or position of mark_as_X() so that it does not get clamped anymore, regardless of the value it gets clamped to

Yes that makes sense, but I don't think that the current message is conveying that information properly (another one of the error messages that should be reported in more detail in the documentation #1294 )

jeromedockes:
circleci failure is unrelated and will be fixed by https://github.com/skrub-data/skrub/pull/1794

GaelVaroquaux:
I'm looking at this PR. It's really amazing, thank you!

I'll give some feedback as I go, as I worry that I'll be interrupt. Sorry for the resulting mess and crowding.

**Feedback 1**: I worry that we are adding an example with a run-time of 4mn. We have another PR doing this. it's not going to scale well, and soon it will impact negatively skrub development. Would it be possible to play a bit wit the parameters of the example (for instance number of trials) to decrease the cost of this new example a bit?

Maybe use a faster learner (I love ExtraTrees for instance, or even a Logistic?).

**Feedback 2**: I see that we are tuning the number of estimators of RandomForests. I don't consider this as the most important hyper parameter. I typically consider the depth as the most important feature to tune.

doc/modules/data_ops/validation/tuning_with_optuna.rst
```diff
@@ -0,0 +1,222 @@
...
+>>> from sklearn.feature_selection import SelectKBest
+>>> from sklearn.ensemble import RandomForestClassifier
+>>> from sklearn.dummy import DummyClassifier
+
+>>> X_a, y_a = make_classification(random_state=0)
+>>> X, y = skrub.X(X_a), skrub.y(y_a)
+>>> selector = SelectKBest(k=skrub.choose_int(4, 20, log=True, name='k'))
+>>> logistic = LogisticRegression(C=skrub.choose_float(0.1, 10.0, log=True, name="C"))
+>>> rf = RandomForestClassifier(
+...    n_estimators=skrub.choose_int(3, 30, log=True, name="N 🌴"),
```
GaelVaroquaux: IMHO, tuning the n_estimator is not very good practice. It typically explores a tradeoff between compute and accuracy, with addition evaluation noise added on top

examples/data_ops/1131_optuna_choices.py
```diff
@@ -0,0 +1,178 @@
...
+from sklearn.ensemble import HistGradientBoostingRegressor, RandomForestRegressor
+
+import skrub
+
+hgb = HistGradientBoostingRegressor(
+    learning_rate=skrub.choose_float(0.01, 0.6, log=True, name="learning_rate")
+)
+rf = RandomForestRegressor(
+    n_estimators=skrub.choose_int(20, 100, name="n_estimators"),
+)
```
GaelVaroquaux, line 48: In the interest of compute time, could we replace one of these two by a logistic regression?

skrub/_data_ops/_evaluation.py
```diff
@@ -950,39 +952,91 @@ def set_params(data_op, params):
...
-    """Helper for `chosen_or_default_outcomes`."""
+def choice_default(choice_id, display_name, choice):
+    if isinstance(choice, _choosing.Choice):
+        return choice.chosen_outcome_idx or 0
+    else:
+        return choice.chosen_outcome_or_default()
 
-    def run(self, data_op):
+
+def random_choice(random_state):
```
GaelVaroquaux, line 962: It would be great to have one-liner doc string on these function to help future maintainers understand the code.

skrub/_data_ops/_evaluation.py
```diff
@@ -950,39 +952,91 @@ def set_params(data_op, params):
...
+
+    def policy(choice_id, display_name, choice):
+        if hasattr(choice, "rvs"):
+            return choice.rvs(random_state=random_state)
+        return int(random_state.randint(len(choice.outcomes)))
+
+    return policy
+
+
+def optuna_suggestion(trial):
```
GaelVaroquaux, line 974: It would be great to have one-liner doc string on these function to help future maintainers understand the code.

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1620,6 +1634,34 @@ def make_learner(self, *, fitted=False, keep_subsampling=False):
...
+              the random seed: for example 'random(0)' sets it to 0. 'random()'
+              is the same as 'random'.
+            - an instance of :class:`numpy.random.RandomState`. Same as 'random',
+              but the provided RandomState is used to sample values.
+            - an instance of :class:`optuna.Trial <optuna.trial.Trial>` or
+              :class:`optuna.FrozenTrial <optuna.trial.FrozenTrial>`. It is
+              used to suggest values for
+              the choices.
+
+            Note none of these options picks the best choice value according to
```
GaelVaroquaux: I'm not sure that I understand this sentence well. Do you mean:
```suggestion
            Note that none of these options picks the best choice value according to
```
?

examples/data_ops/1131_optuna_choices.py
jeromedockes: I used a ExtraTreesRegressor and a Ridge and reduced the number of trials

jeromedockes:
Thanks for the review @GaelVaroquaux !

- example time: I reduced the number of trials, and now use extra trees & ridge. it is still close to 2 minutes, LMK what you think. we can look for another dataset, subsample this one, or tune only the ridge's alpha instead of also including a better model. 
- tuning n_estimators: yes you are right it's not a good hyperparam to tune, I think I just picked the first from the estimator's parameter list 😅 . I now use the `min_samples_leaf` (it also controls the depth in a way, as you suggested, but it is a bit simpler to use and visualize in the example because it is always a number whereas depth has the None option). The n_estimators is tuned in a few other places in the main branch that we can tackle in a separate documentation-only pr

I also edited a bit the user guide page (@rcap107 wrote the original version actually, thanks Riccardo!) . Part of it is very similar to the sphinx gallery example but that may not be a problem 🤔 .

GaelVaroquaux:
Thanks heaps for your changes!! `min_sample_leaf` is great!

Regarding compute time:

> it is still close to 2 minutes, LMK what you think.

A factor 2 is really good to get. Thanks a lot. If we can get a bit more, it's better

> we can look for another dataset, subsample this one

Do you want to try this to see if you can get something that is still convincing and closer to 1mn?

jeromedockes:
ok I just sampled half of the dataset, the examples takes just under a minute now on circleci

GaelVaroquaux:
Thank you so much @jeromedockes

jeromedockes:
thank you both for the reviews & discussion :) looking forward to the upcoming release!

PR #1664, opened 2025-10-07
MAINT - Using pixi for nightlies and fixing failing nightly tests

### LINKED ISSUE #1173: pixi environment using the nightly builds?
not sure if it is possible as I see in the testing.yml the nightly builds are installed with pip but @glemaitre do you think we could have a pixi environment for those?

### LINKED ISSUE #1586: Test nightly dependencies fails all the time
The CI test for nightly dependencies always fails with this error: 

```sh
Run pytest -vsl --cov=skrub --cov-report=xml skrub -n 3
  pytest -vsl --cov=skrub --cov-report=xml skrub -n 3
  shell: /usr/bin/bash {0}
  env:
    pythonLocation: /opt/hostedtoolcache/Python/3.11.13/x64
    PKG_CONFIG_PATH: /opt/hostedtoolcache/Python/3.11.13/x64/lib/pkgconfig
    Python_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    Python2_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    Python3_ROOT_DIR: /opt/hostedtoolcache/Python/3.11.13/x64
    LD_LIBRARY_PATH: /opt/hostedtoolcache/Python/3.11.13/x64/lib
/home/runner/work/_temp/462554e7-474a-4532-a961-d286de1e4390.sh: line 1: pytest: command not found
```

auguste-probabl:
The strategy is to define a new pixi environment where the PyPI indexes provide the nightly builds. This is made difficult by pixi giving priority to conda channels rather than PyPI indexes when both are available. The solution I've found is to explicitly disable and redefine the dependencies, which is repetitive and thus error-prone.

Maybe pixi's [`conda-pypi-map`](https://pixi.sh/latest/reference/pixi_manifest/#conda-pypi-map-optional) could help; alternatively, maybe it's possible to find nightly dependencies in conda channels (after all, the scientific python ones do come from Anaconda)?

Closes #1173 and likely also closes #1586

### DIFF
--- a/skrub/tests/test_on_subframe.py
+++ b/skrub/tests/test_on_subframe.py
@@ -71,8 +71,12 @@ def test_empty_output(df_module, use_fit_transform):
         out = transformer.fit_transform(df)
     else:
         out = transformer.fit(df).transform(df)
+    # Selecting no columns to have an empty dataframe
     selected = s.select(df, ())
 
+    # I need to add a special case for pandas 3.0 here because the type of the
+    # empty dataframe with pandas 3.0 is different from that of out, but here
+    # we don't care about that dtype.
     if sbd.is_pandas(df) and parse(pd.__version__).major >= parse("3.0.0").major:
         out = sbd.to_numpy(out)
         selected = sbd.to_numpy(selected)


auguste-probabl:
I can't tell why nightly tests fail; my first guess would be that some dependency is missing from the `nightly-dependencies` environment, but I don't have evidence supporting this.

rcap107:
Hi @auguste-probabl, I think this PR is already useful even if it is not perfect. I'm already glad I can run nightly tests with it! 

I can reproduce the tests failing on my machine, I'll look into them

rcap107:
Looking at the failed tests, it seems like the vast majority are doctests or unit tests that are failing because a RejectColumn error is raised. From what I can tell, the reason is the migration to `str` in pandas 3.0. At the moment, the dataframe API is checking for "object", so we'll have to update that for the new version (not looking forwards to that...).

https://pandas.pydata.org/pandas-docs/stable/user_guide/migration-3-strings.html#string-migration-guide

There are some additional tests that are failing and that might just be issues with other dependencies. 

So what's happening is that nightly tests are failing because nightly dependencies will break things, so the tests are working as intended :see_no_evil: 

I think we should merge this and start hammering whatever tests are being broken afterwards

WDYT @jeromedockes @GaelVaroquaux @glemaitre

rcap107:
I'll be using this PR to fix the bugs in the nightlies and merge this PR once the bugs have been addressed

rcap107:
Unsuprisingly, fixing some tests broke some others. 

Noting down things that are still failing:
- A bunch of doctests where we print pandas series or dataframes, where the dtype is `object` on old versions and `str` in pandas 3.0
- `test_estimator` fails in all cases (I think because of some change in scikit-learn 1.8)
- `test_parallel_coord_plot` is failing for some reason I haven't figured out yet
- `test_common` is failing because of the pandas changes
- `selectors` fail because of the pandas changes
- Same for `test_clean_categories`, `test_empty_output`, `test_table_vectorizer`, `test_to_str`

I'd like to have a way of replacing the values of certain docstrings depending on the version of pandas, but I couldn't find a decent way of doing it, which means I might have to use doctest +SKIP instead.

GaelVaroquaux:
> I'd like to have a way of replacing the values of certain docstrings depending on the version of pandas, but I couldn't find a decent way of doing it, which means I might have to use doctest +SKIP instead.

I would first try to use ELLIPSIS

jeromedockes:
maybe `convert_dtypes()` can help to have more consistent outputs by using the same ("nullable") dtypes across versions?

jeromedockes:
we could also consider running the doctests only in some environments (e.g. those with pandas < 3 for now), while making sure that all the behavior shown in doctests is also covered by unit tests (which run in all the environments, but can afford more boilerplate to handle discrepancies between versions)

see the `--doctest-modules` pytest option currently set in pyproject.toml

rcap107:
I had to update the version of python because with py311 I was not able to pull the latest package versions (also it makes more sense to have a later version of python to test nightlies).

rcap107:
This PR is blocked by #1618: the failing tests are in `test_table_vectorizer`, and the other PR is making a lot of changes in the same file so we should merge that one first.

auguste-probabl:
:raised_hands:

skrub/_dataframe/_common.py
```diff
@@ -803,8 +803,20 @@ def is_string(col):
 
 @is_string.specialize("pandas", argument_type="Column")
 def _is_string_pandas(col):
+    # Pandas 3.0 introduces str dtypes, which are not parsed as "object"
+    if (
+        parse_version(pd.__version__).base_version
+        >= parse_version("3.0.0").base_version
+    ):
+        # (╯°□°)╯︵ ┻━┻
```
GaelVaroquaux:
```suggestion
```

skrub/_dataframe/tests/test_common.py
```diff
@@ -92,7 +93,18 @@ def test_is_column(df_module):
 
 def test_to_list(df_module):
     col = ns.col(df_module.example_dataframe, "str-col")
-    assert ns.to_list(col) == ["one", None, "three", "four"]
+    if ns.is_pandas(col) and parse(pd.__version__).major >= parse("3.0.0").major:
+        # (╯°□°)╯︵ ┻━┻
```
GaelVaroquaux:
```suggestion
```

skrub/datasets/tests/test_utils.py
```diff
@@ -58,3 +64,46 @@ def test_get_data_home_with_envar(monkeypatch, tmp_path):
 
         assert get_data_home() == dirpath
         assert dirpath.exists()
+
+
+def test_extract_archive_exception_unlink_called():
```
GaelVaroquaux, line 69: I'm a bit confused as to what these have to do with the subject of the PR, but that's not very important

skrub/tests/test_on_subframe.py
```diff
@@ -70,7 +71,16 @@ def test_empty_output(df_module, use_fit_transform):
         out = transformer.fit_transform(df)
     else:
         out = transformer.fit(df).transform(df)
-    df_module.assert_frame_equal(out, s.select(df, ()))
+    selected = s.select(df, ())
```
GaelVaroquaux, line 75: It would be useful to add a comment explaining what the above line is doing (I cannot tell)

skrub/datasets/tests/test_utils.py
rcap107: They were an attempt at getting coverage to pass, but it did not work

I can remove the tests if you prefer

GaelVaroquaux:
Merged, thanks!

PR #1666, opened 2025-10-09
FIX - Implementing the default `get_feature_names_out` method

rcap107:
This PR adds a default `get_feature_names_out` method to `SingleColumnTransformer`. The default method returns the value of `self.all_outputs_`. I added `all_outputs_` to all public skrub transformers according to their specific use. 

To do: adding a proper test that checks that all new estimators have get_feature_names_out.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -39,7 +39,8 @@ Changes
 
 Bugfixes
 --------
-- Added a ``get_feature_names_out`` method to all transformers.
+- Fixed various issues with some transformers by adding ``get_feature_names_out``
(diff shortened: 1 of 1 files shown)


auguste-probabl:
Nice!

rcap107:
> Great!!
> 
> All good to me, but needs a changelog entry, because this is actually a fairly important fix (we had so many issues with this)

There is already one, I've expanded it a bit

GaelVaroquaux:
I must apologize for not seeing it

GaelVaroquaux:
Merged, thanks!

PR #1671, opened 2025-10-12
sanitize behavior when estimator passed to apply() is a dataop

jeromedockes:
this PR fixes 2 issues that could happen when the estimator passed to `.skb.apply()` is a data_op: when checking the existence of some methods and when accessing the scikit-learn tags. it is kind of an edge case but there are valid use cases, like switching to a dummy or faster estimator in preview mode or setting a name on the estimator to retrieve it more easily on an unfitted learner.

### DIFF
--- a/.github/workflows/check_stub_files_diff.yaml
+++ b/.github/workflows/check_stub_files_diff.yaml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
-      - uses: prefix-dev/setup-pixi@v0.9.1
+      - uses: prefix-dev/setup-pixi@v0.9.2
         with:
           pixi-version: v0.41.3
           frozen: true
--- a/.github/workflows/run-code-format-checks.yaml
+++ b/.github/workflows/run-code-format-checks.yaml
@@ -15,7 +15,7 @@ jobs:
 
     steps:
       - uses: actions/checkout@v5
-      - uses: prefix-dev/setup-pixi@v0.9.1
+      - uses: prefix-dev/setup-pixi@v0.9.2
         with:
           pixi-version: v0.41.3
           frozen: true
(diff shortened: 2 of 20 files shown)


skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -639,3 +633,92 @@ def test_get_vars():
...
+        X_a, y_a = make_regression(random_state=0)
+    else:
+        X_a, y_a = make_classification(random_state=0)
+    X_df = pd.DataFrame(X_a).rename(columns=str)
+    if has_preview:
+        X, y = skrub.X(X_df), skrub.y(y_a)
+    else:
+        X, y = skrub.X(), skrub.y()
+    if needs_data:
+
```
rcap107, line 658: Could you explain this condition?

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -639,3 +633,92 @@ def test_get_vars():
...
+            return {}
+
+        def set_params(self):
+            return self
+
+    learner = skrub.var("a").skb.apply(Twice()).skb.make_learner()
+    assert learner.fit_transform({"a": 1}) == 2
+    xy_learner = learner.__skrub_to_Xy_pipeline__({})
+    assert xy_learner._estimator_type == "transformer"
+    if hasattr(xy_learner, "__sklearn_tags__"):
```
rcap107, line 722: is `hasattr` here for compatibility? I don't see conditions that should change this

rcap107:
>  it is kind of an edge case but there are valid use cases, like switching to a dummy or faster estimator in preview mode or setting a name on the estimator to retrieve it more easily on an unfitted learner.

I feel like there should be an example of those use cases in the user guide because otherwise I think almost nobody will ever figure out those use cases on their own

It could be added at a later time, but still it should be there

jeromedockes:
Thanks @rcap107 , I tried to answer the questions directly in comments in the code. I agree about the need for advanced examples (but about this PR specifically, I think someone asked about something similar in the skrub discord)

Issue #1428, opened 2025-06-05
Reconsider supporting numpy arrays


GaelVaroquaux:
I'm starting to realize that skrub provides a lot of value even on numpy arrays. Two examples that I stumbled upon recently:

* Using expressions on numpy arrays, eg to answer the following stackoverflow question https://stackoverflow.com/questions/37685412/avoid-scaling-binary-columns-in-sci-kit-learn-standsardscaler
* Using the TableReport to have a quick view of a numpy array.

I think that we should reconsider supporting numpy arrays as input. I would convert them to pandas dataframe and consider this just as a convenience tool. However, I believe that it would help us get users.

rcap107:
Adding support to numpy arrays to the TableReport should not be too hard and would be useful.

For other use cases, I think it's easier if the user does the conversion to pandas or polars and then works on the data as a dataframe.

Nismamjad1:
working on it

PR #1676, opened 2025-10-15
FEAT - add support for numpy arrays to the TableReport

Nismamjad1:
This PR addresses https://github.com/skrub-data/skrub/issues/1428 by allowing the TableReport to handle 1d and 2d numpy arrays. 

This is done by converting the arrays to pandas dataframes.

### DIFF
--- a/skrub/_reporting/_table_report.py
+++ b/skrub/_reporting/_table_report.py
@@ -194,7 +194,9 @@ class TableReport:
                 dataframe = pd.Series(dataframe, name="0")
 
             elif dataframe.ndim == 2:
-                dataframe = pd.DataFrame(dataframe)
+                dataframe = pd.DataFrame(
+                    dataframe, columns=[str(i) for i in range(dataframe.shape[1])]
+                )
 
             else:
                 raise ValueError(


skrub/_reporting/_table_report.py
```diff
@@ -48,7 +51,7 @@ class TableReport:
 
     This class summarizes a dataframe, providing information such as the type
     and summary statistics (mean, number of missing values, etc.) for each
```
rcap107:
```suggestion
    This class summarizes a dataframe or numpy array, providing information such as
    the type and summary statistics (mean, number of missing values, etc.) for each
```

skrub/_reporting/_table_report.py
```diff
@@ -48,7 +51,7 @@ class TableReport:
...
-    column.
+    column. It converts numpy array to pandas DataFrame or Series.
```
rcap107:
```suggestion
    column. Numpy arrays are converted to pandas DataFrame or Series.
```

skrub/_reporting/_table_report.py
```diff
@@ -186,6 +189,19 @@ def __init__(
         max_plot_columns=None,
         max_association_columns=None,
     ):
+        if isinstance(dataframe, np.ndarray):
+            if dataframe.ndim == 1:
+                dataframe = pd.Series(dataframe, name="0")
+
+            elif dataframe.ndim == 2:
+                dataframe = pd.DataFrame(dataframe)
```
rcap107: It's better to provide explicit names for the columns, because otherwise pandas defines them as an integer range which is annoying to deal with, so something like

```python
dataframe = pd.DataFrame(dataframe, columns=[str(_) for _ in range(dataframe.shape[1]))
```

PR #1678, opened 2025-10-15
DOC: Cleaner, clarify numeric_dtype

GaelVaroquaux:
Clarify that numeric_dtype also applies to string columns that represent numbers


rcap107 (approved):
Thanks for the PR, this clarifies things  @GaelVaroquaux

PR #1682, opened 2025-10-16
fix test after scikit-learn 1.8

jeromedockes:
fix the nightly build test @rcap107


rcap107:
thanks a lot @jeromedockes

PR #1685, opened 2025-10-21
enhancement: clarify TableReport behavior and improve user guidance when plotting is skipped due to max_plot_columns

### LINKED ISSUE #1679: Expand the message when TableReport gives up because of too many columns
With too many columns, we get:
Image

We could:
- Give specific lines of code to change the setting, including the call to "skrub.set_option"
- Mention that there are two arguments that control associations and distributions

These should be done in additional different paragraphs, to space out things visually

basile-desjuzeur:
Closes #1679

Clarifies the user-facing alert in TableReport when plotting is skipped due to `max_plot_columns` exceeding the configured limit.


skrub/_reporting/_data/templates/base.css
```diff
@@ -346,3 +346,13 @@ article.wrapper {
     stroke: var(--color-text-info);
     stroke-width: var(--border-s);
 }
+
+.alert-info-dismissable {
+    position: relative;
```
jeromedockes: I think this is already the case (see the rule on line 313 above)

skrub/_reporting/_data/templates/base.css
```diff
@@ -346,3 +346,13 @@ article.wrapper {
...
+}
+
+.dismiss-container {
```
jeromedockes, line 351: not a very big deal but the button overlaps the text

Image


But actually as this is only shown in the tab that is otherwise empty, I think we don't need it to be dismissable anymore: we could just remove the button altoghether. Also, we probably don't need the blue background which makes links hard to read as you can see on the screenshot. so it could just be regular black text on white bg (or vice-versa in dark mode) without a dismiss button and the css can be simplified. Also, currently it is in a `flex` element, which results in some of the text being cropped in narrow screens. part of the css simplification should remove this `flex` here: 
```
<div id="skipped-plot-alert" class="alert-info-dismissable flex space-between"
```

all this can be done in a separate PR

skrub/_reporting/_data/templates/base.css
```diff
@@ -346,3 +346,13 @@ article.wrapper {
     stroke: var(--color-text-info);
     stroke-width: var(--border-s);
 }
+
+.alert-info-dismissable {
+    position: relative;
+}
```
jeromedockes:
```suggestion
```

jeromedockes (approved):
thanks a lot @basile-desjuzeur ! LGTM, there are further improvements that could be handled either in separate PRs or in this one

GaelVaroquaux:
Thanks a lot for the PR!

Small comments on form:
* We have a color issue, as the text of the link is not visible on the blue background
image
* To me (I have an attention deficit and struggled in school as a young kid :D ), this is a big block of scary text that I don't want to read. I wonder it we can break it a bit, varying the color. I suggest that we have the git of the message in the blue box, and the rest of the of message below. Maybe close the first div after "during report creation." and before "If this result was not intended"... And enclose the rest of the message in a second div with a "happy" color, say light green, or yellow (ideally we reuse color that we're already using elsewhere, as defined by a CSS variable

skrub/_reporting/_data/templates/column-summaries.html
```diff
@@ -20,16 +20,37 @@
...
-
     {% if summary.plots_skipped %}
     <div id="skipped-plot-alert" class="alert-info-dismissable flex space-between"
         data-manager="AlertDismissable">
         <div class="alert-content shrinkable-text">
             Plotting was skipped as the dataframe exceeded the
             <a href="https://skrub-data.org/stable/reference/generated/skrub.TableReport.html"><code>max_plot_columns</code></a>
-            limit set for the TableReport during report creation.
+            limit set for the <code>TableReport</code> during report creation.
+            <p>If this result was not intended, you can adjust this behavior in several ways:</p>
```
GaelVaroquaux:
```suggestion
            <p>You can adjust this behavior in several ways:</p>
```
I'm a huge believer in keeping the messages as short as possible

skrub/_reporting/_data/templates/column-summaries.html
```diff
@@ -20,16 +20,37 @@
...
+                <li>To change the default number of columns to be plotted during the current Python session, use <code>skrub.set_config</code>:
+                    <pre><code class="language-python">    from skrub import set_config
+    set_config(max_plot_columns=50)</code></pre>
+                </li>
+                <li>To make the change permanent, use an environment variable:
+                    <pre><code class="language-shell">    export SKB_MAX_PLOT_COLUMNS=50</code></pre>
+                </li>
+            </ul>
+            <p>
+            If performance or memory usage is a concern, note that <code>TableReport</code> accepts two parameters:
```
GaelVaroquaux:
```suggestion
            Note that <code>TableReport</code> accepts two parameters:
```
(that's me trying to make things shorter focusing on the most important messages)

rcap107:
I think we can merge this PR. Thanks @basile-desjuzeur

PR #1686, opened 2025-10-22
Ajout d'un exemple Skrub

GardeniaLIU:
test pour TableVectorizer de skrub


GaelVaroquaux:
What problem does this solve?
What is the goal of this pull request?

Issue #1680, opened 2025-10-16
ENH - change the behavior of the `numeric_dtype` parameter of the `Cleaner`


rcap107:
Currently, when a column with "string" dtype that contains only numbers (`["1", "2", "3"]`) is transformed by the `TableVectorizer` it is converted automatically to float32. In the `Cleaner` this behavior is off by default, and it can be enabled by setting `numeric_dtype="float32"`. 

The current behavior is however unclear, and the string "float32" does not give a clear idea of what is being done. 

As a result, the behavior should be changed so that it is possible to turn off the automatic parsing in some way. We should use the numeric_dtype parameter to specify the behavior, and rename the values accepted by the argument.

The three behaviors we need to have are: 
- "ignore" => numbers with string dtype are ignored and kept as strings, numerical columns keep their dtype
- "parse" => numbers with string dtype are converted to float32 (current behavior), numerical columns keep their dtype
- "convert" => same as "parse", but convert numerical columns to float32 (the current behavior when numeric_dtype is set to "float32"

The names I'm suggesting here are not necessarily final and can be discussed. 

<details>
<summary>Original issue</summary>
I'm opening this issue to discuss the current behavior of the Cleaner and the TableVectorizer when they encounter string columns that contain only values that can be parsed as numbers. 
By default, the Cleaner leaves the column alone, but when numerical features are converted to float32 (by either the Cleaner or TableVectorizer), string columns are converted silently to numbers. 

```python
import pandas as pd
from skrub import TableVectorizer, Cleaner
data = {"a": ["1", "2", "3"]}
df = pd.DataFrame(data)
cleaned = Cleaner().fit_transform(df)
cleaned.dtypes
```
```
a    object
dtype: object
```

Setting `numeric_dtype="float32"`: 
```python
cleaned = Cleaner(numeric_dtype="float32").fit_transform(df)
cleaned.dtypes
```
gives 
```
a    float32
dtype: object
```

By default, the TableVectorizer converts everything to float32
```python
vec = TableVectorizer().fit_transform(df)
vec.dtypes
```

```
a    float32
dtype: object
```

This can be confusing to the user, especially if they use a numeric column converted to string to consider it categorical column (without actually converting the dtype). 

I think the automated conversion in its current form is often redundant with the pandas/polars parsers, and if it's done on a dataframe prepared by the user it might be misleading, as IDs may be treated as numeric when they should instead be considered categorical. 

[...]

PR #1687, opened 2025-10-22
feat: add a public to float

### LINKED ISSUE #1677: Make a public "ToFloat"
Expose a public "ToFloat" transformer, similar to ToCategorical and ToDateTime

MarieSacksick:
from the private to float.  

Closes #1677   

It is said in https://github.com/skrub-data/skrub/issues/1680 that it's not clear in the documentation of Cleaner & TableVectorizer the transformations applied. I found it quite explicit already actually, I added a bit of text to emphasize what was already said.

### DIFF
--- a/skrub/tests/test_table_vectorizer.py
+++ b/skrub/tests/test_table_vectorizer.py
@@ -27,6 +27,7 @@ from skrub._table_vectorizer import (
     _get_preprocessors,
 )
 from skrub._to_float import ToFloat
+from skrub.conftest import _POLARS_INSTALLED
 
 MSG_PANDAS_DEPRECATED_WARNING = "Skip deprecation warning"
 


rcap107:
> Thanks a lot @MarieSacksick , LGTM!
> 
> the only question I have is if we should rename it ToFloat (instead of ToFloat32), to leave open the possibility of later adding a parameter that controls if it outputs single or double-precision floats. (not sure if there is a usecase for that though)

I think we should, that way we won't have to deprecate things if we ever do need to support other float types. It's not a big change and I think the current behavior should be plenty clear from the docs.

MarieSacksick:
> I think we should, that way we won't have to deprecate things if we ever do need to support other float types. It's not a big change and I think the current behavior should be plenty clear from the docs.

Yes it's pretty clear from the docs, it's like the first line of the description of the transformer.  

I wondered if we should change also https://github.com/skrub-data/skrub/blob/main/skrub/_dataframe/_common.py#L777 from `to_float32` to `to_float`, but since it's private, I'd argue that it can be kept as is, at least for now.

rcap107:
> I wondered if we should change also [`main`/skrub/_dataframe/_common.py#L777](https://github.com/skrub-data/skrub/blob/main/skrub/_dataframe/_common.py?rgh-link-date=2025-10-27T12%3A57%3A43Z#L777) from `to_float32` to `to_float`, but since it's private, I'd argue that it can be kept as is, at least for now.

I think we can keep it as is, and worry about renaming it if we ever decide to add `to_float64` or something

rcap107:
Blocked by #1666 (I'd rather wait for that to be merged to avoid conflicts)

rcap107:
Could you please add a changelog entry @MarieSacksick?

MarieSacksick:
converting this back to draft to wait for https://github.com/skrub-data/skrub/pull/1666 to be merged before.

rcap107:
Thanks a lot @MarieSacksick

PR #1688, opened 2025-10-22
Improve some docstrings cross-referencing (see also) and error message

GaelVaroquaux:
Minor changes to improve usability

### DIFF
--- a/skrub/_select_cols.py
+++ b/skrub/_select_cols.py
@@ -117,7 +117,7 @@ class DropCols(TransformerMixin, BaseEstimator):
     See Also
     --------
     SelectCols : Selecting cols by name, dtypes, or general skrub selectors.
-    Cleaner: Can be used to drop columns with too many NaNs.
+    Cleaner: Can be used to drop columns with too many nulls (or NaNs).
 
     Examples
     --------


skrub/_select_cols.py
```diff
@@ -109,6 +114,11 @@ class DropCols(TransformerMixin, BaseEstimator):
         ``str``: ``"col_name"`` is the same as ``["col_name"]``. See the
         :ref:`selectors <user_guide_selectors>` user guide for more info on selectors.
 
+    See Also
+    --------
+    SelectCols : Selecting cols by name, dtypes, or general skrub selectors.
+    Cleaner: Can be used to drop columns with too many NaNs.
```
rcap107:
```suggestion
    Cleaner: Can be used to drop columns with too many nulls (or NaNs).
```
not sure about the wording (maybe just nulls?), but the Cleaner drops either, not only NaNs

GaelVaroquaux: I'm ok with adding "nulls", but to me, this is database speech, and I've never heard it in statistics, maths, physics, medicine... It's a term that doesn't mean anything to many users.

GaelVaroquaux:
This should now be good to go

Issue #1681, opened 2025-10-16
FEA - Add a `has_dtype` selector


rcap107:
Sometimes dataframe columns contain lists of values. 

```python
import pandas as pd
data = {"a": [["1", "2", "3"], ["4", "5", "6"]]}
df = pd.DataFrame(data)
```

We should have a selector that checks whether a series contains lists. By default, this kind of columns are converted to strings and encoded like strings, which is a behavior I am fine with. 

However, some users may want to perform specific transformations to list columns, and this is what the selector would help with.

kudos07:
Hi, I’d like to take this issue.
My plan is to add a select_list selector that detects columns containing list-like values (lists or tuples).
It’ll follow the same structure as existing selectors like select_categorical, with a simple .apply(isinstance(x, (list, tuple))) check and unit tests in test_selectors.py.

rcap107:
> Hi, I’d like to take this issue. My plan is to add a select_list selector that detects columns containing list-like values (lists or tuples). It’ll follow the same structure as existing selectors like select_categorical, with a simple .apply(isinstance(x, (list, tuple))) check and unit tests in test_selectors.py.

Hi @kudos07, thanks for taking this up. A few pointers:
- You should implement this by adding a`is_list` dispatched function in `skrub/__dataframe/_common.py` (similar to `is_categorical`), and add a test for it. 
- After that is done, you can add the selector and the proper test
- For this PR, we can just keep it simpler, and ignore lists of lists: so long as a cell contains a list, it counts (regardless of the content). It's the user's responsibility to deal with it. 
- A column is considered to be of type "list" only if *all* non-null values in the column are lists 
- At a quick glance, polars should be easy to implement because list columns have the List dtype, while with pandas you'll have to figure out a way to check the dtype. If you're working with numpy dtypes, then lists should be represented as ndarray, but for nullable dtypes it may be a whole different thing. 

I hope that's enough to get started.

rcap107:
To expand a bit on the final point, Pandas has both nullable and numpy dtypes (let's forget about pandas 3 for the moment...). Off the top of my head I don't know what's the difference between the two when it comes to lists. 

When tests are run, you should use the `df_module` provided in `conftest.py` to test all three types (polars, pandas nullable, pandas numpy) to make sure that everything is working as intended with all methods. 

That's for later, however, and I think we can look at that part in more detail when you get there.

jeromedockes:
[...]

PR #1690, opened 2025-10-23
FEA: Add select_list selector and is_list dispatch for list-like columns

kudos07:
Hi @rcap107, I’ve opened the PR for this feature.
It adds the dispatched is_list function (with pandas and polars specializations) and the new select_list selector to identify columns containing list-like values.
Tests for both have been added and pass across all backends. Please let me know if you’d like any tweaks or naming changes.


kudos07:
Hi @rcap107, I think the doctest for select_list might be failing because from skrub.selectors import select_list doesn’t find the symbol in __init__.py.

I guess the fix would be to explicitly add

```python

from ._selectors import select_list
```

in skrub/selectors/__init__.py.

I can push that small update if you’d like, but I wasn’t sure if you’d prefer to handle it later (since import * might already cover it after internal sync). Just let me know what you’d prefer.

rcap107:
Hi @kudos07, and thanks for the PR. 

After some discussions with @jeromedockes, I realized I underestimated how annoying this PR would be. The main issue is that "what is a list" is not a question with a straightforward answer depending on the platform (arrow/polars/pandas etc), and there are way too many variants and edge cases that should be considered. One solution would be what you did, which is using `isinstance` to check for what's inside each cell, but that is an expensive operation, and it might not be good enough.

Unfortunately, I think that we should drop this PR as it is not defined well, and may cause all sorts of trouble in dealing with the different libraries and all the possible definitions of "list".

kudos07:
Thanks @rcap107, that makes sense. I had not considered the full range of backend-specific edge cases, especially with Arrow and Polars.

It is completely fine to drop this if it is not well-defined right now. I really appreciate the discussion and your detailed feedback. It was still a great learning experience to understand how dispatch and selectors are structured.

If you ever revisit this idea with a clearer spec, maybe tied to Arrow list dtype or something similar, I would be happy to take another shot at it.

GaelVaroquaux:
@kudos07 , you have a really great attitude. Thanks a lot. We'll see about the future of this issue. 
In the mean time, prototyping is probably a good way or understanding the ways forward

rcap107:
> Thanks @rcap107, that makes sense. I had not considered the full range of backend-specific edge cases, especially with Arrow and Polars.
> 
> It is completely fine to drop this if it is not well-defined right now. I really appreciate the discussion and your detailed feedback. It was still a great learning experience to understand how dispatch and selectors are structured.
> 
> If you ever revisit this idea with a clearer spec, maybe tied to Arrow list dtype or something similar, I would be happy to take another shot at it.

Thank you for the effort! We can definitely revisit the issue in the future. 

I'll close this for now, so we should continue the discussion in #1681

PR #1693, opened 2025-10-24
MAINT - PEP 621 compliance

DimitriPapadopoulos:
The README does not need to be dynamic.

https://packaging.python.org/en/latest/specifications/pyproject-toml/


glemaitre (approved):
Indeed, we don't have any tools modifying this file.

Thanks. LGTM.

PR #1694, opened 2025-10-24
MAINT - let ruff find the target version

DimitriPapadopoulos:
Indeed ruff infers the target version from `project.requires-python`.

https://docs.astral.sh/ruff/settings/#target-version

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -6,11 +6,10 @@ repos:
     -   id: end-of-file-fixer
     -   id: trailing-whitespace
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    # Ruff version.
-    rev: v0.4.8
+    rev: v0.14.2
     hooks:
-    -   id: ruff
-        args: ["--fix", "--output-format=full"]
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
This PR has a conflict and cannot be merged.

However, before I consider merging it, what does the version change entail? Will it create conflicts with existing PRs because of changed rules? How do we check that?

DimitriPapadopoulos:
* This PR will update Ruff from 0.4.8 to current 0.14.2, which could result in more issues raised by Ruff. In practice, none have occurred in the current code base.
* The target Python version changed from 3.8 to 3.9, ~~will remain 3.9. No changes here.~~ which could result in more issues raised by Ruff. In practice, none have occurred in the current code base. In the future, the target Python version will be bumped automatically with `project.requires-python`.

**Edit:** The only exception to the above is [E721](https://docs.astral.sh/ruff/rules/type-comparison/), which I have added to the ignore list.

In short, chances there are more conflicts in PRs that in the current code base appear pretty dim. If needed, individual rules can be ignored, and applied at a later time.

Issue #1626, opened 2025-09-22
DOC - Switch from using rst to markdown in docs


rcap107:
Using RST makes updating the documentation more complicated than needed in a lot of cases. I think we should move from RST to Markdown once the documentation is in a "relatively stable" shape.

rcap107:
Listing here some resources:
- https://docsforadobe.dev/updating-legacy-docs/converting-rst-to-md/
- https://docs.readthedocs.com/platform/stable/guides/migrate-rest-myst.html
- https://stackoverflow.com/questions/2471804/using-sphinx-with-markdown-instead-of-rest
- https://www.sphinx-doc.org/en/master/usage/markdown.html

PR #1696, opened 2025-10-24
MAINT - Use PyGrep hooks

DimitriPapadopoulos:
Apply Repo-Review suggestion PC170:
https://learn.scientific-python.org/development/guides/repo-review/?repo=skrub-data%2Fskrub&ref=HEAD


DimitriPapadopoulos:
Might be not be needed if #1626 is adressed.

DimitriPapadopoulos:
If you agree with the spirit of this PR, I can fix incorrect back-quoting in RST files. However, if you plan on migrating to Markdown anyway, it might be better to close this PR instead of spending time on it.

Note that most issues occur in [`CHANGES.rst`](https://github.com/skrub-data/skrub/blob/main/CHANGES.rst) because commit messages are copied "as is" although they usually contain Markdown markup because that's what GitHub expects. I would therefore recommend migrating this specific file to Markdown instead of fixing it.

Other files are quite easy to fix.

rcap107:
> Might be not be needed if #1626 is adressed.

#1626 is unlikely to be worked on anytime soon, so until then this PR would be quite useful. I don't think we need to convert changes.rst to markdown either, so we can move on with his.

.pre-commit-config.yaml
```diff
@@ -1,15 +1,23 @@
 repos:
 -   repo: https://github.com/pre-commit/pre-commit-hooks
-    rev: v6.0.0
+    rev: v4.3.0
```
rcap107: PRs were merged out of order 
```suggestion
    rev: v6.0.0
```

.pre-commit-config.yaml
```diff
@@ -1,15 +1,23 @@
...
+    rev: v1.10.0
+    hooks:
+    -   id: rst-backticks
+        exclude: CHANGES.rst
+    -   id: rst-directive-colons
+    -   id: rst-inline-touching-normal
 -   repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.14.2
+    # Ruff version.
+    rev: v0.4.8
```
rcap107:
```suggestion
    rev: v0.14.2
```

.pre-commit-config.yaml
```diff
@@ -1,15 +1,23 @@
...
     hooks:
-    -   id: ruff-check
-        args: ["--fix", "--show-fixes", "--output-format=full"]
+    -   id: ruff
+        args: ["--fix", "--output-format=full"]
```
rcap107:
```suggestion
    -   id: ruff-check
        args: ["--fix", "--show-fixes", "--output-format=full"]
```

.pre-commit-config.yaml
```diff
@@ -20,3 +28,5 @@ repos:
     hooks:
     -   id: codespell
         exclude: .*/package-lock.json
+    additional_dependencies:
+    -   tomli
```
rcap107: Same here, was fixed in another PR 
```suggestion
```

rcap107 (approved):

PR #1701, opened 2025-10-24
chore: Remove dead code

### LINKED ISSUE #1702: Dead code used only in benchmark: what should be done with it?
Thanks to codecov, I found some deadcode in _utils:  
https://app.codecov.io/gh/skrub-data/skrub/blob/MarieSacksick%2Fskrub%3Aadd_tests/skrub%2F_utils.py 

The function [check_input](https://github.com/skrub-data/skrub/blob/60c2d20542586cdf8842465be7369a6e7d65e89d/skrub/_utils.py#L79) is not used anywhere except in an old [benchmark about the min hash encoder](https://github.com/skrub-data/skrub/blob/0.6.2/benchmarks/bench_minhash_batch_number.py#L25) which is a bit odd because it (re?)defines the class `MinHashEncoder`.  

Do you have a policy about this? I guess that the benchmark should still be runnable, and therefore the process would be to recode the benchmark to remove this function and either rely on new ones, or option 2, actually rely on the actual class `MinHashEncoder` instead of recoding it.

MarieSacksick:
As I was going through the code, found this comment that is not relevant anymore.   
Later, thanks to codecov, found that the function `check_duplicated_column_names` was not used anymore.  

Closes #1702.

### DIFF
--- a/skrub/_utils.py
+++ b/skrub/_utils.py
@@ -9,7 +9,6 @@ from typing import Iterable
 import numpy as np
 import sklearn
 from sklearn.base import BaseEstimator, clone
-from sklearn.utils import check_array
 from sklearn.utils.fixes import parse_version
 
 from skrub import _dataframe as sbd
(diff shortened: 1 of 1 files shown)


skrub/_utils.py
```diff
@@ -163,16 +135,6 @@ def get_duplicates(values):
     return duplicates
 
 
-def check_duplicated_column_names(column_names, table_name=None):
```
jeromedockes, line 166: indeed we now add a suffix to column names instead of raising an error when there are duplicates

jeromedockes:
thanks a lot @MarieSacksick , very useful cleanup to avoid the codebase gathering dust over time!

PR #1703, opened 2025-10-24
Enforce ruff/flake8-bugbear rules (B)

DimitriPapadopoulos:
Apply Repo-Review suggestion RF101:
https://learn.scientific-python.org/development/guides/repo-review/?repo=skrub-data%2Fskrub&ref=HEAD

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -710,7 +710,7 @@ environments:
       - pypi: https://files.pythonhosted.org/packages/8a/67/f95b5460f127840310d2187f916cf0023b5875c0717fdf893f71e1325e87/plotly-6.5.2-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/0a/49/737c1a6273c585719858261753da0b688454d1b634438ccba8a9c4eb5aab/polars-1.38.1-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/ae/a2/a00defbddadd8cf1042f52380dcba6b6592b03bac8e3b34c436b62d12d3b/polars_runtime_32-1.38.1-cp310-abi3-macosx_10_12_x86_64.whl
-      - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/pyarrow/24.0.0.dev105/pyarrow-24.0.0.dev105-cp314-cp314-macosx_12_0_x86_64.whl
+      - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/pyarrow/24.0.0.dev124/pyarrow-24.0.0.dev124-cp314-cp314-macosx_12_0_x86_64.whl
       - pypi: https://files.pythonhosted.org/packages/7e/32/a7125fb28c4261a627f999d5fb4afff25b523800faed2c30979949d6facd/pydot-4.0.1-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/10/bd/c038d7cc38edc1aa5bf91ab8068b63d4308c66c4c8bb3cbba7dfbc049f9c/pyparsing-3.3.2-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/ec/57/56b9bcc3c9c6a792fcbaf139543cee77261f3651ca9da0c93f5c1221264b/python_dateutil-2.9.0.post0-py2.py3-none-any.whl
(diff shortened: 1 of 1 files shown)


skrub/_data_ops/_skrub_namespace.py
```diff
@@ -621,7 +623,7 @@ def select(self, cols):
         Examples
         --------
         >>> import skrub
-        >>> from skrub import selectors as s
+        >>> from skrub import selectors
```
rcap107: The intended way to call selectors is the following, so this instance and all the other changes in the docstrings like this should be reverted 
```suggestion
        >>> from skrub import selectors as s
```

skrub/_apply_to_cols.py
```diff
@@ -13,6 +13,7 @@
 
 __all__ = ["ApplyToCols", "SingleColumnTransformer", "RejectColumn"]
 
+_SELECTORS = selectors.all()
```
rcap107: `selctors.all()` isn't selecting all columns, it's selecting the `all()` selector, which selects all columns. 

```suggestion
# By default, select all columns
_SELECT_ALL_COLUMNS = selectors.all() 
```

rcap107: Same change in all the other files

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -41,6 +41,8 @@
 from ._subsampling import SubsamplePreviews, env_with_subsampling
 from ._utils import KFOLD_5, NULL, attribute_error
 
+_SELECTORS = selectors.all()
```
```suggestion
_SELECT_ALL_COLUMNS = selectors.all()
```

rcap107:
As a more general comment for this and all the other PRs that you opened, formatting rules should be chosen by the maintainers of the project, rather than added in a dozen of separate PRs that need to be reviewed one by one. 

In this PR, a large part of the diff is "whatever", the `_SELECTORS` change was made while misunderstanding the code, and while reviewing it I ended up misunderstanding the code myself, meaning that if it had been a more critical change it would have introduced a bug in the code. 

The suggested rules are way too restrictive for new contributors, so that part of the PR is also not going to be merged, and should _definitely_ have been discussed in an issue rather than added here. 

And this is only one of 20 very similar PRs that have been opened, and that I have to spend time on reviewing, making sure that none of the small changes that have been made across a bunch of files can lead to issues of some kind. 

In the future, _please_ consider the time and effort that goes into reviewing all of this. I will still try to review the PRs that have already been opened, and track the suggested changes in an issue for discussion and possible implementation. 

The proper way of contributing to the repo would be opening a meta-issue that collects a series of sub-issues, so that it is possible to decide whether specific changes are warranted. This would take less time than reviewing PRs, and would not risk introducing bugs in the code. 

If it does not look like future PRs take this comment into consideration, I don't exclude closing any new PR like this without even looking at what's inside.

DimitriPapadopoulos:
* I would recommend keeping [B004](https://docs.astral.sh/ruff/rules/unreliable-callable-check/), it does result in more correct code, although this should seldom be an issue in practice:
  > Using `hasattr` is an unreliable mechanism for testing if an object is callable. If `obj` implements a custom `__getattr__`, or if its `__call__` is itself not callable, you may get misleading results.
* [B007](https://docs.astral.sh/ruff/rules/unused-loop-control-variable/) is indeed debatable.
* [B010](https://docs.astral.sh/ruff/rules/set-attr-with-constant/) improves code consistency and readability, but I can understand you'd rather ignore it to avoid frustrating new contributors.

skrub/_data_ops/_skrub_namespace.py
DimitriPapadopoulos: Both versions were available, I didn't know which one to choose. I'll switch to the one you prefer.

~~**Edit:** I may be mistaken, I cannot find a simple occurrence of `from skrub import selectors` any more, but the rationale was somehow similar.~~

DimitriPapadopoulos:
> Other changes do not look like they improve the code in a meaningful way (B007, B004, B010); we can stil implement the changes, however.

I have kept the changes in this PR, but these rules will be ignored in the future.

DimitriPapadopoulos:
Still need to fix this new error in a test:
```
FAILED skrub/tests/test_multi_agg_joiner.py::test_default_cols[pandas-nullable-dtypes] - AssertionError: assert [['rating', '...', 'movieId']] == [['rating', '...ng', 'genre']]
  
  At index 1 diff: ['rating', 'genre', 'movieId'] != ['movieId', 'rating', 'genre']
```
Columns should be sorted in the same order as in the initial `main_table`, shouldn't they?
https://github.com/skrub-data/skrub/blob/20598ed4ab41588f187af22a001ebdcf59a033df/skrub/tests/test_multi_agg_joiner.py#L9-L19

DimitriPapadopoulos:
Note that Ruff can automatically apply some rules, and pre-commit will enforce that without any additional work from contributors. Therefore, rules with automatic fixes are not a barrier, they result in consistent and readable code which can only help maintenance and new contributions.

Ruff provides automatic fixes for many [B](https://docs.astral.sh/ruff/rules/#flake8-bugbear-b) rules, including [B004](https://docs.astral.sh/ruff/rules/unreliable-callable-check/), [B007](https://docs.astral.sh/ruff/rules/unused-loop-control-variable/), [B010](https://docs.astral.sh/ruff/rules/set-attr-with-constant/).

DimitriPapadopoulos:
It looks like CI failed because https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/pandas/ was not available. Need to retry...

skrub/tests/test_multi_agg_joiner.py
```diff
@@ -348,7 +348,10 @@ def test_default_cols(main_table, df_module):
         keys=[["userId", "movieId"], ["userId"]],
     )
     multi_agg_joiner.fit(main_table)
-    multi_agg_joiner._cols == [["rating", "genre"], ["movieId", "rating", "genre"]]
+    assert multi_agg_joiner._cols == [
```
jeromedockes: [here](https://github.com/skrub-data/skrub/blob/eb7788242bdf71361e46d0b8ca56bd28a80716b3/skrub/_multi_agg_joiner.py#L353) is why this test randomly fails, add a sort

DimitriPapadopoulos:
Thank you for fixing the CI issues, I couldn't find the time to investigate.

PR #1704, opened 2025-10-24
Enforce ruff/flake8-comprehensions rules (C4)

DimitriPapadopoulos:


GaelVaroquaux:
Do we want to do that, or is this just adding friction to later devs?

I'm not convinced of the benefit

DimitriPapadopoulos:
The benefits are (i) readability through consistency and simplicity, and (ii) (automated) micro-optimisations all over the code base. While manual micro-optimisations are not worth it, applying rule sets automatically is well accepted or at least less frowned upon. I acknowledge this might deviate from the strict definition of lint, which would focus on actual errors, but the term has evolved. For example, [according to Wikipedia](https://en.wikipedia.org/wiki/Lint_(software)):
> Lint is the [computer science](https://en.wikipedia.org/wiki/Computer_science) term for a [static code analysis](https://en.wikipedia.org/wiki/Static_program_analysis) tool used to flag programming errors, [bugs](https://en.wikipedia.org/wiki/Software_bug), stylistic errors and suspicious constructs.[[4]](https://en.wikipedia.org/wiki/Lint_(software)#cite_note-4)

As for friction for later devs, it remains possible to disable a rule. Actually, (limited) friction with automatic tools looks like a good thing to me: it improves programming style and may even help understand what the code is actually doing or implies.

DimitriPapadopoulos:
[C408](https://docs.astral.sh/ruff/rules/unnecessary-collection-call/#unnecessary-collection-call-c408) (unnecessary-collection-call) is probably the most debatable rule,especially in terms of frequency and resulting dev friction, and I am happy to ignore it.

rcap107:
I'm fine with adding these rules, but I would like to add a few more ignored rules: 
- 403 and 405 remove the literal `set` when it's not needed, but sets are not widely used in the code base so I'd argue that having the literal makes the intent more evident at a glance
- Same goes for 409 and tuples

Other than that, I think the rules are useful.

DimitriPapadopoulos:
I would rather keep `C409` for this case  which really doesn't make sense (feed a tuple to a tuple):
`tuple((1, 2, 3))` →`(1, 2, 3)`
but then I presume you would like to allow this (feed a list to a tuple):
`tuple([1, 2, 3])`

DimitriPapadopoulos:
Rebased and ignoring C403, C405, C409.

PR #1705, opened 2025-10-24
Start applying ruff/flake8-simplify rules (SIM)

DimitriPapadopoulos:


skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -363,7 +363,7 @@ def test_apply_bad_params(why_no_wrap, bad_param):
     else:
         cols = s.all()
     how = "cols" if bad_param == "how" else how
-    allow_reject = True if bad_param == "allow_reject" else False
+    allow_reject = bad_param == "allow_reject"
```
rcap107: the current version is less readable 
```suggestion
    allow_reject = True if bad_param == "allow_reject" else False
```

skrub/tests/test_table_vectorizer.py
```diff
@@ -45,12 +45,10 @@ def type_equality(expected_type, actual_type):
...
-    if (isinstance(expected_type, object) or isinstance(expected_type, str)) and (
-        isinstance(actual_type, object) or isinstance(actual_type, str)
-    ):
-        return True
-    else:
-        return expected_type == actual_type
+    return (
+        isinstance(expected_type, (object, str))
+        and isinstance(actual_type, (object, str))
+    ) or expected_type == actual_type
```
rcap107: please revert this to the original as the suggested change is less readable

skrub/_data_ops/tests/test_data_ops.py
DimitriPapadopoulos: I don't find it less readable, perhaps because I'm used to C.

With that said, I have to agree the `SIM` rules are among the most debatable - and debated :smile:

skrub/tests/test_table_vectorizer.py
DimitriPapadopoulos: I will revert it, but I disagree.

DimitriPapadopoulos: Following the same logic, one would change:
```python
    else:
        return expected_type == actual_type
```
to:
```python
    elif expected_type == actual_type:
        return True
    else:
        return False
```

rcap107: That's a fair comment, but I think splitting the two conditions, rather than grouping them with the `or` is more readable

rcap107:
Merged, thanks @DimitriPapadopoulos

PR #1706, opened 2025-10-24
Start applying ruff/Perflint rules (PERF)

DimitriPapadopoulos:
PERF401 Use `list.extend` to create a transformed list


rcap107:
This PR makes the code less readable, so it's not needed

DimitriPapadopoulos:
@rcap107 This is not about readability. See for example https://github.com/pypa/setuptools/pull/4449#issuecomment-2239482683:
> For me, the issue isn't about performance or readability, but about logical and semantic safety.

PR #1707, opened 2025-10-24
Apply ruff/refurb rules (FURB)

DimitriPapadopoulos:
FURB122 Use of `write` in a for loop


GaelVaroquaux:
What problem is this fixing? What value is this adding?

Relevant read:
https://rgommers.github.io/2019/06/the-cost-of-an-open-source-contribution/

DimitriPapadopoulos:
While it's a micro-optimisation, enforcing the rule consistently over the whole code base may slightly improve I/O performance and will certainly improve consistency.

It occurs to me that I haven't enforced the relevant rule BTW.

PR #1713, opened 2025-10-28
chore: Add a pixi task to open documentation

MarieSacksick:
No related issue.  
As I was working with the docs, I found it more useful to have this.   
A bit nitpick, can be discarded if you don't find it relevant.


rcap107:
Hi @MarieSacksick, as far as I know `open` is a Mac-only command. On Linux, you'd use something like `xdg-open` instead, and I have no idea what's the equivalent on Windows. Aside from that, I think this is more of a command that should be set up by the user using something like `alias`, rather than being added to pixi.

MarieSacksick:
`open` works on linux too. However, indeed, I don't know for windows. I'll close this PR then.

Issue #1660, opened 2025-10-06
DOC: Add different versions to the switcher


GaelVaroquaux:
The pydata sphinx theme has a version switcher, but right now we are using it only to switch between dev and stable. We should add a few older versions here.

rcap107:
I think we should go back to 0.4 at most

MarieSacksick:
I can try this one!

PR #1716, opened 2025-10-29
chore: Enable switch to old versions

MarieSacksick:
Closes https://github.com/skrub-data/skrub/issues/1660.  

(test to see if this is a good way to do. It's not how it's done in scikit-learn nor in skore, but it's how it's documented in [sphinx](https://pydata-sphinx-theme.readthedocs.io/en/stable/user_guide/version-dropdown.html))

### DIFF
--- a/RELEASE_PROCESS.rst
+++ b/RELEASE_PROCESS.rst
@@ -58,7 +58,7 @@ Meanwhile, preparing the post-released PR
     - VERSION.txt: update to ``0.3.dev0`` (the next minor).
     - CHANGES.rst: create a header for the new entries ("ongoing development").
     - doc/version.json: update the version numbers of the stable release and dev branch.
-      Remove the oldest version to keep the 2 previous ones only.
+      Don't forget to add an entry for the previously stable version.
 
 
 The doc update has succeeded


MarieSacksick:
After testing, I can check that indeed the switcher adapts and displays correctly the former versions. However, as the former versions exist in the server rendering the docs, for now I have a 404.  

image

MarieSacksick:
PR can be checked here: https://output.circle-artifacts.com/output/job/5a7bdf86-8c52-4c65-bd46-30cdea989927/artifacts/0/doc/index.html. It cannot be tested directly because of [this line](https://github.com/skrub-data/skrub/blob/main/doc/conf.py#L249) which make it rely on main.  

One trouble is that once we switch to a former version, for instance 0.5, this former version doesn't have the other former versions in the switcher. I would say that it's fine.

GaelVaroquaux:
> One trouble is that once we switch to a former version, for instance 0.5, this former version doesn't have the other former versions in the switcher. I would say that it's fine.

Yes, it's a cost to pay

GaelVaroquaux:
Actually, one more thing: we need to update the release process to add a step on adding updating the version switcher
https://skrub-data.org/dev/RELEASE_PROCESS.html#process

MarieSacksick:
> we need to update the release process to add a step on adding updating the version switcher  

There was already something pointing at it [here](https://github.com/skrub-data/skrub/blob/0.6.X/RELEASE_PROCESS.rst?plain=1#L60), but I added a note to remove the old version.

GaelVaroquaux:
Yes, good point. Thanks!

MarieSacksick:
better?

GaelVaroquaux (approved):
LGTM, merging

Issue #1490, opened 2025-07-07
Cleaner (and TableVectorizer) fail when there is a polars empty column name


jeromedockes:
### Describe the bug

When an input polars dataframe has '' (empty string) as one of the column names, scikit-learn's `set_output` changes the output column name to `column_0`. This causes an error in the Cleaner which expects the output column names to be the same as the inputs, ie the same as those it sets in the dataframe it returns from `transform` and `fit_transform`

The problem is that the skrub transformer inherits from sklearn TransformerMixin, which wraps transform in set_output, which does `return pl.DataFrame(X_output, schema=columns, orient="row")` [here](https://github.com/scikit-learn/scikit-learn/blob/cfd5f7833dfb3794e711e79e4a3373e599d5a1f0/sklearn/utils/_set_output.py#L165). But `pl.DataFrame` when given a schema replaces '' with `column_0` (whereas it does not have that behavior when given a dict):

```
>>> df = pl.DataFrame({'': [1], 'b': [2]}) # ok
>>> df
shape: (1, 2)
┌─────┬─────┐
│     ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 2   │
└─────┴─────┘
>>> pl.DataFrame(df) # ok
shape: (1, 2)
┌─────┬─────┐
│     ┆ b   │
│ --- ┆ --- │
│ i64 ┆ i64 │
╞═════╪═════╡
│ 1   ┆ 2   │
└─────┴─────┘
>>> pl.DataFrame(df, schema=df.columns) # column_0
shape: (1, 2)
┌──────────┬─────┐
│ column_0 ┆ b   │
│ ---      ┆ --- │
│ i64      ┆ i64 │
╞══════════╪═════╡
│ 1        ┆ 2   │
└──────────┴─────┘
```

I'll also maybe try to open an issue in scikit-learn.

We don't need the TransformerMixin because all it does is do the `set_output` which changes the column name in this case but is unnecessary, as the skrub transformer already produces a dataframe of the correct type with the correct column names, and provide a default `fit_transform` but we define it anyway. So the simple fix is to not inherit from the TransformerMixin.

If we absolutely want to inherit from it, we can pass `auto_wrap_output_keys=()` in the class definition eg

```
class OnEachColumn(TransformerMixin, BaseEstimator, auto_wrap_output_keys=()):
```

to disable this wrapping.

[...]

PR #1722, opened 2025-10-31
Fix: Cleaner (and TableVectorizer) fail when there is a polars empty column name

MarieSacksick:
Fixes https://github.com/skrub-data/skrub/issues/1490.

### DIFF
--- a/skrub/_dataframe/_common.py
+++ b/skrub/_dataframe/_common.py
@@ -290,7 +290,7 @@ def _make_dataframe_like_polars(obj, data):
         return pl.DataFrame(data)
     else:
         # not pretty workaround for polars issue https://github.com/pola-rs/polars/issues/25204
-        return pl.DataFrame(data, schema={unit.name:None for unit in data})
+        return pl.DataFrame(data, schema={unit.name: None for unit in data})
 
 
 @dispatch


skrub/_table_vectorizer.py
```diff
@@ -350,6 +350,9 @@ def fit_transform(self, X, y=None):
         self.all_processing_steps_ = {col: [] for col in input_names}
         for step in all_steps[1:]:
             for col, transformer in step.transformers_.items():
+                if col == "column_0" and col not in input_names:
+                    # for polars empty column name, which get renamed by scikit-learn
+                    col = ""
```
MarieSacksick: Other possible solutions: 
1. try/except caching the error (KeyError) but this seems too broad
2. instead of `col not in input_names`, it can be checked in another way, `col not in X.columns`
3. checking if X is a polars dataframe, and if it has an empty column name. 

It may seem obvious: X cannot have two empty column names, there is no "column_1" possible.

MarieSacksick: I also tried doing a correspondance with a dict between input_names and step.transformers_.keys(), but it's not always the same length.

jeromedockes: thanks @MarieSacksick ! I realized the `make_dataframe_like` in `_common.py` calls `pl.DataFrame(data)` so that column_0 can get inserted whenever we use it -- in particular when creating the output of ApplyToCols.transform

maybe it would make the most sense to fix it there? one difficulty is that the input can be several things (dict, list of Series, dataframe, ...) so it might be a bit annoying to figure out if there was an empty column name in the input

for the transformers that inherit from TransformerMixin, if the issue occurs I think it can be fixed by adding `auto_wrap_output_keys=()` kwarg to the class declaration, but that could be discussed separately as it does not seem to be needed to fix #1490 specifically (the original issue description is a bit misleading as I had misunderstood where that column was being added)

jeromedockes: > one difficulty is that the input can be several things (dict, list of Series, dataframe, ...) so it might be a bit annoying to figure out if there was an empty column name in the input

actually it is only allowed to be a dict or a list of Series (with names), so it's not so bad there are only 2 cases to handle looking for the name ''

MarieSacksick: it looks like it's a bug from polars actually: https://github.com/pola-rs/polars/issues/25204.

jeromedockes:
great, thank you @MarieSacksick ! and also for opening the polars issue

Issue #1576, opened 2025-08-27
Add a new way to compute column correlations


rcap107:
As was suggested on mattermost by @dholzmueller (so I don't lose track of it)

A very fast (linear algebra, so vectorizable) method to get (nonlinear) dependency scores between features: https://www.pnas.org/doi/10.1073/pnas.2509860122
Maybe with some adjustment for using categoricals it could be a candidate to replace Cramer's V in skrub. (Apparently ordinal encoding categoricals already works well, but I suspect that's not for high-cardinality.)

JadeAffolabi:
Hi, I would like to work on this issue. 
Do you have some specific guidelines or recommendations ?

rcap107:
Hi @JadeAffolabi, thanks for checking out the issue. Be aware that this looks like a fairly complicated issue to start with. 

Aside from the usual considerations that are reported in https://skrub-data.org/stable/CONTRIBUTING.html, you can start by looking at `skrub/_column_associations.py` to see how we implemented Cramer's V and Pearson. 

As with everything, the code that is used for the feature should support both pandas and polars, so you might need to add a function to `skrub/_dataframe/_common.py`. 

There is also the problem of adding this to the table report once the new association scores are computed, but that may also be tackled in a separate PR.

JadeAffolabi:
Hi, I think I can do it.
Thanks for the indications.

rcap107:
> Hi, I think I can do it. Thanks for the indications.

@JadeAffolabi  I realized late I pasted the wrong link, I meant to refer to the contributing guide

dholzmueller:
I think besides the raw implementation there are two main questions to me that need to be solved:
- How to handle categoricals? For a first draft one could ordinal-encode them and handle them as numericals, but I think for higher-cardinality categoricals it should be better to directly create embeddings instead of using the numerical kernel. Maybe using OneHotEncoder with some value for max_categories could be good.
- How to avoid reporting overfitted scores? IIRC the paper mentions something about computing p-values or so, maybe with permutation tests. I don't know how big of a problem it is, but at least we'd want to avoid giving higher scores to categorical variables because they're easier to overfit.

cc'ing one of the authors, @aradha

JadeAffolabi:
Yes, the one hot encoding is definitely a better approach than ordinal-encode.
The original paper does mention computing p-values on the permutation tests in order to decide whether two variables should be declared dependent.

JadeAffolabi:
Hello, 
[...]

PR #1724, opened 2025-11-02
FEAT - Add interdependence score to column associations

JadeAffolabi:
**Description**
This pull request is linked to issue #1576 .

I implemented the numpy version of the Interdependence Score (IDS) from the paper [Efficiently quantifying dependence in massive scientific datasets using InterDependence Scores](https://www.pnas.org/doi/10.1073/pnas.2509860122#eqn4).

**New file**
The new functions are in the file `skrub/_interdependence_score.py`. 
The main function of the file is _ids_matrix(). It can be called directly.

**TO DO**
If the PR is accepted, I can modified the file `skrub/_column_associations.py`  so that the IDS could be displayed in the output of column_associations, along with the pvalues associated with the IDS.

### DIFF
--- a/examples/0110_interdependence_score.py
+++ b/examples/0110_interdependence_score.py
@@ -41,7 +41,7 @@ v8 = pd.qcut(np.random.randn(n), q=3, labels=["cat2_0", "cat2_1", "cat2_2"]).to_
 v_num = np.column_stack([v0, v1, v2, v3, v4, v5, v6])
 v_cat = np.column_stack((v7, v8))
 X_dict = {f"v{i}": v_num[:, i] for i in range(v_num.shape[1])}
-X_dict.update({f"v{i+7}": v_cat[:, i] for i in range(v_cat.shape[1])})
+X_dict.update({f"v{i + 7}": v_cat[:, i] for i in range(v_cat.shape[1])})
 X = pd.DataFrame(X_dict)
 # %%
 # +----------+-------------+---------------+
--- a/skrub/_interdependence_score.py
+++ b/skrub/_interdependence_score.py
@@ -99,10 +99,10 @@ def _onehot_encode_categoricals(df: DataFrame) -> tuple[DataFrame, dict[int, lis
     for col_idx in range(n_cols):
         col = sbd.col_by_idx(df, col_idx)
         one_hot_cols = None
-        if sbd.is_duration(col):
-            col = sbd.total_seconds(col)
-        if sbd.is_numeric(col) or sbd.is_any_date(col):
-            col = sbd.to_float32(col)
+        if sbd.is_numeric(col):
+            col = (
+                sbd.total_seconds(col) if sbd.is_duration(col) else sbd.to_float32(col)
+            )
             if sbd.n_unique(col) >= _CATEGORICAL_THRESHOLD:
                 new_cols.append(col.to_numpy().reshape(-1, 1))
             else:
--- a/skrub/tests/test_interdependence_score.py
+++ b/skrub/tests/test_interdependence_score.py
@@ -33,7 +33,7 @@ def variables():
     X_cat = np.column_stack((f7, f8))
 
     X_dict = {f"Num-{i}": X_num[:, i] for i in range(X_num.shape[1])}
-    X_dict.update({f"Cat-{i+1}": X_cat[:, i] for i in range(X_cat.shape[1])})
+    X_dict.update({f"Cat-{i + 1}": X_cat[:, i] for i in range(X_cat.shape[1])})
 
     return X_dict
 


rcap107:
Hi @JadeAffolabi, a few points about this PR. 

As I was expecting, this will be a very complex PR that will need various iterations of discussions and implementation, and will take a long time to review. If your expectation is for this PR to be merged in the short term (e.g., for some kind of assignment...), that will most likely not be the case. 

Then, there is no need to open and close the PR multiple times. If you need to re-run the CI, you need to push a new commit to trigger the execution of the CI again. 

The PR is missing all tests (which is also why coverage is not passing), so they should be added. You can refer to https://github.com/skrub-data/skrub/pull/1310 for a PR that adds a similar feature, and how it is tested. 

Similarly, the PR is missing proper examples in the docstrings, and a relative example in the documentation. Again, check #1310 for more info on how this should be done.

So this PR needs a lot more work before it can be merged, and if you happen to need to meet a deadline with it you might want to reconsider working on this. Very often, contributions to open source software can take many months to be merged, and this looks like one that won't be merged anytime soon.

JadeAffolabi:
Hi @rcap107 ,

Thank you for your feedback. I will keep working on the PR.
This work was indeed part of an assignment, but it doesn't matter, I will do my best to complete the PR.

Have a nice day.

rcap107:
> Hi @rcap107 ,
> 
> Thank you for your feedback. I will keep working on the PR. This work was indeed part of an assignment, but it doesn't matter, I will do my best to complete the PR.
> 
> Have a nice day.

That sounds good! What matters to me is that working on this may take a long time, but if that is ok with you then you are of course more than welcome to continue. Thanks for the effort, and feel free to ask if you need further help.

JadeAffolabi:
Hi @rcap107 ,
You said that the "PR is missing proper examples in the docstrings, and a relative example in the documentation."
It seems you a talking about two types of examples. I assume that one is the type of example that shows a snippet of code to demonstrate how to use the new function and that example will be in the main function docstrings. But for the other type of example I don't get it.

I was also wondering if I should add the interdependence score (IdS) to the output of the column_associations function or if I should make the function that implements the IdS public.

JadeAffolabi:
Hello,
I realized that using norm 1 or norm 2 to calculate the IdS does not respect the following property : Ids(x,x) = 1, where a variable has perfect dependence with itself. That's why I propose implementing only the norm infinity (I had implemented all three).

What do you think?

rcap107:
> Hi @rcap107 , You said that the "PR is missing proper examples in the docstrings, and a relative example in the documentation." It seems you a talking about two types of examples. I assume that one is the type of example that shows a snippet of code to demonstrate how to use the new function and that example will be in the main function docstrings. But for the other type of example I don't get it.

In the docstring there should be a snippet of code that shows how you're supposed to call the function and what the result looks like on an example dataset. Then, there should be a script in the "examples" folder that explains gives some additional context on the operation. #1310 is a very good example to take inspiration from because it's a very similar feature and has both the docstring examples and the full script to check out. 

You can also use [this](https://skrub-data.org/stable/tutorial_example.html) document to find more info on how to write an example. 

It should be fairly short, and it's important to have one the gallery. 

> I was also wondering if I should add the interdependence score (IdS) to the output of the column_associations function or if I should make the function that implements the IdS public.

The function should be public, and the output should also be added to the Associations tab in the TableReport (although this part can be done in a separate PR).

From a quick look at the code, it seems like the output of the ids function is a square matrix. If that is the case, it should be modified so that it is a dataframe that contains the list of all the associations between the columns. Check out the `_column_associations.py` file where the `_melt` operation is done for what I mean.

rcap107:
Small note, this pr should get an entry in the changelog.

rcap107:
Hi @JadeAffolabi, I really appreciate the effort you're putting in this PR, especially considering it is quite a complex issue. 

Unfortunately, as a result of it being a complex issue it is also something that takes quite a bit of time to review, to make sure it is working as intended and properly implements the original paper. 

For your information, I won't be able to allocate the time necessary for a while more (likely, not until January), though if I can I'll try to leave some feedback before then.

JadeAffolabi:
Hello,
No worries, take your time.
If there are any other changes to be made, let me know.

examples/0110_interdependence_score.py
```diff
@@ -0,0 +1,101 @@
...
+# --------------------------------------------------------------
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+from skrub import interdependence_score
+
+np.random.seed(42)
+n = 1000
+
```
rcap107, line 28: Here it would be better to write explicitly that some of the variables are dependent on each other (v1 depends on v0 etc.), and that we would expect them to have a higher IDS with each other than the other variables

examples/0110_interdependence_score.py
```diff
@@ -0,0 +1,101 @@
+"""
+interdependence_score : A dependence measure between variables.
+===================================================================
+
+The following example illustrates the use of ``:func:`~skrub.interdependence_score```,
+a function that measures linear and various nonlinear dependencies between variables.
+
+Its basic idea is to approximate the HSIC (Hilbert Schmidt Independence Criterion)
+by computing the first k terms of an infinite dimensional feature map
+for the universal gaussian kernel.
```
rcap107, line 10: Add here a reference to the paper this was taken from

examples/0110_interdependence_score.py
```diff
@@ -0,0 +1,101 @@
...
+for i in range(len(ids_mat)):
+    for j in range(len(ids_mat)):
+        text = ax.text(
+            j, i, f"{ids_mat.iloc[i, j]:.2f}", ha="center", va="center", color="w"
+        )
+plt.title("Interdependence score matrix")
+plt.tight_layout()
+plt.show()
+# %%
+# First of all each variable have perfect dependence with itself (ids = 1).
```
rcap107, line 92:
```suggestion
# First of all, each variable has perfect dependence with itself (IDS = 1).
```

examples/0110_interdependence_score.py
```diff
@@ -0,0 +1,101 @@
...
+# The linearly dependent variables (v0, v1) have ids greater than 0.90,
+# and the non-linear variables as well (v2, v3, v4).
```
rcap107, line 94: Rewording a bit, formatting variables  
```suggestion
# Both linearly dependent (``v1`` wrt ``v0``) and non-linearly dependent 
# variables (``v3`` and ``v4`` wrt ``v2``) have IDS greater than 0.90, signaling
# a strong dependency on the starting variable.
```

examples/0110_interdependence_score.py
```diff
@@ -0,0 +1,101 @@
...
+# The variable v6  which is tanh(v0*v2), has a high score with v0 and v2,
```
rcap107, line 95:
```suggestion
# ``v6``, which is defined as ``tanh(v0*v2)``, has a high score with ``v0`` and ``v2``,
```

examples/0110_interdependence_score.py
```diff
@@ -0,0 +1,101 @@
...
+# but also with the variables related to them v1, v3, v4.
```
rcap107, line 96:
```suggestion
# as well as with ``v1``, ``v3``, ``v4``, which are still related to ``v0``
# (``v1``) and ``v2`` (``v3`` and ``v4``).
```

examples/0110_interdependence_score.py
```diff
@@ -0,0 +1,101 @@
...
+# The ids also captures the relation between the categorical variable v7 and v1,
+# and by extension with v0 because v1 depends on v0.
+# Both independent variables v5 and v8 have ids less than 0.1 with all other variables.
```
rcap107, line 99: rewording
```suggestion
# The IDS also captures the relation between the categorical variable ``v7`` and ``v1``,
# and by extension with ``v0``, since ``v1`` depends on ``v0``.
# Both variables ``v5`` and ``v8`` are independent from all other variables and 
# each other, and their IDS is lower than 0.1 with respect to all other variables.
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+
+DataFrame = TypeVar("DataFrame")
+_N_BINS = 10
+_CATEGORICAL_THRESHOLD = 30
+EPSILON = sys.float_info.epsilon
+
+
+def _gaussian_feature_map(
+    x: numpy.ndarray, k_terms=6, bandwidth_term=1 / 2
+) -> numpy.ndarray:
```
rcap107, line 23: Removing type hints for consistency with the rest of the codebase
```suggestion
    x, k_terms=6, bandwidth_term=1 / 2
)
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+        result : numpy array
+            The resulting array of shape (n_samples, n_features * k_terms).
+    """
+    B = bandwidth_term
+    exp = np.exp(-B * x**2)
+    terms = []
+    for i in range(k_terms):
+        # In their implementation the authors of IDS chose to set B = 1.
+        terms.append(exp * x**i / (math.sqrt(math.factorial(i)) * 1.0))
+    return np.concatenate(terms, axis=1)
```
rcap107, line 49: I don't understand if this function is meant to implement equation 6 of the paper. I see that this is taken directly from the original repository, but in any case it is not clear to me why the value of B is assigned to be `bandwidth_term`, which is set to 1/2 by default, and then in the expansion it is replaced by 1.0. If it's a different B, then it should have a different name, if it's meant to be kept as 1 all the time there is no need to pass it as a parameter. 

I might be missing something in the math here, but it does not seem to me that this code is consistent with what is written in the paper.

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+    Returns
+    -------
+        output: numpy array
+            The resulting array with one hot encoded features.
+        old_idx_matches: dict[int, list]
+    """
+    _, n_cols = sbd.shape(df)
+    new_cols = []
+    old_idx_matches = dict()
+    one_hot_encoder = OneHotEncoder(max_categories=_N_BINS, sparse_output=False)
```
rcap107, line 97: Here we are fitting and transforming with a OneHotEncoder, which means that this becomes a stateful transformation. 

At the moment, this function is supposed to be used exclusively as a metric, so it _may_ be fine, but it's worth looking at again.

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+    current_col_idx = 0
+    for col_idx in range(n_cols):
+        col = sbd.col_by_idx(df, col_idx)
+        one_hot_cols = None
+        if sbd.is_numeric(col):
+            col = (
+                sbd.total_seconds(col) if sbd.is_duration(col) else sbd.to_float32(col)
+            )
+            if sbd.n_unique(col) >= _CATEGORICAL_THRESHOLD:
+                new_cols.append(col.to_numpy().reshape(-1, 1))
```
rcap107, line 107: `col.to_numpy` should be replaced by `sbd.to_numpy(col)` here and anywhere else it's used

```suggestion
                new_cols.append(sbd.to_numpy(col).reshape(-1, 1))
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+        result[:, j] = X[rng.permutation(n), j]
+    return result
+
+
+def _onehot_encode_categoricals(df: DataFrame) -> tuple[DataFrame, dict[int, list]]:
+    """
+    Encode the categorical variables into one-hot encoding.
+    Numerical variables which values are fewer than _CATEGORICAL_THRESHOLD are treated
+    as categorical.
+
```
rcap107, line 83:
```suggestion
    Encode the categorical variables into one-hot encoding.
    Numerical variables with fewer than _CATEGORICAL_THRESHOLD unique values are treated
    as categorical.

```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+
+def _compute_p_val(
+    ids: numpy.ndarray,
+    X: numpy.ndarray,
+    feature_map_function=_gaussian_feature_map,
+    k_terms=6,
+    n_tests=100,
+    bandwidth_term=1 / 2,
+) -> numpy.ndarray:
```
rcap107, line 140: removing type hints
```suggestion
def _compute_p_val(
    ids,
    X,
    feature_map_function=_gaussian_feature_map,
    k_terms=6,
    n_tests=100,
    bandwidth_term=1 / 2,
):
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+            current_col_idx += 1
+
+    output = np.hstack(new_cols)
+    return output, old_idx_matches
+
+
+def _compute_p_val(
+    ids: numpy.ndarray,
+    X: numpy.ndarray,
+    feature_map_function=_gaussian_feature_map,
```
rcap107, line 136: the value of `feature_map_function` is always the same here and in other functions, and it is not exposed to the user, so this parameter should be removed as it is not needed
```suggestion
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+            The p-value matrix associated to the interdependence score matrix.
+    """
+    ids_obs = ids
+    count = np.zeros((X.shape[1], X.shape[1]))
+    for i in range(n_tests):
+        X_permuted = _permute_rows(X)
+        ids_perm, _ = _compute_ids_numpy(
+            X_permuted,
+            k_terms=k_terms,
+            feature_map_function=feature_map_function,
```
rcap107, line 177: same here, `feature_map_function` is always the same so the parameter should be removed
```suggestion
            feature_map_function=feature_map_function,
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+            bandwidth_term=bandwidth_term,
+        )
+        count += np.where(ids_perm > ids_obs, 1, 0)
+    p_vals = count / n_tests
+    return p_vals
+
+
+def _compute_ids_numpy(
+    X: numpy.ndarray,
+    feature_map_function=_gaussian_feature_map,
```
rcap107, line 187:
```suggestion
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+    k_terms=6,
+    p_val=False,
+    num_tests=100,
+    bandwidth_term=1 / 2,
+) -> tuple[np.ndarray, np.ndarray] | tuple[np.ndarray, None]:
```
rcap107, line 192: removing type hints
```suggestion
def _compute_ids_numpy(
    X,
    feature_map_function=_gaussian_feature_map,
    k_terms=6,
    p_val=False,
    num_tests=100,
    bandwidth_term=1 / 2,
):
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+def _ids_matrix(
+    X: DataFrame,
+    feature_map_function=_gaussian_feature_map,
+    k_terms=6,
+    p_val=False,
+    num_tests=100,
+    bandwidth_term=1 / 2,
+) -> tuple[np.ndarray, np.ndarray] | tuple[np.ndarray, None]:
```
rcap107, line 293:
```suggestion
def _ids_matrix(
    X,
    feature_map_function=_gaussian_feature_map,
    k_terms=6,
    p_val=False,
    num_tests=100,
    bandwidth_term=1 / 2,
):
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+
+
+def interdependence_score(
+    X: DataFrame,
+    k_terms=6,
+    p_val=False,
+    num_tests=100,
+    bandwidth_term=1 / 2,
+    return_matrix=False,
+) -> DataFrame | tuple[DataFrame, DataFrame]:
```
rcap107, line 351: type hints
```suggestion
def interdependence_score(
    X,
    k_terms=6,
    p_val=False,
    num_tests=100,
    bandwidth_term=1 / 2,
    return_matrix=False,
):
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+    Returns
+    -------
+    DataFrame or (DataFrame, DataFrame)
+        If ``return_matrix=False`` (default), returns a DataFrame containing each
+        pair of columns, their interdependence score, and optionally the p-value.
+        If ``return_matrix=True``, returns also a square matrix of interdependence scores.
+
+    Notes
+    -----
+    The result is a dataframe with columns: ``['left_column_name', 'right_column_name', 'cramer_v', 'pearson_corr']``
```
rcap107, line 396:
```suggestion
    The result is a dataframe with columns: ``['left_column_name', 'right_column_name', 'interdependence_score', 'pvalue']``
```

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+
+    if p_val:
+        pval_table = _stack_symmetric_associations(
+            pval_matrix, X, statistic_name="pvalue"
+        )
+        on = ["left_column_name", "right_column_name"]
+        ids_table = _join_utils.left_join(
+            ids_table, pval_table, right_on=on, left_on=on
+        )
+        ids_table = drop_columns_with_substring(ids_table, "skrub")
```
rcap107, line 461: This can be replaced by skrub selectors

add `import skrub.selectors as s` at the top of the file 

then
```suggestion
        ids_table = s.select(ids_table, ~s.glob("*skrub*"))
```

`drop_columns_with_substring` can be removed

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
...
+
+    ids_table = drop_columns_with_substring(ids_table, "idx")
```
rcap107, line 463: same here as above 
```suggestion
        ids_table = s.select(ids_table, ~s.glob("*idx*"))  
```

rcap107 (commented):
Here is a first review. As I expected, this will take time and iterations to get right. 

Overall, I have two main points I am concerned by:

- A lot of this code is taken directly from the original repository (https://github.com/aradha/interdependence_scores), and it's not clear to me whether the original author is aware of this, and how we're going to deal with the licenses. 
- The function `interdependence_score` is fit_transforming a OneHotEncoder, meaning that there is a state that should be kept track of. However, since this function is just meant to give a metric, it may be fine to discard it. 

Some other points: 
- For consistency with the code base I removed type hints. 
- `feature_map_function` is never used as a parameter, so it should be removed and `_gaussian_feature_map` should be used directly instead. 

There are some other cosmetic fixes that need to be handled.

skrub/_interdependence_score.py
rcap107: After discussing this IRL with @dholzmueller, we figured out what is happening here: with default values, this is equivalent to setting B=1, but as it is written, it is not clear that the 1.0 is in fact B.

It should be rewritten more explicitly, like:
```python
def _gaussian_feature_map(
    x, k_terms=6, bandwidth_term=1
):
    B = bandwidth_term
    exp = np.exp(- (x**2) / (2 * B ** 2))
    terms = []
    for i in range(k_terms):
        terms.append(exp * x**i / (math.sqrt(math.factorial(i)) * B))
```

Having B in one place and 1.0 elsewhere is inconsistent and makes it hard to understand what is actually happening

skrub/_interdependence_score.py
```diff
@@ -0,0 +1,492 @@
+"""Detect which columns have strong statistical dependence using InterDependenceScore"""
```
rcap107, line 1: Please add here a link to the original repository and explain that this is an adaptation of the original implementation so that it can be used in skrub.

rcap107:
A couple of additional notes after IRL discussions:
> * A lot of this code is taken directly from the original repository ([aradha/interdependence_scores](https://github.com/aradha/interdependence_scores)), and it's not clear to me whether the original author is aware of this, and how we're going to deal with the licenses.

Given that the original license is MIT, it should be fine to reuse the code, provided that it gets referenced properly in the code. 

> * The function `interdependence_score` is fit_transforming a OneHotEncoder, meaning that there is a state that should be kept track of. However, since this function is just meant to give a metric, it may be fine to discard it.

Since this is a metric, fitting the OneHotEncoder is not a problem.

rcap107:
Hi @JadeAffolabi, 
The more I think about this PR, the less convinced I am it should be merged after all. It's unclear how useful it would be in practice, and beside the comments in my review, there needs to be some extensive performance benchmarking. Basically, it's a lot of work left to be done for very unclear gains. 

I'm discussing with other maintainers whether the PR should be merged after all, and for the time being it's better if you spend any more time on this particular problem. I should have stopped you before you embarked in this, I apologize for the mess-up. In any case, thank you very much for the effort you put into this.

JadeAffolabi:
Hi @rcap107 ,

I apologize for the slow response.
Thank you for the code review and for the honest update. I completely understand your concerns.
I will stop working this issue. I enjoyed working on it, since it was my first attempt of contribution.

Thanks,

rcap107:
> Hi @rcap107 ,
> 
> I apologize for the slow response. Thank you for the code review and for the honest update. I completely understand your concerns. I will stop working this issue. I enjoyed working on it, since it was my first attempt of contribution.
> 
> Thanks,

Unfortunately, sometimes contributions don't pan out. I'll close this PR now. Thanks again for the time and effort you put in this.

You're more than welcome to look for a new issue if you're still interested in contributing.

PR #1726, opened 2025-11-03
FEAT - adding a heuristic for parsing units in string columns

### LINKED ISSUE #1542: Add a transformer that uses heuristics to parse string columns that contain mixed strings like 10$/100 kg
### Problem Description

Sometimes dataframes contain numerical data that includes a string to denote the unit of measurement, like a column about money that has "$ 10", or a column that contains percentages like "29%".

As a result, it is necessary to write a custom function that parses the string, then either trims out the unit, or splits the column into two. 

### Feature Description

It would be nice to have a transformer that implements a few heuristics to do this automatically. 

For example, it could use a regex to detect `numbers<space> string`, or `string<space>numbers`. 

We might want to allow the user to provide the regex themselves. 

I think this would be especially convenient with ApplyOnCols. 

It could also be an optional step in the Cleaner, e.g., by "extending" the ToFloat transformer to also handle common formats. If we go in this direction, it probably should not take custom formats and just use whatever heuristics we come up with. 

### Alternative Solutions

_No response_

### Additional Context

_No response_

gabrielapgomezji:
Fixes #1542


rcap107:
This issue was discussed IRL during the WiMLDS sprint at Probabl on oct 28th

Issue #1675, opened 2025-10-15
BUG - TableReport fails on lazy polars dataframes


rcap107:
If I try to open a lazy dataframe, the table report fails. 

```python
import polars as pl
data = {"a": ["1", "2", "3"]}
df = pl.DataFrame(data).lazy()
from skrub import TableReport
TableReport(df)
```
Gives:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[6], line 1
----> 1 TableReport(df)

File /tmp/scratch/lib/python3.13/site-packages/skrub/_reporting/_table_report.py:206, in TableReport.__init__(self, dataframe, n_rows, order_by, title, column_filters, verbose, max_plot_columns, max_association_columns)
    200 self.max_plot_columns, self.max_association_columns = _check_max_cols(
    201     max_plot_columns, max_association_columns
    202 )
    203 self.dataframe = (
    204     sbd.to_frame(dataframe) if sbd.is_column(dataframe) else dataframe
    205 )
--> 206 self.n_columns = sbd.shape(self.dataframe)[1]

File ~/.local/share/uv/python/cpython-3.13.0rc2-linux-x86_64-gnu/lib/python3.13/functools.py:929, in singledispatch.<locals>.wrapper(*args, **kw)
    926 if not args:
    927     raise TypeError(f'{funcname} requires at least '
    928                     '1 positional argument')
--> 929 return dispatch(args[0].__class__)(*args, **kw)

File /tmp/scratch/lib/python3.13/site-packages/skrub/_dataframe/_common.py:503, in _shape_polars(obj)
    501 @shape.specialize("polars")
    502 def _shape_polars(obj):
--> 503     return obj.shape

AttributeError: 'LazyFrame' object has no attribute 'shape'
```

How should we handle this? Do we just collect automatically, or raise a more informative exception?

rcap107:
After discussions, we decided that the better approach is to raise an informative exception so that the user can decide on their own whether they should collect or not. 

Collecting automatically may lead to all sorts of issues, so it's better that it's the user who decides.

fatiben2002:
Working on this one

rcap107:
Closed by #1767

PR #1732, opened 2025-11-04
FIX - Improving the error message raised when the TableReport is called on a lazy polars dataframe

fatiben2002:
This pull request aims to fix the polars exception issue #1675.

Through out this pull request, I have edited these files:
**- skrub/_reporting/_table_report.py:** In this file I created a function ( lazy_dataframe_exception) checks if a Dataframe is both a polar and Lazyframe, if so an exception is raised. Then, I add a line within the "class TableReport" that eventually checks if the dataframe is polar and lazyframe.
**- skrub/_reporting/tests/test_table_report.py:** Within this file, I import the polars library while also adding a test. 

--> Let me know if this is a good solution.

### DIFF
--- a/skrub/_reporting/_table_report.py
+++ b/skrub/_reporting/_table_report.py
@@ -25,7 +25,6 @@ def _check_max_cols(max_plot_columns, max_association_columns):
             "'max_plot_columns' must be a positive scalar or 'all', got"
             f" {max_plot_columns!r}."
         )
-
     max_association_columns = (
         max_association_columns
         if max_association_columns is not None
@@ -42,6 +41,11 @@ def _check_max_cols(max_plot_columns, max_association_columns):
 
     return max_plot_columns, max_association_columns
 
+def lazy_dataframe_exception(dataframe):
+    if sbd.is_polars(dataframe) and sbd.is_lazyframe(dataframe):
+        raise ValueError(
+            "Cannot automatically collect a lazyframe, this might lead to issues!"
+        )
 
 class TableReport:
     r"""Summarize the contents of a dataframe.
@@ -203,6 +207,7 @@ class TableReport:
         self.dataframe = (
             sbd.to_frame(dataframe) if sbd.is_column(dataframe) else dataframe
         )
+        lazy_dataframe_exception(self.dataframe)
         self.n_columns = sbd.shape(self.dataframe)[1]
 
     def _set_minimal_mode(self):
--- a/skrub/_reporting/tests/test_table_report.py
+++ b/skrub/_reporting/tests/test_table_report.py
@@ -8,12 +8,12 @@ from pathlib import Path
 import numpy as np
 import pytest
 from sklearn.utils import Bunch
+import polars as pl
 
 from skrub import TableReport, ToDatetime
 from skrub import _dataframe as sbd
 from skrub._reporting._sample_table import make_table
 from skrub.conftest import skip_polars_installed_without_pyarrow
-
 
 @pytest.fixture
 def simple_df(df_module):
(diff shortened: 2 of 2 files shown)


skrub/_reporting/_table_report.py
```diff
@@ -42,6 +41,11 @@ def _check_max_cols(max_plot_columns, max_association_columns):
 
     return max_plot_columns, max_association_columns
 
+def lazy_dataframe_exception(dataframe):
```
rcap107, line 44: I think there is no need to use a function here, since it's only checking for a condition and raising an exception: if it's wrapped in a function, we have to go to the definition to understand what it's doing.

skrub/_reporting/_table_report.py
```diff
@@ -42,6 +41,11 @@ def _check_max_cols(max_plot_columns, max_association_columns):
...
+    if sbd.is_polars(dataframe) and sbd.is_lazyframe(dataframe):
+        raise ValueError(
+            "Cannot automatically collect a lazyframe, this might lead to issues!"
+        )
```
rcap107, line 48: Maybe we can have a more explicit error message:
```suggestion
        raise ValueError(
            "The TableReport does not support lazy dataframes. Please call `.collect()` to use the TableReport on the current dataframe."
        )
```

(note that if you accept this suggestion it'll break the pre-commit checks, so you'll have to break the line I wrote so it's on multiple lines)

Also a clarification: there is nothing that prevents us from collecting the dataframe ourselves, but as we discussed, this could lead to problems so we intentionally avoid collecting the lazyframe.

skrub/_reporting/_table_report.py
```diff
@@ -203,6 +207,7 @@ def __init__(
         self.dataframe = (
             sbd.to_frame(dataframe) if sbd.is_column(dataframe) else dataframe
         )
+        lazy_dataframe_exception(self.dataframe)
```
rcap107, line 210: As I mentioned in the previous comment, you can put the check directly here, rather than using a function

skrub/_reporting/tests/test_table_report.py
```diff
@@ -8,13 +8,13 @@
 import numpy as np
 import pytest
 from sklearn.utils import Bunch
+import polars as pl
```
rcap107, line 11: Polars is not a required package, which means that importing it here will lead to errors and failed tests in some environments: it's what is failing the CI tests in the current version of the code. 

You can deal with this using `pytest.importorskip("polars")`: see https://docs.pytest.org/en/4.6.x/reference.html#pytest-importorskip

skrub/_reporting/tests/test_table_report.py
```diff
@@ -99,6 +97,18 @@ def test_empty_dataframe(df_module):
     assert "The dataframe is empty." in html
 
 
+def test_lazyframe_exception():
+    lazy_df = pl.DataFrame({
```
rcap107, line 101: This should fix the test
```suggestion
def test_lazyframe_exception():
    pl = pytest.importorskip("polars")
    lazy_df = pl.DataFrame({
```

skrub/_reporting/tests/test_table_report.py
```diff
@@ -99,6 +97,18 @@ def test_empty_dataframe(df_module):
...
+        "a": ["1", "2", "3"]
+    }).lazy()
+
+    with pytest.raises(
+        ValueError,
+        match=r"Cannot automaticaly collect a lazyframe, this might lead to issues!"
```
rcap107, line 107: This message will need to be updated according to the updated message.

You can also use regular expressions to select only a part of the message, rather than having to paste everything.

rcap107:
This PR should also be documented in the changelog as a fix, something like "Improved error message when a polars lazyframe is passed to the TableReport"

rcap107:
Hey @fatiben2002, are you still interested in working on this PR, or would you rather have someone else take over?

fatiben2002:
@rcap107 Hey Ricardo! Sorry for the delay, I just saw your comments. Yes I am absolutely interested to finish this!

rcap107:
Closing this in favor of #1767

PR #1734, opened 2025-11-04
Doc fixes

GaelVaroquaux:
- Fix links on the front page
- Fix a variable name in an example


rcap107:
I wish we didn't have to update those links manually every time, it's not the first time they slip through

rcap107 (approved):

Issue #1476, opened 2025-06-27
Add an example dedicated to showing the features of the `TableReport`


rcap107:
### Describe the issue linked to the documentation

The `TableReport` has a lot of features, some of which are either hidden, or not easy to show in docstring examples. 

There should be an example that shows all the features of the TableReport on different datasets. 

It should also include a mention of the `max_association_columns` and `max_plot_columns`  and what they look like on large datasets (see #1469)

GaelVaroquaux:
I'm +1 on that suggestion.

Maybe something to leave up for the sprint?

rcap107:
> Maybe something to leave up for the sprint?

yes that was the plan

rcap107:
Regarding this issue, my first idea was having different datasets that let us highlight different features of the TableReport.

One problem there might be with this approach is that we would have to write down all the features and then tell the user to try them out by themselves, which would be tedious and might lose people (it's also why I thought of adding a default view of the table in #1489). 

One alternative would be using screenshots of the different features, in addition to generating the reports, so the user would have an idea of where to look and what the resulting view should be. The problem with this is that the example would not be generated automatically, and adding new features to the TableReport may render the screenshots obsolete. 

What do you think?

Vincent-Maladiere:
What about building an example, which renders different TableReports, and adding a big callout at the beginning saying "You can click on the report :)" followed by a GIF showing clicks on the table report? It's ok if the GIF becomes obsolete, its goal is only to show interactivity

GaelVaroquaux:
> What about building an example, which renders different TableReports, and adding a big callout at the beginning saying "You can click on the report :)" followed by a GIF showing clicks on the table report? It's ok if the GIF becomes obsolete, its goal is only shows interactivity

Yes, totally!!

rcap107:
Yes, sounds good to me

alamehm:
Here is how a GIF may look like. 
There is a small issue when the content of table report change in length, which forse scrolling in jupyter notebooks. 

![Image](https://github.com/user-attachments/assets/25e89ede-584a-4d01-b487-a8f04e7740c7)

rcap107:
Some ideas for this issue:
- Start with a gif and a big message that advises the reader to click on each TableReport
- Cover the main features of the report with a simple dataframe 
- Highlight outliers, sorted columns, columns with many missing values, weird distributions
- Show the config parameters
- Show how to save the Table Report
- If possible, use multiple different example datasets to show different behaviors

[...]

PR #1737, opened 2025-11-05
ENH - Adding a parameter to select the default TableReport tab to show

rcap107:
This is a relatively small change that simplifies a lot writing examples with the TableReport.

The idea is having the parameter "default_tab" to let the user select the specific tab in the report (either the preview, stats, distributions, or associations) so that the TableReport is rendered with the given tab already pre-selected.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -30,8 +30,8 @@ New features
 - :meth:`DataOp.skb.full_report` now accepts a new parameter, title, that is displayed
   in the html report.
   :pr:`1654` by :user:`Marie Sacksick <MarieSacksick>`.
-- :class:`TableReport` now includes the ``default_tab`` parameter, which lets the
-  user select which tab should be shown by default when the ``TableReport`` is
+- :class:`TableReport` now includes the ``open_tab`` parameter, which lets the
+  user select which tab should be opened when the ``TableReport`` is
   rendered. :pr:`1737` by :user:`Riccardo Cappuzzo<rcap107>`.
 
 


rcap107:
This would make #1476 easier to write as well

generated_for_index/baskets.html
```diff
@@ -0,0 +1,30 @@
+<table border="1" class="dataframe">
```
jeromedockes: how are those 3 home page files changes related to the PR ?

skrub/_reporting/_html.py
```diff
@@ -17,6 +17,13 @@
 
 _HIGH_ASSOCIATION_THRESHOLD = 0.9
 
+_TAB_MAPPING = {
```
jeromedockes: maybe tab_name_to_id?

skrub/_reporting/_html.py
```diff
@@ -108,7 +115,13 @@ def _get_column_filters(summary):
     return filters
 
 
-def to_html(summary, standalone=True, column_filters=None, minimal_report_mode=False):
+def to_html(
+    summary,
+    standalone=True,
+    column_filters=None,
+    minimal_report_mode=False,
+    default_tab="table",
```
jeromedockes: I'm not sure it's really a "default" -- it is the one that is open when loading the page. not sure we can find a better name though 🤔 , maybe open_tab, selected_tab

skrub/_reporting/_html.py
```diff
@@ -153,6 +172,8 @@ def to_html(summary, standalone=True, column_filters=None, minimal_report_mode=F
             "base64_column_filters": _b64_encode(column_filters),
             "report_id": f"report_{secrets.token_hex()[:8]}",
             "minimal_report_mode": minimal_report_mode,
+            "default_tab": default_tab,
+            "default_panel_id": default_panel_id,
```
jeromedockes:
```suggestion
            "default_panel_id": default_panel_id,
```

default_tab doesn't seem to be used in the template, and that's good because it's best not to have 2 different parameters to encode the same info

skrub/_reporting/_table_report.py
```diff
@@ -243,6 +262,9 @@ def _set_minimal_mode(self):
         self._to_html_kwargs["minimal_report_mode"] = True
         self.max_association_columns = 0
         self.max_plot_columns = 0
+        # In minimal mode, fall back to 'table' if user selected unavailable tabs
+        if self.default_tab in ["distributions", "associations"]:
```
jeromedockes: maybe check not in [the tabs that still exist] instead as that's what you really need to know (in case other tabs are added in the future)

generated_for_index/baskets.html
rcap107: I am not entirely sure how those files ended up there, I removed them

skrub/_reporting/_html.py
rcap107: I think `open_tab` works

skrub/_reporting/_table_report.py
rcap107: I'm not sure what you mean by this, or how to check this

skrub/_reporting/_data/templates/tabs.html
```diff
@@ -15,23 +15,23 @@
         <div class="tab-list tab-list-scroller" data-manager="TabList" >
             <div>
                 <button type="button" data-target-panel-id="dataframe-sample-panel"
-                    data-role="tab" data-is-selected data-test="sample-tab" class="tab"
+                    data-role="tab" {% if default_panel_id == "dataframe-sample-panel" %}data-is-selected{% endif %} data-test="sample-tab" class="tab"
```
jeromedockes:
```suggestion
                    data-role="tab" {% if open_panel_id == "dataframe-sample-panel" %}data-is-selected{% endif %} data-test="sample-tab" class="tab"
```

As you renamed the name of the variable in the template environment you also need to change it here (and everywhere it's used in the template)

jeromedockes:
I guess we could also add a test in the js test suite to check that the correct tab is actually opened when opening the report

skrub/_reporting/_data/templates/tabs.html
rcap107: good catch

CHANGES.rst
```diff
@@ -30,6 +30,10 @@ New features
 - :meth:`DataOp.skb.full_report` now accepts a new parameter, title, that is displayed
   in the html report.
   :pr:`1654` by :user:`Marie Sacksick <MarieSacksick>`.
+- :class:`TableReport` now includes the ``default_tab`` parameter, which lets the
```
jeromedockes:
```suggestion
- :class:`TableReport` now includes the ``open_tab`` parameter, which lets the
```

CHANGES.rst
```diff
@@ -30,6 +30,10 @@ New features
...
+  user select which tab should be shown by default when the ``TableReport`` is
```
jeromedockes:
```suggestion
  user select which tab should be opened when the ``TableReport`` is
```

jeromedockes (approved):
thanks @rcap107 !

PR #1739, opened 2025-11-06
DOC - improving docstring and user guide for `deduplicate`

rcap107:
I added an example with pandas and polars on how to replace values in a column using `deduplicate`.

### DIFF
--- a/doc/modules/configuration_and_utils/deduplicate_categorical_data.rst
+++ b/doc/modules/configuration_and_utils/deduplicate_categorical_data.rst
@@ -52,6 +52,8 @@ Deduplicating values in a dataframe
 -----------------------------------
 
 |deduplicate| can be used to replace values in a dataframe that contains typos.
+This can be done with ``deduplicate_correspondence`` and the ``map`` function in
+pandas, or the ``replace`` function in polars.
 >>> import pandas as pd
 >>> df = pd.DataFrame({'color': duplicated, 'value': range(10)})
 >>> df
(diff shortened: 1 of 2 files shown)


doc/modules/configuration_and_utils/deduplicate_categorical_data.rst
```diff
@@ -47,3 +47,59 @@ dtype: object
...
+3  black      3              black
+4  black      4              black
+5  uhibe      5              white
+6  white      6              white
+7  white      7              white
+8  white      8              white
+9  white      9              white
+
+With polars:
+>>> import polars as pl  # doctest: +SKIP
```
GaelVaroquaux, line 86: The skip doctest here is leading the line below to fail.

Maybe the line below should also be skipped?

GaelVaroquaux (commented):
Super useful!!

Two things:
- The doctests are failing (presumably because of the "skip" uniformly applied)
- I think that I would add somewhere 'Using the `deduplicate_correspondence` dict defined above'. Maybe just before it is used

doc/modules/configuration_and_utils/deduplicate_categorical_data.rst
rcap107: yes, I thought I fixed that but I must have forgotten

doc/modules/configuration_and_utils/deduplicate_categorical_data.rst
```diff
@@ -47,3 +47,61 @@ dtype: object
 'white', 'white', 'white', 'white', 'white']
 
 See the |deduplicate| documentation for caveats and more detail.
+
+Deduplicating values in a dataframe
+-----------------------------------
+
+|deduplicate| can be used to replace values in a dataframe that contains typos.
+This can be done with ``deduplicate_correspondence`` and the ``map`` function in
+pandas, or the ``replace`` function in polars.
```
GaelVaroquaux:
```suggestion
This can be done with ``deduplicate_correspondence`` computed above and the
``map`` function in pandas, or the ``replace`` function in polars.
```

GaelVaroquaux (approved):
LGTM. Will merge when tests are done (I committed a minor modification)

PR #1745, opened 2025-11-10
DOC: iterate on selectors doc:

GaelVaroquaux:
* Add some seealso and some structure

* Attempts to have titles that summarize the content

* Move a subsection to group more topics and to reading order

* Cosmetic: more explicit code blocks (indented, and starting with "::")

### DIFF
--- a/doc/modules/multi_column_operations/selectors.rst
+++ b/doc/modules/multi_column_operations/selectors.rst
@@ -204,4 +204,3 @@ do this like this:
 2  History -1.069045
 3  Science  0.267261
 4      Art -1.069045
-


GaelVaroquaux:
> I'm not sure why the code blocks are needed. Aren't doctests supposed to be already rendered as code blocks?

They are not needed, as the sphinx engines tries to detect what is code and what is not. But 1) it can fail, 2) I'm one who thinks that "explicit is better than implicit"

Finally, some editors don't detect as well as sphinx, and thus do not do the syntax highlighting without these

rcap107:
> I'm not sure why the code blocks are needed. Aren't doctests supposed to be already rendered as code blocks?
> They are not needed, as the sphinx engines tries to detect what is code and what is not. But 1) it can fail, 2) I'm one who thinks that "explicit is better than implicit" Finally, some editors don't detect as well as sphinx, and thus do not do the syntax highlighting without these

Makes sense, I'll keep that in mind for the future. 

Merging the PR, thanks a lot

PR #1750, opened 2025-11-11
DOC: see also pointing to ApplyToCols/Frame

GaelVaroquaux:
Make it easier for people who are using TableVectorizer or Cleaner to discover ApplyToCols/Frame

I'm doing this because I find that these two objects (ApplyToCols / ApplyToFrame) are incredibly useful and we need to help users discovering them.

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -211,11 +211,11 @@ class Cleaner(TransformerMixin, BaseEstimator):
 
     ApplyToCols :
         Apply a given transformer separately to each column in a selection of columns.
-        Useful to complement the default heuristics of the Cleaner.
+        Useful to complement the default heuristics of the ``Cleaner``.
 
     ApplyToFrame :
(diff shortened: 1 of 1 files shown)


skrub/_table_vectorizer.py
```diff
@@ -529,6 +537,14 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
         Preprocesses each column of a dataframe with consistency checks and
         sanitization, e.g., of null values or dates.
 
+    ApplyToCols :
+        Apply a given transformer separately to each column in a selection of columns.
+        Useful to complement the default heuristics of the TableVectorizer.
```
rcap107: "TableVectorizer" is formatted in the See also above for the `tabular_pipeline`. I think it should either be formatted in both places, or nowhere for consistency.

GaelVaroquaux: Done, thanks

rcap107:
Looks good to me, I'll merge after the CI is done

Thanks @GaelVaroquaux

PR #1751, opened 2025-11-12
Start applying ruff/Perflint rules (PERF)

DimitriPapadopoulos:
PERF401 Use `list.extend` to create a transformed list


rcap107:
Thanks for the PR, but I do not think this is needed

PR #1699, opened 2025-10-24
MAINT - black → ruff format

DimitriPapadopoulos:
Fixes #1665.

Supersedes #1695 

Please consider merging #1694 first, so that I can reformat with a recent version of ruff.

The default line length is 88, both for black and ruff. How about making explicit implicit, removing `line-length = 88`?

PR #1752, opened 2025-11-12
Update black config

DimitriPapadopoulos:
In case you'd want to avoid #1699.


rcap107:
We're going to switch to ruff, so we can close this in favor of #1699

PR #1753, opened 2025-11-12
Apply Repo-Review rules PP304 to PP308: improve pytest configuration

DimitriPapadopoulos:
Improves pytest configuration to catch more issues.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -333,7 +333,6 @@ log_cli_level = "INFO"
 xfail_strict = true
 addopts = [
   "--doctest-modules",
-  "-ra",
   "--strict-config",
   "--strict-markers",
 ]


rcap107:
Thanks for the PR. 

Note that we use `xfail` in various places to prevent the CI from failing due to downloading issues, or for some specific configurations. Setting `xfail_strict = true` without addressing those tests will prevent the CI from passing in any case, so that should be done before merging this PR.

DimitriPapadopoulos:
> Note that we use `xfail` in various places to prevent the CI from failing due to downloading issues

Any suggestion how to identify such tests?

rcap107:
Hi @DimitriPapadopoulos, looking at the tests that fail, tests in `test_fetching` are what I was talking about: now they're failing with strict because the download did in fact not fail. Those tests should not be strict.

Some other tests that need to be skipped are those in `test_docstrings`: they're marked as todo and I've opened #1882 to track work done there. 

 I also don't think that printing info about the skipped tests adds much, rather it's making the already very long log more verbose, so I'd rather remove that.

DimitriPapadopoulos:
What do you mean exactly by _that_? Remove `"--strict-config"` or remove the `pytest.xfail()` messages?

rcap107:
> What do you mean exactly by _that_? Remove `"--strict-config"` or remove the `pytest.xfail()` messages?

I mean remove the `-ra` flag from the pytest config. I don't think it brings useful information.

rcap107:
I think the pre-commit check is failing because there is a mismatch between lockfiles, it should be fixed after merging with main

rcap107 (approved):
Thanks for the PR, this will improve the tests

PR #1759, opened 2025-11-17
Make compute_ngram_distance private

### LINKED ISSUE #1755: Make the function `compute_ngram_distance` private
> I'm in favor of making it private. It is too low-level and does not do enough to be useful for users IMO. I has just 2 lines of code to call TfidfVectorizer then pdist, and leaves it up to the caller to make sure input strings are unique and actually do something of the resulting distances, so it really looks like a small private helper to me. 

 _Originally posted by @jeromedockes in [#1730](https://github.com/skrub-data/skrub/issues/1730#issuecomment-3538417347)_

ghost:
Renamed `compute_ngram_distance` to `_compute_ngram_distance` to make it private.
This is because this function is a low-level helper and should not be part of the public API.
I wasn’t able to run the full test suite locally due to missing dependencies, but all references and imports have been updated.
Closes #1755

### DIFF
--- a/examples/0050_deduplication.py
+++ b/examples/0050_deduplication.py
@@ -159,9 +159,9 @@ translation_table.head()
 
 from scipy.spatial.distance import squareform
 
-from skrub import compute_ngram_distance
+from skrub import _compute_ngram_distance
 
-ngram_distances = compute_ngram_distance(unique_examples)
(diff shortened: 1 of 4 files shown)


rcap107:
This PR should also have a changelog entry

PR #1762, opened 2025-11-19
FIX - adding get_feature_names_out to Cleaner

rcap107:
# Bug Fix Pull Request

## Description
The `Cleaner` did not have a `get_feature_names_out` method, and did not expose `all_outputs_`. This PR fixes the problem by adding both, to be consistent with the TableVectorizer. 

## How Has This Been Tested?
A new test has been added to check the behavior of `get_feature_names_out` in the Cleaner.


skrub/_table_vectorizer.py
```diff
@@ -198,6 +198,9 @@ class Cleaner(TransformerMixin, BaseEstimator):
         Maps the name of each column to a list of all the processing steps that were
         applied to it.
 
+    all_outputs_ : list of str
```
GaelVaroquaux, line 201:
```suggestion
    all_outputs_names_ : list of str
```

skrub/_table_vectorizer.py
```diff
@@ -355,6 +358,7 @@ def fit_transform(self, X, y=None):
         )
         self._pipeline = make_pipeline(*all_steps)
         result = self._pipeline.fit_transform(X)
+        self.all_outputs_ = sbd.column_names(result)
```
GaelVaroquaux, line 361:
```suggestion
        self.all_outputs_names_ = sbd.column_names(result)
```

skrub/_table_vectorizer.py
```diff
@@ -398,6 +402,22 @@ def fit(self, X, y=None):
...
+        input_features : array-like of str or None, default=None
+            Ignored.
+
+        Returns
+        -------
+        list of strings
+            The column names.
+        """
+        check_is_fitted(self, "all_outputs_")
+        return np.asarray(self.all_outputs_)
```
GaelVaroquaux, line 419:
```suggestion
        return np.asarray(self.all_outputs_names_)
```

GaelVaroquaux (commented):
I think that another name would be more understandable. What do you think?

rcap107:
> I think that another name would be more understandable. What do you think?

We are using `all_outputs_` in most other places in the library, so if we want to change the name we should do it everywhere. I don't think it's worth it.

GaelVaroquaux:
> We are using all_outputs_ in most other places in the library, so if we want to change the name we should do it everywhere. I don't think it's worth it.

OK

GaelVaroquaux (approved):
LGTM. Merging, thanks!!

PR #1764, opened 2025-11-20
FIX - Proper escaping of special characters for data ops tooltips

### LINKED ISSUE #1760: Drawing of DataOps plan fails for scripts containing backslashes
### Describe the bug

Using line breaks via backslash in DataOp plan definition results in failures during drawing the plan. 

Since the "\\" is not properly handled as special character in the html file.

### Steps/Code to Reproduce

```python
import skrub
out = skrub.as_data_op(0) + \
    1
g = out.skb.draw_graph()
g.svg.decode("utf-8")
```

### Expected Results

No error is thrown.

### Actual Results

```
"dot" with args ['-Tsvg', '/var/folders/6h/lrv_1lzs6835g_mwvby2hc1w0000gn/T/tmp33rymr10/tmps_4n9g36'] returned code: 1

stdout, stderr:
 b''
b'Error: /var/folders/6h/lrv_1lzs6835g_mwvby2hc1w0000gn/T/tmp33rymr10/tmps_4n9g36: syntax error in line 4 near \'"\'\n'

  File "/Users/elias/PycharmProjects/skrub/examples/tmp.py", line 10, in <module>
    main()
  File "/Users/elias/PycharmProjects/skrub/examples/tmp.py", line 7, in main
    g.svg.decode("utf-8")
  File "/Users/elias/PycharmProjects/skrub/skrub/_data_ops/_inspection.py", line 258, in svg
    svg = self.graph.create_svg(encoding="utf-8")
  File "/Users/elias/PycharmProjects/skrub/.venv/lib/python3.9/site-packages/pydot/core.py", line 181, in __create_method
    return self.create(format=f, prog=prog, encoding=encoding)
  File "/Users/elias/PycharmProjects/skrub/.venv/lib/python3.9/site-packages/pydot/core.py", line 1844, in create
    assert process.returncode == 0, (
AssertionError: "dot" with args ['-Tsvg', '/var/folders/6h/lrv_1lzs6835g_mwvby2hc1w0000gn/T/tmp33rymr10/tmps_4n9g36'] returned code: 1
```

### Versions

```shell
System:
    python: 3.9.6 (default, Feb  3 2024, 15:58:27)  [Clang 15.0.0 (clang-1500.3.9.4)]
executable: /Users/elias/PycharmProjects/skrub/.venv/bin/python
   machine: macOS-26.0.1-arm64-arm-64bit

Python dependencies:
      sklearn: 1.6.1
          pip: None
   setuptools: 80.9.0
        numpy: 2.0.2
        scipy: 1.13.1
       Cython: None
       pandas: 2.3.3
   matplotlib: 3.9.4
       joblib: 1.5.2
threadpoolctl: 3.6.0

Built with OpenMP: True

threadpoolctl info:
       user_api: blas
   internal_api: openblas
    num_threads: 12
         prefix: libopenblas
       filepath: /Users/elias/PycharmProjects/skrub/.venv/lib/python3.9/site-packages/scipy/.dylibs/libopenblas.0.dylib
        version: 0.3.27
threading_layer: pthreads
   architecture: neoversen1

       user_api: openmp
   internal_api: openmp
    num_threads: 12
         prefix: libomp
       filepath: /Users/elias/PycharmProjects/skrub/.venv/lib/python3.9/site-packages/sklearn/.dylibs/libomp.dylib
        version: None
0.7.dev0
```

rcap107:
# Bug Fix Pull Request

## Description
The PR fixes a bug that occurs when a data op contains a line break..

Special characters are now escaped properly with HTML-safe equivalents. 

Fixes #1760

```python
import skrub
out = skrub.as_data_op(0) + \
    1
g = out.skb.draw_graph()
g.svg.decode("utf-8")
```
This raises an exception due to the presence of special characters (`\` and the new line).


skrub/_data_ops/_inspection.py
```diff
@@ -309,6 +309,11 @@ def _node_kwargs(data_op, url=None):
     tooltip = html.escape(data_op._skrub_impl.creation_stack_last_line())
     if description := data_op._skrub_impl.description:
         tooltip = f"{tooltip}\n\n{html.escape(description)}"
+    # Escape special characters in tooltip for DOT format compatibility
+    # Backslashes must be escaped first to avoid double-escaping
+    tooltip = (
+        tooltip.replace("\\", "\\\\").replace("\n", "&#10;").replace('"', "&quot;")
+    )
```
jeromedockes:
```suggestion
    # Also escape backslahses inserted in the .dot file
    # otherwise they are interpreted as escape sequences by graphviz
    tooltip = tooltip.replace('\\', '\\\\')
```

jeromedockes:
thanks a lot @rcap107 ! Looks good but I think we could add a test, for example the snippet you provided in the description?

skrub/_data_ops/tests/test_inspection.py
```diff
@@ -170,6 +170,14 @@ def test_draw_graph_open(monkeypatch):
     mock.assert_called_once()
 
 
+@pytest.mark.skipif(not _inspection._has_graphviz(), reason="report requires graphviz")
+def test_escaping_characters():
+    out = skrub.as_data_op(0) + 1
```
jeromedockes: this doesn't test the handling of backslashes as intended... I'm guessing you originally had something like 

```
... \
+ 1
```

and the pre-commit reformatted it to put it all on one line. and actually we will need to put this test out of the data_ops module because stack frames inside that module are discarded from the displayed traceback so it wouldn't show up anyway. Basically we need to make sure the added test fails with the old version of the code and passes with the new one -- this one would pass even without the fix

jeromedockes:
I opened a pr on your branch to fix the test @rcap107 : https://github.com/rcap107/skrub/pull/1

jeromedockes (approved):
thanks a lot @rcap107 :)

PR #1767, opened 2025-11-20
 FIX - Improving the error message raised when the TableReport is called on a lazy polars dataframe

fatiben2002:
- Removed, in `_table_report.py`, the extra function for raising an exception and added the latter directly in the class `TableReport`.
- Removed the import in the file `test_table_report.py`
- Used this line to import polars `pl = pytest.importorskip("polars")`
- Improved the message in the `test_table_report.py` for the test function  
- Added a comment in the changelog 

P.S: There was an old comment about the same fix in the changelog by another user, I did not delete it. Let me know if I should :) or if I need any other changes. Sorry for the delay again!

Note: Due to some issue with the branch I had to create a new PR, [here](https://github.com/skrub-data/skrub/pull/1732) is the link to the last one.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,12 +41,12 @@ Changes
   :pr:`1644` by :user:`Simon Dierickx <dierickxsimon>`
 - The transformer :class:`ToFloat`, previously called `ToFloat32`, is now public.
   :pr:`1687` by :user:`Marie Sacksick <MarieSacksick>`.
-
-Bugfixes
---------
 - Improved the error message raised when a Polars lazyframe is passed to
   :class:`TableReport`, clarifying that `.collect()` must be called first.
   :pr:`1767` by :user:`Fatima Ben Kadour <fatiben2002>`
+
+Bugfixes
+--------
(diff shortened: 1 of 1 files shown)


rcap107:
Hi @fatiben2002, is there a reason why you opened this PR rather than continuing with #1732? It's not a big deal, just making sure we keep the work in one place.

fatiben2002:
> Hi @fatiben2002, is there a reason why you opened this PR rather than continuing with #1732? It's not a big deal, just making sure we keep the work in one place.

Hi @rcap107, I was encountering some issues with my branch and my software so I couldn't push to the old request. Hope it's okay, also I joined the link of the last PR in the current PR description. Thank u !

rcap107:
Hi @fatiben2002, thanks a lot for the PR. I pushed a few formatting fixes just so you wouldn't have to bother with that.

If you're curious, all I had to do was running `pre-commit run --all-files` from my dev environment, and push the result. 

I also moved the changelog entry from "Bugfixes" to "Changes". Other than that, the code looks good, and we can merge once the tests pass.

PR #1772, opened 2025-11-24
FEAT -  Adding `decimal` as parameter for `ToFloat32`

### LINKED ISSUE #1728: ToFloat fails when trying to parse numbers with "," decimal separators
### Describe the bug

I am trying to convert to float a string number with comma (",") decimal separators. I am trying to use ToFloat but it is failing to parse.

### Steps/Code to Reproduce

```
>>> import skrub
import pandas as pd
>>> import pandas as pd
>>>
>>> tf = skrub.ToFloat()
>>> s = pd.Series(["1,1", "2,2"])
>>> s
0    1,1
1    2,2
dtype: object
>>> tf.fit_transform(s)
```

### Expected Results

```
>>> tf.fit_transform(s)
0    1.1
1    2.2
```

### Actual Results

```
  File "pandas/_libs/lib.pyx", line 2407, in pandas._libs.lib.maybe_convert_numeric
ValueError: Unable to parse string "1,1"

During handling of the above exception, another exception occurred:

  File "/Users/Shared/work/skrub/skrub/_to_float.py", line 194, in fit_transform
    numeric = sbd.to_float32(column, strict=True)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rcap/.local/share/uv/python/cpython-3.12.8-macos-aarch64-none/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Shared/work/skrub/skrub/_dataframe/_common.py", line 784, in _to_float32_pandas
    col = pd.to_numeric(col, errors="raise" if strict else "coerce")
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/rcap/venvs/skrub-dev/lib/python3.12/site-packages/pandas/core/tools/numeric.py", line 235, in to_numeric
    values, new_mask = lib.maybe_convert_numeric(  # type: ignore[call-overload]
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas/_libs/lib.pyx", line 2449, in pandas._libs.lib.maybe_convert_numeric
ValueError: Unable to parse string "1,1" at position 0

The above exception was the direct cause of the following exception:

  File "/Users/Shared/work/skrub/skrub/_apply_to_cols.py", line 176, in fit_transform
    return f(self, X, y=y, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/Shared/work/skrub/skrub/_to_float.py", line 197, in fit_transform
    raise RejectColumn(
skrub._apply_to_cols.RejectColumn: Could not convert column None to numbers.
```

### Versions

```shell
System:
    python: 3.12.8 (main, Dec  6 2024, 19:42:06) [Clang 18.1.8 ]
executable: /Users/work/venvs/skrub-dev/bin/python
   machine: macOS-26.0.1-arm64-arm-64bit

Python dependencies:
      sklearn: 1.7.1
          pip: None
   setuptools: 80.9.0
        numpy: 2.3.2
        scipy: 1.16.1
       Cython: None
       pandas: 2.3.2
   matplotlib: 3.10.5
       joblib: 1.5.1
threadpoolctl: 3.6.0

Built with OpenMP: True

threadpoolctl info:
       user_api: openmp
   internal_api: openmp
    num_threads: 8
         prefix: libomp
       filepath: /Users/work/venvs/skrub-dev/lib/python3.12/site-packages/sklearn/.dylibs/libomp.dylib
        version: None
0.7.dev0
```

gabrielapgomezji:
This issue addresses #1728 

The ToFloat transformer now includes a `decimal` parameter that lets the user specify the decimal separator to use for the given column. Then, all the possible thousands separators are removed, and the decimal separator is converted to a `.` before the column is passed to `to_float32`.


doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,112 @@
...
+
+- the **Cleaner** (|Cleaner|), to normalize all numeric-like columns before modeling
+- the **|TableVectorizer|**, ensuring a consistent numeric dtype across all numeric features
+
+This makes |ToFloat| a core building block of skrub’s handling of heterogeneous
+tabular data.
+
+``ToFloat`` ensures that downstream machine-learning models receive numeric data
+that is clean, consistent, lightweight, and free of locale-specific quirks or
+string-encoded values.
```
rcap107: This section can be removed
```suggestion
```

skrub/_to_float.py
```diff
@@ -1,8 +1,34 @@
...
+@dispatch
+def _str_replace(col, strict=True):
+    raise_dispatch_unregistered_type(col, kind="Series")
+
+
+@_str_replace.specialize("pandas", argument_type="Column")
+def _str_replace_pandas(col, decimal):
+    pattern = POSSIBLE_SEPARATORS.copy()
+    pattern.remove(decimal)
+    col = col.str.replace(r"^\((.*)\)$", r"-\1", regex=True)
```
emassoulie: Should this transformation be systematic? There may be edge cases where a parsed number just happens to be in parentheses, although they may be too few to justify adding a rule.

skrub/tests/test_to_float.py
```diff
@@ -43,3 +43,49 @@ def test_rejected_columns(df_module):
...
+        ("-1.234,56", -1234.56, ","),
+        ("(1,234.56)", -1234.56, "."),
+        ("(1.234,56)", -1234.56, ","),
+        ("1,23,456.78", 123456.78, "."),
+        ("12,3456.78", 123456.78, "."),
+        (".56", 0.56, "."),
+        (",56", 0.56, ","),
+    ],
+)
+def test_number_parsing(input_str, expected_float, decimal, df_module):
```
emassoulie: Might it be worth adding tests for the code's behaviour in case of an invalid entry?

skrub/_to_float.py
rcap107: For the parentheses in particular, I think that the main use case (numbers used in accounting) should be parsed by default. I think most users would just remove the parentheses as part of the parsing operation. We should probably make this behavior more evident in the docstring, however. 

I am starting to think that these heuristics should have a flag that can turn them on/off.

skrub/tests/test_to_float.py
rcap107: yes, we should check a few weird cases and make sure they fail as expected

rcap107:
After some discussion, I think this PR needs some more time before it can be merged, and unfortunately won't be part of the next release.

The current implementation is removing all thousands separators other than what is specified as the "decimal" separator, which is quite risky and may leads to problems. It's better to follow what pandas is doing, i.e., have both `decimal` and `thousands` as separators. By default, the `thousands` separator should be `None` (so no replacement).

If there is some kind of weird string like `1,2.3,4`, it should not be parsed as a number. I am not sure how far we should do to parse something like `1,2.34` with decimal `.` and thousands `,`: it's not a format I recognize, but it would still be recognized as `12.34` rather than being rejected. 

Another check that may be considered is counting the number of `decimal` separators, and reject any case where there is more than one. 

Some additional comments:
- While it's impossible to test all possible scenarios, tests should also include as many weird edge cases as we can come up with to see what could be the result.
- The ToFloat docstring needs some more work to explain in more detail the behavior when `decimal` and `thousands` are set. 

I'll convert this back to draft and keep an eye on this for the next PR.

gabrielapgomezji:
When talking about the tests, it was mentioned to include 3 tests:
- A test for Good inputs
- A test for Bad Inputs
- A test for bad parameters
I merged the last two tests including also bada parameters in the test. If it's better to have the 3 tests individually instead of the 2, I will modify it.

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,133 @@
...
+- **Automatic parsing of decimal and thousands separators**, regardless of locale:
+  - The decimal separator must be specified explicitly and can be either ``.`` or ``,``
+  - The thousands separator can be one of ``.``, ``,``, space (``" "``), apostrophe (``'``),
+  or None (no thousands separator)
+  - The transformer supports integers, decimals (including leading-decimal forms such as .56 or ,56), scientific notation
+  and negative numbers (including parentheses)
+  - Decimal and thousands separators must be different characters
+
+- **Parentheses interpreted as negative numbers**, a common format in financial datasets:
+  - ``(1,234.56)`` → ``-1234.56``
```
rcap107:
```suggestion
  - The transformer supports integers, decimals (including leading-decimal forms such as .56 or ,56), scientific notation
  and negative numbers
  - Numbers in parentheses are interpreted as negative numbers (``(1,234.56)`` → ``-1234.56``). This format is more common in financial datasets.
  - Decimal and thousands separators must be different characters
```

this will break formatting

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,133 @@
...
+As with all skrub transformers, |ToFloat| behaves like a standard
+scikit-learn transformer and is fully compatible with pipelines.
+
+How to use |ToFloat|
+--------------------
+The |ToFloat| transformer must be applied to individual columns. It behaves like
+a standard scikit-learn transformer.
+Each column is expected to use a single decimals and thousands separator, which is
+specified through the ``decimal`` and ``thousand`` parameter. If this parameter is not provided,
+the default decimal separators are ``'.'`` and ``None``.
```
rcap107:
```suggestion
The |ToFloat| transformer must be applied to individual columns, and it behaves
like a standard scikit-learn transformer. 
|ToFloat| requires a ``decimal`` and a ``thousands`` separator, which are ``'.'`` and 
``None`` (no thousands separator) by default.
Each column is expected to use a single separator for decimals, and one for thousands: 
if any characters other than the provided selectors are encountered in the column, it will not 
be converted. 
```

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,133 @@
...
+
+During ``fit``, |ToFloat| attempts to convert all values in the column to
+numeric values after automatically removing other possible thousands separators
+(``,``, ``.``, space, apostrophe). If any value cannot be converted, the column
+is rejected with a ``RejectColumn`` exception.
```
rcap107: I think this is not how the current version of the code is working: the regex pattern should reject anything that contains characters different from either the decimal or thousands separator. 

There should also be an explanation of how the check is done (checking if there are parentheses, checking if thousands are separated by groups of 3 digits, adding the scientific notation)

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,133 @@
...
+>>> s = pd.Series(['1.1', None, '3.3'], name='x')
+>>> ToFloat().fit_transform(s)
+0    1.1
+1    NaN
+2    3.3
+Name: x, dtype: float32
+
+Locale-dependent decimal separators can be handled by specifying the
+``decimal`` and ``thousand`` parameter. Here we use comma as decimal separator, and
+a spaces as thousands separators:
```
rcap107:
```suggestion
a space as thousands separators:
```

skrub/_to_float.py
```diff
@@ -1,9 +1,61 @@
+import re
+
 from . import _dataframe as sbd
 from ._apply_to_cols import RejectColumn, SingleColumnTransformer
+from ._dispatch import dispatch, raise_dispatch_unregistered_type
 
 __all__ = ["ToFloat"]
 
 
+def _build_number_regex(decimal, thousand):
```
rcap107: Please add comments explaining what each part of this regex is doing, it is quite complex and without comments it will be hard to maintain

skrub/_to_float.py
```diff
@@ -1,9 +1,61 @@
...
+
+
+@dispatch
+def _str_replace(col, strict=True):
+    raise_dispatch_unregistered_type(col, kind="Series")
+
+
+@_str_replace.specialize("pandas", argument_type="Column")
+def _str_replace_pandas(col, decimal, thousand):
+    col = col.str.replace(r"^\((.*)\)$", r"-\1", regex=True)
```
rcap107: Please add a comment explaining that this is replacing the parentheses with a `-` sign

skrub/_to_float.py
```diff
@@ -22,6 +74,17 @@ class ToFloat(SingleColumnTransformer):
...
+    Parameters
+    ----------
+    decimal : str, default='.'
+        Character to recognize as the decimal separator when converting from
+        strings to floats. Other possible decimal separators are removed from
+        the strings before conversion.
+    thousand : str or None, default=None
+    Character used as thousands separator. Supported values are ``"."``,
+    ``,``, space (``" "``), apostrophe (``"'"``), or ``None`` (no thousands
+    separator). The decimal and thousands separators must differ.
```
rcap107:
```suggestion
        Character used as thousands separator. Supported values are ``"."``,
        ``,``, space (``" "``), apostrophe (``"'"``), or ``None`` (no thousands
        separator). The decimal and thousands separators must differ.
```

I think doctests are failing because this is not tabbed properly

skrub/_to_float.py
```diff
@@ -165,8 +228,35 @@ class ToFloat(SingleColumnTransformer):
     >>> s = pd.Series([1.1, None], dtype='float32')
     >>> to_float.fit_transform(s) is s
     True
+
+    Handling parentheses around negative numbers
```
rcap107:
```suggestion
    Negative numbers represented using parentheses are converted
    so they use "-" instead.
```

skrub/_to_float.py
```diff
@@ -165,8 +228,35 @@ class ToFloat(SingleColumnTransformer):
...
+    >>> s = pd.Series(["-1,234.56", "1,234.56", "(1,234.56)"], name='parens')
+    >>> ToFloat(decimal=".", thousand=",").fit_transform(s)
+    0   -1234.5...
+    1    1234.5...
+    2   -1234.5...
+    dtype: float32
+
+    Scientific notation
```
rcap107:
```suggestion
    Numbers that use scientific notation are converted:
```

skrub/_to_float.py
```diff
@@ -165,8 +228,35 @@ class ToFloat(SingleColumnTransformer):
...
+    >>> s = pd.Series(["1.23e+4", "1.23E+4"], name="x")
+    >>> ToFloat(decimal=".").fit_transform(s)
+    0    12300.0
+    1    12300.0
+    Name: x, dtype: float32
+
+    Space as thousand separator
```
rcap107:
```suggestion
    It is possible to specify the thousands separator, e.g., to use " "
```

skrub/tests/test_to_float.py
```diff
@@ -43,3 +43,55 @@ def test_rejected_columns(df_module):
...
+        ("1.2.3.4,0", ",", "."),
+        ("12,3456.78", ".", ","),
+        ("1 234,567.34", ".", ","),
+        ("1'234,567.34", ".", ","),
+        ("1'234'234,567.34", ",", "'"),
+        ("123.45.67", ".", None),
+        ("1,,234", ".", ","),
+        ("1.23,45", ".", ","),
+        # decimal == thousand
+        ("123,456,789", ",", ","),
```
rcap107: Here we are testing  that `RejectColumn` is raised as expected when it encounters values that should not be converted. This case should be moved to a separate test that verifies that the correct exception is raised if the parameters are incorrect. The same (new) test should also check that a ValueError is raised if `decimal` is None.

rcap107 (dismissed):

rcap107:
Thanks a lot for the PR @gabrielapgomezji! This will be very useful for parsing data that is not in the usual locale.

My comments are mostly about improving clarity in the documentation and adding comments in the code. I think the actual content of the PR is in a good shape, it's just a matter of polishing at this point.

skrub/_to_float.py
```diff
@@ -185,12 +276,28 @@ def fit_transform(self, column, y=None):
         """
         del y
         self.all_outputs_ = [sbd.name(column)]
+        if self.thousand is None:
+            self.thousand = ""  # No thousand separator
```
rcap107: This should be moved to the init, parameters should not be modified in the fit

skrub/_to_float.py
```diff
@@ -1,9 +1,99 @@
...
+    if not col.fillna("").str.match(number_re, na=False).all():
+        raise RejectColumn(f"Could not convert column {sbd.name(col)!r} to numbers.")
+    return True
+
+
+@_str_is_valid_number.specialize("polars", argument_type="Column")
+def _str_is_valid_number_polars(col, number_re):
+    # Check if all values in the column match the number regex.
+    # - Fill NaN values with empty string to avoid match errors.
+    # - Use `str.match` with `na=False` to treat empty/missing values as non-matching.
```
rcap107:
```suggestion
    # - Use `str.contains` with `literal=False` to treat empty/missing values as non-matching.
```

skrub/_to_float.py
```diff
@@ -22,6 +112,17 @@ class ToFloat(SingleColumnTransformer):
     During ``transform``, entries for which conversion fails are replaced by
     null values.
 
+    Parameters
+    ----------
+    decimal : str, default='.'
+        Character to recognize as the decimal separator when converting from
+        strings to floats. Other possible decimal separators are removed from
+        the strings before conversion.
```
rcap107, line 61: not the case anymore
```suggestion
        strings to floats.
```

skrub/_to_float.py
```diff
@@ -165,8 +266,36 @@ class ToFloat(SingleColumnTransformer):
...
+    dtype: float32
+
+    Numbers that use scientific notation are converted:
+    >>> s = pd.Series(["1.23e+4", "1.23E+4"], name="x")
+    >>> ToFloat(decimal=".").fit_transform(s)
+    0    12300.0
+    1    12300.0
+    Name: x, dtype: float32
+
+    It is possible to specify the thousands separator, e.g., to use " "
```
rcap107:
```suggestion
    It is possible to specify the thousands separator, e.g., to use ``" "``
```

skrub/_to_float.py
```diff
@@ -165,8 +266,36 @@ class ToFloat(SingleColumnTransformer):
...
+    >>> s = pd.Series(["4 567,89", "12 567,89"], name="x")
+    >>> ToFloat(decimal=",", thousand=" ").fit_transform(s) # doctest: +ELLIPSIS
```
rcap107: ELLIPSIS is enabled by default
```suggestion
    >>> ToFloat(decimal=",", thousand=" ").fit_transform(s)
```

rcap107 (commented):
A few more cosmetic fixes, but I think we're almost done here. Thanks @gabrielapgomezji

rcap107:
I did a very quick and dirty benchmark comparing the performance of the ToFloat transformer in the latest version of skrub,  and the ToFloat in this PR. 

I generated a dataframe with 10M rows and 30 columns. 

There is a small, but noticeable difference in time when there is no conversion to be done, so we might want to add a condition where the formatting check is skipped in the default case (decimal="." and thousands=None). 
```
skrub version: 0.8.dev0
Elapsed time for ToFloat transformation: 0.3026 seconds
skrub version: 0.7.1
Elapsed time for ToFloat transformation: 0.2914 seconds
```

<details>
<summary>Code</summary>

```python
import time
import polars as pl
import numpy as np
import skrub
from importlib import metadata
import polars.selectors as cs

version = metadata.version("skrub")
print(f"skrub version: {version}")

# Set random seed for reproducibility
np.random.seed(42)

# Generate random float data
data = np.random.uniform(1000, 1000000, size=(10_000_000, 30))

# Convert floats to strings with space as thousands separator
df = pl.DataFrame(data)

from skrub import ToFloat, ApplyToCols
def benchmark_tofloat(df):
    tic = time.time()
    transformer = ApplyToCols(ToFloat())
    transformed_df = transformer.fit_transform(df)
    toc = time.time()

    return toc - tic

# Run benchmark
times = []
for run in range(100):
    elapsed_time = benchmark_tofloat(df)
    times.append(elapsed_time)

elapsed_time = np.median(times)
print(f"Elapsed time for ToFloat transformation: {elapsed_time:.4f} seconds")
```
</details>

rcap107 (commented):
Note from IRL meeting on this: the regex should also check for blocks of 4 digits, not only blocks of 3 (like now). The easiest way to do that is likely to have two separate regex. If the first (checking groups of 3) fails, then the second is tested, and if the second also fails then the column is rejected.

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,133 @@
...
+The |ToFloat| transformer is used internally by both the |Cleaner| and the
+|TableVectorizer| to guarantee that downstream estimators receive clean and
+uniform numeric data.
+
+What |ToFloat| does
+-------------------
+
+The |ToFloat| transformer provides:
+
+- **Automatic conversion to 32-bit floating-point values (`float32`).**
```
rcap107:
```suggestion
- **Automatic conversion to 32-bit floating-point values (``float32``).**
```

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,133 @@
...
+  This dtype is lightweight and fully supported by scikit-learn estimators.
+
+- **Automatic parsing of decimal and thousands separators**, regardless of locale:
+  - The decimal separator must be specified explicitly and can be either ``.`` or ``,``
+  - The thousands separator can be one of ``.``, ``,``, space (``" "``), apostrophe (``'``),
+  or None (no thousands separator)
```
rcap107:
```suggestion
  or ``None`` (no thousands separator)
```

skrub/_to_float.py
```diff
@@ -165,8 +336,63 @@ class ToFloat(SingleColumnTransformer):
...
+    1    12567.8...
+    Name: x, dtype: float32
+
+
+    Number grouping patterns supported:
+
+    >>> import pandas as pd
+    >>> from skrub._to_float import ToFloat
+
+    # 1) Groups of 3 digits (Western style)
```
rcap107:
```suggestion
    # 1) Groups of 3 digits
```
I'm thinking that it may be better to remove "Western style" here (and "Indian style" and "Chinese style" elsewhere). The numerical examples should remain.

skrub/_to_float.py
```diff
@@ -185,12 +411,28 @@ def fit_transform(self, column, y=None):
...
+            raise ValueError("The thousand and decimal separators must differ.")
+
         if sbd.is_any_date(column) or sbd.is_categorical(column):
             raise RejectColumn(
                 f"Refusing to cast column {sbd.name(column)!r} "
                 f"with dtype '{sbd.dtype(column)}' to numbers."
             )
         try:
+            if sbd.is_string(column):
+                if self.decimal == "." and self.thousand == "":
```
rcap107: Maybe add here a comment to explain this is the default case, so we skip the regex checks for performance reasons

CHANGES.rst
```diff
@@ -109,6 +112,9 @@ Bug Fixes
 
 Release 0.7.2
 =============
+- :class:`ToFloat32` has the parameter decimal to let the user specify whether they use ',' or '.'
+  as decimal separator and it also handles negative numbers indicated with parentheses.
+  :pr:`1772` by :user:`Gabriela Gómez Jiménez <gabrielapgomezji>`.
```
```suggestion
```

This part was duplicated for some reason

CHANGES.rst
```diff
@@ -11,6 +11,9 @@ Ongoing Development
 
 New Features
 ------------
+- :class:`ToFloat32` has the parameter decimal to let the user specify whether they use ',' or '.'
+  as decimal separator and it also handles negative numbers indicated with parentheses.
+  :pr:`1772` by :user:`Gabriela Gómez Jiménez <gabrielapgomezji>`.
```
```suggestion
- :class:`ToFloat32` now allows users to specify ``decimal`` and ``thousand`` separators to parse numerical columns that use formatting different from the default formatting used in Python, such as ``1'234,5``.  Multi-group formats (``1,23,456.78``) and groups of 4 digits (``12,3456.78``) are also supported. Additionally, negative numbers indicated with parentheses are converted to the regular numeric format (``(432)`` becomes ``-432``). :pr:`1772` by :user:`Gabriela Gómez Jiménez <gabrielapgomezji>`.
```

rcap107: This will likely break the formatting

rcap107 (commented):
Hi @gabrielapgomezji, thanks a lot. I think we're almost done with this: i have mostly comments on style or wording, but I think the functionality and tests are what we want.

skrub/_to_float.py
```diff
@@ -1,9 +1,40 @@
 from . import _dataframe as sbd
+from ._dispatch import dispatch, raise_dispatch_unregistered_type
 from ._single_column_transformer import RejectColumn, SingleColumnTransformer
 
 __all__ = ["ToFloat"]
 
 
+@dispatch
+def _str_replace(col, strict=True):
```
jeromedockes:
```suggestion
def _str_replace(col, decimal, thousand, parentheses):
```

skrub/_to_float.py
```diff
@@ -1,9 +1,40 @@
...
+@_str_replace.specialize("polars", argument_type="Column")
+def _str_replace_polars(col, decimal, thousand, parentheses):
+    # Replace parentheses around numbers with a leading minus sign
+    # e.g., "(123.45)" → "-123.45"
+    if parentheses:
+        col = col.str.replace_all(r"^\((.*)\)$", r"-$1")
+    # Remove thousand separators
+    col = col.str.replace_all(thousand, "", literal=True)
+    # Replace decimal separator with '.'
+    return col.str.replace_all(f"[{decimal}]", ".")
```
jeromedockes: why not `replace_all(decimal, ".", literal=True)` similar to the line above?

skrub/_to_float.py
```diff
@@ -165,8 +207,64 @@ class ToFloat(SingleColumnTransformer):
     >>> s = pd.Series([1.1, None], dtype='float32')
     >>> to_float.fit_transform(s) is s
     True
+
+    Negative numbers represented using parentheses are converted
```
jeromedockes:
```suggestion
    If ``parentheses=True``, negative numbers represented using parentheses are converted
```

skrub/_to_float.py
```diff
@@ -22,6 +53,17 @@ class ToFloat(SingleColumnTransformer):
...
+    Parameters
+    ----------
+    decimal : str, default='.'
+        Character to recognize as the decimal separator when converting from
+        strings to floats. Other possible decimal separators are removed from
+        the strings before conversion.
+    thousand : str or None, default=None
+        Character used as thousands separator. Supported values are ``"."``,
+        ``,``, space (``" "``), apostrophe (``"'"``), or ``None`` (no thousands
+        separator). The decimal and thousands separators must differ.
```
jeromedockes, line 65: i think we are missing the entry for `parentheses`?

skrub/_to_float.py
```diff
@@ -165,8 +207,64 @@ class ToFloat(SingleColumnTransformer):
     >>> s = pd.Series([1.1, None], dtype='float32')
     >>> to_float.fit_transform(s) is s
     True
+
+    Negative numbers represented using parentheses are converted
+    so they use "-" instead.
```
jeromedockes:
```suggestion
    so they use "-" instead.
    
```

otherwise the doctest does not get rendered correctly, see https://output.circle-artifacts.com/output/job/00938aa8-187c-43c2-a4db-3fedfddc4f10/artifacts/0/doc/reference/generated/skrub.ToFloat.html

skrub/_to_float.py
```diff
@@ -165,8 +207,64 @@ class ToFloat(SingleColumnTransformer):
...
+    Name: x, dtype: float32
+
+    It is possible to specify the thousands separator, e.g., to use " "
+    >>> s = pd.Series(["4 567,89", "12 567,89"], name="x")
+    >>> ToFloat(decimal=",", thousand=" ").fit_transform(s) # doctest: +ELLIPSIS
+    0    4567.8...
+    1    12567.8...
+    Name: x, dtype: float32
+
+    Note
```
jeromedockes: I think we must follow the sections described in numpydoc or our sphinx template will not render them, see the docs rendered by the ci

https://numpydoc.readthedocs.io/en/latest/format.html#docstring-standard

skrub/_to_float.py
```diff
@@ -165,8 +207,64 @@ class ToFloat(SingleColumnTransformer):
...
+    Multiple grouping separators may be collapsed:
+
+    >>> s = pd.Series(["1.2.3.4,0"], name="x")
+    >>> ToFloat(decimal=",", thousand=".").fit_transform(s)
+    0    1234.0
+    Name: x, dtype: float32
     """  # noqa: E501
 
+    def __init__(self, decimal=".", thousand=None, parentheses=False):
+        super().__init__()
```
jeromedockes:
```suggestion
```

skrub/_to_float.py
```diff
@@ -165,8 +207,64 @@ class ToFloat(SingleColumnTransformer):
...
+        self.decimal = decimal
+        self.thousand = "" if thousand is None else thousand
```
jeromedockes: the `__init__` should not do any transformations or validation but just store the parameters


```suggestion
        self.thousand = thousand
```

the reason is that if someone then changes the parameters with `set_params`, `__init__` is not called again so any parameter validation and transformation must be done in fit and fit_transform, not init

skrub/_to_float.py
jeromedockes: we can change it slightly so we do not perform `is_string` unless we need to

```
if (self.decimal != "." or self.thousand not in (None, "")) and sbd.is_string(column):
    column = _str_replace(...)
numeric = sbd.to_float32(column, strict=True)
```

jeromedockes: we also need to do the string replacement in transform. so we can add a small helper `self._str_replace()` that checks if replacement is necessary and performs it

jeromedockes (commented):
thank you @gabrielapgomezji it is much simpler now!!

skrub/_to_float.py
```diff
@@ -181,13 +285,34 @@ def fit_transform(self, column, y=None):
...
+        if self._thousand == self.decimal:
+            raise ValueError("The thousand and decimal separators must differ.")
+
         if sbd.is_any_date(column) or sbd.is_categorical(column):
             raise RejectColumn(
                 f"Refusing to cast column {sbd.name(column)!r} "
                 f"with dtype '{sbd.dtype(column)}' to numbers."
             )
         try:
+            self._is_string = False
```
jeromedockes: let's use a more explicit name, like `_needs_str_replace` or something, because it's not just a question of being a string or not

skrub/_to_float.py
```diff
@@ -181,13 +285,34 @@ def fit_transform(self, column, y=None):
...
+            # Default case it's directly converted to float32.
+            # Regex validation and string replacement is skipped.
+            # Optimization: allows us to avoid manipulation for
+            # formats that Python can parse natively.
+            if (
+                self.decimal != "." or self._thousand not in (None, "")
```
jeromedockes: or self.parentheses

skrub/_to_float.py
```diff
@@ -162,8 +208,66 @@ class ToFloat(SingleColumnTransformer):
...
+    Name: x, dtype: float32
+
+    Multiple grouping separators may be collapsed:
+
+    >>> s = pd.Series(["1.2.3.4,0"], name="x")
+    >>> ToFloat(decimal=",", thousand=".").fit_transform(s)
+    0    1234.0
+    Name: x, dtype: float32
+
+    Notes
```
jeromedockes: the notes section has to come before the examples one I think. also it is not clear in the paragraph below that it is only about the 3 new  parameters. id does validate number formatting, once those replacements have been made

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,141 @@
+.. |ToFloat| replace:: :class:`~skrub.ToFloat`
```
jeromedockes, line 2: I think this page can be made a lot shorter or maybe even replaced by inserting a little paragraph in some existing section -- we have to keep the user guide short or users won't read it :)

skrub/_to_float.py
```diff
@@ -162,8 +208,66 @@ class ToFloat(SingleColumnTransformer):
...
+    >>> ToFloat(decimal=".", thousand=",").fit_transform(s)
+    0    1.2345
+    Name: x, dtype: float32
+
+    Multiple grouping separators may be collapsed:
+
+    >>> s = pd.Series(["1.2.3.4,0"], name="x")
+    >>> ToFloat(decimal=",", thousand=".").fit_transform(s)
+    0    1234.0
+    Name: x, dtype: float32
```
jeromedockes: i think all those 3 points and the notes can be collapsed into 1 sentence + example to keep things short. e.g. "this does not do advanced processing, just replaces thousands with '' and decimal with '.' then tries to parse as numbers: tofloat(...).fit_transform(['1,,23,4', ...])

jeromedockes (approved):
thank you very much @gabrielapgomezji ! LGTM aside from some comments on the docs

CHANGES.rst
```diff
@@ -48,7 +54,6 @@ Deprecations
 
 
 Release 0.9.0
-=============
```
jeromedockes, line 51: undo this change?

doc/column_level_featurizing.rst
```diff
@@ -15,5 +15,5 @@ directly by machine learning models.
 
    modules/column_level_featurizing/feature_engineering_categorical
    modules/column_level_featurizing/feature_engineering_datetimes
-   modules/column_level_featurizing/robust_scaling
```
jeromedockes, line 18: why are we removing robust_scaling from here?

doc/modules/column_level_featurizing/feature_engineering_numerical.rst
```diff
@@ -0,0 +1,106 @@
...
+.. |TableVectorizer| replace:: :class:`~skrub.TableVectorizer`
+.. |Cleaner| replace:: :class:`~skrub.Cleaner`
+.. |RobustScaler| replace:: :class:`~sklearn.preprocessing.RobustScaler`
+.. |RejectColumn| replace:: :class:`~skrub.core.RejectColumn`
+
+.. _user_guide_feature_engineering_numeric_to_float:
+
+Parsing and scaling numeric features
+==========================================================
```
jeromedockes:
```suggestion
====================================
```

doc/modules/column_level_featurizing/robust_scaling.rst
```diff
@@ -1,6 +1,110 @@
 .. |SquashingScaler| replace:: :class:`~skrub.SquashingScaler`
+.. |ToFloat| replace:: :class:`~skrub.ToFloat`
+.. |TableVectorizer| replace:: :class:`~skrub.TableVectorizer`
+.. |Cleaner| replace:: :class:`~skrub.Cleaner`
 .. |RobustScaler| replace:: :class:`~sklearn.preprocessing.RobustScaler`
 
+.. _user_guide_feature_engineering_numeric_to_float:
+
+Parsing and scaling numeric features with skrub
+==========================================================
```
jeromedockes: this content seems duplicated with feature_engineering_numerical?

skrub/_to_float.py
```diff
@@ -181,13 +276,36 @@ def fit_transform(self, column, y=None):
...
             )
         try:
+            self._needs_str_replace = False
+            # Default case it's directly converted to float32.
+            # Regex validation and string replacement is skipped.
+            # Optimization: allows us to avoid manipulation for
+            # formats that Python can parse natively.
+            if (
+                self.decimal != "."
+                or self._thousand not in (None, "")
```
jeromedockes:
```suggestion
                or self._thousand != ""
```

self._thousand can never be None i think

doc/modules/column_level_featurizing/robust_scaling.rst
rcap107: it is, I forgot to remove the old file

doc/column_level_featurizing.rst
rcap107: the content has been moved to feature_engineering_numerical

jeromedockes (approved):
Looks great now, thank you for all this work @gabrielapgomezji  and @rcap107 !! :)

PR #1775, opened 2025-11-27
FEAT - Use sampling seed in `summarize_dataframe` for reproducibility

thomass-dev:
Only for reproducibility: `summarize_dataframe` now accepts a new parameter `seed`, used to set the sampling seed (`sbd.sample`) when computing the top associations.

I wonder if we should reuse `SKB_SUBSAMPLING_SEED` from [skrub.DataOp.skb.subsample()](https://skrub-data.org/stable/reference/generated/skrub.DataOp.skb.subsample.html#skrub.DataOp.skb.subsample),


jeromedockes (approved):
thanks a lot @thomass-dev @glemaitre !

PR #1778, opened 2025-11-28
DOC - adding a new example on image classification with data ops

rcap107:
This PR adds a new example for handwritten digit classification using skrub's Data Ops and updates several broken links in the documentation to point to the correct examples.

### DIFF
--- a/doc/modules/data_ops/basics/building_data_ops_plan.rst
+++ b/doc/modules/data_ops/basics/building_data_ops_plan.rst
@@ -90,5 +90,5 @@ By working only on Data Ops we ensure that all the operations done on the data
 are added correctly to the computational graph, which then allows the resulting
 learner to execute all steps as intended.
 
-See :ref:`sphx_glr_auto_examples_data_ops_0110_data_ops_intro.py` for an introductory
+See :ref:`sphx_glr_auto_examples_data_ops_1110_data_ops_intro.py` for an introductory
 example on how to use skrub DataOps on a single dataframe.
--- a/examples/data_ops/1151_digits.py
+++ b/examples/data_ops/1151_digits.py
@@ -1,0 +1,165 @@
+"""
+Handwritten digit classification with skrub
+============================================
+
+This example demonstrates how to use skrub's Data Ops to build a
+machine learning pipeline for handwritten digit classification. We'll cover:
+
+1. Loading and exploring the UCI ML hand-written digits dataset
+2. Building a basic classification model with skrub Data Ops
+3. Performing hyperparameter tuning with multiple algorithms
+4. Comparing model performance before and after tuning
+
+This example is inspired by the official scikit-learn example:
+https://scikit-learn.org/stable/auto_examples/classification/plot_digits_classification.html
+"""
+
+# %%
+# Loading the Data
+# ================
+#
+# We use the UCI ML hand-written digits dataset provided by scikit-learn.
+# This dataset contains 8x8 pixel images of handwritten digits (0-9) that have
+# been flattened to 64-dimensional feature vectors. Our goal is to build a
+# classifier that can accurately predict the digit from the pixel values.
+
+from sklearn.datasets import load_digits
+
+digits = load_digits()
+X, y = digits.data, digits.target
+print(f"Dataset shape: {X.shape}")
+print(f"Number of classes: {len(set(y))}")
+# %%
+# Visualizing Sample Digits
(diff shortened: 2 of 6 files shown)


rcap107:
From skrub's meeting: hpo should take less time so I need to look into the grid to speed things up a bit

doc/modules/data_ops/basics/building_data_ops_plan.rst
```diff
@@ -90,5 +90,5 @@ By working only on Data Ops we ensure that all the operations done on the data
 are added correctly to the computational graph, which then allows the resulting
 learner to execute all steps as intended.
 
-See :ref:`sphx_glr_auto_examples_data_ops_0110_data_ops_intro.py` for an introductory
+See :ref:`sphx_glr_auto_examples_data_ops_1110_data_ops_intro.py` for an introductory
```
jeromedockes, line 93: We should start always using some explicitly named anchors like [this](https://github.com/skrub-data/skrub/blob/b19fd494197272ede1d785daa59616897b90ac92/examples/data_ops/1130_choices.py#L5) otherwise those links are likely to get broken again

examples/data_ops/1151_digits.py
```diff
@@ -0,0 +1,165 @@
+"""
+Handwritten digit classification with skrub
+============================================
+
+This example demonstrates how to use skrub's Data Ops to build a
```
jeromedockes, line 5: maybe this sentence could say why this example is here / what it adds wrt to the other examples? e.g. illustrate dataops can be useful on non-dataframish data or something like that

examples/data_ops/1151_digits.py
```diff
@@ -0,0 +1,165 @@
...
+# ~~~~~~~~~~~~~~~~~~~~~~~~~~~
+#
+# We begin with a Support Vector Machine (SVM) classifier. The skrub
+# Data Ops framework allows us to define our model and apply it to the data
+# in a declarative way using the apply method.
+
+from sklearn.svm import SVC
+
+model = SVC()
+predictions = X.skb.apply(model, y=y)
```
jeromedockes, line 71: here also we could insist on what distinguishes this example / what it adds wrt to the other hyperparam search examples. is it that we are using numpy arrays rather than dataframes?

maybe we could go one step further and use pytorch instead of numpy+sklearn, given that we have it installed for the example gallery anyway due to the textencoder. it's more often used for processing images and "can I use this other framework" is a question that comes up

examples/data_ops/1151_digits.py
```diff
@@ -0,0 +1,165 @@
...
+
+# %%
+# Comparing Results
+# ~~~~~~~~~~~~~~~~~
+#
+# Finally, we compare the performance of our baseline model with the
+# best model found during hyperparameter tuning:
+
+best_learner = search.best_learner_
+cv_results_search = skrub.cross_validate(
```
jeromedockes, line 150: technically we are overfitting the hyperparameters. maybe we should split away a test set at the beginning that we don't use for hyperparam search?

jeromedockes:
> From skrub's meeting: hpo should take less time so I need to look into the grid to speed things up a bit

is HPO necessary in this example (given that it is shown in a couple of other examples)? What is the main point you want to get across with this one? Deciding that might help find what can be cut to speed it up

jeromedockes:
Also, if you want to work more on the added example, there is the option to move the fix of broken links to a separate PR so that this part can be merged sooner

doc/modules/data_ops/basics/building_data_ops_plan.rst
rcap107: I was doing that, but then at some point for some reason it started breaking the rendering because the autoexamples were adding their own label, so there were two and the gallery was not rendering properly

rcap107:
> Also, if you want to work more on the added example, there is the option to move the fix of broken links to a separate PR so that this part can be merged sooner

I think I'll do that so I can also deal with #1793

doc/modules/data_ops/basics/building_data_ops_plan.rst
jeromedockes: the usual sphinx fun :)

rcap107:
Links will be fixed in #1798, the rest of the PR needs some work and will likely not be merged for a while

rcap107:
I think this PR can be closed in favor of #1880

PR #1789, opened 2025-12-04
Add add_tostr flag to Cleaner to control ToStr transformation

PilliSiddharth:
# Add `add_tostr` flag to `Cleaner` to control `ToStr` transformation

## Summary
This PR introduces an `add_tostr` parameter to the `Cleaner` transformer, allowing users to disable the automatic `ToStr()` conversion applied to non-numeric, non-categorical, and non-datetime columns. This prevents list/object-like columns from being converted to strings unless explicitly requested.

## What’s changed
- Add `add_tostr` argument to `Cleaner` (default: `True`).
- Update `_get_preprocessors` to include `ToStr()` only when `add_tostr=True`.
- Update `Cleaner` docstring (parameter documented; notes updated).
- Add `test_add_tostr` to validate both `add_tostr=True` and `add_tostr=False`.

## Motivation
`Cleaner` should not silently change complex dtypes (e.g., `list`, `object`) to strings. This flag gives users control and mirrors the existing pattern used for `numeric_dtype="float32"`.

## Test coverage
- New unit test `skrub/tests/test_cleaner.py::test_add_tostr` added.
- Local test run: relevant tests pass.

### DIFF
--- a/skrub/_table_vectorizer.py
+++ b/skrub/_table_vectorizer.py
@@ -189,16 +189,16 @@ class Cleaner(TransformerMixin, BaseEstimator):
         to ``np.float32`` dtype thanks to the transformer ``ToFloat``.
         If ``None``, numerical columns are not modified.
 
-    n_jobs : int, default=None
-        Number of jobs to run in parallel.
-        ``None`` means 1 unless in a joblib ``parallel_backend`` context.
-        ``-1`` means using all processors.
-
     cast_to_str : bool, default=False
         If ``True``, apply the ``ToStr`` transformer to non-numeric,
         non-categorical, and non-datetime columns, converting them to strings.
         If ``False``, this step is skipped and such columns retain their
         original dtype (e.g., lists, structs).
+
+    n_jobs : int, default=None
+        Number of jobs to run in parallel.
+        ``None`` means 1 unless in a joblib ``parallel_backend`` context.
+        ``-1`` means using all processors.
 
     Attributes
     ----------


CHANGES.rst
```diff
@@ -79,6 +79,9 @@ Bugfixes
   ``.skb.set_name()``, ``.skb.set_description()``, ``.skb.mark_as_X()`` or
   ``.skb.mark_as_y()`` used to raise an error, this has been fixed in :pr:`1782`
   by :user:`Jérôme Dockès <jeromedockes>`.
+- Added ``add_tostr`` parameter to :class:`Cleaner` to prevent unintended
```
rcap107: Thinking about it, even though I suggested it I don't like `add_tostr` as a parameter name for the Cleaner 🤔 we might have to change this before merging. 

A user doesn't know what `tostr` is (since it's a private transformer), so `add_tostr` does not give enough information.

skrub/_table_vectorizer.py
```diff
@@ -134,10 +135,11 @@ def _get_preprocessors(
...
         transformers.append(ToFloat())
-    transformers += [
-        CleanCategories(),
-        ToStr(),
-    ]
+
+    if add_tostr:
+        transformers.append(ToStr())
+
+    transformers.append(CleanCategories())
```
rcap107, line 139: The order of the transformers is important. `CleanCategories` should go before `ToStr`, so this part needs to be fixed.

skrub/_table_vectorizer.py
```diff
@@ -192,6 +194,13 @@ class Cleaner(TransformerMixin, BaseEstimator):
         ``None`` means 1 unless in a joblib ``parallel_backend`` context.
         ``-1`` means using all processors.
 
+    add_tostr : bool, default=True
+        If ``True``, apply the ``ToStr`` transformer to non-numeric,
+        non-categorical, and non-datetime columns, converting them to strings.
+        If ``False``, this step is skipped and such columns retain their
+        original dtype (e.g., lists, objects). This behaves similarly to how
+        ``numeric_dtype='float32'`` controls whether ``ToFloat`` is applied.
```
rcap107:
```suggestion
        If ``False``, this step is skipped and such columns retain their
        original dtype (e.g., lists, structs).
```

skrub/_table_vectorizer.py
```diff
@@ -243,8 +252,11 @@ class Cleaner(TransformerMixin, BaseEstimator):
     - ``CleanCategories()``: process categorical columns depending on the dataframe
       library (Pandas or Polars) to force consistent typing and avoid issues downstream.
 
-    - ``ToStr()``: convert columns to strings, unless they are numerical,
-      categorical, or datetime.
+    - ``ToStr()``: convert columns to strings unless they are numerical,
+    categorical, or datetime. This step is controlled by the ``add_tostr``
+    parameter. When ``add_tostr=True`` (default), string conversion is applied.
```
rcap107: as mentioned above, the name `add_tostr` will probably be changed

skrub/_table_vectorizer.py
```diff
@@ -313,13 +325,15 @@ def __init__(
         datetime_format=None,
         numeric_dtype=None,
         n_jobs=1,
+        add_tostr=True,
```
rcap107:
```suggestion
        add_tostr=False,
```

In the Cleaner, conversion to string should be off by default.

skrub/tests/test_cleaner.py
rcap107, line 1: Please move this test to `tests/test_table_vectorizer.py`, which is where the other tests for the Cleaner are.

At some point we might consider splitting the code, but that's out of the scope of this PR.

rcap107 (dismissed):
Thanks for the PR @PilliSiddharth. Overall, it looks good, but there are a few things that should be changed.

We should also rethink the name of the `add_tostr` parameter, but at the moment I don't have a very good alternative in mind.

PilliSiddharth:
Hey @rcap107, I moved the `test_add_tostr` test into `tests/test_table_vectorizer.py`, set the default to False, fixed the ordering so CleanCategories runs before ToStr, and fixed the docstring text. About the name, `add_tostr` works technically, but I agree it doesn’t feel very descriptive. Something like `convert_objects` might be clearer, but happy to go with whatever you prefer. Let me know if this looks right or if you want me to adjust anything else before I update the PR again.

rcap107:
Hi @PilliSiddharth, I think you forgot to push your latest commit. 

In the end, I'd like the parameter to be called `cast_to_str`. Also, the `test_add_tostr` should be modified slightly:
- Change the name to `cast_to_str`
- Test that the default behavior of the Cleaner does not change the type, and that setting `cast_to_str=True` does change the type as expected. In the current version, we're not testing the default behavior.

PilliSiddharth:
Hey @rcap107, I pushed the updates: renamed the parameter to cast_to_str, updated the test and docstring, and fixed the example output.

rcap107 (approved):

skrub/_table_vectorizer.py
```diff
@@ -313,13 +323,15 @@ def __init__(
...
         n_jobs=1,
+        cast_to_str=False,
     ):
         self.drop_null_fraction = drop_null_fraction
         self.drop_if_constant = drop_if_constant
         self.drop_if_unique = drop_if_unique
         self.datetime_format = datetime_format
         self.numeric_dtype = numeric_dtype
         self.n_jobs = n_jobs
+        self.cast_to_str = cast_to_str
```
rcap107, line 333:
```suggestion
        self.cast_to_str = cast_to_str
        self.n_jobs = n_jobs
```

skrub/_table_vectorizer.py
```diff
@@ -313,13 +323,15 @@ def __init__(
         datetime_format=None,
         numeric_dtype=None,
         n_jobs=1,
+        cast_to_str=False,
```
rcap107, line 325:
```suggestion
        cast_to_str=False,
        n_jobs=1,
```

rcap107:
Hi @PilliSiddharth, thanks a lot for the PR. I made a small change in the order of parameters, swapping `cast_to_str` and `n_jobs`. Aside from that, it looks good to me and I'll merge it once the CI is done. 

An aside, if you are having issues with doctests failing, it's often easier to use the ELLIPSIS (`...`) to skip parts of the test. In some cases the same test can fail with specific package/python versions because the string representations change from one version to another, and using `...` is the simplest way to get the tests to pass.

rcap107:
Whoops, the order change broke the CI, I'll fix it

PilliSiddharth:
Sounds good, glad I could be a part! 😄

PR #1795, opened 2025-12-07
fix docstring rendering

jeromedockes:
The docstring had a  section that is not one of the official numpy ones: "advanced examples". As a result, that part was being omitted from the rendered docs


rcap107:
wonderful

PR #1799, opened 2025-12-10
Added type-specific error message to cross_validate function

### LINKED ISSUE #1773: Improve error message when a data op is passed to `skrub.cross_validate` rather than a learner or search
While working with Data Ops I realized I have been making the same mistake relatively often: 

```python
pred = X.skb.apply(estimator, y=y)

skrub.cross_validate(pred, environment=pred.skb.get_data()
```

This gives the following error message:
```
RuntimeError: Evaluation of '.data_op' failed.
You can see the full traceback above. The error message was:
AttributeError: 'Series' object has no attribute 'data_op'
```

The error is caused by the fact that `cross_validate` is expecting either a learner, or a search created with `make_*_search`. The error message does not explain this at all.

The error message should be clarified so that, if the user passes a data op rather than a learner, a proper suggestion is provided (e.g., "you passed a data op, cross_validate requires a learner created with make_learner" etc).

emassoulie:
# Bug Fix Pull Request

## Description

Fixes #1773, giving a more verbose error message when the cross_validate function receives an incorrect input.

## How Has This Been Tested?

Running cross_validate with expected and unexpected types.

Run skrub.cross_validate with a data_op type as input rather than a learner.

### DIFF
--- a/skrub/_data_ops/tests/test_estimators.py
+++ b/skrub/_data_ops/tests/test_estimators.py
@@ -174,7 +174,11 @@ def test_cross_validate_type_error():
     learner = data_op.skb.make_learner()
     skrub.cross_validate(learner, data)
 
-    with pytest.raises(ValueError):
+    with pytest.raises(
+        ValueError,
+        match="`cross_validate` function requires either a Learner object or "
+        "a ParamSearch object, got *.",
+    ):
         skrub.cross_validate(data_op, data)
 
 


skrub/_data_ops/_estimator.py
```diff
@@ -646,6 +646,12 @@ def cross_validate(learner, environment, *, keep_subsampling=False, **kwargs):
     4    0.85
     Name: test_score, dtype: float64
     """
+    if not hasattr(learner, "__skrub_to_Xy_pipeline__"):
+        raise ValueError(
+            f"`cross_validate` function requires either a Learner object \
+                         or a ParamSearch object, got {type(learner)}."
+        )
```
jeromedockes, line 653:
```suggestion
        raise ValueError(
            "`cross_validate` function requires either a Learner object or "
            f"a ParamSearch object, got {type(learner)}."
        )
```

formatting nitpick :)

skrub/_data_ops/tests/test_estimators.py
```diff
@@ -167,6 +167,17 @@ def test_cross_validate(data_op, data, n_jobs):
...
 
 
+def test_cross_validate_type_error():
+    # Checks that cross_validate fails as expected when passed anything
+    # other than a Learner, ParamSearch or OptunaParamSearch object
+    data_op, data = get_data_op_and_data("simple")
+    learner = data_op.skb.make_learner()
+    skrub.cross_validate(learner, data)
+
+    with pytest.raises(ValueError):
```
jeromedockes: could you use `match` to check the error message is the one we expect here? thanks!

PR #1800, opened 2025-12-10
BUG - Fixing a few edge cases in the generation of the parallel coordinate plot

rcap107:
# Bug Fix Pull Request

This PR addresses a bug that may occur when using `plot_parallel_coord` if the results of the crossvalidation are not correct. 

## Description

The PR addresses the problem by adding a small eps and removing nan values from what is used in `_add_jitter`. 

## How Has This Been Tested?

A new test has been added to `test_parallel_coord.py` to check the values that are passed to `_add_jitter`

The bug first appeared in the CI testing the optuna example: 

https://app.circleci.com/pipelines/github/skrub-data/skrub/7618/workflows/eef20e21-776e-42d2-b24a-3031c0c05614/jobs/14823

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -58,6 +58,9 @@ Changes
 - Computing the associations in `TableReport` is now deterministic and can be controlled
   by the new parameter ``subsampling_seed`` of the global configuration.
   :pr:`1775` by :user:`Thomas S. <thomass-dev>`.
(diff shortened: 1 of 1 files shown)


rcap107:
It seems like the problem occurs if the column is integer and all values are np.nan

rcap107:
In the end, the problem was in the function `_prepare_numeric_column`. In some edge cases NaNs were not handled correctly, and now they are.

PR #1808, opened 2025-12-13
DOC : Clarified imputation scope in vision.rst

### LINKED ISSUE #1797: DOC - Add somewhere in the docs that imputation is outside skrub's scope
This is a possible question for an FAQ page (once we have one). 

> Is skrub going to implement transformers that do imputation? 

No, imputation is out of skrub's scope: scikit-learn implements transformers that perform imputation. Academic work has also shown that imputation is expensive and often does not improve prediction results https://arxiv.org/pdf/2407.19804

Something along these lines.

Soham-47:
Fixes #1797 
This PR updates the "Vision statement" in the documentation to clarify that missing value imputation is outside of skrub's scope.It explicitly recommends using scikit-learn's transformers (e.g., SimpleImputer, IterativeImputer) for this purpose. It also references recent academic work (https://arxiv.org/pdf/2407.19804) highlighting that imputation is often computationally expensive and may not improve predictive performance with modern models.

### DIFF
--- a/doc/vision.rst
+++ b/doc/vision.rst
@@ -34,11 +34,6 @@ To make things simpler, skrub uses defaults that are chosen empirically to
 give good machine learning, even though these are sometimes heuristic, as
 in the :class:`TableVectorizer`.
 
-Imputation is out of skrub's scope: scikit-learn implements transformers
-that perform imputation. Academic work has also shown that imputation is
-expensive and often does not improve prediction results
-(https://arxiv.org/pdf/2407.19804).
-
 
 Roadmap
 =======
@@ -61,3 +56,9 @@ From shorter term to longer term:
 
 - Automatic feature synthesis in databases, building on the assembling
   features
+
+
+Imputation is out of skrub's scope: scikit-learn implements transformers
+that perform imputation. Academic work has also shown that imputation is
+expensive and often does not improve prediction results
+(https://arxiv.org/pdf/2407.19804).


doc/vision.rst
```diff
@@ -34,6 +34,11 @@ To make things simpler, skrub uses defaults that are chosen empirically to
 give good machine learning, even though these are sometimes heuristic, as
 in the :class:`TableVectorizer`.
 
+Imputation is out of skrub's scope: scikit-learn implements transformers
```
rcap107: I think this should be moved at the end of the document, after the roadmap

rcap107:
Hi @Soham-47, thanks for checking out the issue. Usually it's better to discuss in the issue before opening a PR, but it's not a big deal. Specifically, in this case it might have been better to discuss where this information would go, but it's quite likely it would have ended where you put it anyway, so it's all good. 

As I commented, I'd rather move that paragraph to the end of the page, but aside from that I think it looks good.

Soham-47:
Hi @rcap107 , thanks for the feedback, and apologies for not discussing it in the issue first. I’ve updated the PR and moved the paragraph to the end of the page as suggested.

rcap107:
> Hi @rcap107 , thanks for the feedback, and apologies for not discussing it in the issue first. I’ve updated the PR and moved the paragraph to the end of the page as suggested.

No worries! It's just as information for the next time 👍

rcap107 (approved):

Issue #1787, opened 2025-12-04
BUG - Clarification of the behavior of DropCols


rcap107:
The behavior of `DropCols` is not consistent with what it says in the docstring.

The docstring states

> The other columns are kept in their original order. A ValueError is raised if any of the provided column names are not in the dataframe.

But the actual behavior is different: if the column to drop is not found in the dataframe, then no exception is raised. 

```python
>>> from skrub import DropCols
>>> import pandas as pd
>>> df = pd.DataFrame({"a": [1,2,3,4], "b":[3,4,5,6]})
>>> drop = DropCols("a")
>>> df_drop = df.drop(columns="a")
>>> df_drop
   b
0  3
1  4
2  5
3  6

>>> drop.fit_transform(df)
   b
0  3
1  4
2  5
3  6

>>> drop.transform(df_drop)
   b
0  3
1  4
2  5
3  6
```

Either the docstring must be fixed, or the behavior of the transformer must be adapted so that not finding the column name raises an exception. 
I think the current behavior of DropCols makes more sense, but I can also see some value in raising an error if the column is not found. Maybe detecting some problem upstream?

rcap107:
There are a few different approaches that we can take:

- We don't raise errors in either fit or transform if we try to drop a column that isn't found in the column names
- We try to enforce the same behavior in fit and transform, so that if a column is found at fit time, but isn't found at transform time, an exception is raise
- We raise an exception in fit if the column isn't found (this is closer to the default behavior of pandas drop https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.drop.html#pandas.DataFrame.drop) 

I think we should proceed with the first solution, as it is simpler to implement and explain in the docs. We might want to re-evaluate this in the future. 

As a result, the current issue becomes one of improving the docstring of `DropCols` so that the current behavior is explained properly.

wenebob:
Hello @rcap107 

Newbie here!

I ran the exemplary code above, and with a  non-existent column `d` 
As expected, both `drop.fit(df)` and `drop.fit_transform(df)` threw the same error: `ValueError: The following columns are requested for selection but missing from dataframe: ['d']`

On the other hand, the `transform` method does not throw an error if the column is not found.  My question is, is it a bad idea given that either `fit` or `fit_transform` will be called first?

Also, will it suffice to specify in the documentation that the `transform` method drops the columns only if they exist, and otherwise ignores them?

rcap107:
[...]

PR #1813, opened 2025-12-17
FIX - Make `SelectCols` and `DropCols` implement sklearn convention

auguste-probabl:
For `check_is_fitted` to work correctly when nested inside Pipelines.

Addresses https://github.com/skrub-data/skrub/issues/1787#issuecomment-3664369896 and https://github.com/scikit-learn/scikit-learn/issues/32910

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -17,6 +17,10 @@ Changes
 
 Bugfixes
 --------
+- :class:`DropCols` and :class:`SelectCols:` attributes were renamed to end
+  with an underscore, in order to follow a scikit-learn convention which is
+  used to determine if an estimator is fitted. :pr:`1813` by :user:`Auguste
+  Baum <auguste-probabl>`.
 
 Release 0.7.0
 =============


GaelVaroquaux:
I'll let @rcap107 merge this one, but I think that this is the right fix.

Maybe a changelog entry, so that if people hit the problem, they know it's fixed by updating skrub?

rcap107:
> I'll let @rcap107 merge this one, but I think that this is the right fix.
> 
> Maybe a changelog entry, so that if people hit the problem, they know it's fixed by updating skrub?

@jeromedockes commented in https://github.com/skrub-data/skrub/issues/1787#issuecomment-3665732607 that we should decide on the interface before merging

Personally I'm fine with making both select_cols_ and kept_cols public, because both give useful information to the user, especially when complex selectors are used, so the current version works for me.

GaelVaroquaux:
> Personally I'm fine with making both select_cols_ and kept_cols public, because both give useful information to the user, especially when complex selectors are used, so the current version works for me.

I agree

rcap107:
Looks good to me, thanks a lot @auguste-probabl

Issue #1779, opened 2025-11-28
DOC - improve the documentation of `SkrubLearner.report()`


rcap107:
The current docstring of `SkrubLearner.report()` is very short and not very informative, and should be expanded to describe how to use the function, including some examples. There should also be an entry in the user guide that explains the use case and expands on that.

wenebob:
Can I work on this?

rcap107:
Hi @Faith-Nchifor, absolutely.

What I would like from this issue is a better docstring for the learner `report`, which looks like this at the moment:
https://skrub-data.org/dev/reference/generated/skrub.SkrubLearner.html#skrub.SkrubLearner.report

The idea of the `report` is that it gives a precise idea of how much time it takes to fit or predict with the learner using the environment that is being provided. So, it should explain that the environment is the dictionary of variables with the values specified by the user, and that the "mode" can be "fit", "predict", "transform". Also, that the timings shown in the report are relative to the given environment and mode. There should also be a small mention that `report` can take the same parameters that can be passed to `.skb.full_report`.
This can be added to the docstring of the learner, which is in `skrub/_data_ops/_estimator.py`.

This feature should also be explained in the user guide, with a short paragraph that explains the reason, how to specify the environment, and that the learner is going to report the timings for the specified mode.

Note that the environment can be extracted from a data op by using `.skb.get_data`, which returns a dictionary that contains the values provided as preview.

rcap107:
Closed by #1815

PR #1815, opened 2025-12-17
An improvement to SkrubLearner.report() method

wenebob:
As mentioned in issue #1779 
I added an example and use cases.

### DIFF
--- a/skrub/_data_ops/_estimator.py
+++ b/skrub/_data_ops/_estimator.py
@@ -147,91 +147,38 @@ class SkrubLearner(_CloudPickleDataOp, BaseEstimator):
 
         Examples
         --------
-        >>> from sklearn.decomposition import PCA
-        >>> from sklearn.dummy import DummyClassifier
+        We start by creating the learner for a simple DataOp:
+
         >>> import skrub
-        >>> from skrub import selectors as s
-
-        >>> orders = skrub.datasets.toy_orders()
-        >>> X, y = skrub.X(), skrub.y()
-        >>> pred = (
-        ...     X.skb.apply(skrub.StringEncoder(n_components=2), cols=["product"])
-        ...     .skb.set_name("product_encoder")
-        ...     .skb.apply(skrub.ToDatetime(), cols = ["date"])
-        ...     .skb.apply(
-        ...            skrub.DatetimeEncoder(add_total_seconds = False),
-        ...            cols=["date"]
-        ...     )
-        ...     .skb.apply(PCA(n_components=2), cols=s.glob("date_*"))
-        ...     .skb.set_name("pca")
-        ...     .skb.apply(DummyClassifier(), y=y)
-        ...     .skb.set_name("classifier")
-        ... )
+        >>> from sklearn.linear_model import LogisticRegression
+        >>> from sklearn.datasets import make_classification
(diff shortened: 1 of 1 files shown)


wenebob:
Hey @rcap107 
Some of my checks are failing here. Can you guide me on what to do?

skrub/_data_ops/_estimator.py
```diff
@@ -144,6 +144,94 @@ def report(self, *, environment, mode, **full_report_kwargs):
...
+        ...         'y' : orders.y
+        ...     }
+        ... )
+        # doctest: +ELLIPSIS
+        {'result': array([[0.75, 0.25],
+            [0.75, 0.25],
+            [0.75, 0.25],
+            [0.75, 0.25]]),
+        'error': None,
+        'report_path': PosixPath('/path/to/reports/index.html')}
```
jeromedockes: Thank you very much for this work @Faith-Nchifor (and the other PR too) and welcome to skrub! One issue is not due to your example but rather to the fact that we execute those doctests in all environments, including some where graphviz (which is needed for creating the reports) may not be installed. So we will need to skip those tests for now. This is done with the `# doctest: +SKIP` directive. `ELLIPSIS` is to allow replacing parts of the output with `...`, but this is already turned on for the whole repo in the configuration in pyproject.toml

Also, we usually try to keep doctests very short and simplisting and illustrating only features of the function we are documenting. More interesting and realistic examples are shown in the example gallery. So here we can shorten it a bit by using a simpler pipeline, as the reporting works the same for simple or complex dataops. Therefore I would suggest something along those lines:

```suggestion
        We start by creating the learner for a simple DataOp:

        >>> import skrub
        >>> from sklearn.linear_model import LogisticRegression
        >>> from sklearn.datasets import make_classification
        >>> pred = skrub.X().skb.apply(LogisticRegression(), y=skrub.y())
        >>> X, y = make_classification(n_samples=20, random_state=0)
        >>> split = pred.skb.train_test_split({'X': X, 'y': y}, shuffle=False)
        >>> learner = pred.skb.make_learner()

        We can now obtain reports for the different methods of the learner such
        as 'fit', 'predict_proba', etc.

        >>> fit_results = learner.report(
        ...     environment=split["train"], mode="fit", open=False
        ... )  # doctest: +SKIP
        >>> fit_results['report_path']  # doctest: +SKIP
        PosixPath('.../skrub_data/execution_reports/full_data_op_report_.../index.html')

        Note that our learner has been fitted now, we can use it for predictions.

        >>> predict_results = learner.report(
        ...     environment=split["train"], mode="predict", open=False
        ... )  # doctest: +SKIP
        >>> predict_results['report_path']  # doctest: +SKIP
        PosixPath('.../skrub_data/execution_reports/full_data_op_report_.../index.html')

        In addition to the report, we can also retrieve the actual output of
        the 'predict' method:

        >>> predict_results['result']  # doctest: +SKIP
        array([0, 1, 0, 1, 1, 1, 0, 0, 0, 1, 1, 1, 0, 1, 0])
```

the doctest suggested above passes even if the +SKIP are removed, when graphviz is installed, and due to the +SKIP it should not fail the CI (I hope 😅 )

wenebob:
Hello @jeromedockes,
Thank you for this explanation and correction. Your example was much clearer.

jeromedockes:
Great, thanks again for this contribution @Faith-Nchifor ! 🎉

PR #1819, opened 2025-12-19
Expose "vocabulary" parameter to "StringEncoder"

emassoulie:
Response to issue 1792

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,9 +14,9 @@ New features
 
 Changes
 -------
-- The :class: `StringEncoder` now exposes the ``vocabulary`` parameter from the parent
-  :class: `TfidfVectorizer`.
-  :pr: `1819` by :user: `Eloi Massoulié <emassoulie>`
+- The :class:`StringEncoder` now exposes the ``vocabulary`` parameter from the parent
+  :class:`TfidfVectorizer`.
+  :pr:`1819` by :user:`Eloi Massoulié <emassoulie>`
 
 
 Bugfixes


MarieSacksick:
hello @emassoulie, thank you for your contribution :tada: !  
Could you add a test to this new option please?   

I think it's also worth to add an entry in the changelog :).

skrub/_string_encoder.py
```diff
@@ -61,6 +61,11 @@ class StringEncoder(TransformerMixin, SingleColumnTransformer):
         Used during randomized svd. Pass an int for reproducible results across
         multiple function calls.
 
+    vocabulary : Mapping or iterable, default=None
```
GaelVaroquaux, line 64:
```suggestion
    vocabulary_ : Mapping or iterable, default=None
```
The scikit-learn convention requires to have an underscore at the end of attributes that are derived from the data

GaelVaroquaux: You'll need to propagate this change :)

emassoulie:
`vocabulary` name reverted following a discussion with @rcap107!
We were also wondering whether it is  necessary to include an example in the docstring for this specific case. It feels like any use case for "vocabulary" would just be passing on another model's vocabulary attribute to the StringEncoder, which doesn't make for particularly compelling code

skrub/_string_encoder.py
rcap107: Here `vocabulary` is a user-provided value though, it's not derived from the data. In fact, we never define a `vocabulary_` in `fit_transform`.

CHANGES.rst
```diff
@@ -87,6 +87,9 @@ Changes
 - Computing the associations in :class:`TableReport` is now deterministic and can
   be controlled by the new parameter ``subsampling_seed`` of the global configuration.
   :pr:`1775` by :user:`Thomas S. <thomass-dev>`.
+- The :class: `StringEncoder` now exposes the ``vocabulary`` parameter from the parent
```
rcap107: The changelog entry was added in the wrong place, it should be at the top of the file in the proper section

skrub/_string_encoder.py
```diff
@@ -165,21 +172,28 @@ def fit_transform(self, X, y=None):
...
         elif self.vectorizer == "hashing":
-            self.vectorizer_ = Pipeline(
-                [
-                    (
-                        "hashing",
-                        HashingVectorizer(
-                            ngram_range=self.ngram_range,
-                            analyzer=self.analyzer,
-                            stop_words=self.stop_words,
+            if self.vocabulary is None:
```
rcap107: I think that having something like 
```python
            if self.vocabulary is not None:
                raise ValueError(...)
```
and then continuing with `self.vectorizer_` would be more readable

Issue #1730, opened 2025-11-04
exposing SingleColumnTransformer and RejectColumn


jeromedockes:
### Problem Description

the ApplyToCols has been made public. it becomes even more useful if user-defined transformers can make full use of its functionality, by handling a single Series (makes code easier to write) and by possibly rejecting columns with a RejectColumn exception.

For example suppose I want to write a transformer that parses zip codes and converts to a latitude and longitude. As it processes a single column at a time it's a bit simpler if I can inherit from SingleColumnTransformer and deal with Series rather than DataFrames in fit() and transform(). Also it is useful to just raise a RejectColumn exception to signal that the column does not contain zip codes after all, in which case the ApplyToCols will leave it unchanged. Auto-detecting the zip code columns before transforming them would be much more annoying to do otherwise, and would basically force me to write a full transformer (meaning I wouldn't be able to use ApplyToCols for this)

### Feature Description

SingleColumnTransformer and RejectColumn would be added to the public API, possibly with improvements or additions to the documentation & examples, without any change to the functionality

Both objects should be added to a new `skrub.core` namespace, which includes more "dev-oriented" objects that should not clutter the "general use case" namespace.

rcap107:
I agree that this would be useful, and I think it would also simplify docstrings in other places, letting us refer to the public docstring of SingleColumnTransformer and RejectColumn directly. 

I also think this would be targeted mostly at advanced users, and it should be reflected in the documentation. 

In any case, this should be done after #1729. Both issues may be addressed at the same time, but given that this issue involves some work on the documentation, it might be better to split them.

GaelVaroquaux:
I'm fine with the suggestion of adding SingleColumnTransformer and RejectColumn to the public API.

My question would however be: should they be exposed in skrub's namespace, or in another namespace, such as "skrub.api"?
I'm asking to avoid polluting the tab-completion on the main namespace with things that are not useful aside for advanced users.

rcap107:
> My question would however be: should they be exposed in skrub's namespace, or in another namespace, such as "skrub.api"? I'm asking to avoid polluting the tab-completion on the main namespace with things that are not useful aside for advanced users.

I think that makes sense. I think the use case is advanced enough that people that are interested in it shouldn't have issues with one additional level of imports.

However, my follow-up question to this is: should we reconsider importing everything at once in skrub, and maybe split some of the lower level objects into specific namespaces? Like "joining", "encoding" etc

[...]

PR #1820, opened 2025-12-21
chore: Refactor SingleColumnTransformer so it's in its own file

### LINKED ISSUE #1729: CHORE - Refactor SingleColumnTransformer so it's in its own file
This is a minor fix, but I think it would be helpful for the readability of the code.

At the moment, the `SingleColumnTransformer` is in the file `_apply_to_cols.py`, with additional code for the docs and the `RejectColumn` exception. I think this has historical reasons, going back to when `_on_each_column` was private.

However now that `ApplyToCols` is out I think we should separate the functions so that `SingleColumnTransformer` is in its own file.

techy4shri:
Closes #1729 
- Refactored SingleColumnTransformer in a separate file
- Updated imports everywhere accordingly
- Updated Doctests to remove the documentation example errors

Now the new file has separate transformer classes, the tests all but pass. I refrained from making a new test file since `test_on_each_column.py` has both `SingleColumnTransformer`  and `RejectColumn` covered in it.
Let me know if any changes are required.
Note: The current implementation is such that both SingleColumnTransformer and RejectColumn are added as an internal libs rather than being part of the public API. I read the issue and the long discussion about making them both public in [this issue](#1730 ) discussion but it seems to me that the internal team would handle it later, after the issue I am assigned to will be closed. Just putting it here :)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,13 @@ New features
 
 Changes
 -------
+- The :class:`StringEncoder` now exposes the ``vocabulary`` parameter from the parent
+  :class:`TfidfVectorizer`.
+  :pr:`1819` by :user:`Eloi Massoulié <emassoulie>`
+
+
+- :func:`compute_ngram_distance` has been renamed to :func:`_compute_ngram_distance` and is now a private function.
+  :pr:`1838` by :user:`Siddharth Baleja <siddharthbaleja>`.
 
 Bugfixes
 --------
--- a/examples/0050_deduplication.py
+++ b/examples/0050_deduplication.py
@@ -148,33 +148,6 @@ translation_table.head()
 # If we want to adapt the translation table, we can
 # modify it manually.
 
-###############################################################################
-# Visualizing string pair-wise distance between names
-# ---------------------------------------------------
-#
-# Below, we use a heatmap to visualize the pairwise-distance between medication
-# names. A darker color means that two medication names are closer together
-# (i.e. more similar), a lighter color means a larger distance.
-#
-
-from scipy.spatial.distance import squareform
-
-from skrub import compute_ngram_distance
-
-ngram_distances = compute_ngram_distance(unique_examples)
-square_distances = squareform(ngram_distances)
-
-import seaborn as sns
-
-fig, ax = plt.subplots(figsize=(14, 12))
-sns.heatmap(
-    square_distances, yticklabels=unique_examples, xticklabels=unique_examples, ax=ax
-)
-plt.show()
-
-###############################################################################
-# We have three clusters appearing - the original medication
-# names and their misspellings that form a cluster around them.
 
 ###############################################################################
 # Conclusion
--- a/skrub/__init__.py
+++ b/skrub/__init__.py
@@ -29,7 +29,7 @@ from ._data_ops import (
     y,
 )
 from ._datetime_encoder import DatetimeEncoder
-from ._deduplicate import compute_ngram_distance, deduplicate
+from ._deduplicate import deduplicate
 from ._drop_uninformative import DropUninformative
 from ._fuzzy_join import fuzzy_join
 from ._gap_encoder import GapEncoder
@@ -77,7 +77,7 @@ __all__ = [
     "Cleaner",
     "DropUninformative",
     "deduplicate",
-    "compute_ngram_distance",
+    "deduplicate",
     "ToCategorical",
     "to_datetime",
     "AggJoiner",
--- a/skrub/_deduplicate.py
+++ b/skrub/_deduplicate.py
@@ -12,7 +12,7 @@ from sklearn.feature_extraction.text import TfidfVectorizer
 from sklearn.metrics import silhouette_score
 
 
-def compute_ngram_distance(
+def _compute_ngram_distance(
     unique_words,
     ngram_range=(2, 4),
     analyzer="char_wb",
@@ -260,7 +260,7 @@ def deduplicate(
     9  white      9              white
     """
     unique_words, counts = np.unique(X, return_counts=True)
-    distance_mat = compute_ngram_distance(
+    distance_mat = _compute_ngram_distance(
         unique_words, ngram_range=ngram_range, analyzer=analyzer
     )
 
--- a/skrub/_string_encoder.py
+++ b/skrub/_string_encoder.py
@@ -61,6 +61,11 @@ class StringEncoder(TransformerMixin, SingleColumnTransformer):
         Used during randomized svd. Pass an int for reproducible results across
         multiple function calls.
 
+    vocabulary : Mapping or iterable, default=None
+        In case of "tfidf" vectorizer, the vocabulary mapping passed to the vectorizer.
+        Either a Mapping (e.g., a dict) where keys are terms and values are
+        indices in the feature matrix, or an iterable over terms.
+
     Attributes
     ----------
     input_name_ : str
@@ -131,6 +136,7 @@ class StringEncoder(TransformerMixin, SingleColumnTransformer):
         analyzer="char_wb",
         stop_words=None,
         random_state=None,
+        vocabulary=None,
     ):
         self.n_components = n_components
         self.vectorizer = vectorizer
@@ -138,6 +144,7 @@ class StringEncoder(TransformerMixin, SingleColumnTransformer):
         self.analyzer = analyzer
         self.stop_words = stop_words
         self.random_state = random_state
+        self.vocabulary = vocabulary
 
     def fit_transform(self, X, y=None):
         """Fit the encoder and transform a column.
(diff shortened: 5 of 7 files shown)


rcap107:
Hi @techy4shri, thanks a lot for this PR! It's already in a good shape.

The tests are failing because doctests are checking that the examples in the documentations are passing. Doctests are failing in some places because they are expecting the RejectColumn exception to still be in skrub._apply_to_cols, while now it has been moved to the new file. Those doctests should therefore be updated with the new path to not fail:
```python
    skrub._apply_to_cols.RejectColumn: 
```
should become
```python
    skrub_single_column_transformer.RejectColumn: 
```

Also, could you move the tests on the single_column_transformer to their own separate test file?

techy4shri:
Understood @rcap107 ! Will do! Thanks for the confirmation.

techy4shri:
@rcap107 the PR is ready for review, kindly let me know if any changes are required.

techy4shri:
okk thanks! I shall be waiting for the final judgement 🫡 @jeromedockes do let me know your thoughts too 👀

skrub/_apply_to_cols.py
```diff
@@ -368,10 +166,10 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
     >>> ToDatetime().fit_transform(df["birthday"])
     0   2024-01-29
     Name: birthday, dtype: datetime64[...]
-    >>> ToDatetime().fit_transform(df["city"])
+    >>> ToDatetime().fit_transform(df["city"])  # doctest: +ELLIPSIS
```
```suggestion
    >>> ToDatetime().fit_transform(df["city"])
```

we have ELLIPSIS in the  doctest_optionflags in pyproject.toml so in theory it shouldn't be needed here; did you add it because the test was failing when you ran it? In my environment it seems to run without

jeromedockes: (and the same question applies to other places where ELLIPSIS was added)

skrub/_clean_null_strings.py
```diff
@@ -159,10 +159,11 @@ class CleanNullStrings(SingleColumnTransformer):
     ``ToFloat`` (only) after being cleaned by ``CleanNullStrings``:
 
     >>> from skrub._to_float import ToFloat
-    >>> ToFloat().fit_transform(s)
+    >>> ToFloat().fit_transform(s)  # doctest: +ELLIPSIS
     Traceback (most recent call last):
         ...
-    skrub._apply_to_cols.RejectColumn: Could not convert column 's' to numbers.
+    skrub._single_column_transformer.RejectColumn:
```
jeromedockes: I think we should not wrap this line but keep it identical to the actual output (no line break before 'Could not convert' ...)

skrub/_single_column_transformer.py
```diff
@@ -0,0 +1,215 @@
...
+    "   ApplyToCol({class_name}())\n\n"
+    "To apply to selected columns::\n\n"
+    "   ApplyToCols({class_name}(), cols=['col_name_1', 'col_name_2'])"
+)
+_SINGLE_COL_PARAGRAPH = textwrap.fill(
+    _SINGLE_COL_LINE, initial_indent="    ", subsequent_indent="    "
+)
+_SINGLE_COL_NOTE = f".. note::\n\n{_SINGLE_COL_PARAGRAPH}\n"
+
+__all__ = ["SingleColumnTransformer", "RejectColumn"]
```
jeromedockes: I think it's good to keep `__all__` right after the imports because it tells us everything the module exposes

skrub/_single_column_transformer.py
```diff
@@ -0,0 +1,215 @@
...
+from sklearn.utils.validation import check_is_fitted
+
+from . import _dataframe as sbd
+
+_SINGLE_COL_LINE = (
+    "``{class_name}`` is a type of single-column transformer. Unlike most scikit-learn"
+    " estimators, its ``fit``, ``transform`` and ``fit_transform`` methods expect a"
+    " single column (a pandas or polars Series) rather than a full dataframe. To apply"
+    " this transformer to one or more columns in a dataframe, use it as a parameter in"
+    " a ``skrub.ApplyToCols`` or a ``skrub.TableVectorizer``."
```
jeromedockes, line 19: @rcap107  #1826 will need updating after merging this PR

skrub/_single_column_transformer.py
```diff
@@ -0,0 +1,215 @@
+"""
+A single place for all the single column transformers.
```
jeromedockes:
```suggestion
Base class for single-column transformers.
```

or any equivalent wording you prefer; I just mean this module doesn't contain all the single column transformers it only provides a base class for them

skrub/tests/test_on_each_column.py
```diff
@@ -9,87 +9,19 @@
 from sklearn.preprocessing import OneHotEncoder
 
 from skrub import _dataframe as sbd
```
jeromedockes, line 11: (not in this PR) this file should be renamed `test_apply_to_cols`

skrub/tests/test_single_column_transformer.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def transform(self, X):
+        assert sbd.is_column(X)
+        X = sbd.make_dataframe_like(X, [X])
+        return super().transform(X)
+
+
+@pytest.mark.parametrize("output_kind", ["single_column", "dataframe", "column_list"])
+@pytest.mark.parametrize("transformer_class", [Mult, SingleColMult])
+def test_single_column_transformer(
```
jeromedockes: I think despite the name this is testing the handling of ApplyToCols of the transformer's different outputs so it should probably stay in the original module, and this will avoid duplicating the code of the Mult dummy class above

skrub/tests/test_single_column_transformer.py
```diff
@@ -0,0 +1,219 @@
...
+    }
+    if output_kind == "single_column":
+        expected_data.pop("a 0 * 2.0")
+        expected_data.pop("a 1 * 2.0")
+    expected = df_module.make_dataframe(expected_data)
+    df_module.assert_frame_equal(out, expected)
+    assert mapper.get_feature_names_out() == list(expected_data.keys())
+
+
+class Rejector(SingleColumnTransformer):
```
jeromedockes: same for all the code below: they are tests for ApplyToCols

skrub/tests/test_single_column_transformer.py
```diff
@@ -0,0 +1,219 @@
...
+
+def test_rejection_forbidden_in_transform(df_module):
+    df = df_module.example_dataframe
+    mapper = ApplyToCols(RejectInTransform(), allow_reject=True)
+    mapper.fit(df)
+    with pytest.raises(ValueError, match=".*failed on.*int-col"):
+        mapper.transform(df)
+
+
+class RenameB(SingleColumnTransformer):
```
jeromedockes: RenameB is unused here (but it is in test_on_each_column, where it is also found)

skrub/_apply_to_cols.py
techy4shri: Yeah, when I first wrote the tests and and ran pytest (and also after simply setting up and no code change), I didn't think it would be needed but the tests kept failing on my side so I added this.  Let me update my branch locally once and run the tests again, hopefully, no errors this time.

skrub/_clean_null_strings.py
techy4shri: That would be the formatter's doing, let me fix that.

skrub/tests/test_on_each_column.py
techy4shri: did think it was a bit confusing, happy to know this is supposed to be that 😭

skrub/tests/test_single_column_transformer.py
techy4shri: 🤔 From what I gathered, the `_single_column_transformer` is ultimately being used with `apply_to_cols` in some cases , but I will revert the changes, thanks!

skrub/_clean_null_strings.py
techy4shri: yup, still getting that when i try to make commits:
```
E501 Line too long (91 > 88)
   --> skrub\_clean_null_strings.py:165:89
    |
163 |     Traceback (most recent call last):
164 |         ...
165 |     skrub._single_column_transformer.RejectColumn: Could not convert column 's' to numbers.
    |                                                                                         ^^^
166 |     >>> ToFloat().fit_transform(cleaner.fit_transform(s))
167 |     0    1.1
    |
```

techy4shri: and the pre-commit check fails :)

skrub/_apply_to_cols.py
techy4shri: ok it passes this time, so removing that.

techy4shri:
Pre-commit check will fail now because of line length.

skrub/_single_column_transformer.py
josselinonduty: Once this PR is accepted, I can merge main into my branch and update #1826 iyw

rcap107: Hi @josselinonduty, yes please do, thanks.

skrub/_clean_null_strings.py
rcap107: @techy4shri if you add `# noqa: 501` pre-commit checks should let the line through

rcap107:
Hi @techy4shri, thanks for the PR, and I'm sorry for the spotty support these past weeks. The PR should be ready to merge once the precommit checks pass, which should be easy to fix with `# noqa: 501`.

techy4shri:
@rcap107 @jeromedockes all checks are passing now!

rcap107:
I saw, thanks a lot @techy4shri. The PR looks good to me, I'm waiting for @jeromedockes to check if his comments have been addressed before merging.

rcap107:
It's been a couple of days and this PR is currently blocking a few others, so I'll merge it. If needed, we can add minor changes in a follow-up.

Thanks a lot @techy4shri for the work and @jeromedockes for the review.

Issue #1761, opened 2025-11-19
DOC - Improve the documentation of `set_config`


rcap107:
The current docstring of `set_config` contains mistakes, and shows the default value to be `None`, without explaining that this means relying on the default value (which is `False`).

As for mistakes, `use_table_report` shows the default value to be `True`, while `use_table_report_data_ops` is shown as `False`, while it's the other way around. 

I think that there should also be a line saying that if the value is `None` (the actual default), then the default configuration is used. Something like

```python
    use_table_report : bool, default=None
        The type of display used for dataframes. The default configuration is ``False``.

        - If ``True``, replace the default DataFrame HTML displays with
          :class:`~skrub.TableReport`.
        - If ``False``, the original Pandas or Polars dataframe HTML representation
          will be used.
		- If ``None``, the current configuration is used. 

        This configuration can also be set with the ``SKB_USE_TABLE_REPORT``
        environment variable.
```

PR #1822, opened 2025-12-25
doc(set_config): fix the doc string of set_config

MarieSacksick:
Closes https://github.com/skrub-data/skrub/issues/1761

### DIFF
--- a/skrub/_config.py
+++ b/skrub/_config.py
@@ -98,7 +98,7 @@ def set_config(
     Parameters
     ----------
     use_table_report : bool, default=None
-        The type of display used for dataframes. If ``None``, fallbacks to the current
+        The type of display used for dataframes. If ``None``, falls back to the current
         configuration, which is ``False`` by default.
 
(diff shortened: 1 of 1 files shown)


skrub/_config.py
```diff
@@ -98,7 +98,8 @@ def set_config(
     Parameters
     ----------
     use_table_report : bool, default=None
-        The type of display used for dataframes. Default is ``True``.
```
MarieSacksick, line 101: I dislike this sentence, because I find it it's very confusing with the line above which says that the default is None. I proposed another formulation below.

skrub/_config.py
```diff
@@ -98,7 +98,8 @@ def set_config(
...
+        The type of display used for dataframes. If ``None``, fallbacks to the current
```
jeromedockes: maybe "falls back", or "defaults"?

jeromedockes (approved):

PR #1824, opened 2025-12-29
FEAT - Add a search input to report w/ regex toggle

### LINKED ISSUE #1625: TableReport: Add a filter by column name
It should be possible to filter columns by a given column name in the TableReport. 

I am not sure if this should be a replacement of the current filters (which are fairly limited as is, but maybe users need them), or if it should be one additional field somewhere.

josselinonduty:
Fixes #1625

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ Ongoing Development
 
 New features
 ------------
+- You can now filter columns in the TableReport using a search box. It supports plain text and
+  regular expressions. :pr:`1824` by :user:`Josselin Dulongcourty <josselinonduty>`.
 
 Changes
 -------


josselinonduty:
As discussed in https://github.com/skrub-data/skrub/issues/1625#issuecomment-3546184405, this feature is not necessary anymore.

I still wanted to share what I had done, even if it is discarded right away.

> [!NOTE]
> Reopening and closing allows me to update the PR (for CI fixes) while keeping it closed, to avoid work overload for maintainers. It doesn't look nice but it works

---

Architecture
---

I have implemented a Model (as in Model View Controller architecture). That way, it is easier to implement new controls, inputs, filters, etc. since data is centralised in a single entity.

Search input
---

Thanks to this model, I was easily able to add a search text input to allow the user to filter through the columns using substrings. At the moment, it filters columns based on if their name contains the query as a single substring. ([code](https://github.com/skrub-data/skrub/pull/1824/changes#diff-25d701a8fddf1b4bc3a0c391b80f5913dfb9e4da250db9d7d002994ad700dbe6R233))

Examples:
Query: `time`
Does match: `time_day`, `month_time`, `time`, `ANYTIME`
Does not match: `ti_me`, `ti me`

Regex
---

This search input also supports regex queries thanks to a toggle button.
If the regex query is invalid, no columns are displayed. *This could be improved by displaying a warning*.

How the UI looks:
image

Examples:
Query: `^(d|e)` (*meaning: any string that starts with either d or e*)
Does match: `department`, `department_any`, `employment`...
Does not match: `x_department`, `_employer`...

Tests
---
I have added a whole bunch of e2e tests to ensure my work is tested and maintainable.
My work has some code comments, where I felt it was needed.

Miscellaneous changes
---
I have appended prefixes/suffixes to variables and classes that had ambiguous names.
I have updated the "No columns found" text/view to display this newly added search query as well.
Finally, I had to make minor changes to the CSS.

Transparency
---
**AI assistance** (github copilot in my case) was used to try not to forget some tests, and to identify where comments would be useful and had to be left. **Nothing AI-written was committed without review and proper testing** (AI struggles with e2e testing anyway). It also gave me some thoughts about the design pattern I had implemented; for instance, [this part](https://github.com/skrub-data/skrub/pull/1824/changes#diff-25d701a8fddf1b4bc3a0c391b80f5913dfb9e4da250db9d7d002994ad700dbe6R106-R119) is the result of an idea it gave me.

PR #1825, opened 2025-12-30
DOC - improve documentation for Drop #1756

### LINKED ISSUE #1756: DOC - improve the documentation of the Drop transformer
Improve the documentation of the `Drop` transformer by adding a docstring and an entry and example in the user guide. 

> > For some reason, Drop (not DropCols) is a public class, when it clearly shouldn't be. This should be fixed.
> 
> they are both public, and both are useful. Drop is the single-column transformer that just drops the column it is applied to. It is public because it is actually used in public attributes of the TableVectorizer when the user passes "drop" as one of the encoders. But it also becomes useful in its own right now that we have the ApplyToCols and the data ops -- for example `.skb.apply(skrub.choose_from([StringEncoder(), Drop()]))`
> 
> If that is the intended way of using `Drop`, then it makes sense to keep it, but if that's the case it needs a proper docstring because at the moment it's not even in the API docs.  

 _Originally posted by @rcap107 in [#1730](https://github.com/skrub-data/skrub/issues/1730#issuecomment-3538435128)_

wikip50700:
This PR improves the documentation for the ```Drop``` class by adding docstrings and a user guide section with an example showing how to use ```Drop``` with ```ApplyToCols```.


skrub/_select_cols.py
```diff
@@ -197,9 +197,70 @@ def get_feature_names_out(self, input_features=None):
 
 
 class Drop(SingleColumnTransformer):
+    """Drop a single column in a column-wise transformation pipeline.
+
+    ``Drop`` is a single-column transformer that removes the column it is
+    applied to and produces no output features. Unlike :class:`DropCols`,
+    which operates on a full DataFrame, ``Drop`` is designed to be used in
+    column-wise contexts such as :class:`ApplyToCols`, data ops, or internally
```
rcap107, line 205:
```suggestion
    column-wise contexts such as :class:`ApplyToCols`, Data Ops, or internally
```

rcap107:
Hi @wikip50700, thanks for this PR. I think it's already in a good shape, however before merging I would like to add also an example with the Data Ops similar to what is written in the original issue #1756:

`.skb.apply(skrub.choose_from([StringEncoder(), Drop()]))`

The reasoning for the `choose_from` there is that it's not clear whether the given column is useful or not, so it would be good to find that out using hyperparmeter tuning. The example should be in `/doc/modules/data_ops/validation/hyperparameter_tuning.rst`, after the section on feature selection.

rcap107:
Hi @wikip50700, are you still working on this?

emassoulie:
I agree with @rcap107, the docstrings are sound with simple enough examples, but this operation would benefit from a concrete, complex use case. I can try and get something together based on the issue you mentioned!

rcap107:
Replaced by #2104

PR #1834, opened 2026-01-12
Add simple DatetimeEncoder example with periodic encoding (#1629)

### LINKED ISSUE #1629: DOC - Write a new example with the DatetimeEncoder
The current DatetimeEncoder example has a number of issues that have also already been pointed out in issues #1394 and #1282

The example itself is not very convincing, so we should come up with a better example (possibly with a different dataset) and use that instead.

If finding a suitable dataset is problematic, we can also have a simple example that is just showcasing what the DatetimeEncoder is doing (similar to the SquashingScaler example), and then replace it later with something else.

TaobaOuathrani:
This PR adds a self-contained example showcasing how
DatetimeEncoder extracts meaningful time-based features in a
regression setting.

- No forecasting, purely illustrative
- Shows periodic encoding for linear models
- Uses a minimal synthetic dataset
- Addresses issues  #1629

Closes #1629

### DIFF
--- a/examples/0030_datetime_encoder.py
+++ b/examples/0030_datetime_encoder.py
@@ -352,3 +352,78 @@ plt.show()
 # features from a datetime column.
 # Also check out the |TableVectorizer|, which automatically recognizes
 # and transforms datetime columns by default.
+
+"""
+Simple example: understanding what DatetimeEncoder extracts
+------------------------------------------------------------
+
+This example demonstrates what the DatetimeEncoder does on a simple,
+non-forecasting regression task.
+
+We compare a naive baseline using a raw numeric timestamp with a model
+that uses DatetimeEncoder with periodic encoding to extract meaningful
+time-based features for a linear regression model.
+"""
+
+import numpy as np
+import pandas as pd
+
+from skrub import DatetimeEncoder, ApplyToCols
+from sklearn.linear_model import LinearRegression
+from sklearn.pipeline import make_pipeline
+from sklearn.metrics import r2_score
+
+# ---------------------------------------------------------------------
+# Create a simple synthetic dataset
+# ---------------------------------------------------------------------
+rng = pd.date_range("2023-01-01", periods=300, freq="h")
+
+X = pd.DataFrame({"date": rng})
+
+# Target depends on hour of day and weekday (cyclic pattern)
+y = (
+    10 * X["date"].dt.hour.isin([8, 9, 17, 18]).astype(int)
+    + 5 * (X["date"].dt.weekday < 5).astype(int)
+    + np.random.normal(0, 1, size=len(X))
+)
+
+# ---------------------------------------------------------------------
+# Baseline model: naive numeric timestamp
+# ---------------------------------------------------------------------
+X_baseline = pd.DataFrame(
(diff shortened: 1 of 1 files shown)


examples/0030_datetime_encoder.py
```diff
@@ -352,3 +352,78 @@
 # features from a datetime column.
 # Also check out the |TableVectorizer|, which automatically recognizes
 # and transforms datetime columns by default.
+
+"""
+Simple example: understanding what DatetimeEncoder extracts
+------------------------------------------------------------
+
+This example demonstrates what the DatetimeEncoder does on a simple,
+non-forecasting regression task.
```
rcap107, line 361: Please expand a bit this section explaining that for the sake of the example we are using a synthetic dataset, and a plot of what the target looks like.

examples/0030_datetime_encoder.py
```diff
@@ -352,3 +352,78 @@
...
+    {"timestamp": X["date"].astype("int64") // 10**9}
+)
+
+model_baseline = LinearRegression()
+model_baseline.fit(X_baseline, y)
+
+y_pred_baseline = model_baseline.predict(X_baseline)
+print("R² with naive timestamp:", r2_score(y, y_pred_baseline))
+
+# ---------------------------------------------------------------------
```
rcap107, line 403: There should also be a section with a DatetimeEncoder that does not include periodic features for the sake of the comparison.

rcap107: Also please add a few lines explaining in words what is happening here as part of the narrative of the example.

examples/0030_datetime_encoder.py
```diff
@@ -352,3 +352,78 @@
...
+print("R² with DatetimeEncoder:", r2_score(y, y_pred_datetime))
+
+# ---------------------------------------------------------------------
+# Inspect generated datetime features
+# ---------------------------------------------------------------------
+apply = model_datetime.named_steps["applytocols"]
+encoder = next(iter(apply.transformers_.values()))
+
+print("Generated features:")
+print(encoder.get_feature_names_out())
```
rcap107, line 429: Please add a conclusion summarizing briefly what is being done here

rcap107 (commented):
Hi @TaobaOuathrani, thanks for this PR. It's a good start, but there is still some work to be done. 

To begin with, this example should be in its own file, and the content of the current file (0030_datetime_encoder.py) should be moved to the FIXME folder in `examples`: we might want to revisit the forecasting example in the future, possibly by using the Data Ops. 

Another general comment is that there should be more narrative, i.e., more text describing what are the steps that are being executed and the reasoning for it. Something like "we want to add periodic features because we are using a linear model and they are useful" etc. 

I am not too keen on using synthetic data as the example, but I am aware that finding a dataset that is suitable for the example (something that has a timeseries but isn't forecasting-based) is hard, so we might be stuck with this for now. 

In any case, thanks for the PR. We can keep on iterating on this.

rcap107:
Oh, and pre-commit checks can be addressed by running either `pixi run lint` or `pre-commit run --all-files` in the project's folder.

rcap107:
Hi @TaobaOuathrani, are you still working on this PR?

Issue #1791, opened 2025-12-05
DOC - Improve the documentation of the selector functions


rcap107:
Selectors and their methods (expand, expand_index) are lacking proper docstrings, so they should be added. 

I think this issue could be more complex than regular documentation issues because selectors need to be handled in a specific way for the documentation to be generated correctly, so it should not be handled by a new contributor.

GaelVaroquaux:
Should we really expose these objects in the docs (the "All", "And"... aka the selector classes rather than the selector constructors or factories)?

I don't think that most users actually need to know about them, and exposing them will make the docs more crowded, and thus harder to find what's important.

rcap107:
When I opened this issue I was thinking specifically about the .expand and .expand_index methods, which can be useful and should be documented properly. I was not thinking of adding docs to the private methods (All And etc.).

In either case, while we address .expand we might want to review the current docs and examples, for example to show that selectors can be composed with each other.

PR #1841, opened 2026-01-14
Added documentation for the Selector object, as well as references to it

emassoulie:
Closes https://github.com/skrub-data/skrub/issues/1791 .

In order to give proper documentation to the Selector class's `expand` and `expand_index` method, added a doc entry to the class itself (which was already public) and references to that documentation in the docstring of each of its inherited classes.


skrub/selectors/_base.py
```diff
@@ -4,7 +4,7 @@
 
 
 def all():
-    """Select all columns.
+    """Select all columns. See also :class:`skrub.selectors.Selector()`.
```
rcap107: I'm not sure anymore there should be a "See also" referring to `Selector` for all these functions, especially since we are saying that the `Selector` should not be used by the users. 

However, we might want to use the See Also field to refer to the user guide, where we explain that the function is returning a selector and that the selector has `expand` and `expand_index`. 

This also means that we need to add a proper mention of this in the user guide, with examples and explanation. 

Also note the formatting of the `See Also`: it's a specific field in the numpy docs. It's used in various places in the library, like the StringEncoder.

skrub/selectors/_base.py
```diff
@@ -219,10 +221,31 @@ def select(df, selector):
...
         raise NotImplementedError()
 
     def expand(self, df):
+        """Lists the column names that the selector would retain if applied to
+        the dataframe `df`.
+
+        Parameters
+        ----------
+        df : dataframe
+
```
rcap107, line 240: There should be an example here showing what `expand` does when applied to a dataframe.

skrub/selectors/_base.py
```diff
@@ -219,10 +221,31 @@ def select(df, selector):
...
+    as :meth:`skrub.selectors.all()` or :meth:`skrub.selectors.make_selector()`.
+    """
+
     def _matches(self, col):
         raise NotImplementedError()
 
     def expand(self, df):
+        """Lists the column names that the selector would retain if applied to
+        the dataframe `df`.
+
```
rcap107, line 236: Either here or in the examples there should be a message mentioning that this can be useful for using the selectors with standard dataframe libraries.

skrub/selectors/_base.py
```diff
@@ -219,10 +221,31 @@ def select(df, selector):
...
+        Parameters
+        ----------
+        df : dataframe
+
+        Returns
+        -------
+        list
+            The list of `df`'s columns that the item would select. In effect,
+            running `df[sel.expand(df)]` should give the exact same result as
+            `sel.transform(df)`.
```
rcap107: The lines I removed here are a better fit for the Notes section, or for an example. 
```suggestion
            The list of `df`'s columns that the item would select. 
```

skrub/selectors/_base.py
```diff
@@ -231,6 +254,21 @@ def expand(self, df):
         return matching_col_names
 
     def expand_index(self, df):
+        """Lists the indices of dataframe `df`'s columns that the selector
+        would retain if applied to `df`.
+
```
rcap107, line 279: The comments I made for the `expand` case also apply here

skrub/selectors/_base.py
emassoulie: Doesn't "See also" necessarily refer to API pages, rather than the User Guide? In which case such a reference might be better suited to the "Notes" section

doc/modules/multi_column_operations/selectors.rst
```diff
@@ -156,7 +185,7 @@ that operate on columns to improve their versatility.
 For example, we can drop columns that have more unique values than a certain amount
 by combining :func:`~skrub.selectors.cardinality_below` with :class:`skrub.DropCols`.
 We first select the columns that have more than 3 unique values, then we invert the
-selector and finally transform the dataframe.
+selector and finally transform the dataframe::
```
rcap107: the :: here is breaking the formatting of the code block below
```suggestion
selector and finally transform the dataframe:
```

doc/modules/multi_column_operations/selectors.rst
```diff
@@ -174,7 +203,7 @@ selector and finally transform the dataframe.
 Selectors can be used in conjunction with :class:`~skrub.ApplyToCols` to transform columns
 based on specific requirements.
 
-Consider the following example:
+Consider the following example::
```
rcap107:
```suggestion
Consider the following example:
```

same here

doc/modules/multi_column_operations/selectors.rst
```diff
@@ -191,7 +220,7 @@ Consider the following example:
 4      Art     3
 
 We might want to apply the :class:`~sklearn.preprocessing.StandardScaler` only to the numeric column. We can
-do this like this:
+do this like this::
```
rcap107:
```suggestion
do this like this:
```

skrub/selectors/_base.py
```diff
@@ -219,10 +220,51 @@ def select(df, selector):
...
+        df : dataframe
+
+        Returns
+        -------
+        list
+            The list of `df`'s columns that the item would select.
+        Examples
+        --------
+        >>> from skrub import selectors as s
+        >>> some_selector = ~s.glob("*_mm") | s.filter(lambda col: 297.0 in col.values)
```
rcap107: I think we can simplify the command here (this will break the doctest)

```suggestion
        >>> some_selector = ~s.glob("*_mm")
```

skrub/selectors/_base.py
```diff
@@ -231,6 +273,44 @@ def expand(self, df):
...
+        Returns
+        -------
+        list
+            The list of indices among `df`'s columns that the item would select.
+
+
+        Examples
+        --------
+        >>> from skrub import selectors as s
+        >>> some_selector = ~s.glob("*_mm") | s.filter(lambda col: 297.0 in col.values)
```
rcap107:
```suggestion
        >>> some_selector = ~s.glob("*_mm")
```

skrub/selectors/_base.py
```diff
@@ -231,6 +273,44 @@ def expand(self, df):
...
+        >>> import polars as pl
+        >>> polars_df = pl.DataFrame(
```
rcap107:
```suggestion
        >>> df = pl.DataFrame(
```

skrub/selectors/_base.py
rcap107: this selector doesn't work in polars by the way

skrub/selectors/_base.py
```diff
@@ -231,6 +273,44 @@ def expand(self, df):
...
+        ...     {
+        ...         "height_mm": [210.0, 297.0],
+        ...         "width_mm": [188.5, 210.0],
+        ...         "kind": ["A5", "A4"],
+        ...         "ID": [5, 4],
+        ...     }
+        ... )
+        >>> some_selector.expand(df)
```
rcap107: this is the docstring for expand_index, not expand

doc/modules/multi_column_operations/selectors.rst
```diff
@@ -145,6 +145,35 @@ following selector won't compute the cardinality of non-categorical columns::
...
+    >>> polars_df = pl.DataFrame(
+    ...     {
+    ...         "height_mm": [210.0, 297.0],
+    ...         "width_mm": [188.5, 210.0],
+    ...         "kind": ["A5", "A4"],
+    ...         "ID": [5, 4],
+    ...     }
+    ... )
+    >>> some_selector.expand(polars_df)
+    ['height_mm', 'kind', 'ID']
```
rcap107: no need to indent and have the :: here, it's already being rendered by sphinx

also, I think it's better if we explain clearly what this selector is doing (no cols with _mm or cols that include the value 297) because it's a bit convoluted and readers may get lost 

```suggestion
useful on complicated combinations of selectors:

>>> some_selector = ~s.glob('*_mm') | s.filter(lambda col: 297.0 in col)
>>> import polars as pl
>>> polars_df = pl.DataFrame(
...     {
...         "height_mm": [210.0, 297.0],
...         "width_mm": [188.5, 210.0],
...         "kind": ["A5", "A4"],
...         "ID": [5, 4],
...     }
... )
>>> some_selector.expand(polars_df)
['height_mm', 'kind', 'ID']
```

PR #1852, opened 2026-01-19
ENH - Add path fields to the bunch objects returned by the dataset fetchers

### LINKED ISSUE #1422: Various features and improvements for the skrub.datasets utilities
I think the skrub.datasets utilities for loading and returning datasets could be improved a bit for for a better user experience:
- skrub supports both pandas and polars, but datasets are always returned as pandas. It would be good to add an option to return the dataframes as polars dataframes instead. This could be implemented as an option for  #1377 ("preferred dataframe engine" for example)
- Currently, datasets are returned as a `Bunch` object, which is pretty much a dict that shows the repr of the dataframes when printed in a notebook cell. I think it would be better to have a more compact repr that shows the name and shape of the dataframes in the bunch, as well as metadata info if available. 
- This was mentioned in another issue (#1252), but it would be useful to have more info on the datasets (at least, their size) on the doc entry for the respective function. 
- It would be nice to have an option to save the datasets in parquet, rather than csv (this could be another option for #1377 )

rcap107:
This PR is improving the dataset fetcher functions. It addresses #1422 by adding the path to the dataset file to the Bunch object returned by the fetcher. 

I am also adding the default data folder of skrub to the configuration file, and I'm deprecating the original name `SKRUB_DATA_DIRECTORY` in favor of `SKB_DATA_DIRECTORY`, to follow the same format as the other environment variables set by skrub.


rcap107:
Something I did not consider is that some datasets have multiple paths (such as `fetch_plane_delays`).

For the moment I am returning a list of paths, but then the result is that then single datasets become clunky to load, like 

```python
data = fetch_employee_salaries()
df = pd.read_csv(data["paths"][0])
```

I'm not sure what's the best way to deal with this

jeromedockes:
thanks a lot @rcap107 !

> For the moment I am returning a list of paths, but then the result is that then single datasets become clunky to load

yes and it will be hard to get wrong as the order is a bit arbitrary (alphabetical). Also, using the paths rather than dataframes will be the default in almost all examples so it would be nice to avoid the extra `["paths"]` indirection. I can  suggest 2 options, there may be better ones:

- put it all at the same level in the result, and add `_file` or `_path` to the keys, eg the `fetch_credit_fraud` would return `{"baskets": baskets_df, "baskets_file": baskets_csv_file_path, "products": ..., "products_file": ...}`
- or add a parameter to control whether in-memory objects or file paths are returned. `fetch_credit_fraud(load_data=True)['baskets']` would be a dataframe and `fetch_credit_fraud(load_data=False)['baskets']` would be a file path. load_data=False would be a better default but unfortunately not backwards compatible.

An advantage of the second option is it avoids using memory for the dataframes if we don't use them. This is also achievable in the first option by overriding item lookup in the bunch to load the dataframes lazily when the keys are accessed but this is slightly more complex.

jeromedockes:
Important: when there are transformations done in the dataframes such as selecting a train or test set, the result must be written in the data dir so we can return the correct file paths. For example in `fetch_credit_fraud` at the moment if we use `split`, the dataframes will contain only the specified split but the paths always point to the full dataset

jeromedockes:
> I am also adding the default data folder of skrub to the configuration file, and I'm deprecating the original name SKRUB_DATA_DIRECTORY in favor of SKB_DATA_DIRECTORY, to follow the same format as the other environment variables set by skrub.

FWIW I think `SKRUB` would have been more natural & easier to remember, I can definitely see myself doing `export SKRUB_DATA_DIRECTORY=...` and wondering why it is being ignored. But I agree that the most important is to be consistent across all options

examples/0020_text_with_string_encoders.py
```diff
@@ -59,6 +63,12 @@
 
 TableReport(X)
 
+# %%
+# We prepare the target variable by mapping the binary labels "Toxic" and "Not Toxic"
+# to 1 and 0, respectively. The target is reused throughout the example.
+
+y = X.pop("is_toxic").map({"Toxic": 1, "Not Toxic": 0})
```
jeromedockes, line 70: not related to this PR but isn't the classifier able to deal with this? or we should use a LabelEncoder? otherwise when we later get some predictions for unseen examples we would need to be able to revert that encoding

rcap107:
I modified the code so that the bunch now has a field named DATASET_path for each dataset, but looking at the examples I am not convinced it's a very ergonomic solution, especially for datasets with a single table.

skrub/datasets/_fetching.py
```diff
@@ -46,6 +48,11 @@ def fetch_employee_salaries(data_home=None, split="all"):
         - ``y`` : pd.DataFrame, target labels. Shape: (9228, 1)
         - ``metadata`` : a dictionary containing the name, description, source and
           target
+        - ``employee_salaries_path`` : str, the path to the employee salaries CSV file
+        - ``employee_salaries_train_path`` : str, the path to the train split CSV file
```
jeromedockes: I think this part of the docstring is outdated? it is always employee_salaries_path now right?
also there is the short "path" for single-table datasets?

skrub/datasets/_fetching.py
```diff
@@ -96,6 +113,7 @@ def fetch_medical_charge(data_home=None):
         - ``y`` : pd.DataFrame, target labels. Shape: (163065, 1)
         - ``metadata`` : a dictionary containing the name, description, source and
           target
+        - ``medical_charge_path`` : str, the path to the medical charge CSV file
```
jeromedockes: same as "path"? maybe for single table datasets we should have only that key 🤔

skrub/datasets/_fetching.py
```diff
@@ -242,10 +264,20 @@ def fetch_credit_fraud(data_home=None, split="train"):
...
         - ``baskets`` : pd.DataFrame, table containing baskets ID and target.
         Shape: (92790, 2)
-        - ``product`` : pd.DataFrame, table containing features about products
+        - ``products`` : pd.DataFrame, table containing features about products
           contained in baskets. Shape: (163357, 7)
         - ``metadata`` : a dictionary containing the name, description, source and
           target
+        - ``baskets_path`` : str, the path to the baskets CSV file
+        - ``products_path`` : str, the path to the products CSV file
+        - ``baskets_train_path`` : str, the path to the train split baskets CSV file
```
jeromedockes: outdated?

skrub/_config.py
```diff
@@ -177,6 +213,18 @@ def set_config(
         This configuration can also be set with the ``SKB_CARDINALITY_THRESHOLD``
         environment variable.
 
+    data_folder : str or pathlib.Path, default=None
```
jeromedockes: as the env var name is directory let us use directory everywhere (or dir for short), and never "folder"

skrub/datasets/_utils.py
```diff
@@ -262,12 +265,24 @@ def load_dataset_files(dataset_name, data_home):
...
         if file_path.suffix == ".csv":
             bunch[file_path.stem] = pd.read_csv(file_path)
+            bunch[file_path.stem + "_path"] = str(file_path)
         elif file_path.suffix == ".json":
             bunch[file_path.stem] = json.loads(file_path.read_text(encoding="utf-8"))
-
+            if file_path.stem != "metadata":
+                bunch[file_path.stem + "_path"] = str(file_path)
+            else:
+                bunch["metadata_path"] = str(file_path)
```
jeromedockes: i'm not sure what this does differently than the if clause?

rcap107: it does nothing different, you're right

examples/data_ops/1120_multiple_tables.py
```diff
@@ -32,25 +32,66 @@
...
 # %%
 # The ``products`` table contains information about the products that have been
 # purchased, and the basket they belong to. A basket contains at least one product.
 # Products can be associated with the corresponding basket through the "basket_ID"
 # column.
 
 # %%
-skrub.TableReport(dataset.products)
-
+products
```
jeromedockes, line 77: I think the text below needs to be updated now that the start of the pipeline has changed. for example the next section that says "we start by building 2 variables"

we can also take the opportunity to trim it as it has the problem we have discussed of mixing discussion of how to use it and motivation / rationale

examples/data_ops/1120_multiple_tables.py
```diff
@@ -194,25 +229,31 @@
...
-            basket_ID="abc",
-            item="COMPUTER",
-            cash_price=200,
-            make="APPLE",
-            model="XXX-X",
-            goods_code="239246782",
-            Nbr_of_prod_purchas=1,
-        )
-    ]
+RocCurveDisplay.from_predictions(
```
jeromedockes: nice addition :)

CHANGES.rst
```diff
@@ -36,6 +36,12 @@ Changes
   :pr:`1819` by :user:`Eloi Massoulié <emassoulie>`
 - :func:`compute_ngram_distance` has been renamed to :func:`_compute_ngram_distance` and is now a private function.
   :pr:`1838` by :user:`Siddharth Baleja <siddharthbaleja>`.
+- The dataset fetcher functions now include the ``path`` and ``paths`` fields, which contain the paths to the datasets.
```
jeromedockes: The "paths" part is outdated?

examples/data_ops/1140_subsampling.py
```diff
@@ -14,10 +14,13 @@
 
 # %%
 
+import pandas as pd
+
 import skrub
 import skrub.datasets
 
-dataset = skrub.datasets.fetch_employee_salaries().employee_salaries
+file_path = skrub.datasets.fetch_employee_salaries().path
```
jeromedockes, line 22: I think in some examples we should start the pipeline from the path, but I'm not sure which. we can revisit that in another PR if you prefer

jeromedockes (approved):
Very cool addition, thanks!! we will have to remember to exploit it in future changes to the examples :) I also like the use of a dropdown in the credit fraud one

LGTM

examples/data_ops/1140_subsampling.py
rcap107: Yes, I think we should just merge this PR and open a separate issue tracking the examples that need to be updated.

jeromedockes:
thanks @rcap107 !

PR #1853, opened 2026-01-20
FIX - CI is failing with latest pixi lock version

rcap107:
related to #1821


rcap107:
It seems like the tests are failing because pyparsing is now version 3.3.1, while the lock file in main is still using 3.2.5. According to the release notes (https://github.com/pyparsing/pyparsing/blob/master/CHANGES): 

> In the pyparsing release 3.3.0, use of many of the pre-PEP8 methods (such as
`ParserElement.parseString`) will start to raise `DeprecationWarnings`. I plan to
completely drop the pre-PEP8 methods in pyparsing 4.0, though we won't see that release
until some time later in 2026. So there is plenty of time to convert existing parsers to
the new function names before the legacy functions are completely removed.

I'm pinning the pyparsing version for the time being

rcap107:
I'm also updating the ci config to remove the `deploy` step as per https://circleci.com/docs/guides/orchestrate/migrate-from-deploy-to-run/

.circleci/config.yml
```diff
@@ -44,17 +44,17 @@ jobs:
       - attach_workspace:
           at: doc/_build/html
       - run: ls -ltrh doc/_build/html
-      - deploy:
+      - run:
           command: |
-            if [[ "${CIRCLE_BRANCH}" =~ ^main$|^[0-9]+\.[0-9]+\.X$ ]]; then
+            if [[ "${CIRCLE_BRANCH}" =~ ^main$|^[0-9]+\.[0-9]+\.X$|^fix-pixi$ ]]; then
```
glemaitre: do you still want to keep this fix-pixi?

pyproject.toml
```diff
@@ -171,6 +171,7 @@ jinja2 = "==3.1.2"
 matplotlib = "==3.4.3"
 pydot = "*"
 requests = "==2.27.1"
+pyparsing = "<3.3"
```
glemaitre, line 176: Maybe put a small comment above to explain why

glemaitre (approved):
Looks good otherwise.

.circleci/config.yml
rcap107: whoops no, I forgot to remove this

glemaitre:
Thanks @rcap107 for the fix :)

Issue #1447, opened 2025-06-18
Improve the documentation of the skrub.datasets fetch functions


rcap107:
At the moment, each dataset has its own version of the documentation. Some have info about the dataset, some have a description of the Bunch object. In general, it's messy and not very uniform. 

I think all datasets should include information on the size of the dataset itself (number of rows/columns, size on disk), a description of the task, and in general the format should be standardized.

rcap107:
This was mostly dealt with in #1503, but the docstring of `fetch_california_housing` is still missing the size on disk so I'll close this once that small change is made.

wenebob:
Hey @rcap107, can I work on this?

rcap107:
Hi @Faith-Nchifor, yes, thanks a lot! This is quite a simple issue, for this you should just fetch the dataset and note down the size of the dataset on disk, as well as some other statistics about it. You can check some of the other fetchers to see how it was formatted in the other cases: 

https://skrub-data.org/stable/reference/generated/skrub.datasets.fetch_movielens.html#skrub.datasets.fetch_movielens

PR #1854, opened 2026-01-20
Included `size on disk` on the california housing dataset

wenebob:
As @rcap107 pointed out in #1447, both datasets had descriptions, such as dataframe sizes and size on disk, except the `fetch_california_housing`, which was missing the size on disk.  This has now been included. I cross-checked to find that nothing else was missing from the dataset docstrings.


rcap107:
Tiny fix, but enough to close an old issue, thanks a lot @Faith-Nchifor! There are a few other issues that can be picked up if you'd like.

PR #1855, opened 2026-01-21
FIX - CI OOM issue and some tests failing due to pandas 3.0

rcap107:
This PR is addressing various issues that were discovered because CircleCI was failing because it was running out of memory. 

The main issue with CircleCI was addressed by adding the `html-split` command to the docs makefile. This command first builds the examples and excludes the rest of the documentation, then builds the documentation without re-executing the examples. Finally, it copies the files generated for the examples to the build folder to have the full documentation build. 

I also removed doctests from the sphinx configuration because they should already be covered by `test` and `test-user-guide`, which should help saving some memory. 

Since I had to update the lock file, I also fixed a few tests that were breaking because of the pandas 3.0 release. 

I also added `memory_profiler` as a doc dependency to show the peak memory for the various examples. 

I also removed all the leftover references to the KEN embeddings that were not removed in the relevant PR.


skrub/_dataframe/tests/test_common.py
```diff
@@ -548,16 +548,6 @@ def test_is_string(df_module):
         assert not ns.is_string(ns.col(df, col))
 
 
-def test_sentinel_is_string_pandas_3(df_module):
```
rcap107, line 551: I could not find anywhere the reason for adding this test, other than acting as a reminder that pandas 3.0 has been released. This is the only test that fails, so I'm just removing it here

doc/modules/multi_column_operations/multi_column_operations.rst
```diff
@@ -19,15 +19,17 @@ is done with the :class:`sklearn.compose.ColumnTransformer`:
 
 
 >>> import pandas as pd
+>>> import warnings
```
rcap107: Even if this works, this is really ugly and I'd rather not do this

The problem is that we are executing the doctests with different versions of pandas. With min-reqs, it's pandas 1.5.3 and these lines (with `object`) work fine, while in the latest versions it's pandas 3 which raises a deprecation warning here and asks to put in `str` to silence it. 

So, there is no way around this problem other than rewriting the example entirely so that it does not depend on the pandas version.

rcap107:
Numpy 2.4.1 is still causing the CI to run out of memory

lesteve:
FWIW in scikit-learn we did bump into a somewhat similar issue, see https://github.com/scikit-learn/scikit-learn/pull/32902.

It's not clear what was causing it, but I suspect some updated package maybe? A shortlist of packages were in https://github.com/scikit-learn/scikit-learn/pull/32902#issuecomment-3685435307 and I tried to reproduce in the CI with SSH debug and it did not seem to be numpy 2.4.

As a work-around, we are now using a bigger CircleCI runner (`medium+` 6GB RAM instead of 4GB) to avoid the issue, see https://github.com/scikit-learn/scikit-learn/blob/62b08044bf7d4d8e96b439247c6d8d6ea0445496/.circleci/config.yml#L57-L60

rcap107:
In the end, it seems like increasing the size of the container solved the issue (at least for the time being), but I suspect we will have to deal with the issue again in the future. Still, I think that in the interest of moving on with this and various other issues, we should just go with the current fix. 

I am also dealing with an unrelated side problem (another one...) with the lock file, but aside from that the PR should be ready to merge soon enough.

doc/api_reference.py
```diff
@@ -306,9 +306,6 @@
                     "datasets.fetch_drug_directory",
                     "datasets.fetch_employee_salaries",
                     "datasets.fetch_flight_delays",
-                    "datasets.fetch_ken_embeddings",
```
rcap107, line 309: these lines should have been removed in #1567

doc/conf.py
```diff
@@ -64,7 +64,6 @@
     # builtin
     "sphinx.ext.autodoc",
     "sphinx.ext.autosummary",
-    "sphinx.ext.doctest",
```
rcap107, line 67: doctests are already being run as part of the test matrix

doc/conf.py
```diff
@@ -396,24 +395,15 @@ def notebook_modification_function(notebook_content, notebook_filename):
             "</div>",
         ]
     )
-    # TODO: remove this when we remove KEN embeddings
```
rcap107, line 399: this should also have been removed in #1567

doc/conf.py
```diff
@@ -467,6 +457,13 @@ def reset_skrub_config(gallery_conf, fname):
     skrub.set_config(**default_global_config)
 
 
+def call_garbage_collector(gallery_conf, fname):
```
rcap107, line 460: this should help reduce a bit the memory pressure

doc/conf.py
```diff
@@ -555,9 +553,6 @@ def reset_skrub_config(gallery_conf, fname):
     "pandas.melt": "pandas.melt",
     "pandas.merge": "pandas.merge",
     # Skrub
-    "fetch_ken_table_aliases": "skrub.datasets.fetch_ken_table_aliases",
```
rcap107, line 558: again #1567

doc/modules/column_level_featurizing/feature_engineering_datetimes.rst
```diff
@@ -35,9 +35,9 @@ Skrub provides helpers to parse datetime string columns automatically:
 >>> s = pd.Series(["2024-05-05T13:17:52", None, "2024-05-07T13:17:52"], name="when")
 >>> s
 0    2024-05-05T13:17:52
-1                   None
+1                   ...
```
rcap107, line 38: pandas 3.0 has NaN here

doc/modules/multi_column_operations/multi_column_operations.rst
```diff
@@ -15,29 +15,7 @@ to be scaled at the same time.
 While the heuristics used by the :class:`TableVectorizer` are usually good enough
 to apply the proper transformers to different datatypes, using it may not be an
 option in all cases. In scikit-learn pipelines, the column selection operation can
-is done with the :class:`sklearn.compose.ColumnTransformer`:
```
rcap107, line 18: this code block can't run without failing because pandas 3.0 raises an exception on the `dtype_include=object` lines, asking to put `str` instead 

min reqs pandas won't accept `str` as parameter, and just fails

pyproject.toml
```diff
@@ -258,7 +259,7 @@ ci-py314-polars-without-pyarrow = ["py314", "test", "polars-without-pyarrow"]
 ci-py314-latest-optional-deps = ["py314", "optional", "test"]
 
 # no-default-feature is necessary for pixi to download from nightly PyPI indexes rather than conda channels
-ci-nightly-deps = { no-default-feature = true, features = [ "py314", "nightly-dependencies", "test" ] }
+# ci-nightly-deps = { no-default-feature = true, features = [ "py314", "nightly-dependencies", "test" ]}
```
rcap107: this is commented out only so I could debug circleCI, it should be uncommented once scipy releases the win64 dev version

skrub/_clean_categories.py
```diff
@@ -85,6 +85,20 @@ class CleanCategories(SingleColumnTransformer):
...
+    Name: c, dtype: category
+    Categories (3, ...): ['one', 'three', 'two']
+    >>> cleaner = CleanCategories()
+    >>> cleaner.fit_transform(s)
+    0      one
+    1      two
+    2    three
+    Name: c, dtype: category
+    Categories (3, ...): ['one', 'three', 'two']
     >>> cleaner.fit_transform(s) is s # doctest: +SKIP
```
rcap107: I think this particular line should be removed, because it's not true with pandas 3.0, even though the end result should be the same

skrub/_data_ops/tests/test_errors.py
```diff
@@ -550,6 +551,7 @@ def test_bad_names():
         skrub.var("_skrub_X")
 
 
+@skip_polars_installed_without_pyarrow
```
rcap107, line 554: this test did not fail with pandas <3.0

skrub/tests/test_multi_agg_joiner.py
```diff
@@ -3,16 +3,20 @@
...
 from skrub import _dataframe as sbd
 from skrub._multi_agg_joiner import MultiAggJoiner
-from skrub.conftest import _POLARS_INSTALLED
+from skrub.conftest import _POLARS_INSTALLED, skip_polars_installed_without_pyarrow
 
 
 @pytest.fixture
-def main_table():
-    df = pd.DataFrame(
+def main_table(df_module):
```
rcap107, line 10: this needed to be switched to `df_module` and was also needed for the switch to pandas 3.0 

as a result, some of the tests failed because of the pandas nullable types

skrub/tests/test_multi_agg_joiner.py
```diff
@@ -3,16 +3,20 @@
...
+    df = df_module.make_dataframe(
         {
             "userId": [1, 1, 1, 2, 2, 2],
             "movieId": [1, 3, 6, 318, 6, 1704],
-            "rating": [4.0, 4.0, 4.0, 3.0, 2.0, 4.0],
+            # Note that the ratings for movie 6 are 4.0 and 3.0, so that the
```
rcap107, line 15: comment explains the reasoning

doc/Makefile
```diff
@@ -29,16 +28,15 @@ html:
 	# the embedding of images more robust
 	rm -rf $(BUILDDIR)/html/_images
 	#rm -rf _build/doctrees/
-	$(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
+	SKB_TABLE_REPORT_VERBOSITY=0 $(SPHINXBUILD) -b html $(ALLSPHINXOPTS) $(BUILDDIR)/html
```
rcap107, line 31: setting verbosity to 0 so that sphinx build docs are not clogged by the tablereport writing "Progressing"

.pre-commit-config.yaml
```diff
@@ -1,26 +1,28 @@
 repos:
--   repo: https://github.com/pre-commit/pre-commit-hooks
```
rcap107, line 2: this is all formatting

.pre-commit-config.yaml
```diff
@@ -1,26 +1,28 @@
...
+      - id: ruff-format
+  - repo: https://github.com/codespell-project/codespell
     # Configuration for codespell is in pyproject.toml
     rev: v2.3.0
     hooks:
-    -   id: codespell
+      - id: codespell
         exclude: .*/package-lock.json
+        additional_dependencies:
+          - tomli # needed to parse pyproject.toml
```
rcap107, line 28: this line lets codespell read the config in pyproject and actually fixes it

skrub/_clean_categories.py
```diff
@@ -85,8 +85,6 @@ class CleanCategories(SingleColumnTransformer):
     2    three
     Name: c, dtype: category
     Categories (3, ...): ['one', 'three', 'two']
-    >>> cleaner.fit_transform(s) is s # doctest: +SKIP
```
rcap107, line 88: this will not pass on pandas 3.0, but the actual behavior is the same, so I think we can remove this line

GaelVaroquaux: What do you mean not pass?
Does it raise an error?

rcap107: on pandas 3.0 

```python
cleaner.fit_transform(s) is s
```
returns False

but aside from that change no tests break and the behavior as far as I can tell is the same

GaelVaroquaux: OK.

GaelVaroquaux (approved):
LGTM. Thank you

PR #1857, opened 2026-01-24
docs: Reorganize the "Development" section

### LINKED ISSUE #1582: DOC - Reorganizing the "Development" section in the website top bar
A general comment: 
in the "Development" section of the website, there are subsections including the one you want to add and "Contributing to skrub". "Contributing to skrub" is then also a section in the drop-down of "More". We might want to add an issue to clear this up.

_Originally posted by @jovan-stojanovic in https://github.com/skrub-data/skrub/pull/1578#pullrequestreview-3173194089_

MarieSacksick:
Closes #1582 

This issue says three things:

1. We might also consider reformatting a bit the Development outline if needed.
2. Also would it make sense to rename this to just "Contributing"? Since the context is already within the skrub repo, "Contributing to skrub" might be a bit redundant. (I agreed, and therefore did the change)
3. we can start by removing "Contributing to skrub" from the drop down menu, since it's already present in the "Development" page.

2 and 3 are tackled.  
For 1, I quite like the contributing guide for now, because it's very detailed. Most of what I did is changing from:  
image

to:  

image

The idea is to go step by step for the contributor. Building the doc locally to me is part of checking the quality of contribution.

### DIFF
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,7 +1,7 @@
 .. _contributing:
 
-Contributing to skrub
-=====================
+Contributing guide
+==================
 
 First off, thanks for taking the time to contribute!
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,7 +1,7 @@
 .. _contributing:
 
 Contributing guide
-==================
+=====================
 
 First off, thanks for taking the time to contribute!
 
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -1,7 +1,7 @@
 .. _contributing:
 
 Contributing guide
-=====================
+==================
 
 First off, thanks for taking the time to contribute!
 
@@ -177,6 +177,8 @@ When contributing, keep these project goals in mind:
     - Document all public functions, methods, variables, and class signatures.
     - The public API refers to all components available for import and use by library users. Anything that doesn't begin with an underscore is considered part of the public API.
 
+Checking the quality of your code contribution
+----------------------------------------------
 
 Testing the code
 ~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.rst
```diff
@@ -419,16 +359,76 @@ You can view it by opening the local ``doc/_build/html/index.html`` file.
...
+
+Finally, if the remote repository was changed, you might need to run
+  ``pre-commit run --all-files`` to make sure that the formatting is
+  correct.
+
+Integration
+^^^^^^^^^^^
+
+Community consensus is key in the integration process. Expect a minimum
+of 1 to 3 reviews depending on the size of the change before we consider
```
MarieSacksick: this is actually unclear to me: isn't it this instead, since there is a mention of community?

```suggestion
of 1 to 3 reviewers depending on the size of the change before we consider
```

CONTRIBUTING.rst
```diff
@@ -276,79 +278,18 @@ can be pushed. Something worth noting is that if the ``pre-commit``
...
---------------------------
+Ensuring the documentation builds
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
 ..
   Inspired by: https://github.com/scikit-learn/scikit-learn/blob/main/doc/developers/contributing.rst
 
 **Before submitting your pull request, ensure that your modifications haven't
 introduced any new Sphinx warnings by building the documentation locally
 and addressing any issues.**
```
MarieSacksick: Given the number of warnings already present, I think it's very hard check to do and to enforce.
  Shouldn't it be removed?


```suggestion
```

MarieSacksick: Or, other solution would be to create a couple of issues to fix the current warnings in documentation building.

rcap107: We can do both. From experience silencing all the sphinx warning would be nigh impossible, so I think it makes sense to remove this sentence.

CONTRIBUTING.rst
```diff
@@ -276,79 +278,18 @@ can be pushed. Something worth noting is that if the ``pre-commit``
...
+..
+  Given the number of warnings already present, I think it's very hard check to do and to enforce.
+  Shouldn't it be removed?
```
rcap107: I think this can be removed

CONTRIBUTING.rst
rcap107: I would rather have something like "Expect a minimum of 1 to 3 reviews from maintainers..."

rcap107 (commented):
This looks good! Thanks a lot @MarieSacksick 👍 

I left a couple of comments, but aside from that I think it's already a big improvement

rcap107 (commented):
This looks good! Thanks a lot @MarieSacksick 👍 

I left a couple of comments, but aside from that I think it's already a big improvement

CONTRIBUTING.rst
MarieSacksick: partly resolved, I didn't create the issues (yet)

rcap107 (approved):
Looks good to me, thanks a lot @MarieSacksick

Issue #1862, opened 2026-01-26
[ENH] Enhance hyperparameter choice system with constraint checking and dependency tracking.


Mohima6:
### Problem Description

I wish I could use skrub to define hyperparameter search spaces that include explicit value constraints and dependencies between hyperparameters.

Currently, skrub’s choice system (choose_int, choose_float, choose_from, optional, etc.) allows declarative hyperparameter selection, but:

- There is no way to enforce constraints on sampled values (e.g., only multiples of 8, only positive values, etc.).
- Hyperparameter dependencies (e.g., one choice depending on another) must be manually handled outside the choice system.
- Only uniform and log-uniform sampling are supported, limiting expressiveness for real-world optimization setups.

As a result, users must implement custom filtering or external logic to avoid invalid configurations, which reduces clarity and wastes search iterations.

### Feature Description

I propose extending the hyperparameter choice system with:

1) Constraint checking on choices

Add an optional `constraints` parameter to numeric choices:

Expected behavior:
- During sampling (rvs), invalid values are automatically resampled until constraints are satisfied.
- Default values are validated at initialization.
- Clear errors are raised if constraints cannot be satisfied.

2) Dependent hyperparameter choices

Introduce a DependentChoice object that defines choices based on previously-sampled ones:

Expected behavior:
- Sampling respects dependency order.
- Generated configurations remain internally consistent.

3) Advanced sampling distributions

Expected behavior:
- Default behavior remains uniform/log-uniform.
- Additional distributions improve Bayesian optimization priors.

All additions should be backward compatible with existing APIs.

### Alternative Solutions

Current workarounds include:

- Post-sampling filtering of invalid configurations, which wastes optimization iterations.
- Writing custom sampling code outside skrub’s declarative choice system.
- Using external libraries such as Optuna or scikit-optimize to handle constraints, which adds dependencies and breaks integration with skrub’s native API.

While these approaches work, they reduce code clarity and make hyperparameter definitions less self-contained.

### Additional Context

This enhancement builds on skrub’s existing choice system used in components such as TableVectorizer.

Similar constraint and dependency features exist in hyperparameter optimization frameworks (e.g., Optuna’s conditional parameters), showing real-world demand for this capability.

I would be happy to contribute a PR implementing constraint checking first, if this proposal aligns with project direction.

Mohima6:
would like to work for pr.

PR #1863, opened 2026-01-26
[ENH] Extend choice system with constraints and dependent hyperparame…

Mohima6:
…ters

This change proposes enhancements to skrub’s hyperparameter choice system to support constraint checking and dependency tracking between hyperparameters.

Currently, numeric and categorical choices allow declarative search space definitions but lack mechanisms to:
- Enforce validity constraints on sampled values.
- Define hyperparameters whose structure depends on previously chosen values.

The proposed extension introduces:
- Optional constraint validation for numeric choices, ensuring sampled and default values satisfy user-defined rules.
- Dependency-aware choices, allowing one hyperparameter to be defined based on the sampled value of another.
- A foundation for supporting richer sampling strategies in the future.

These additions improve expressiveness, prevent invalid configurations during optimization, and remain backward compatible with existing APIs.

### DIFF
--- a/skrub/_data_ops/update_choosing.py
+++ b/skrub/_data_ops/update_choosing.py
@@ -1,0 +1,1376 @@
+import dataclasses
+import functools
+import numbers
+import typing
+from collections.abc import Callable, Sequence
+from typing import Any, Optional as TypingOptional, Union, TypeVar
+
+import numpy as np
+from scipy import stats
+from sklearn.utils import check_random_state
+
+from .. import _utils
+from ._utils import NULL, OPTIONAL_VALUE
+
+
+T = TypeVar('T')
+DistributionType = Union[str, stats.rv_continuous, stats.rv_discrete]
+
+
+def _with_fields(obj, **fields):
+    """
+    Make a copy of a dataclass instance with different values for some of the attributes
+    """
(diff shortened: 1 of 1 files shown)


rcap107:
Hi @Mohima6, thanks for checking out skrub. 

While the idea is interesting, unfortunately the current execution is not acceptable. 

To begin with, such a modification should have been discussed in an issue before working on it, since it touches a critical part of the codebase and it is important to have an idea of what is the plan before opening a PR. 

Secondly, the PR in its current form is impossible to review: by creating a brand new file that copy/pastes the old `_choosing.py` file it is now impossible to have a clean diff between the old file and the new. As a result, reviewing this becomes a scavenger hunt trying to find all the changes that may have been done in the file. 

Thirdly (and this follows the first point), data ops by themselves are a part of the codebase that is particularly tricky to navigate, and for this reason we prefer to leave such modifications to the maintainers. This is something that would have been brought up during in a discussion on the issue. 

Given all this, I will be closing both this and #1865, given it has the same problems.

rcap107:
I did not realize that #1862 was opened, then immediately closed as "completed" without any actual discussion or waiting for feedback. The PR remains closed.

PR #1861, opened 2026-01-26
ENH - Add configurable distance metric for fuzzy join nearest‑neighbor matching

sabasiddique1:
REF: #1847 
**Goal**  
Allow users to choose the distance metric used for nearest‑neighbor matching in `fuzzy_join` / `Joiner`, enabling alternatives like cosine or manhattan for different data types.

**Context / Issue**  
Addresses enhancement request to support non‑Euclidean distance metrics for fuzzy join nearest neighbors.

**What changed (step‑by‑step)**  
1. Added a `metric` parameter to `fuzzy_join` and `Joiner` with default `'euclidean'`.  
2. Passed the `metric` through the matching classes to `sklearn.neighbors.NearestNeighbors`.  
3. Documented the new parameter in docstrings with common metric examples.  
4. Added tests covering multiple metrics and invalid metric handling.

**Tests**  
- `pytest skrub/tests/test_fuzzy_join.py -v`

PR #1870, opened 2026-01-29
DOC: explain unpacking results from deferred functions

sabasiddique1:
**Goal**
Add a short user‑guide note explaining why deferred functions returning multiple values cannot be unpacked directly, and how to handle it.

**What changed**
Added a small “Unpacking multiple outputs from deferred functions” section to control_flow.rst

### DIFF
--- a/doc/modules/data_ops/basics/control_flow.rst
+++ b/doc/modules/data_ops/basics/control_flow.rst
@@ -42,7 +42,7 @@ see a result for that value:
 <GetAttr 'columns'>
 Result:
 ―――――――
-Index(['item', 'price', 'qty'], dtype='str')
+Index(['item', 'price', 'qty'], dtype=...)
 
 The "result" we see is an *example* result that the computation produces for the
 data we provided. But we want to fit our pipeline and apply it to different


doc/modules/data_ops/basics/control_flow.rst
```diff
@@ -42,7 +42,7 @@ see a result for that value:
 <GetAttr 'columns'>
 Result:
 ―――――――
-Index(['item', 'price', 'qty'], dtype='object')
+Index(['item', 'price', 'qty'], dtype='str')
```
rcap107:
```suggestion
Index(['item', 'price', 'qty'], dtype=...)
```

Adding an ELLIPSIS here should let the tests pass.

rcap107:
Hi @sabasiddique1, thanks a lot for the PR. 

I wanted to point out that it's better to ask in the original issue before opening a PR. For a tiny one like this, it's not a big deal, but for a larger one like #1861 it's very much appreciated.

sabasiddique1:
> Hi @sabasiddique1, thanks a lot for the PR.
> 
> I wanted to point out that it's better to ask in the original issue before opening a PR. For a tiny one like this, it's not a big deal, but for a larger one like #1861 it's very much appreciated.

Hi @rcap107 , thanks for the reminder! I should have checked in the issue first. I’ll do that going forward, especially for larger changes like #1861. Appreciate the guidance!

doc/modules/data_ops/basics/control_flow.rst
```diff
@@ -103,6 +103,25 @@ Result:
...
 3  fork    2.2    4
 
+Unpacking multiple outputs from deferred functions
+--------------------------------------------------
+
+When a deferred function returns more than one value, you cannot unpack the
+result directly because unpacking iterates over the result. Iteration is not
+supported on DataOps until evaluation.
+
+Instead, keep the result as a single DataOp and index into it:
```
rcap107, line 117: Maybe we should be more explicit here and tell users to prefer deferred functions that return only a single value, and use this strategy only when it's really required.

sabasiddique1: Thanks for the suggestion. I’ve clarified that deferred functions should return a single value when possible, and that multiple outputs should only be used when necessary.

rcap107:
Hi @sabasiddique1, the tests are still failing because there is no ELLIPSIS in the docstring. The problem there is that pandas 3.0 will output `str`, while older versions of pandas still output `object`. Adding the ELLIPSIS should fix the issue.

PR #1875, opened 2026-01-30
DOC: clarify handling of numeric strings in TableVectorizer

### LINKED ISSUE #1874: DOC - Add an entry in the user guide on how to deal with numbers with str dtype
There have been various discussions and questions on how series that contain only numbers, but with string as their dtype, should be dealt with.

By default, the TableVectorizer will convert these series to numerical and encode them as such. However, users may want to use the str dtype as a way to denote each number as a category. This behavior is discussed in #1680.

This issue is specifically about adding to the user guide a short explanation on what is the proper way to encode such columns using skrub, that is by converting them to categorical and then feeding them to the table vectorizer. The doc entry should show both an example of what happens when the column is not converted, and one when the column is converted. 

```python
import pandas as pd
from skrub import ToCategorical, ApplyToCols, TableVectorizer

s = pd.Series([str(i) for i in range(10)], name='c')

pipeline = make_pipeline(
    ApplyToCols(ToCategorical(), cols=['c']),
    TableVectorizer()
)
X_transformed = pipeline.fit_transform(pd.DataFrame(s))
````

### PR DESCRIPTION
**Pull request by sabasiddique1**
## Summary

Adds a short section to the TableVectorizer user guide clarifying how numeric strings are handled.

By default, numeric strings are parsed as numeric features. This PR documents how to encode them as categorical by converting the column to pandas' `category` dtype using `ToCategorical` with `ApplyToCols` before passing it to `TableVectorizer`.

Includes a minimal before/after example demonstrating both behaviors.

Closes #1874.

## Type of change

Documentation update

## Test plan

- Ran the example locally.
- Confirmed:
  - Default: numeric strings → numeric column.
  - With `ToCategorical` + `ApplyToCols`: one-hot encoded categorical columns (e.g., `c_0`, `c_1`, `c_2`).

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -9,14 +9,25 @@ Release history
 Ongoing Development
 ===================
 
-New features
+New Features
 ------------
+- The ``eager_data_ops`` :ref:`configuration
+  <user_guide_configuration_parameters>` option has been added. When set to
+  False, no previews are computed and validation is deferred until the DataOp is
+  actually used (e.g. with ``.skb.eval()``) rather than as soon as it is
+  defined. This can make the definition of complex DataOps with many nodes
+  faster (the overhead it removes typically becomes noticeable only in DataOps
+  with 50-100 nodes or more). Moreover, the evaluation of large DataOps has also
+  become faster. :pr:`1890` by :user:`Jérôme Dockès <jeromedockes>`.
 
 Changes
 -------
 
-Bugfixes
+Bug Fixes
 --------
+- The :class:`TableVectorizer` now correctly handles the case where one of the
+  provided encoders is a scikit-learn Pipeline that starts with a skrub
+  single-column transformer. :pr:`1899` by :user:`Jérôme Dockès <jeromedockes>`.
 
 Release 0.7.2
 =============
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -3,7 +3,7 @@
 Contributing guide
 ==================
 
-First off, thanks for taking the time to contribute!
+First off, thank you for taking the time to contribute!
 
 Below are some guidelines to help you get started.
 
@@ -20,7 +20,7 @@ What to know before you begin
 -----------------------------
 
 To understand the purpose and goals behind skrub, please read our
-`vision statement. <https://skrub-data.org/stable/vision.html>`_
+`vision statement <https://skrub-data.org/stable/vision.html>`_.
 
 If you're interested in the research behind skrub,
 we encourage you to explore these papers:
--- a/README.rst
+++ b/README.rst
@@ -18,10 +18,10 @@ skrub
 
 
 **skrub** (formerly *dirty_cat*) is a Python
-library that facilitates doing machine learning with dataframes.
+library that facilitates machine learning with dataframes.
 
 If you like the package, spread the word and ⭐ this repository!
-You can also join the `discord server <https://discord.gg/ABaPnm7fDC>`_.
+You can also join the `Discord server <https://discord.gg/ABaPnm7fDC>`_.
 
 Website: https://skrub-data.org/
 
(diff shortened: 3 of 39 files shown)

### CODE REVIEW

File: doc/modules/default_wrangling/table_vectorizer.rst [hunk 1]
```diff
@@ -36,6 +36,33 @@ The same parameters used for the |Cleaner| can also be set when creating the
 (``drop_null_fraction`` etc.), and a ``datetime_format`` parameter for the
 datetime parsing step.
 
+Numeric strings and categorical encoding
```
rcap107: I would put this section at the end of this file, rather than here.

doc/modules/default_wrangling/table_vectorizer.rst
```diff
@@ -36,6 +36,33 @@ The same parameters used for the |Cleaner| can also be set when creating the
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, numeric values stored as strings are parsed and treated as numeric
+features. If you need to treat those values as categories (for example, ID
+columns or codes), explicitly convert them to pandas' ``category`` dtype before
+calling |TableVectorizer|.
+
+.. code-block:: python
```
rcap107: Rather than a code block, this section should be written using doctests: by using doctests we can make sure that the output of the code is what we're expecting. Doctests are also run on the ci so that we can check that the code is working for the different environments. This can't be done with the code blocks.

rcap107 (commented):
Thanks for the PR @sabasiddique1.  The code snippet is what I wanted to see, but it should be rewritten so that it uses doctests (https://docs.python.org/3/library/doctest.html) rather than a plain code block. You can find examples of how this is done in the documentation.

sabasiddique1:
Thanks for the review! I moved the section to the end of the file and rewrote the example as doctests with deterministic feature-name checks. Verified locally that default parsing yields ['c'] and categorical conversion yields ['c_1', 'c_2', 'c_3'].

doc/modules/default_wrangling/table_vectorizer.rst
```diff
@@ -104,3 +105,36 @@ of control over which operations are applied to which columns.
...
+Default behavior: numeric strings are parsed as a single numeric column (feature
+names are not one-hot encoded):
+
+>>> import pandas as pd
+>>> from skrub import TableVectorizer
+>>> df = pd.DataFrame({"c": ["1", "2", "3"]})
+>>> tv = TableVectorizer().fit(df)
+>>> list(map(str, sorted(tv.get_feature_names_out())))
+['c']
+
```
rcap107, line 133: I think it's fine to simplify this and show directly that the result is a float column

(this may break the formatting)

```suggestion
>>> tv = TableVectorizer()
>>> tv.fit_transform(df)
c
0  1.0
1  2.0
2  3.0
```

doc/modules/default_wrangling/table_vectorizer.rst
```diff
@@ -104,3 +105,36 @@ of control over which operations are applied to which columns.
...
+>>> from skrub import ApplyToCols, ToCategorical
+>>> from sklearn.pipeline import make_pipeline
+>>> pipe = make_pipeline(
+...     ApplyToCols(ToCategorical(), cols=["c"]),
+...     TableVectorizer(),
+... )
+>>> pipe.fit(df)  # doctest: +ELLIPSIS
+Pipeline(steps=[('applytocols', ...), ('tablevectorizer', ...)])
+>>> list(map(str, sorted(pipe.named_steps["tablevectorizer"].get_feature_names_out())))
+['c_1', 'c_2', 'c_3']
```
rcap107: The same here: 
```suggestion
>>> pipe.fit_transform(df)
   c_1  c_2  c_3
0  1.0  0.0  0.0
1  0.0  1.0  0.0
2  0.0  0.0  1.0
```

rcap107: Sidenote: ELLIPSIS is active by default everywhere in the repo, so there is no need to add it

doc/modules/default_wrangling/table_vectorizer.rst
```diff
@@ -127,12 +127,10 @@ names are not one-hot encoded):
 >>> tv = TableVectorizer()
 >>> tv.fit_transform(df)
 c
```
rcap107, line 129: @sabasiddique1 the duplicated line here broke the doctests, I fixed it so they'd pass

PR #1877, opened 2026-02-02
DOC - Correct a few wording mistakes

star1327p:
Correct a few wording mistakes

### DIFF
--- a/examples/FIXME/08_join_aggregation_full.py
+++ b/examples/FIXME/08_join_aggregation_full.py
@@ -191,8 +191,9 @@ results["Dummy Negative"] = get_results(
 # Low effort estimator
 # --------------------
 #
-# Next, we use the |TableVectorizer| and an |HGBC| to create a very simple baseline model
-# that uses the sparse dataset directly. Note that due to the large number of high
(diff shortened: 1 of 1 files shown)


rcap107:
Hi @star1327p, it looks like formatting checks are failing for a line too long, and the pixi lock file is also misbehaving. 

There is one ongoing PR (#1855) that should address some issues with the lock file (including formatting shenanigans), so this PR should be put on hold until the other is merged.

star1327p:
@rcap107 Thanks for the heads up! I have fixed the line that was too long, so the PR should be ready to merge after the lock file issue gets resolved.

examples/FIXME/08_join_aggregation_full.py
```diff
@@ -191,9 +191,10 @@ def get_results(model, X_test, y_test, threshold, amount, time_to_fit):
 # Low effort estimator
 # --------------------
 #
-# Next, we use the |TableVectorizer| and a |HGBC| to create a very simple baseline model
```
rcap107, line 194: This is not a big deal, but for your information examples in FIXME are there because they need extensive rewriting. Fixing typos there may be useful, but be aware that the content may be changed at some point, or we might get rid of the examples entirely.

PR #1880, opened 2026-02-02
DOC - new example for using PyTorch w/ DataOps via skorch

e-strauss:
This PR adds a new documentation example demonstrating how to use PyTorch models with the Skrub DataOps framework by wrapping them with the skorch library. The example fetches the MNIST dataset, defines a small convolutional neural network for digit classification, and integrates it into a DataOps pipeline for training and evaluation. To support this example, the commit adds `skorch` as a new doc dependency.

### DIFF
--- a/examples/data_ops/1160_pytorch.py
+++ b/examples/data_ops/1160_pytorch.py
@@ -1,0 +1,218 @@
+"""
+Using PyTorch (via skorch) in DataOps
+======================================
+
+This example shows how to wrap a PyTorch model with skorch and plug it into a
+skrub DataOps plan.
+
+.. note::
+    This example requires the optional dependencies ``torch`` and ``skorch``.
+
+The main goal here is to show the *integration pattern*:
+
+- **PyTorch** defines the model (an ``nn.Module``)
+- **skorch** wraps it as a scikit-learn compatible estimator
+- **skrub DataOps** builds a plan and can tune skorch (and therefore PyTorch)
+  hyperparameters using the skrub choices.
+"""
+
+# %%
+# Loading the data
+# =================
+#
+# We use scikit-learn's digits dataset because it is small and ships with
+# scikit-learn. Each sample is an 8x8 grayscale image of a
+# handwritten digit, encoded as 64 pixel intensity values and displays a
+# number from 0 to 9.
+from sklearn.datasets import load_digits
+
+digits = load_digits()
+X, y = digits.data, digits.target
+print(f"Dataset shape: {X.shape}")
+print(f"Number of classes: {len(set(y))}")
+
+# %%
+# Start of the DataOps plan
+# ==========================
+#
+# We start the DataOps plan by creating the skrub variables X and y.
+import skrub
+
+X = skrub.X(X)
+y = skrub.y(y)
+
+# %%
+# Data preprocessing
+# ==================
+#
+# We start by normalizing the pixel values to [0, 1] by first
+# computing the global max value and then dividing the pixel values
+# by this max value. Importantly, we freeze the max value (scaling factor)
+# after fitting so that the same rescaling is applied later when we use our
+# dataop for prediction on new (test) data.
+#
+# A convolutional network expects images with shape (N, C, H, W) where:
+#
+# - N: number of samples
+# - C: number of color channels (1 for grayscale)
+# - H, W: image height and width
+#
+# So we reshape the images to (N, 1, 8, 8) for the CNN. The -1 means the first
+# dimension (N) is inferred automatically from the array size.
+#
+# The advantage of using DataOps is that the preprocessing steps are tracked
+# in the plan and will be automatically applied during prediction.
+
+max_value = X.max().skb.freeze_after_fit()
+X_scaled = X / max_value
+X_reshaped = X_scaled.reshape(-1, 1, 8, 8).astype("float32")
+X_reshaped.skb.draw_graph()
+
+
+# %%
+# Building a NN Classifier
+# =========================
+#
+# We'll build a tiny CNN using PyTorch and wrap it with skorch to make it
+# scikit-learn compatible. The architecture uses a single convolution + pooling
+# stage and a small MLP head. The architectural choices below are meant to be:
+#
+# - **standard**: 3x3 convolutions and 2x2 max-pooling are very common
+# - **small**: the dataset and images are tiny, so we keep the model tiny too
+#
+# If you want more background on CNN building blocks and how convolution/pooling
+# changes tensor shapes, see the CS231n notes:
+# https://cs231n.github.io/convolutional-networks/
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+
+
+class TinyCNN(nn.Module):
+    def __init__(self, conv_channels: int = 8, hidden_units: int = 32):
+        super().__init__()
+        self.conv_channels = conv_channels
+        self.hidden_units = hidden_units
+
+        # 2-level CNN with 2x2 max-pooling
+        self.conv1 = nn.Conv2d(
+            in_channels=1, out_channels=conv_channels, kernel_size=3, padding=1
+        )
+        self.conv2 = nn.Conv2d(conv_channels, conv_channels, kernel_size=3, padding=1)
+        self.pool = nn.MaxPool2d(kernel_size=2)
+
+        # input shape = (8,8) -> conv1: (8,8) -> conv2: (8,8) -> pool: (4,4)
+        image_shape_after_conv = 4 * 4
+
+        # MLP head
+        self.fc1 = nn.Linear(conv_channels * image_shape_after_conv, hidden_units)
+        self.dropout = nn.Dropout(p=0.25)  # Regularization to avoid overfitting
+        self.fc2 = nn.Linear(hidden_units, 10)  # 10 digit classes (0..9)
+
+    def forward(self, x):
+        x = F.relu(self.conv1(x))
+        x = self.pool(F.relu(self.conv2(x)))
+        x = x.flatten(start_dim=1)
+        x = self.dropout(F.relu(self.fc1(x)))
+        return self.fc2(x)
+
+
+# %%
+# Skorch provides scikit-learn compatible wrappers around torch training loops.
+# That makes the torch model usable by skrub DataOps (and scikit-learn tools in
+# general).
+#
+# We use :func:`skrub.choose_from()` to define hyperparameters that the DataOps
+# grid search will tune: conv_channels, hidden_units, and max_epochs.
+# The other parameters are set to common choices for this task and training data size.
+
+from skorch import NeuralNetClassifier
+
+device = "cpu"  # use "cuda" or "mps" if available
+
+net = NeuralNetClassifier(
+    module=TinyCNN,
+    # These choices are intentionally small so the example runs quickly.
+    module__conv_channels=skrub.choose_from([8, 16], name="conv_channels"),
+    module__hidden_units=skrub.choose_from([8, 16, 32], name="hidden_units"),
+    max_epochs=skrub.choose_from([10, 15], name="max_epochs"),
+    optimizer__lr=0.01,
+    optimizer=optim.Adam,
+    criterion=nn.CrossEntropyLoss,
+    device=device,
+    train_split=None,  # We'll use skrub's grid search for validation
+    verbose=0,
+)
+
+
+# %%
+# Tuning the model's hyperparameters with DataOps
+# ===============================================
+#
+# We integrate the model into the DataOps plan. First, we
+# convert the target labels to integers for the loss computation
+# and apply the model to the preprocessed X and y.
+
+y_int = y.astype("int64")
+predictor = X_reshaped.skb.apply(net, y=y_int)
+predictor.skb.draw_graph()
+
+# %%
+# Finally, we use 4-fold cross-validation for the hyperparameter
+# tuning on our DataOps plan.
+
+from sklearn.model_selection import KFold
+
+cv = KFold(n_splits=4, shuffle=True, random_state=42)
+search = predictor.skb.make_grid_search(
+    cv=cv,
+    fitted=True,
+    n_jobs=-1,
+)
+print("\nSearch results:")
+print(search.results_.to_string(index=False))
+
+# %%
+# Let's take a better look at the well-performing models by looking
+# at the parallel coordinates plot. We filter to models with
+# score >= 0.94 to focus on the top-performing configurations.
+
+fig = search.plot_results(min_score=0.94)
+fig
+
+# %%
+# Interpreting the results
+# ========================
+#
+# Looking at the search results, we can observe several patterns:
+#
+# - **Model capacity matters**: Larger configurations with ``conv_channels=16``
+#   and ``hidden_units=32`` tend to perform best. Smaller models with
+#   ``conv_channels=8`` and/or ``hidden_units=8`` perform significantly worse,
+#   indicating that the task benefits from increased model capacity.
+# - **More epochs generally help**: Configurations with ``max_epochs=15`` tend to
+#   perform slightly better than those with ``max_epochs=10``, though the gains
+#   are modest compared to architectural changes.
+
+# %%
+# Conclusion
+# ==========
+#
+# In this example, we've shown how to use **PyTorch** and **skorch** within
+# skrub DataOps. The key steps were:
+#
+# 1. Define a PyTorch ``nn.Module`` (our ``TinyCNN``)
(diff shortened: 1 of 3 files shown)


e-strauss:
the example runtimes locally (on cpu) were:

```sh
pixi run -e doc python examples/data_ops/1151_pytorch.py
Preview:
  epoch    train_loss    valid_acc    valid_loss     dur
-------  ------------  -----------  ------------  ------
      1        2.2811       0.3500        2.1826  0.1035
      2        2.0903       0.4833        1.9407  0.0808
      3        1.7689       0.4833        1.5819  0.0810
      
Learner train:
  epoch    train_loss    valid_acc    valid_loss      dur
-------  ------------  -----------  ------------  -------
      1        0.2753       0.9712        0.0897  14.2956
      2        0.0618       0.9807        0.0624  14.3741
      3        0.0429       0.9848        0.0503  13.9364
```
      
is that a problem for the CI?

rcap107:
Hi @e-strauss, thanks for the PR, this will be useful.

You can check the runtime of each example in the rendered docs. This one took about 1min30s to run, which is a bit long but still manageable. 

To merge this PR you will need to merge main and pay some attention to the lock file because we changed some of the dependencies. I suggest you delete your local lock file before merging, then create a new one after the merge is done.

e-strauss:
@rcap107 okk, got it. I reduced the train size, and it reduced the overall runtime to 48s.

I rebased on main and squashed the commits. The lock is updated based on the new doc dep

rcap107:
Hi @e-strauss, I looked at the PR and I think it'll take some more work before it can be merged. The main problem is that it lacks almost all the necessary detail for people to understand what the code is doing. 

As a rule of thumb, we want to minimize the number of lines of code and the complexity of the example, so that every line is either there for a reason, or it is removed. Then, every line of code should also be explained. 

For example, the `Net` module is not explained at all. How did you decide the numbers? What are they based on? If they're the "typical parameters", add a reference to, for example, a torch example that explains them. 

The intent of the example is showing how torch can be used with skrub, and here most of the attention of the user will be spent trying to understand the why and the what of the parameters of the network. I know because I'm not familiar with torch, and that's exactly what I had to do to to understand what is going on here. 

Given that there are so many parameters, it would be useful to show how they play with the choose functions, so I'd add at least one or two choices to show that it is possible to use them with skorch objects (and if they can't, then it's something we need to fix).

I was also working on a similar image classification example that does not use torch in #1778. If you find it useful, you can recycle most of the narrative from there. 

I am also quite concerned that the memory usage is very high. It may cause issues with the example running on the CI. I'm not sure if that part can be optimized though.

e-strauss:
@rcap107 I rewrote the script to use the scikit-learn digits dataset, following an approach similar to your example. I also simplified the network significantly, which is sufficient given that the images are only 8×8. This change reduced the memory usage to 668 MB.

I added explanations for the different components of the neural network, along with a pointer to further information on CNNs.

I really like the idea of adding choices—it turns the example into a Neural Architecture Search use case built on skrub’s DataOps.

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+X = skrub.X(X)
+y = skrub.y(y)
+
+# %%
+# Data preprocessing
+# ==================
+#
+# A convolutional network expects images with shape (N, C, H, W) where:
+# - N: number of samples
+# - C: number of channels (1 for grayscale)
```
rcap107:
```suggestion
# - C: number of color channels (1 for grayscale)
```

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+# - H, W: image height and width
+X_reshaped = X.reshape(-1, 1, 8, 8).astype("float32")
```
rcap107: Can you add a small line explaining that the -1 means that the final dimension is inferred?

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+
+# The digit pixels are in [0, 16]. Normalizing to [0, 1] makes optimization
+# easier and is standard in vision models.
+X_reshaped /= X_reshaped.max()
```
rcap107: We want people to always create new data ops rather than reassigning variables

```suggestion
X_scaled = X_reshaped / X_reshaped.max()
```

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+    train_split=None,
+    verbose=0,
+)
+
+
+# %%
+# Integrating the model into the Data Ops plan
+# ===================================================
+# Finally, we convert the target labels to integers for the loss
+# computation andintegrate the model into the Data Ops plan by applying
```
rcap107:
```suggestion
# computation and integrate the model into the Data Ops plan by applying
```

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+# it to the variables X and y.
+
+y_int = y.astype("int64")
+predictor = X_reshaped.skb.apply(net, y=y_int)
```
rcap107:
```suggestion
predictor = X_scaled.skb.apply(net, y=y_int)
```

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+
+- **PyTorch** defines the model (an ``nn.Module``)
+- **skorch** wraps it as a scikit-learn compatible estimator
+- **skrub Data Ops** builds a plan and can tune skorch (and therefore PyTorch)
+  hyperparameters using the skrub choices.
+"""
+
+# %%
+# Reproducibility
+# ===============
```
rcap107: Using `===` as the header separator is setting the section formatting to having all top-level titles. 
```suggestion
# ---------------
```

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+import random
+
+import numpy as np
+
+SEED = 42
+random.seed(SEED)
+np.random.seed(SEED)
+
+# %%
+# Loading the data
```
rcap107, line 20: I think it's better if the first three sections are combined in a single set up section that sets the random seeds, loads the data and creates the variables

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+import skrub
+
+X = skrub.X(X)
+y = skrub.y(y)
+
+# %%
+# Data preprocessing
+# ==================
+#
+# A convolutional network expects images with shape (N, C, H, W) where:
```
rcap107, line 54: There needs to be a line break for lists to be rendered properly (rst things)

```suggestion
# A convolutional network expects images with shape (N, C, H, W) where:

```

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+# ===============================
+#
+# skorch provides scikit-learn compatible wrappers around torch training loops.
+# That makes the torch model usable by skrub Data Ops (and scikit-learn tools in
+# general).
+from skorch import NeuralNetClassifier
+
+# Pick a device that exists on the current machine.
+device = "cpu"  # "cuda" or "mps"
+
```
rcap107, line 132: There should be a mention to the skrub choose functions in the text to make them more evident. Can you also add a line like "the other parameters are set to common choices for this task", or remove them if they're already the defaults? (again to keep the number of lines people need to read as small as possible)

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+# - C: number of channels (1 for grayscale)
+# - H, W: image height and width
+X_reshaped = X.reshape(-1, 1, 8, 8).astype("float32")
+
+# The digit pixels are in [0, 16]. Normalizing to [0, 1] makes optimization
+# easier and is standard in vision models.
+X_reshaped /= X_reshaped.max()
+
+# %%
+# Building a tiny CNN in PyTorch
```
rcap107: I would also combine this and the next section into a "building a nn classifier" section (or something along those lines)

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+    iterator_train__shuffle=False,
+    # We disable skorch's internal validation split because we'll evaluate with
+    # cross-validation in the Data Ops plan.
+    train_split=None,
+    verbose=0,
+)
+
+
+# %%
+# Integrating the model into the Data Ops plan
```
rcap107: This and the next section can be combined into something like "tuning the model's hyperparameters with the data ops"

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,193 @@
...
+# The grid search results table shows one row per hyperparameter configuration.
+#
+# - **Bigger models tend to do better**: increasing ``conv_channels`` and/or
+#   ``hidden_units`` increases the model capacity, which helps on digits.
+# - **Tiny hidden_units value is a bottleneck**: when ``hidden_units=1``, the
+#   network must compress the whole image into a single number before the
+#   final 10-way classifier. In practice this often collapses to near-chance
+#   performance (\(\approx 0.10\) accuracy for 10 classes).
+# - **More epochs can help**: higher ``max_epochs`` can improve accuracy, at
+#   the cost of longer training time, but it can also overfit the training data.
```
rcap107: There should be a small section with conclusions ("in this example we've shown how to use torch and skorch etc."), as well as some crossreferences to other examples. The most relevant are probably the example on optuna and the example on choices

rcap107 (commented):
Hi @e-strauss, thanks a lot for the improvements, the example is much better now! 
I left some more comments on formatting and the structure of the example. I think the level of detail is quite good now, it's just a matter of cosmetic changes. 

And thanks for reducing the memory usage, that's very important.

e-strauss:
I resolved the comments, simplified the code slightly and added a Conclusion section. lmk what you think bout it

GaelVaroquaux:
Looks really nice. Thanks a lot. 

Could we add a small plot, to give a visual message to the example. For instance a parallel coordinate plot using DataOps?

examples/data_ops/1151_pytorch.py
```diff
@@ -0,0 +1,215 @@
...
+# - C: number of color channels (1 for grayscale)
+# - H, W: image height and width
+#
+# We reshape the images and normalize the pixel values to [0, 1].
+# The -1 means the first dimension (N) is inferred automatically
+# from the array size.
+
+# %%
+X_reshaped = X.reshape(-1, 1, 8, 8).astype("float32")
+X_scaled = X_reshaped / X_reshaped.max()
```
jeromedockes: we probably want:

```suggestion
X_scaled = X_reshaped / X_reshaped.max().skb.freeze_after_fit()
```

or some equivalent change so that `max()` is computed once on the train set and is not recomputed on the test set. for example we could have a scaler transformer (scikit-learn MinMaxScaler is not adapted as it is for 2d inputs + does the rescaling per-feature not globally)

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,215 @@
...
+# - H, W: image height and width
+#
+# We reshape the images and normalize the pixel values to [0, 1].
+# The -1 means the first dimension (N) is inferred automatically
+# from the array size.
+
+# %%
+X_reshaped = X.reshape(-1, 1, 8, 8).astype("float32")
+X_scaled = X_reshaped / X_reshaped.max()
+
```
jeromedockes, line 71:
```suggestion
X_scaled

```

this will show X_scaled as the output of the cell, with the preview and a drawing of the graph so far. we may want to point out here that the reshaping and scaling is recorded as part of our pipeline, another benefit of using the DataOps in addition to the hyperparameter tuning mentioned later

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,215 @@
+"""
```
jeromedockes, line 1: could we name it `1160` instead of `1151`? this will make it easier to insert other examples in-between without renaming the file later (unfortunately documentation page URLs reflect the source file names so renaming the python file breaks links to the online example). thanks!

jeromedockes (commented):
thank you very much @e-strauss this is a very neat addition!

examples/data_ops/1151_pytorch.py
GaelVaroquaux: Add a comment explaining what this does. It's subtle

e-strauss: ah thanks good catch

jeromedockes: Indeed. Usually whenever something needs to be fitted I use an estimator. For reference here it would look something like:

```python
class GlobalMaxScaler(TransformerMixin, BaseEstimator):
    def fit(self, X, y=None):
        self.max_ = X.max()

    def transform(self, X):
        return X / self.max_

X_scaled = X_reshaped.skb.apply(GlobalMaxScaler())
```

but it is more verbose in this case

e-strauss: ya, I agree, I kept the `.skb.freeze` for simplicity and added clarification in the text section above

examples/data_ops/1160_pytorch.py
e-strauss: done, I used `X_scaled[0,0]` to get a usable output (otherwise you get 4d tensor). To avoid the GetItem in the displayed plan, I tried to use `X_scaled.skb.draw_graph()` but it seemed to distort the code section in the rendered html

e-strauss: nvm, must have been something else, it renders fine now.
i will switch to `X_scaled.skb.draw_graph()`, i thinks its a bit more tidy

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,217 @@
...
+y = skrub.y(y)
+
+# %%
+# Data preprocessing
+# ==================
+#
+# We start by normalizing the pixel values to [0, 1] by first
+# computing the global max value and then dividing the pixel values
+# by this max value. Importantly, we freeze the max value after fitting to
+# avoid data leakage during prediction.
```
jeromedockes: This is not so much about leakage but rather about remembering the scaling we used when fitting the model and re-applying the same transformation, so maybe something like "we freeze the max value (scaling factor) after fitting so that the same rescaling is applied later when we use our dataop for prediction on new (test) data" -- WDYT?

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,217 @@
...
+        x = self.dropout(F.relu(self.fc1(x)))
+        return self.fc2(x)
+
+
+# %%
+# Skorch provides scikit-learn compatible wrappers around torch training loops.
+# That makes the torch model usable by skrub DataOps (and scikit-learn tools in
+# general).
+#
+# We use **skrub.choose_from()** to define hyperparameters that the DataOps
```
jeromedockes:
```suggestion
# We use :func:`skrub.choose_from()` to define hyperparameters that the DataOps
```

e-strauss: ah nice, i added the link there

examples/data_ops/1160_pytorch.py
```diff
@@ -0,0 +1,217 @@
...
+# %%
+# Conclusion
+# ==========
+#
+# In this example, we've shown how to use **PyTorch** and **skorch** within
+# skrub DataOps. The key steps were:
+#
+# 1. Define a PyTorch ``nn.Module`` (our ``TinyCNN``)
+# 2. Wrap it with skorch's ``NeuralNetClassifier`` to make it scikit-learn compatible
+# 3. Use ``skrub.choose_from()`` to specify hyperparameters for tuning
```
e-strauss: and here

examples/data_ops/1160_pytorch.py
e-strauss: yes, you're right, i was thinking bout it in the wrong way. sounds good to me :)

GaelVaroquaux:
Great!
I wanted to approve and merge, but the PR is still marked as draft.
@e-strauss , do you want to mark it as ready?

rcap107:
> Great! I wanted to approve and merge, but the PR is still marked as draft. @e-strauss , do you want to mark it as ready?

I was the one who set this to draft, but I think all the comments have been addressed so it can be merged

e-strauss:
thanks for the review, ready to merge from my side :)

GaelVaroquaux:
Thanks, merged!

GaelVaroquaux:
Nice, by the way, this is very cool!!

PR #1881, opened 2026-02-02
ENH - Extend the `has_nulls` selector so that it can take a threshold of null values 

### LINKED ISSUE #1736: FEAT - Extend the `has_nulls` selector so it can take a fraction of null values
At the moment, the `has_nulls` selector selects only columns that have at least one null value. It would be useful to add a threshold (similar to that used in `DropUninformative`) to select all columns that have a number of null values larger than the given threshold. 

This issue is related to #1723

gabrielapgomezji:
Fixes #1736


skrub/selectors/_selectors.py
```diff
@@ -508,7 +509,32 @@ def has_nulls():
...
+    def null_count_check(column, threshold=None):
+        return sum(sbd.is_null(column)) / len(column) >= threshold
+
+    if threshold == 0:
+        return Filter(sbd.has_nulls, name="has_nulls")
+    else:
+        if not isinstance(threshold, numbers.Number) or not 0.0 < threshold <= 1.0:
+            raise ValueError(
+                f"Threshold {threshold} is invalid. Threshold"
+                " should be a number in the range [0, 1], or None."
```
rcap107:
```suggestion
                " should be a number in the range (0, 1], or None."
```

skrub/selectors/tests/test_selectors.py
```diff
@@ -79,13 +78,6 @@ def test_dtype_selectors(df_module):
         assert s.any_date().expand(df) == ["datetime-col"]
 
 
-def test_dtype_pandas_object():
```
rcap107, line 82: Why was this test removed?

CHANGES.rst
```diff
@@ -19,6 +19,9 @@ Changes
   :pr:`1819` by :user:`Eloi Massoulié <emassoulie>`
 - :func:`compute_ngram_distance` has been renamed to :func:`_compute_ngram_distance` and is now a private function.
   :pr:`1838` by :user:`Siddharth Baleja <siddharthbaleja>`.
+- :func:`selectors.has_nulls` threshold parameter was added to drop columns with a
+  percentage of null values above the given threshold.
+  :pr:`1881` by :user:`Gabriela Gómez Jiménez <gabrielapgomezji>`.
```
```suggestion
- :func:`selectors.has_nulls` now takes a ``threshold`` parameter, which allows selecting columns that have a fraction of null values above the given threshold.
  :pr:`1881` by :user:`Gabriela Gómez Jiménez <gabrielapgomezji>`.
```

rcap107: this will likely break the formatting

doc/modules/multi_column_operations/advanced_selectors.rst
```diff
@@ -84,3 +84,114 @@ and checks if any data points extend further than 2 IQRs of the lower and upper
 2   14   31
 3   15   32
 4  100  300
+
+
+Select columns with null values
+--------------------------------
+Selectors :func:`has_nulls` and :func:`DropUninformative` can be used to get information
```
rcap107: Rather than mentioning explicitly DropUninformative, you can add a reference to the user guide page for that object, and simplify this section so it's only talking about has_nulls.

doc/modules/multi_column_operations/advanced_selectors.rst
```diff
@@ -84,3 +84,114 @@ and checks if any data points extend further than 2 IQRs of the lower and upper
...
+>>> cleaned_df
+   patient_id   age treatment
+0           1  25.0     med_A
+1           2  30.0     med_B
+2           3   ...     med_C
+3           4  45.0     med_D
+4           5  50.0     med_E
+5           6   ...     med_F
+6           7  60.0     med_G
+7           8  65.0     med_H
```
rcap107: Remove this example in favor of a link to the dropuninformative page

```suggestion
```

skrub/selectors/_selectors.py
```diff
@@ -175,7 +176,7 @@ def numeric():
     i8          int8
     bool_       bool
     Bool_    boolean
-    str_     ...
```
rcap107, line 179: removing this ellipsis will fail doctests because different pandas versions output different things

skrub/selectors/tests/test_selectors.py
```diff
@@ -106,6 +98,16 @@ def test_has_nulls(df_module):
...
 
 
+def test_has_nulls_threshold(df_module):
+    df = df_module.make_dataframe(
+        dict(a=[0, 1, 2, None], b=[0, None, 2, None], c=["a", None, None, None])
+    )
+    assert s.has_nulls(threshold=0.25).expand(df) == ["a", "b", "c"]
+    assert s.has_nulls(threshold=0.5).expand(df) == ["b", "c"]
+    assert s.has_nulls(threshold=0.75).expand(df) == ["c"]
+    assert s.has_nulls(threshold=1.0).expand(df) == []
```
rcap107: there should also be a test for threshold=0

rcap107 (changes requested):
Hi @gabrielapgomezji, thanks for the PR. I left a few comments that should be handled before we proceed

skrub/selectors/_selectors.py
```diff
@@ -508,7 +509,32 @@ def has_nulls():
...
+    >>> s.select(df2, s.has_nulls(threshold=0.25))
+       few_nulls  many_nulls
+    0        1.0         1.0
+    1        2.0         ...
+    2        3.0         ...
+    3        ...         ...
     """
-    return Filter(sbd.has_nulls, name="has_nulls")
+
+    def null_count_check(column, threshold=None):
```
rcap107: This function should be moved outside of `has_nulls` because it's not needed here and can cause issues with pickling

skrub/selectors/_selectors.py
```diff
@@ -503,7 +504,11 @@ def cardinality_below(threshold):
     return Filter(_cardinality_below, args=(threshold,), name="cardinality_below")
 
 
-def has_nulls():
+def null_count_check(column, threshold=None):
```
jeromedockes: can we rename to `_null_count_check` (or maybe `_null_proportion_above`?) -- starting with an underscore indicates it is private and not supposed to be useful outside of this module.

Also, there should not be a default value for threshold here, as a value is always validated and passed by the caller (and anyway the None default value would cause an error)

skrub/selectors/_selectors.py
```diff
@@ -503,7 +504,11 @@ def cardinality_below(threshold):
...
+    return sum(sbd.is_null(column)) / len(column) >= threshold
```
jeromedockes: not sure if it's important but this is slightly inconsistent with drop_uniformative which does a strict comparison and treats 1 as a special case, whereas here we do an inclusive comparison and treat 0 as a special case 🤔

skrub/selectors/_selectors.py
```diff
@@ -503,7 +504,11 @@ def cardinality_below(threshold):
...
+
+
+def has_nulls(threshold=0):
```
jeromedockes:
```suggestion
def has_nulls(threshold=0.0):
```

if only proportions are accepted the default should be a float

skrub/selectors/_selectors.py
```diff
@@ -516,7 +521,29 @@ def has_nulls():
     >>> s.select(df, s.has_nulls())
           b     c
     0   0.0     a
-    1   NaN     b
+    1   ...     b
     2  20.0  ...
+
+    Use the ``threshold`` parameter to filter columns by null percentage:
```
jeromedockes: when reading 'percentage' I am not sure if I should pass 25 or 0.25 if I mean 25%. maybe 'proportion'?

skrub/selectors/_selectors.py
```diff
@@ -516,7 +521,29 @@ def has_nulls():
...
+
+    >>> df2 = pd.DataFrame(dict(
+    ...     few_nulls=[1, 2, 3, None],
+    ...     many_nulls=[1, None, None, None],
+    ...     no_nulls=[1, 2, 3, 4]))
+    >>> s.select(df2, s.has_nulls(threshold=0.25))
```
jeromedockes: perhaps use a threshold of 0.5 for the example so we see one of the columns with nulls selected and the other not?

skrub/selectors/_selectors.py
```diff
@@ -516,7 +521,29 @@ def has_nulls():
...
     """
-    return Filter(sbd.has_nulls, name="has_nulls")
+
+    if threshold == 0:
+        return Filter(sbd.has_nulls, name="has_nulls")
+    else:
+        if not isinstance(threshold, numbers.Number) or not 0.0 < threshold <= 1.0:
+            raise ValueError(
+                f"Threshold {threshold} is invalid. Threshold"
+                " should be a number in the range (0, 1], or None."
```
jeromedockes:
```suggestion
                " should be a number in the range [0.0, 1.0]."
```

0.0 is accepted, and it doesn't seem like None is

PR #1890, opened 2026-02-06
DataOps performance improvements

jeromedockes:
This improves the speed of DataOps when they become rather large (in number of nodes) following feedback from @e-strauss @phaniarnab @sscdotopen @rcap107 

- some checks are done whenever a DataOp is created (ie a node is added to the graph), and some of those checks require traversing the whole graph -> quadratic complexity in the number of nodes, which can become noticeable when we have very complex pipelines. This PR allows through the skrub config (and its context manager) to disable those eager checks and previews. they are instead performed when using the final dataop eg by calling eval() or make_learner(). the behavior is unchanged unless the config option is used
- when we evaluate a dataop intermediate results are cached until we know they are not needed anymore. after each step, results that are not needed anymore are discarded. this PR changes the strategy used to find which results can be discarded, by using reference counts instead of tracing from the final results to identify nodes that are not reachable anymore without going through already-available results. again the benefit is to reduce complexity and speed up larger dataops. this part is not an option but an implementation detail change with no visible effects for users (except the speedup)

this results in visible improvements on real examples reported by @phaniarnab and @e-strauss (and it is easy to construct pathological cases that make the speedup arbitrarily large, due to the change in asymptotic complexity)

### DIFF
--- a/doc/modules/data_ops/basics/using_previews.rst
+++ b/doc/modules/data_ops/basics/using_previews.rst
@@ -50,8 +50,8 @@ Disabling previews and eager checks
 -----------------------------------
 
 By default, as soon as a DataOp is defined, some validity checks are performed
-and the preview results are computed eagerly. In very complex Data Ops plans
-(100+ nodes), running checks after adding each node adds noticeable overhead.
+and the preview results are computed eagerly. In very complex DataOps plans
+(100+ nodes), running checks after adding each node can cause a noticeable overhead.
 To avoid this, it is possible to disable eager checks with the ``"eager_data_ops"``
 is easily achieved with the ``"eager_data_ops"`` :ref:`configuration
 <user_guide_configuration_parameters>` option.
--- a/skrub/_data_ops/_data_ops.py
+++ b/skrub/_data_ops/_data_ops.py
@@ -325,8 +325,8 @@ def checked_data_op_constructor(f=None, /, *, allow_skipping=True, eval_preview=
     are eagerly computed (to make them available and trigger possible errors)
     as soon as they are defined.
 
-    However, eager checks and preview computation may be disabled in some
-    situations (e.g. the config option eager_data_ops has the value False).
+    However, eager checks and preview computation may be disabled
+    via the eager_data_ops config option.
 
     We decorate all the functions that create DataOps rather than perform
     checks in ``__init__`` to make tracebacks as short as possible: the second
--- a/skrub/_data_ops/_skrub_namespace.py
+++ b/skrub/_data_ops/_skrub_namespace.py
@@ -92,6 +92,20 @@ def _check_grid_search_possible(data_op):
 
 
 def _check_before(f):
+    """
+    Decorator to perform validation of a DataOp before calling a function.
+
+    Usually some checks such as no duplicate names are performed whenever a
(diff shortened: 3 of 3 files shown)


doc/modules/data_ops/basics/using_previews.rst
```diff
@@ -45,3 +45,23 @@ one example dataset.
...
 It is not necessary to provide a value for every variable: it is however advisable
 to do so when possible, as it allows to catch errors early on.
+
+Disabling previews and eager checks
+-----------------------------------
+
+By default, as soon as a DataOp is defined, some validity checks are performed
+and the preview results are computed eagerly. In some cases we may want to
+disable this, for example if we are defining a very complex DataOp with over 100
+nodes and running checks after each node is added adds noticeable overhead. This
```
rcap107: Reworded a bit for clarity. This is definitely going to break the formatting. 

```suggestion
and the preview results are computed eagerly. In very complex Data Ops plans
(100+ nodes), running checks after adding each node adds noticeable overhead.
To avoid this, it is possible to disable eager checks with the ``"eager_data_ops"``
```

skrub/_data_ops/_data_ops.py
```diff
@@ -313,49 +317,116 @@ def _find_dataframe(data_op, func_name):
...
 
-    We must decorate all the functions that create DataOps rather than do
-    it in ``__init__`` to make tracebacks as short as possible: the second
+    This decorator must be applied to all functions that create a DataOp, so
+    that they are eagerly checked (e.g. for duplicate names) and their previews
+    are eagerly computed (to make them available and trigger possible errors)
+    as soon as they are defined.
+
+    However, eager checks and preview computation may be disabled in some
+    situations (e.g. the config option eager_data_ops has the value False).
```
rcap107:
```suggestion
    situations (e.g., the Data Ops graph is very large). 
    Checks can be disabled by setting the config option 
    eager_data_ops to False.
```

skrub/_data_ops/_data_ops.py
```diff
@@ -313,49 +317,116 @@ def _find_dataframe(data_op, func_name):
...
+    @checked_data_op_constructor(allow_skipping=False)
+    def make_a_data_op(a, b): ...
+
+    Parameters
+    ----------
+    f : function
+        The function to decorate. Passed implicitly when using the @decorator
+        syntax.
+
+    allow_skipping : bool, default = True
```
rcap107, line 356: I'm not sure I understand what this parameter is for. I get that it's there to disable skipping for functions like `eval`, but then `eval` and other functions like it get decorated with `@_check_before` instead. Doesn't that bypass `allow_skipping`? If that's the case, I don't understand why this is needed 🤔

rcap107 (approved):
Thanks for the PR @jeromedockes, I left a couple of comments on the docs. There's is also something that's not clear to me about a parameter, but aside from that I think this is good.

skrub/_data_ops/_data_ops.py
rcap107: I didn't notice that `allow_skipping` is set to False in check_before, it makes sense now

rcap107:

PR #1892, opened 2026-02-09
Adding shuffled version of the toxicity dataset

### LINKED ISSUE #1234: shuffling the toxicity dataset?

the dataset in [this example](https://skrub-data.org/stable/auto_examples/02_text_with_string_encoders.html#sphx-glr-auto-examples-02-text-with-string-encoders-py) is sorted by label: the first 500 tweets are toxic and the rest non-toxic

the example does not explicitly shuffle it, but it still works because `cross_validate` for classification uses a stratified k fold by default. However that is not immediately obvious, and I think there has been some discussion recently in scikit-learn that it might be better not to stratify by default. So I think it might be good to actually shuffle the dataset.

Not sure where would be the best place to do it -- in the hosted zip file, in the fetcher, in the example, or by passing shuffle=True to cross_validate

WDYT @Vincent-Maladiere 

_No response_

rcap107:
Closes #1234 

This PR changes adds a shuffled version of the toxicity dataset to the list of datasets used by the fetchers. The relative dataset has already been added to the dataset repository. 

TODO:
- [x] Upload dataset on figshare
- [x] Upload dataset on osf


jeromedockes (approved):
thanks @rcap107 !!

PR #1851, opened 2026-01-16
Exposing SingleColumnTransformer and RejectColumn

emassoulie:
Fixes https://github.com/skrub-data/skrub/issues/1730

Exposes classes from the newly-created `_single_column_transformer.py`, and moves them to the `skrub.core` package.

PR #1900, opened 2026-02-12
Detect single-column transformers when they are the first step of a Pipeline

jeromedockes:
when we inspect an estimator to find out if it is a single-column transformer, special-case Pipelines and look at the first step

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -27,7 +27,8 @@ Bug Fixes
 --------
 - The :class:`TableVectorizer` now correctly handles the case where one of the
   provided encoders is a scikit-learn Pipeline that starts with a skrub
-  single-column transformer. :pr:`1899` by :user:`Jérôme Dockès <jeromedockes>`.
+  single-column transformer. :pr:`1899` by :user:`Jérôme Dockès <jeromedockes>`
+  and :pr:`1900` by :user:`Jérôme Dockès <jeromedockes>`.
 
 Release 0.7.2
 =============


skrub/_single_column_transformer.py
```diff
@@ -209,3 +210,36 @@ def _insert_after_first_paragraph(document, text_to_insert):
...
     return "".join(output_lines)
+
+
+def is_single_column_transformer(transformer):
+    """
+    Check if the provided transformer is a single-column transformer.
+
+    This is done by checking the special attribute
+    __single_column_transformer__ (thus inheriting from the
+    SingleColumnTransformer class is not mandatory). We treat scikit-learn
```
rcap107, line 247: In what situations would it be better to just set __single_column_transformer__ rather than inheriting from the class?

jeromedockes: for example you don't want the checks and default methods provided by the base class, you're a third-party module that wants to interoperate with skrub but not depend on it, ... there can be a few reasons so a kind of "protocol" like this attribute can be less cumbersome than imposing inheritance. in any case that is already what is in place so not really related to this PR

rcap107: Makes sense. I wonder if this should be explained somewhere in the documentation (for very advanced users).

jeromedockes: probably -- we can polish the SingleColumnTransformer docs in #1851

skrub/_single_column_transformer.py
```diff
@@ -209,3 +210,36 @@ def _insert_after_first_paragraph(document, text_to_insert):
...
+    -------
+    bool
+        Whether the transformer is a single-column transformer, meaning that it
+        should be passed columns rather than dataframes.
+    """
+    if hasattr(transformer, "__single_column_transformer__"):
+        return True
+    if isinstance(transformer, Pipeline):
+        try:
+            return is_single_column_transformer(transformer.steps[0][1])
```
rcap107, line 269: There should be a comment here to explain why the function is calling itself, and why the 1  `transformers.steps[0][1]` 

It's because each step is a tuple (name, transformer), but it's not clear from the code

PR #1901, opened 2026-02-12
DOC - Adding how-to guides

rcap107:
Tentative PR to have an idea of how the homepage and docs would look like if we split the user guide in two: the current user guide would be more discoursive and be more of an introduction and explanation of what skrub is about, while the how-to guides are instead very to-the-point and attempt to explain how to address a single specific operation.

This is a part of the ongoing effort to implement the doc framework explained in https://diataxis.fr

Besides having an idea of where to put the content for the how-to guides, I'll also move some of the material from the current user guide to the how-to section. I'm not planning to do everything in this PR, rather I want to get an idea of what the user guide may look like.

### DIFF
--- a/doc/_templates/index.html
+++ b/doc/_templates/index.html
@@ -47,7 +47,7 @@ still full-width (unlike the article) #}
       We directly connect database tables to machine learning.
     </p>
     <div class="text-center mt-4">
-        <a class="btn sk-btn-orange mb-1 d-inline-block" href="{{ pathto('auto_examples/0000_getting_started') }}" role="button">Get started →</a>
+        <a class="btn sk-btn-orange mb-1 d-inline-block" href="{{ pathto('auto_tutorials/0000_getting_started') }}" role="button">Get started →</a>
     </div>
   </div>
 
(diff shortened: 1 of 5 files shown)


rcap107:
From IRL meeting: 

We should merge this PR relatively quickly, but before doing that I need to move some more material from the current user guide to the how-to guide so that it does not look quite as barren as it does now

rcap107:
I shuffled around some material and I added a collapsible admonition. I think we should merge this PR relatively soon and iterate over it, rather than do a bunch of work here and have conflicts down the line.

lisaleemcb:
So nice to have a How to section! What would you think about adding subcategories so that users can easily find the topics relating to a specific features? For example, I'm thinking the first three links would go under a Table Report header.

rcap107:
> So nice to have a How to section! What would you think about adding subcategories so that users can easily find the topics relating to a specific features? For example, I'm thinking the first three links would go under a Table Report header.

I completely agree, this pr was mostly a proof of concept to see what a how-to section would look like, I moved only the bare minimum there to not have a completely empty page. My idea was to migrate iteratively more material to the how-to index, and set proper section names as part of that process.

lisaleemcb:
Ah, make sense. That's a sensible approach, and I think this is a great start!

PR #1902, opened 2026-02-12
DOC- Let Copilot suggest documentation improvements

DimitriPapadopoulos:
This is the result of asking Copilot to improve the documentation
```
Please fix the documentation, starting with small issues: grammar, typos, format, clarity. 
```

I had to revert a few Copilot suggestions, but overall they made sense. This is a manual PR, but obviously this could be automated using a GitHub workflow. Probably no need to run the workflow on every PR, rather once every N months or before minor/major releases.

### DIFF
--- a/doc/vision.rst
+++ b/doc/vision.rst
@@ -1,27 +1,27 @@
 ===============================
-Vision: where is skrub heading?
+Vision: Where is skrub heading?
 ===============================
 
 .. currentmodule:: skrub
 
-Vision statement
+Vision Statement
 ================
 
 The goal of skrub is to facilitate machine learning on tables:
 `pandas <https://pandas.pydata.org>`__
-and `polars <https://pola.rs>`__ dataframes, SQL databases...
+and `polars <https://pola.rs>`__ dataframes, SQL databases, and more.
 
 |
 
-Skrub is high-level, with a philosophy and an API matching that of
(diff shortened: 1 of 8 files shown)


doc/vision.rst
```diff
@@ -1,27 +1,27 @@
...
+and `polars <https://pola.rs>`__ dataframes, SQL databases, and more.
 
 |
 
-Skrub is high-level, with a philosophy and an API matching that of
+Skrub is high-level, with a philosophy and API matching that of
 `scikit-learn <http://scikit-learn.org>`_. It strives to bridge the world
-of databases to that of machine-learning, **enabling imperfect assembly and
-representations of the data when it is noisy**, using the downstream
-target to predict to guide assembly when possible (supervised learning for
```
DimitriPapadopoulos, line 20: I hope Copilot and myself are not misunderstanding this sentence when removing `to predict`.

MarieSacksick (approved):
Thanks @DimitriPapadopoulos.  
I see no reason to object this PR, but I find no change interesting enough to create a bot to check regularly (whatever the frequency).

Actually, I like these little documentation changes for good first issues and encourage new contributors. They could be dealt with faster and more efficiently, but it removes a step in the contribution stairs and makes the introduction more difficult and impressive for new comers. Since they are not bugs and do not create misunderstanding for users, I'm very fine keeping that kind of things slow.

PR #1903, opened 2026-02-12
MAINT - Dependency Groups in pyproject.toml

DimitriPapadopoulos:
Compliance with [PEP 735](https://peps.python.org/pep-0735/).


rcap107:
Thanks for the PR 👍

PR #1904, opened 2026-02-12
MAINT - Bump Ruff and get rid of Black

DimitriPapadopoulos:
31320e2 replaced Black with Ruff.


rcap107 (approved):
Thanks for the cleanup

PR #1907, opened 2026-02-14
ENH: Be more explicit about max_*_columns in TableReport

### LINKED ISSUE #1721: Be more explicit about the max_*_columns in the TableReport
During the sprint yesterday, with @JulietteBgl we found that the attributes max_*_columns were not clear. In particular, the attribute max_plot_columns. Indeed, as the attribute is an int, we expect to have at least this number of columns plotted, while actually it behaves as a bool: plot or do not plot.  
I understand the necessity of having a maximum though, to avoid unexpected computational slowness.  

**plotting**  

I see two possible solutions here.  
1. keep the API and the configuration as is, but display the first N = max_plot_columns in the table report, with a message saying that not all columns have been plotted (if relevant) to avoid misleading
2. keep the configuration settings, to still have a max, but change the attribute from `max_plot_columns (int)` to `plot_columns (bool)`. There would be no default: the columns display if nb_columns < config.max_plot_columns. If set to True, it displays everything no matter the number of columns, if set to False, it plots nothing no matter the number of columns.

I personally like option 1 best, as I write option 2 I see that it takes longer to explain --> it's confusing.  

**association**

This one is less obvious, because it doesn't make sense to compute the association on the N first columns. This is where the option 2 above makes more sense, to have something similar between plotting and association.  

Related to https://github.com/skrub-data/skrub/issues/1523

JulietteBgl:
Closes #1721 

This PR:
- replaces the old `max_plot_columns` and `max_association_columns` by `columns_threshold`
- creates 2 new arguments (bool): `compute_association` and `plot_distribution`


jeromedockes:
Thank you very much for working on this @JulietteBgl  :tada: !

can I suggest some small tweaks to the options:

- keeping 2 thresholds for plots and associations. the plots take linear time and little memory, whereas associations have quadratic time and space requirements so one may want to set a much lower threshold for the associations. also I can see someone generally not caring about either of them while wanting the other.
- not having the compute_associations and plot_distributions in the config. they are redundant with the thresholds (you can set the max to 0). what I had understood from the issue is that the TableReport has options True, False or maybe, and the configuration allows tweaking the threshold that controls the "maybe"
- in the TableReport, the default should not be True or False but something like "auto". True always plots, False never does, and the default "auto" plots only if the size is smaller that the threshold (which can be configured)

ie something along the lines of

```
if isinstance(plot_distributions, str) and plot_distributions == "auto":
    do_plots = df.shape[1] < get_config()["max_plot_columns"]
elif isinstance(plot_distributions, Integral):
    do_plots = bool(Integral)
else:
    raise ValueError("...")
```

WDYT?

JulietteBgl:
Thank you so much for your feedback, @jeromedockes ! 
It looks like I didn't understood the initial logic, I'll work on it again soon and let you know

rcap107:
Hi @JulietteBgl, thanks for the PR! 

I just wanted to remind you that there should be a changelog entry for this PR

jeromedockes:
thanks @JulietteBgl  :) I think what you have is not very far, if anything it should make things a bit easier

rcap107:
Hi @JulietteBgl, it seems like this PR hasn't been merged with upstream in a very long time. This means that there are a lot of conflicts, and in particular the TableReport has been altered in various places while you were working on this.

I'm addressing the conflicts manually. I'll do my best to avoid that, but I may end up breaking some stuff that you worked on. For the time being, please don't push any more commits. I'll let you know when it's safe to work on the PR again. Once that happens, remember to pull to update your local branch before going back to work. 

In general, it's very important to keep branches up to date with upstream, because otherwise conflicts become a huge headache to deal with. 

I should have noticed this earlier, sorry about that.

JulietteBgl:
Hi @rcap107, sorry about that! I started resolving the conflicts but got interrupted and didn't get a chance to push my commits. Thanks for taking care of it. I’ll run some final tests and checks, then move the PR to 'Ready for Review' once I’m done.

rcap107:
> Hi @rcap107, sorry about that! I started resolving the conflicts but got interrupted and didn't get a chance to push my commits. Thanks for taking care of it. I’ll run some final tests and checks, then move the PR to 'Ready for Review' once I’m done.

Conflicts were getting a bit out of hand so I thought it'd be easier if I dealt with them, also because the TableReport has been changed in the meantime as well. I think now the workspace should be cleaner for you to continue.

Thanks for the work, by the way

JulietteBgl:
The new version of the code now does the following to improve the clarity of TableReport:

- Introduces explicit parameters: `plot_distributions` and `compute_associations` parameters. These accept True, False, or "auto". If True: always plot/compute associations, if False: never plot/compute associations. "auto" compute associations only when the number of columns does not exceed  ``plot_threshold`` or ``associations_threshold``.
- Renames threshold parameters to `plots_threshold` and `associations_threshold`.

jeromedockes:
thanks a lot @JulietteBgl ! the new parameters or their behavior are great.

I think we should remove the old ones `plots_threshold` and `associations_threshold` from `TableReport`, so that we only have the boolean parameters in the TableReport class itself, and only the thresholds in the configuration. so when `plot_distributions='auto'` we directly look at the configuration; there is no `plots_threshold` parameter that can override the configured threshold. Does this make sense?

If you agree, we just need to decide when we look up the configuration to resolve "auto" into a boolean: during the `__init__` of the report, or when we actually compute the summary. I guess the simplest, which is also consistent with the current behavior and with the handling of `verbose` is to do it in `__init__`, so maybe your `_validate_plot_and_association` function can also accept the dataframe (or its number of columns) and return 2 booleans `plot_distributions` and `compute_associations` instead of 4 values? The alternative is to do this validation and look up of the configuration in `_summarize`

regarding the configuration option names, given that the configuration has options related to diverse parts of the library and not only the table report, maybe we can make the names a bit longer to help users recognize what those options are about. as we have `table_report_verbosity` maybe it could be `table_reports_plots_threshold` instead of just `plots_threshold`? I agree that's a bit long though :thinking: 

apart from those 2 details I think we are almost done, thank you so much!

JulietteBgl:
Hello @jeromedockes,
Many thanks for your detailed feedback. I like the idea of resolving "auto" into a boolean earlier in the `init` method. Now the code should now reflect your remarks, let me know if I missed anything!

jeromedockes (approved):
LGTM, thank you very much @JulietteBgl ! this is a cool improvement, the TableReport is one of the most popular features and this makes the interface of the TableReport and the config much more reasonable :)

rcap107:
Hi @JulietteBgl, thanks a lot for this PR! It's already looking great

Since this is a breaking change (various parameters have been changed), we should have a deprecation cycle for the old parameters. This means modifying slightly the code so that the pre-change behavior is maintained if users use the old parameter names. 

Given it's a bit more work and more on the maintenance side, it could be done by us in a separate PR if you want to move on from this. Of course, if you're interested in working a bit more on this PR you're more than welcome to do that. 

What would you prefer?

JulietteBgl:
Hello @rcap107 , that's a good point. I can tackle this on Saturday the 25th. If it's urgent, do not hesitate to take it over before then!

rcap107:
> Hello @rcap107 , that's a good point. I can tackle this on Saturday the 25th. If it's urgent, do not hesitate to take it over before then!

Hi @JulietteBgl, we won't be releasing for a couple more weeks (at least) anyway, so feel free to work on it then. Thanks a lot!

JulietteBgl:
Hello @rcap107 , the old parameter names are now kept with warnings to avoid breaking existing users!

jeromedockes (approved):
thanks! 🚀

rcap107:
Thanks a lot @JulietteBgl for the PR!

MarieSacksick:
woop woop!! Congrats!

PR #1908, opened 2026-02-14
fix issue 1828 about a warning

### LINKED ISSUE #1828: Discussion - Remove useless warning?
### Problem Description

When using this simple piece of code, I have a warning:  

```
from skrub import tabular_pipeline

df, y = make_classification(random_state=42)
estimator = tabular_pipeline("classification")
estimator.fit(df, y)
```

The warning says:  
```
UserWarning: Only pandas and polars DataFrames are supported, but input is a Numpy array. Please convert Numpy arrays to DataFrames before passing them to skrub transformers. Converting to pandas DataFrame with columns ['0', '1', …].
  warnings.warn(
```

I don't really understand the point of the warning, because it doesn't seem to be a problem at all. 

### Feature Description

What about removing the warning since having a numpy array as input doesn't seem to be a problem, at least in that use case with only one column, and doesn't call for a user action?  
I'm also surprised by the fact that it seems to mention only the input y - why isn't df also a problem?

### Alternative Solutions

Be more explicit about the consequences of automatic conversion. Otherwise as a user, I'm very tempted to turn off warning notifications.

### Additional Context

_No response_

sandrineh:
According to the conversation about the issue, I propose to delete the warning message in the file called "-check_input".
This message brings confusion regarding the use of array for the tabular_pipeline function, whereas the conversion of the array to dataframe is carried out later in the code and does not prevent the code from executing correctly overall.


jeromedockes:
Thank you very much for working on this @sandrineh ! The change looks great; I think the only part left will be removing the "with pytest.warns ..." checks in [this test](https://github.com/skrub-data/skrub/blob/6ff71712b3fcc7601af07a435a7ee685e2a7d2d6/skrub/tests/test_check_input.py#L19) (just removing the context managers, but still keeping the statements that are in those blocks)

jeromedockes (approved):

jeromedockes:
thank you very much @sandrineh !! sorry it took so long for us to get back to this

PR #1910, opened 2026-02-16
[ENH]Refine Cleaner numeric_dtype modes

### LINKED ISSUE #1680: ENH - change the behavior of the `numeric_dtype` parameter of the `Cleaner`

Varshith-Yadav:
## Summary
This PR updates `Cleaner.numeric_dtype` to use explicit behavior modes and keeps `TableVectorizer` behavior unchanged.

`Cleaner.numeric_dtype` now accepts:
- `"passthrough"`: do not parse numeric-looking strings; keep numeric dtypes unchanged.
- `"parse"`: parse numeric-looking string columns to float32; keep numeric dtypes unchanged.
- `"float32"`: `"parse"` + convert numeric columns to float32.

## What changed
- Updated `Cleaner` logic and validation in `skrub/_table_vectorizer.py`.
- Added support in preprocessors to apply `ToFloat` to selected columns only (used for `"parse"` mode on string columns).
- Updated Cleaner docstring to clarify all modes.
- Added tests for all three modes and invalid values in `skrub/tests/test_table_vectorizer.py`.
- Added changelog entry in `CHANGES.rst`.

## Notes
- `TableVectorizer` behavior is intentionally unchanged.


rcap107:
Hi @Varshith-Yadav, I'll review the PR sometime during the week. 

The CI has failed for reasons that are unrelated to your PR, we'll deal with it. 

Something else I noticed is that the changelog is not correct: I think you might have forgotten to update your branch so that it is up to date with upstream/main, because the entry you added is in a section that has already been released. Another thing is that the changelog should include the number of your PR, like the other entries in the changelog. 

As I said, I'll do a proper review in the next few days. That said, thanks for opening this.

skrub/_table_vectorizer.py
```diff
@@ -356,20 +373,25 @@ def fit_transform(self, X, y=None):
...
             raise ValueError(
-                "`numeric_dtype` must be one of"
-                f"[`None`, `'float32'`]. Found {self.numeric_dtype}."
+                "`numeric_dtype` must be one of "
+                "[`'passthrough'`, `'parse'`, `'float32'`]. "
+                f"Found {self.numeric_dtype}."
             )
 
+        add_tofloat32 = self.numeric_dtype in ("parse", "float32")
+        tofloat_cols = s.string() if self.numeric_dtype == "parse" else None
```
rcap107: Rather than creating `tofloat_cols` here and passing it to _get_preprocessors, I think it would be simpler to just reuse `add_tofloat32` and define the selector directly in _get_preprocessors

In the way this is written, it's hard to follow what tofloat_cols is and what it does. Also, I think that the name tofloat_cols is a bit misleading, because this is a selector rather than a list of columns

skrub/_table_vectorizer.py
```diff
@@ -252,6 +263,13 @@ class Cleaner(TransformerMixin, BaseEstimator):
       actual datetimes with the correct dtype. If ``datetime_format`` is provided,
       it is forwarded to ``ToDatetime()``. Otherwise, the format is inferred.
 
+    - ``ToFloat()``: apply numeric parsing and conversion according to
+      ``numeric_dtype``:
+
+      - ``"passthrough"``: skipped.
+      - ``"parse"``: applied to string columns only.
+      - ``"float32"``: applied to all columns.
```
rcap107: There should be an example here to show the behavior of the three different parameters

skrub/_table_vectorizer.py
```diff
@@ -120,32 +120,37 @@ def _get_preprocessors(
...
+                drop_null_fraction=drop_null_fraction,
+                drop_if_constant=drop_if_constant,
+                drop_if_unique=drop_if_unique,
+            ),
+            cols,
         ),
-        ToDatetime(format=datetime_format),
+        (ToDatetime(format=datetime_format), cols),
     ]
     if add_tofloat32:
```
rcap107: This condition should check what is the value of add_tofloat32, which should be one of the three (passthrough, parse and float32), then the selector should be defined accordingly here, rather than it being passed from outside.

skrub/tests/test_table_vectorizer.py
```diff
@@ -446,6 +446,41 @@ def test_convert_float32(df_module):
...
 
+@pytest.mark.parametrize(
+    "numeric_dtype, parse_numeric_strings, convert_numeric_columns",
+    [
+        ("passthrough", False, False),
+        ("parse", True, False),
+        ("float32", True, True),
+    ],
+)
+def test_cleaner_numeric_dtype_modes(
```
rcap107: I think this test can be simplified by putting in the parametrize the expected transformed dataframe and testing the equality of the real and the expected dataframes rather than having all the different combinations of parameters

rcap107 (commented):
Hi @Varshith-Yadav, thank a lot for this PR.  I think it's in a good shape, but there are some parts that I think could be simplified.

Varshith-Yadav:
Thanks for the review @rcap107.

What I changed
- Refactored _get_preprocessors so it now handles numeric_dtype directly ("passthrough", "parse", "float32"), instead of receiving an external selector.
- Removed tofloat_cols from Cleaner.fit_transform; Cleaner now just validates and forwards numeric_dtype.
- Added a concrete docstring example in Cleaner showing behavior for all three numeric_dtype values.
- Simplified test_cleaner_numeric_dtype_modes to parametrize expected transformed dataframes and assert full dataframe equality.

rcap107:
Hi @Varshith-Yadav, after discussing IRL with other maintainers, we decided that it would be better to tweak a bit the behavior of the parameter. What you did implements correctly what's in the issue, but that might not be the best way of dealing with it. 

The new idea is to have only a single boolean parameter called `parse_strings`: if this parameter is true, then the ToFloat32 transformer is applied only to string columns, while if it's false then no parsing is done, and ToFloat32 isn't added to the preprocessing at all. 

In a separate PR we might want to add back the parameter for converting to Float32, but for this PR having only parse_strings is good enough. 

That means that you'll have to undo some of what you've worked on, but unfortunately that happens at times. 

Thanks a lot for the work so far!

Varshith-Yadav:
Thanks  @rcap107 for the clarification and for discussing this with the maintainers.

I updated the PR to follow the new direction:

- Replaced `Cleaner.numeric_dtype` with `Cleaner.parse_strings` (bool).
- `parse_strings=True`: applies `ToFloat` only to string columns.
- `parse_strings=False`: no `ToFloat` step is added.
- Kept `TableVectorizer` behavior unchanged (still uses float32 conversion in its own preprocessing path).
- Updated tests to `parse_strings` and simplified expectations.
- Updated docs/changelog to remove `numeric_dtype` wording and describe `parse_strings`.

Tests:
- `python -m pytest skrub/tests/test_table_vectorizer.py` passes.
- `python -m pre_commit run --all-files` passes.

rcap107:
Hi @Varshith-Yadav, after further discussions with the maintainers the conclusion is that there should still be a way of converting to float32 in the Cleaner, but that it should be with a separate parameter which I think we can still call `numeric_dtype`. 

However, differently from the original implementation and what is done in the TableVectorizer, here the conversion to float32 should only be done on columns that are already float: the idea is that if they are float64, they should be converted to float32 to save memory. Integer columns should be left alone and keep whatever dtype they're using (int32, int64 or anything else). 

Another thing: the changelog still needs to be fixed.

Thanks a lot!

Varshith-Yadav:
Thanks for the guidance.

I aligned the implementation with the final behavior we discussed:

- `Cleaner(parse_strings: bool)` controls parsing of numeric-looking string columns.
- `Cleaner(numeric_dtype="float32")` now downcasts only floating-point columns to `float32`.
- Integer columns are preserved as integer dtypes.
- `TableVectorizer` behavior is unchanged.

I also updated tests to cover:
- `parse_strings` on/off
- `numeric_dtype=None` vs `"float32"` for float-only downcasting
- combined `parse_strings=True` + `numeric_dtype="float32"`
- validation errors for invalid parameter values

Docs and changelog were updated accordingly.

Validation:
- `pre-commit run --all-files` passes
- `pytest skrub/tests/test_table_vectorizer.py` passes

doc/modules/default_wrangling/cleaning_dataframes.rst
```diff
@@ -70,31 +71,52 @@ dtype:  ...
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+Parsing numeric-looking strings with the |Cleaner|
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-By default, when the |Cleaner| encounters numeric dtypes (e.g., ``int8``,
-``float64``), it leaves them as-is. In some cases, it may be beneficial to have
-the same numeric dtype for all numeric columns to guarantee compatibility between
-values.
+By default, when the |Cleaner| encounters numeric-looking strings, it leaves
+them unchanged.
```
rcap107: It's not clear from this what a "numeric-looking string" is, and we are actually talking about series that contain such strings. I think it would be better to move the example dataframe from a bit below here, so that what we are talking about is clearer.

doc/modules/default_wrangling/cleaning_dataframes.rst
```diff
@@ -70,31 +71,52 @@ dtype:  ...
...
 
-The |Cleaner| allows conversion of numeric features to ``float32`` by setting
-the ``numeric_dtype`` parameter:
+The |Cleaner| can parse those values by setting ``parse_strings=True``:
```
rcap107: There should be a small comment about how strings are parsed to float32 (rather than some other dtype)

doc/modules/default_wrangling/cleaning_dataframes.rst
```diff
@@ -70,31 +71,52 @@ dtype:  ...
...
-ensures that all missing values have the same representation. This also ensures
-compatibility with scikit-learn transformers.
+When ``parse_strings=False`` (default), both columns keep their original dtypes.
+
+Downcasting float dtypes to ``float32`` with the |Cleaner|
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+By default, floating-point columns (e.g. ``float64``) keep their original dtype.
+To downcast only floating-point columns to ``float32``, set
+``numeric_dtype="float32"``:
```
rcap107: I think it would be better to write explicitly that integer colums are left untouched: it's already in the example, but readers may not notice it.

skrub/_table_vectorizer.py
```diff
@@ -119,33 +119,56 @@ def _get_preprocessors(
...
+        if numeric_dtype == "float32":
+            float_cols = cols & s.float()
+            tofloat_cols = (
+                float_cols if tofloat_cols is None else tofloat_cols | float_cols
+            )
+        elif numeric_dtype is not None:
+            raise ValueError(
+                "`numeric_dtype` must be one of "
+                f"[`None`, `'float32'`]. Found {numeric_dtype}."
+            )
```
rcap107: This check is already performed in Cleaner.fit so it's not needed here, and I think it's also why the coverage check is failing
```suggestion
```

skrub/_table_vectorizer.py
```diff
@@ -184,10 +207,17 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
+
+        - ``False``: no numeric parsing is attempted.
+        - ``True``: apply :class:`ToFloat` to string columns only. String columns
+          that contain only numerical values are converted to ``np.float32``.
+
     numeric_dtype : "float32" or None, default=None
-        If set to ``float32``, convert columns with numerical information
-        to ``np.float32`` dtype thanks to the transformer ``ToFloat``.
-        If ``None``, numerical columns are not modified.
+        Whether to downcast floating-point columns to ``np.float32``.
```
rcap107: `float16` columns should be cast to `float32` for consistency
```suggestion
        Whether to cast floating-point columns to ``np.float32``.
```

skrub/_table_vectorizer.py
```diff
@@ -252,6 +282,12 @@ class Cleaner(TransformerMixin, BaseEstimator):
       actual datetimes with the correct dtype. If ``datetime_format`` is provided,
       it is forwarded to ``ToDatetime()``. Otherwise, the format is inferred.
 
+    - ``ToFloat()``:
+      - if ``parse_strings=True``, apply ``ToFloat()`` on string columns only,
+        converting numeric-looking strings to ``np.float32``;
```
rcap107: Here it should be explained again what "numeric-looking" means

skrub/tests/test_table_vectorizer.py
```diff
@@ -439,11 +459,87 @@ def test_convert_float32(df_module):
...
+        (
+            True,
+            lambda X: sbd.with_columns(
+                X,
+                num_str=ToFloat().fit_transform(X["num_str"]),
+            ),
+        ),
+    ],
+)
+def test_cleaner_parse_strings(df_module, parse_strings, expected_df):
```
rcap107: These tests could be combined back into a single parametrized test that is checking all the four combinations, and additionally that parametrizes the float dtype to make sure that it gets converted properly:

(after the pars_strings/numeric_dtype parametrization)
```python
@pytest.mark.parametrize("float_dtype", [np.float16, np.float32, np.float64], ids=["float16", "float32", "float64"])
def test_cleaner_parse_numbers_all(df_module, parse_strings, numeric_dtype, expected_df, float_dtype):
    X = df_module.make_dataframe(
        {
            "num_str": ["1", "2", "3"],
            "int_col": [1, 2, 3],
            "float_col": np.array([1.5, 2.5, 3.5], dtype=float_dtype),
        }
    )```

It would be very similar to your original implementation, I'm sorry that you had to do the same work twice 😅

**Review by rcap107 — COMMENTED** [review]
Hi @Varshith-Yadav, the PR is looking pretty good, thanks a lot for that! 

I added a few comments on the docstrings to clarify some parts, and suggested a change in the tests. It's pretty much going back to the original implementation that you first submitted, sorry about that.

I think that after addressing the this review the PR should be good to go.

**Comment by Varshith-Yadav** [comment]
@rcap107  Implemented the requested changes and addressed the related docs updates.

Code changes in `skrub/_table_vectorizer.py`:
1. Removed the redundant `numeric_dtype` validation from `_get_preprocessors` (validation is already handled in `Cleaner.fit`).
2. Updated wording from “downcast” to “cast” for `numeric_dtype` to reflect that floating dtypes (including `float16`) are cast to `float32`.
3. Clarified `parse_strings` documentation by explicitly defining parsing behavior: string columns whose non-missing values can all be parsed as numbers are converted to `np.float32`.

Doc changes in `doc/modules/default_wrangling/cleaning_dataframes.rst`:
1. Clarified what “numeric-looking strings” means with concrete context and examples.
2. Added/updated explanation that parsed string numerics are cast to `float32`.
3. Made explicit that `numeric_dtype="float32"` only affects floating-point columns, while integer columns are left unchanged.

Validation run locally:
1. `python -m pytest skrub/tests/test_table_vectorizer.py -k "cleaner and (numeric_dtype or parse_strings)"`
2. `python -m pytest skrub/tests/test_table_vectorizer.py`
3. `python -m pre_commit run --files skrub/_table_vectorizer.py doc/modules/default_wrangling/cleaning_dataframes.rst`

Also, I accidentally closed and reopened this PR during branch cleanup earlier. That was my mistake, apologies for the noise.

File: skrub/_table_vectorizer.py [hunk 12]
```diff
@@ -119,33 +119,51 @@ def _get_preprocessors(
...
+            cols,
         ),
-        ToDatetime(format=datetime_format),
+        (ToDatetime(format=datetime_format), cols),
     ]
+
     if add_tofloat32:
-        transformers.append(ToFloat())
+        transformers.append((ToFloat(), cols))
+    else:
```
jeromedockes: maybe we can simplify a bit, along the lines of

```python
        tofloat_cols = s.cols(())
        if parse_strings:
            tofloat_cols = tofloat_cols | s.string()
        if numeric_dtype == "float32":
            tofloat_cols = tofloat_cols | s.float()
        transformers.append((ToFloat(), cols & tofloat_cols))

```

jeromedockes: or if we want to avoid the empty transformer when none of the conditions apply it might be clearer to list the cases here:

```python
match add_to_float_32, parse_strings, numeric_dtype == "float32":
    case False, False, False:
        tofloat_cols = None
    case False, False, True:
        tofloat_cols = cols & s.float()
    case False, True, False:
        tofloat_cols = cols & s.string()
    case False, True, True:
        tofloat_cols = cols & (s.string() | s.float())
    case True, *_:
        tofloat_cols = cols
if tofloat_cols is not None:
    transformers.append((ToFloat(), tofloat_cols))
```

rcap107:
Hi @Varshith-Yadav, it looks like coverage is complaining because one of the lines of the match is not being covered by tests: the `        case False, True, True:` line needs to be handled. That can be handled either by adding a test that is checking specifically that condition, or by refactoring the test_cleaner_parse_strings and test_cleaner_numeric_dtype tests so that they test all the combinations. 

Also, I wanted to thank you again for the time spent on this PR. It's been taking a few iterations to get to the best interface possible, but that happens sometimes and the result is having a better package overall.

rcap107:
Hi @Varshith-Yadav, I wrapped up the PR with the final changes. It was a longer process than we expected, but your work helped with coming up with a better and clearer API, so the time we took to reach a consensus was an investment.

Thanks a lot for doing most of the work!

jeromedockes:
@rcap107 from IRL discussion

```python
match parse_numbers, cast_numbers_to_float32:
    case False, False:
        tofloat_cols = None
    case False, True:
        tofloat_cols = cols & s.numeric()
    case True, False:
        tofloat_cols = cols & s.string()
    case True, True:
        tofloat_cols = cols
```

when casting all numbers go to float32
parse_numbers rather than parse strings (the input of parsing is always strings, the useful info is what we are looking for in the strings)
we don't need the add_tofloat32 param anymore

skrub/_table_vectorizer.py
```diff
@@ -189,17 +208,33 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
 
-    numeric_dtype : "float32" or None, default=None
-        If set to ``float32``, convert columns with numerical information
-        to ``np.float32`` dtype thanks to the transformer ``ToFloat``.
-        If ``None``, numerical columns are not modified.
+    parse_numbers : bool, default=False
+        Whether to parse strings that represent numeric values.
+
+        - ``False``: no numeric parsing is attempted.
+        - ``True``: apply :class:`ToFloat` to string columns only. String columns
```
jeromedockes:
```suggestion
        - ``True``: apply :class:`ToFloat` to string columns. String columns
```

skrub/_table_vectorizer.py
```diff
@@ -189,17 +208,33 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
+          whose non-missing values can all be parsed as numbers are converted to
+          ``np.float32``.
```
jeromedockes:
```suggestion
          ``float32``.
```

(in polars for example it will not be the numpy dtype)

skrub/_table_vectorizer.py
```diff
@@ -189,17 +208,33 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
+
+    cast_to_float : bool, default=False
```
jeromedockes: `cast_numbers_to_float32` ?

skrub/_table_vectorizer.py
```diff
@@ -189,17 +208,33 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
+        Whether to cast floating-point columns to ``np.float32``.
+        If set to ``True``, only columns that are already floating-point
+        are converted to ``np.float32``.
```
jeromedockes:
```suggestion
        If set to ``True``, numeric columns are converted to ``float32``.
```

skrub/_table_vectorizer.py
```diff
@@ -189,17 +208,33 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
 
     cast_to_str : bool, default=False
         If ``True``, apply the ``ToStr`` transformer to non-numeric,
         non-categorical, and non-datetime columns, converting them to strings.
         If ``False``, this step is skipped and such columns retain their
         original dtype (e.g., lists, structs).
 
+    numeric_dtype : "float32" or None, default=None
+        If set to "float32", this parameter has the same effect as
+        ``cast_to_float=True``: it casts floating-point columns to ``np.float32``.
```
jeromedockes: as cast_to_float=True and parse_numbers=True

and integer columns are affected

skrub/_table_vectorizer.py
```diff
@@ -189,17 +208,33 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
+        Integer columns are not affected.
+
+        .. deprecated:: 0.9.0
+            Use ``cast_to_float=True`` instead.
```
jeromedockes: and `parse_numbers`

skrub/_table_vectorizer.py
```diff
@@ -254,9 +285,16 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
       missing values.
 
-    - ``ToDatetime()``: parse datetimes represented as strings and return them as
+    - :class:`ToDatetime`: parse datetimes represented as strings and return them as
       actual datetimes with the correct dtype. If ``datetime_format`` is provided,
-      it is forwarded to ``ToDatetime()``. Otherwise, the format is inferred.
+      it is forwarded to :class:`ToDatetime`. Otherwise, the format is inferred.
+
+    - :class:`ToFloat`:
+      - if ``parse_numbers=True``, apply :class:`ToFloat` on string columns only,
```
jeromedockes:
```suggestion
      - if ``parse_numbers=True``, apply :class:`ToFloat` on string columns,
```

skrub/_table_vectorizer.py
```diff
@@ -254,9 +285,16 @@ class Cleaner(TransformerMixin, BaseEstimator):
...
+        converting strings whose non-missing values can all be parsed as numbers
+        to ``np.float32``;
```
jeromedockes:
```suggestion
        to ``float32``;
```

skrub/_table_vectorizer.py
```diff
@@ -377,11 +429,29 @@ def fit_transform(self, X, y=None):
...
+            if self.numeric_dtype == "float32":
+                cast_to_float = True
+            else:
+                raise ValueError(
+                    f"Unsupported value for `numeric_dtype`: {self.numeric_dtype!r}. "
+                    "The only supported value is 'float32'."
+                )
+
+        if not isinstance(self.parse_numbers, bool):
+            raise ValueError(
```
jeromedockes: typeerror

skrub/_table_vectorizer.py
```diff
@@ -377,11 +429,29 @@ def fit_transform(self, X, y=None):
...
+                f"`parse_numbers` must be a boolean. Found {self.parse_numbers!r}."
+            )
+        if not isinstance(cast_to_float, bool):
             raise ValueError(
-                "`numeric_dtype` must be one of"
-                f"[`None`, `'float32'`]. Found {self.numeric_dtype}."
+                f"`cast_to_float` must be a boolean. Found {cast_to_float!r}."
```
jeromedockes: typeerror

doc/modules/default_wrangling/cleaning_dataframes.rst
```diff
@@ -42,10 +37,10 @@ parsing datetimes and numbers.
   - Convert to strings: Convert columns to strings unless they have a more informative
     dtype, such as numeric, categorical, or datetime.
 
-If ``numeric_dtype`` is set to ``float32``, the ``Cleaner`` will also convert
-numeric columns to ``np.float32`` dtype, ensuring a consistent representation
-of numbers and missing values. This can be useful if the ``Cleaner``
-is used as a preprocessing step at the beginning of an ML pipeline.
+If ``parse_numbers`` is set to ``True``, the ``Cleaner`` will parse
+string columns that contain only numbers and convert them to ``np.float32``.
```
jeromedockes:
```suggestion
string columns that contain only numbers and convert them to ``float32``.
```

doc/modules/default_wrangling/cleaning_dataframes.rst
```diff
@@ -42,10 +37,10 @@ parsing datetimes and numbers.
...
+If ``cast_to_float32=True``, the ``Cleaner`` will also numeric columns
```
jeromedockes:
```suggestion
If ``cast_to_float32=True``, the ``Cleaner`` will also convert numeric columns
```

jeromedockes (approved):

PR #1911, opened 2026-02-17
ENH - Add a generic ApplyToCols

### LINKED ISSUE #1754: FEAT -  Add an `Apply` transformer that automatically chooses the correct ApplyTo depending on the transformer

rcap107:
This PR changes the behavior of ApplyToCols so that it can choose between the current behavior (apply the same transformer separately to multiple columns, then combine the result in a dataframe) and the behavior of ApplyToFrame (apply the same transformer to a subset of columns in the dataset and combine the result in a dataframe) without requiring the user to decide. 

This simplifies the user experience a lot, while retaining the versatility of the transformers. The default behavior is to detect automatically which behavior to adopt, but the users can also specify manually what they need: this is the same behavior adopted by .skb.apply in the data ops plan. 

Closes #1754 

TODO
- [ ] Implement the new ApplyToCols class
- [ ] Adapt the tests and add new tests to check the behavior
- [ ] Update the doctring of the new class
- [ ] Update the examples to reflect the new behavior


rcap107:
Closing this to simplify the diff.

PR #1913, opened 2026-02-17
ENH - Renaming the Apply transformers

rcap107:
# Refactoring Pull Request

## Description
This is the first of two PRs that will be merged to address #1754. The overall plan is to have a new transformer called ApplyToCols that chooses automatically between the functionality of the current ApplyToCols and the current ApplyToFrame, to simplify the user experience: users won't need to understand the difference between the two transformers anymore.

Since we are going to reuse the same name (ApplyToCols), this complicates the implmentation and will cause issues with backwards compatibility, which is why this PR and the folllowing need to be treated with care and merged close to each other.

In this PR I'm going to rename all instances of ApplyToCols to ApplyOnEachCol, and all instances of ApplyToFrame to ApplySubFrame, and rename the respective files accordingly.

This is to simplify the diff, because I will be working on a separate ApplyToCols object in a different PR, and having both renaming and a new object in the same place is going to cause issues.


jeromedockes (approved):
cool, thanks @rcap107 ! To be continued ... :)

PR #1914, opened 2026-02-17
WIP - trying to fix nightly dependencies not being available

rcap107:
In the past week or so, the CI has been failing constantly for the same reason: pixi tries to build the lockfile for all the environments, fails to fetch some of the nightly dependencies and errors out. This problem usually goes away after a few hours, but it still prevents tests from running. 

I've been talking with the pixi devs about the problem, and apparently it's caused by downtime on anaconda's side, which is something they can't do anything about. 

This PR is trying to throw things at the wall and see if anything sticks.


rcap107:
closing this, the problem is likely on conda's side

PR #1916, opened 2026-02-19
better errors for lazyframes

### LINKED ISSUE #1915: BUG - polars LazyFrames are not supported by data ops
### Describe the bug

Polars LazyFrames are not supported by data ops, and must be converted to eager frames in order to execute most operations. 

So far I have encountered two main problematic situations:
- `.skb.apply` does not support lazyframes because it is slicing and slicing is not available for lazyframes
- `.skb.make_randomized_search` does not support lazyframes because it forwards whatever is inside the variable to the scikit-learn randomized search object, and the random search does not support lazyframes. 

These are just the first two cases that I found so far, but I imagine there are more problematic situations. These settings represent a major part of what a user would be doing.

This issue needs to be dealt with by maintainers. 

### Steps/Code to Reproduce

_No response_

### Expected Results

_No response_

### Actual Results

_No response_

### Versions

```shell
.
```

jeromedockes:
closes #1915 

we now get a better error saying that lazyframes are not yet supported in skrub and to collect() whenever we try to pass a lazyframe where a dataframe is expected

Currently CheckInput collects lazyframes with a warning; I guess to be consistent that should be removed and become an error (probably best to tackle that in a separate pr)

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,6 +22,9 @@ New Features
 
 Changes
 -------
+- ``ApplyToCols`` and ``ApplyToFrame`` have been renamed :class:`ApplyToEachCol`
+  and :class:`ApplyToSubFrame` respectively.
+  :pr:`1913` by :user:`Riccardo Cappuzzo <rcap107>`.
 
 Bug Fixes
 --------
--- a/doc/api_reference.py
+++ b/doc/api_reference.py
@@ -66,8 +66,8 @@ API_REFERENCE = {
                     "TableVectorizer",
                     "SelectCols",
                     "DropCols",
-                    "ApplyToCols",
-                    "ApplyToFrame",
+                    "ApplyToEachCol",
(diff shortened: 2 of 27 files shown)


skrub/_check_input.py
```diff
@@ -64,20 +64,17 @@ def _check_not_pandas_sparse_pandas(df):
 
 
 def _check_is_dataframe(df):
+    if sbd.is_lazyframe(df):
```
rcap107, line 67: I thought the idea was to never collect automatically to avoid memory issues, here we are collecting though?

jeromedockes: yes exactly; see my comment in the pr description. It is the current behavior, I think this should be removed in another PR

rcap107: ah whoops I missed that, mb

jeromedockes:
> We should open an issue to remove the collect from check_input

see #1925

PR #1917, opened 2026-02-19
Feat-generic-apply-to-cols

### LINKED ISSUE #1754: FEAT -  Add an `Apply` transformer that automatically chooses the correct ApplyTo depending on the transformer

rcap107:
# New Feature Pull Request

## Description

This PR adds a new ApplyToCols object that uses the wrap_transformer function to automatically select between ApplyToEachCol (the old ApplyToCol) and ApplyToSubFrame (the old ApplyToFrame).

Fixes #1754


rcap107:
messed up the git tree, I'll reopen

PR #1919, opened 2026-02-19
FEAT - add new apply to cols

### LINKED ISSUE #1754: FEAT -  Add an `Apply` transformer that automatically chooses the correct ApplyTo depending on the transformer

rcap107:
This PR adds a generic ApplyToCols that automatically chooses between ApplyToEachCol (the old ApplyToCols) and ApplyToSubFrame (the old (ApplyToFrame). 

This addresses #1754 and is a follow up to  #1913

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,10 +23,11 @@ New Features
 Changes
 -------
 - :class:`ApplyToCols` has been modified so that now it can detect automatically
-  whether the provided transformer should be applied on each column, or on the
-  entire dataframe. In most cases, this replaces the original ``ApplyToCols``
-  and ``ApplyToFrame``. As a result, ``ApplyToCols`` and ``ApplyToFrame`` have
-  been renamed :class:`ApplyToEachCol` and :class:`ApplyToSubFrame` respectively.
+  whether the provided transformer should be applied independently on each column,
+  or on all selected columns as a single dataframe. In most cases, this replaces
+  the original ``ApplyToCols`` and ``ApplyToFrame``. As a result, ``ApplyToCols``
+  and ``ApplyToFrame`` have been renamed :class:`ApplyToEachCol` and
+  :class:`ApplyToSubFrame` respectively.
   The behavior of the old ``ApplyToCols`` can be replicated by setting the parameter
   ``how`` to ``cols``.
   :pr:`1913` and :pr:`1919` by :user:`Riccardo Cappuzzo <rcap107>`.
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,7 +28,7 @@ Changes
   and ``ApplyToFrame``. As a result, ``ApplyToCols`` and ``ApplyToFrame`` have
   been renamed :class:`ApplyToEachCol` and :class:`ApplyToSubFrame` respectively.
   The behavior of the old ``ApplyToCols`` can be replicated by setting the parameter
-  ``columnwise`` to ``True``.
+  ``how`` to ``cols``.
   :pr:`1913` and :pr:`1919` by :user:`Riccardo Cappuzzo <rcap107>`.
 
 Bug Fixes
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -22,9 +22,14 @@ New Features
 
 Changes
 -------
-- ``ApplyToCols`` and ``ApplyToFrame`` have been renamed :class:`ApplyToEachCol`
-  and :class:`ApplyToSubFrame` respectively.
-  :pr:`1913` by :user:`Riccardo Cappuzzo <rcap107>`.
+- :class:`ApplyToCols` has been modified so that now it can detect automatically
+  whether the provided transformer should be applied on each column, or on the
+  entire dataframe. In most cases, this replaces the original ``ApplyToCols``
+  and ``ApplyToFrame``. As a result, ``ApplyToCols`` and ``ApplyToFrame`` have
+  been renamed :class:`ApplyToEachCol` and :class:`ApplyToSubFrame` respectively.
+  The behavior of the old ``ApplyToCols`` can be replicated by setting the parameter
+  ``columnwise`` to ``True``.
+  :pr:`1913` and :pr:`1919` by :user:`Riccardo Cappuzzo <rcap107>`.
 
 Bug Fixes
 --------
--- a/examples/0090_apply_to_cols.py
+++ b/examples/0090_apply_to_cols.py
@@ -7,6 +7,19 @@ In previous examples, we saw how skrub provides powerful abstractions like
 
 In this new example, we show how to create more flexible pipelines by selecting
 and transforming dataframe columns using arbitrary logic.
+
+.. |ApplyToCols| replace:: :class:`~skrub.ApplyToCols`
+.. |ApplyToEachCol| replace:: :class:`~skrub.ApplyToEachCol`
+.. |ApplyToSubFrame| replace:: :class:`~skrub.ApplyToSubFrame`
+.. |StringEncoder| replace:: :class:`~skrub.StringEncoder`
+.. |SelectCols| replace:: :class:`~skrub.SelectCols`
+.. |DropCols| replace:: :class:`~skrub.DropCols`
+.. |TableVectorizer| replace:: :class:`~skrub.TableVectorizer`
+.. |OrdinalEncoder| replace:: :class:`~sklearn.preprocessing.OrdinalEncoder`
+.. |PCA| replace:: :class:`~sklearn.decomposition.PCA`
+.. |Pipeline| replace:: :class:`~sklearn.pipeline.Pipeline`
+.. |ColumnTransformer| replace:: :class:`~sklearn.compose.ColumnTransformer`
+
 """
 
 # %%
@@ -21,21 +34,21 @@ X, y = data.X, data.y
 X
 
 # %%
-# Our goal is now to apply a :class:`~skrub.StringEncoder` to two columns of our
+# Our goal is now to apply a |StringEncoder| to two columns of our
 # choosing: ``division`` and ``employee_position_title``.
 #
-# We can achieve this using :class:`~skrub.ApplyToEachCol`, whose job is to apply a
+# We can achieve this using |ApplyToCols|, whose job is to apply a
 # transformer to multiple columns independently, and let unmatched columns through
 # without changes.
 # This can be seen as a handy drop-in replacement of the
-# :class:`~sklearn.compose.ColumnTransformer`.
+# |ColumnTransformer|.
 #
 # Since we selected two columns and set the number of components to ``30`` each,
-# :class:`~skrub.ApplyToEachCol` will create ``2*30`` embedding columns in the dataframe
+# |ApplyToCols| will create ``2*30`` embedding columns in the dataframe
 # ``Xt``, which we prefix with ``lsa_``.
-from skrub import ApplyToEachCol, StringEncoder
+from skrub import ApplyToCols, StringEncoder
 
-apply_string_encoder = ApplyToEachCol(
+apply_string_encoder = ApplyToCols(
     StringEncoder(n_components=30),
     cols=["division", "employee_position_title"],
     rename_columns="lsa_{}",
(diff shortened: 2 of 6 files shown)


skrub/tests/test_apply_to_cols.py
```diff
@@ -0,0 +1,162 @@
...
+
+import numpy as np
+import pytest
+from sklearn.preprocessing import OrdinalEncoder, StandardScaler
+
+from skrub import ApplyToCols, ApplyToEachCol
+from skrub import _dataframe as sbd
+from skrub import selectors as s
+from skrub._apply_sub_frame import ApplyToSubFrame
+from skrub._to_datetime import ToDatetime
```
jeromedockes: those 2 can also be imported directly from `skrub` right?

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+from ._wrap_transformer import wrap_transformer
+
+_SELECT_ALL_COLUMNS = selectors.all()
+
+
+class ApplyToCols(TransformerMixin, BaseEstimator):
+    """
+    Map a transformer to columns in a dataframe.
+
+    A separate clone of the transformer is applied to each column separately.
```
jeromedockes: this is only in the "cols" case

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+    transformer : transformer instance
+        The transformer to apply to the selected columns.
+
+    cols : str, sequence of str, or skrub selector, optional
+        The columns to attempt to transform. Only the selected columns will have
+        the transformer applied. Columns outside this selection are passed
+        through unchanged (``fit_transform`` is not called on them) and remain
+        unmodified in the output. The default is to attempt transforming all
+        columns.
+        See the documentation of
```
jeromedockes: as ApplyToCols is the main entry point I think any further explanation should be copied here

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+        :class:`ApplyToEachCol` or :class:`ApplyToSubFrame` for details.
+
+    allow_reject : bool, default=False
+        Whether to allow rejecting all columns. See the documentation of
```
jeromedockes: "whether to allow refusing to transform columns for which the provided transformer is not suited, for example rejecting non-datetime columns if transformer is a DatetimeEncoder." and here it is ok to keep the pointer to applytoeachcol as you did because it is a less-frquently-used option

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+        ``ApplyToEachCol`` for details.
+
+    how : "auto", "cols" or "frame", optional
+        How the transformer is applied. In most cases the default "auto"
+        is appropriate.
+        - "cols" means `transformer` is wrapped in a :class:`ApplyToEachCol`
+            transformer, which fits a separate clone of `transformer` each
+            column in `cols`. `transformer` must be a transformer (have a
```
jeromedockes: the 'transformer' must be a transformer part can be removed I think

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+            ``fit_transform`` method).
+        - "frame" means `transformer` is wrapped in a :class:`ApplyToSubFrame`
+            transformer, which fits a single clone of `transformer` to the
+            selected part of the input dataframe. `transformer` must be a
+            transformer.
```
jeromedockes: also no ned for must be a transformer

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+    All columns not listed in ``cols`` remain unmodified in the output.
+    Moreover, if ``allow_reject`` is ``True`` and the transformers'
+    ``fit_transform`` raises a ``RejectColumn`` exception for a particular
+    column, that column is passed through unchanged. If ``allow_reject`` is
+    ``False``, ``RejectColumn`` exceptions are propagated, like other errors
+    raised by the transformer.
+    Wrapper that applies a transformer to columns selected by a selector. The
+    wrapper automatically selects the correct class between ``ApplyToEachCol`` and
+    ``ApplyToSubFrame`` based on the type of the input transformer, but this can
+    be overridden with the ``how`` parameter.
```
jeromedockes: this part seems left-over after some editing?

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+        return self
+
+    def fit_transform(self, X, y=None):
+        if self.how not in ("auto", "cols", "frame"):
+            raise ValueError(
+                f"Invalid value for 'how': {self.how}. "
+                "Expected one of 'auto', 'cols', or 'frame'."
+            )
+
+        if self.allow_reject not in (True, False):
```
jeromedockes: in (True, False) rather than checking if it is a bool is to also accept numpy bools?

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,134 @@
...
+            raise ValueError(
```
jeromedockes: maybe typeerror here

skrub/tests/test_apply_to_cols.py
```diff
@@ -0,0 +1,161 @@
...
+from skrub import selectors as s
+from skrub._to_datetime import ToDatetime
+
+
+def test_single_column_transformer_becomes_apply_to_each_col(df_module):
+    """SingleColumnTransformer should be wrapped as ApplyToEachCol."""
+    at = ApplyToCols(ToDatetime(), cols=s.all())
+    X = df_module.make_dataframe({"date_col": ["2020-01-01", "2020-01-02"]})
+    at.fit(X)
+    assert hasattr(at._wrapped_transformer, "transformers_")
```
jeromedockes:
```suggestion
    assert hasattr(at, "transformers_")
```

jeromedockes: (same for other tests)

jeromedockes (commented):
nice! I have mostly nitpicks :)

skrub/_apply_to_cols.py
rcap107: that's all the detail there is, actually 

the rest would go in the notes

skrub/_apply_to_cols.py
rcap107: is this what you mean?

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,125 @@
...
+        Whether to allow refusing to transform columns for which the provided
+        transformer is not suited, for example rejecting non-datetime columns if
+        transformer is a DatetimeEncoder. See the documentation of
+        :class:`ApplyToEachCol` for details.
+
+    how : "auto", "cols" or "frame", optional
+        How the transformer is applied. In most cases the default "auto"
+        is appropriate.
+        - "cols" means `transformer` is wrapped in a :class:`ApplyToEachCol`
+            transformer, which fits a separate clone of `transformer` each
```
```suggestion
            transformer, which fits a separate clone of `transformer` to each
```

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,125 @@
...
+
+        if isinstance(self._wrapped_transformer, ApplyToEachCol):
+            self.transformers_ = self._wrapped_transformer.transformers_
+        else:
+            self.transformer_ = self._wrapped_transformer.transformer_
+
+        return X_transformed
+
+    def transform(self, X):
+        return self._wrapped_transformer.transform(X)
```
jeromedockes, line 213: maybe we should also add something like

```python
def get_feature_names_out(self, ...): return self._wrapped_transformer.get_feature_names_out(...)
```

jeromedockes:
you also need to add it to the reference docs

examples/0090_apply_to_cols.py
```diff
@@ -44,40 +57,61 @@
...
+#    classes.
+#    The former applies a transformer to each column independently, while the latter
+#    applies a transformer to a sub-dataframe.
+#    Normally, users don't need to worry about these two classes, but they can be useful
+#    when more control is needed.
+
 # %%
 # Note that selectors also come in handy in a pipeline to select or drop columns, using
-# :class:`~skrub.SelectCols` and :class:`~skrub.DropCols`!
+# |SelectCols| and |DropCols|!
```
jeromedockes:
```suggestion
# |SelectCols| and |DropCols|.
```

while we're at it :)

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,230 @@
...
+        - "cols" means `transformer` is wrapped in a :class:`ApplyToEachCol`
+            transformer, which fits a separate clone of `transformer` to each
+            column in `cols`.
+        - "frame" means `transformer` is wrapped in a :class:`ApplyToSubFrame`
+            transformer, which fits a single clone of `transformer` to the
+            selected part of the input dataframe.
+        - "auto" chooses the wrapping depending on the input and transformer.
+            If the transformer has a ``__single_column_transformer__`` attribute,
+            "cols" is chosen. Otherwise "frame" is chosen.
+
```
jeromedockes, line 76: we are missing the documentation for keep_original, rename_columns and n_jobs

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,230 @@
...
+        rename_columns="{}",
+        how="auto",
+        n_jobs=None,
+    ):
+        self.transformer = transformer
+        self.cols = cols
+        self.allow_reject = allow_reject
+        self.keep_original = keep_original
+        self.rename_columns = rename_columns
+        self.n_jobs = n_jobs
```
jeromedockes: nitpick: can you assign  the attributes in the same order that parameters appear in the signature? this makes it easier to check they are all there. Also we may want to make all except transformer and cols keyword-only (add a `, *` in-between), and list them in the order of likelihood to be of interest -- not sure what that is but I think "how" would go near the top. maybe how, n_jobs, allow_reject, keep_original, rename

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,230 @@
...
+        if not isinstance(self.keep_original, bool):
+            raise TypeError(
+                f"Invalid value for 'keep_original': {self.keep_original}. "
+                "Expected a boolean."
+            )
+
+        columnwise = {"auto": "auto", "cols": True, "frame": False}[self.how]
+
+        # To make sure transform() raises if fit_transform() hasn't been called
+        self._wrapped_transformer = None
```
```suggestion
```

I don't think we need this

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,230 @@
...
+        ):
+            raise NotFittedError(
+                "This ApplyToCols instance is not fitted yet. Call 'fit' with "
+                "appropriate arguments before using this estimator."
+            )
+        if isinstance(self._wrapped_transformer, ApplyToEachCol):
+            check_is_fitted(self, "transformers_")
+        else:
+            check_is_fitted(self, "transformer_")
+
```
jeromedockes, line 212:
```suggestion
        check_is_fitted(self)
```

just `check_is_fitted(self)` should be enough. it will pass if self contains any attributes ending with `_` and not starting with `__`, and fail otherwise.

jeromedockes: (the suggested diff shows just 1 line for some reason but the suggestion is to replace all the code above with just `check_is_fitted(self)`

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,230 @@
...
+            or self._wrapped_transformer is None
+        ):
+            raise NotFittedError(
+                "This ApplyToCols instance is not fitted yet. Call 'fit' with "
+                "appropriate arguments before using this estimator."
+            )
+        if isinstance(self._wrapped_transformer, ApplyToEachCol):
+            check_is_fitted(self, "transformers_")
+        else:
+            check_is_fitted(self, "transformer_")
```
jeromedockes: here also just `check_is_fitted(self)`

CHANGES.rst
```diff
@@ -22,9 +22,14 @@ New Features
 
 Changes
 -------
-- ``ApplyToCols`` and ``ApplyToFrame`` have been renamed :class:`ApplyToEachCol`
-  and :class:`ApplyToSubFrame` respectively.
-  :pr:`1913` by :user:`Riccardo Cappuzzo <rcap107>`.
+- :class:`ApplyToCols` has been modified so that now it can detect automatically
+  whether the provided transformer should be applied on each column, or on the
+  entire dataframe. In most cases, this replaces the original ``ApplyToCols``
```
jeromedockes: independently on each column, or on all selected columns as a single dataframe

skrub/tests/test_apply_to_cols.py
```diff
@@ -0,0 +1,219 @@
...
+
+    at.fit(X_train)
+    X_train_transformed = at.transform(X_train)
+    X_test_transformed = at.transform(X_test)
+
+    assert sbd.shape(X_train_transformed) == (3, 2)
+    assert sbd.shape(X_test_transformed) == (2, 2)
+
+
+def test_correct_transformer_attributes(df_module):
```
jeromedockes: maybe this could be merged with the first 2 tests at the top of the module

skrub/tests/test_apply_to_cols.py
```diff
@@ -0,0 +1,219 @@
...
+        {
+            "date": [datetime.datetime(2020, 1, 1), datetime.datetime(2020, 1, 2)],
+            "value": [1, 2],
+        }
+    )
+
+    df_module.assert_frame_equal(X_transformed, X_expected)
+
+    # A RejectColumn exception should be raised
+    with pytest.raises(ValueError):
```
jeromedockes, line 135: raises(RejectColumn)?

skrub/tests/test_apply_to_cols.py
```diff
@@ -0,0 +1,219 @@
...
+    with pytest.raises(NotFittedError):
+        at.transform(X)
+
+
+def test_check_is_fitted_get_feature_names_out():
+    """Test that get_feature_names_out raises NotFittedError before fitting."""
+    at = ApplyToCols(ToDatetime(), cols=s.all())
+
+    # Should raise NotFittedError when calling get_feature_names_out before fit
+    with pytest.raises(NotFittedError):
```
jeromedockes, line 165: can we check a case where get_feature_names_out does work?

skrub/_apply_to_cols.py
```diff
@@ -0,0 +1,230 @@
...
+        Whether to allow refusing to transform columns for which the provided
+        transformer is not suited, for example rejecting non-datetime columns if
+        transformer is a DatetimeEncoder. See the documentation of
+        :class:`ApplyToEachCol` for details.
+
+    how : "auto", "cols" or "frame", optional
+        How the transformer is applied. In most cases the default "auto"
+        is appropriate.
+        - "cols" means `transformer` is wrapped in a :class:`ApplyToEachCol`
+            transformer, which fits a separate clone of `transformer` to each
```
jeromedockes: extra leading whitespace gets rendered as a quote, check the html documentation

examples/0090_apply_to_cols.py
```diff
@@ -44,40 +57,61 @@
...
 model = make_pipeline(
     apply_string_encoder,
     apply_pca,
 ).fit_transform(X)
 
+# %%
+# .. admonition:: Under the hood of |ApplyToCols|
+#   :collapsible: closed
+#
+#    |ApplyToCols| is implemented using the |ApplyToEachCol| and |ApplyToSubFrame|
```
jeromedockes: also here, check the rendering sphinx understands it as a quotation

jeromedockes (approved):
LGTM, thanks @rcap107 🎉

GaelVaroquaux:
> Merged #1919 into main.

Oh my God, so cool!! Thanks!!

PR #1920, opened 2026-02-19
Show values provided in environment in dataop report

jeromedockes:
when the evaluation of a node is short-circuited because it is provided directly in the environment (eg we are injecting the split X and y) the evaluation report would just say it was bypassed without showing the value; now it is displayed. also it now displays the key used to retrieve it (or if it was X or y).

as a side effect due to slightly changing how such "bypassed" nodes are handled, the cache pruning can now clear nodes in an edge case where the result would have been kept before (when it is needed by both a node that is bypassed and one that needs computing)

for example given this script which generates reports for train and predict on a train_test_split generated by the dataop:

```python
import skrub
from skrub import selectors as s

df = skrub.var("df", fetch_employee_salaries().employee_salaries)
data_op = (
    df.skb.select(s.all() - "current_annual_salary")
    .skb.mark_as_X()
    .skb.apply(skrub.TableVectorizer())
    .skb.apply(
        HistGradientBoostingRegressor(), y=df["current_annual_salary"].skb.mark_as_y()
    )
)

split = data_op.skb.train_test_split()
learner = data_op.skb.make_learner()
learner.report(environment=split["train"], mode="fit", title="fit")
learner.report(environment=split["test"], mode="predict", title="predict")
```

in predict for the node marked as X the main branch produces:

screenshot_2026-02-19T18:18:10+01:00

while the pr shows its value

screenshot_2026-02-19T18:18:31+01:00

### DIFF
--- a/skrub/_reporting/_data/templates/data_ops/node.html
+++ b/skrub/_reporting/_data/templates/data_ops/node.html
@@ -79,14 +79,7 @@
                 </div>
                 {% endif %}
 
-                {% if env_key is not none %}
-                <div>
-                    <h3>Provided in environment</h3>
-                    The value for this step was provided directly in the environment (as the value for <strong><code>{{ {'_skrub_X': 'X', '_skrub_y': 'y'}.get(env_key, env_key.__repr__()) }}</code></strong>).
-                </div>
-                {% endif %}
-
-                {% if estimator_html_repr %}
+                {% if estimator_html_repr and env_key is none %}
                 <div>
                     <h3>Estimator fitted in this step:</h3>
                     {{ estimator_html_repr | safe }}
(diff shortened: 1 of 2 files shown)


GaelVaroquaux:
It's nice.

But also, I wonder how many users are going to be hitting these issues. They are clearly very advanced topics

skrub/_reporting/_data/templates/data_ops/node.html
```diff
@@ -72,12 +72,20 @@ <h3>This step was used by:</h3>
...
                 </div>
                 {% endif %}
+
+                {% if env_key is not none %}
+                <div>
+                    <h3>Provided in environment</h3>
+                    The value for this step was provided directly in the environment (as the value for <strong><code>{{ {'_skrub_X': 'X', '_skrub_y': 'y'}.get(env_key, env_key.__repr__()) }}</code></strong>).
+                </div>
+                {% endif %}
+
```
rcap107: I think this part should be made more evident, and I think it would be clearer to have a message similar to the previous version of the code, like

"This step was not executed because its value was provided directly in the `environment` as the value etc.

I also think that this should have a different formatting, such as a callout block. That would make it more evident that this behavior is different from how the learner typically works.

jeromedockes: ok here is a new screenshot

screenshot_2026-03-05T15:11:27+01:00

rcap107: I think it looks great, thanks!

Issue #1605, opened 2025-09-12
Reconsider the plots shown in the homepage and add a carousel


rcap107:
We have been de-prioritizing the GapEncoder a lot, but it's still one of the first things that people see when they open the homepage. 

I think we should replace it with something else:
- A simple change would be using the SquashingScaler as the "poster encoder"
- A more complex alternative would be using something like the sphinx [card-carousel](https://sphinx-design.readthedocs.io/en/latest/cards.html#card-carousels) to have multiple figures, which would allow to potentially have the SquashingScaler, DatetimeEncoder and StringEncoder all in the front page. 

My suggestion is to start with the SquashingScaler and then move to the card carousel in a later PR.

rcap107:
From meeting: the DatetimeEncoder with periodic features is better than the SquashingEncoder. We can also go with the carousel, but later.

rcap107:
We could use this for the carousel: https://sphinx-carousel.readthedocs.io/en/1.2.0/theme_pydata_sphinx_theme/index.html

rcap107:
The GapEncoder part of this issue was addressed in #1924, but the current homepage can still be improved. We can add a carousel, or alter what's shown in some other way.

PR #1924, opened 2026-02-23
DOC - removing GapEncoder from the homepage

rcap107:
Addresses part of  #1605

For the moment I am replacing the GapEncoder with the DatetimeEncoder. Later, we should replace this with a carousel. 

I'm not sure what to put in the "sample code", because adding the code used for generating this plot is not feasible.


jeromedockes (approved):
thanks!

PR #1927, opened 2026-02-24
small implementation simplification

jeromedockes:
when collecting X and y for splitting, CV etc. they are now computed jointly instead of one after the other with clearing of intermediate results disabled. the result is the same, but it is very slightly simpler and could use less memory in some cases due to turning on the intermediate result garbage collection for this operation


rcap107 (approved):
Thanks for the improvement @jeromedockes

PR #1928, opened 2026-02-24
add test

jeromedockes:


skrub/_data_ops/tests/test_estimators.py
```diff
@@ -569,6 +569,21 @@ def test_train_test_split(with_y):
...
+    """
+    df = skrub.var(
+        "df", pd.DataFrame({"a": list(range(1_000)), "b": list(range(1_000))})
+    )
+    shuffled = df.sample(frac=1)
+    X = shuffled[["a"]].skb.mark_as_X()
+    y = shuffled["b"].skb.mark_as_y()
+    data_op = X.skb.apply(Ridge(), y=y)
+    split = data_op.skb.train_test_split()
+    assert (split["X_train"]["a"].to_numpy() == split["y_train"].to_numpy()).all()
```
rcap107, line 584: To make sure I'm understanding: 

This is asserting that, after shuffling, the order of the training samples of X and the training samples of y are the same?

So that if the samples are [[1,2,3,4], [1,2,3,4]]

after shuffling it becomes [[3,4,1,2], [3,4,1,2]]

then X_train is [3,4,1] and it needs to be equal to y_train which is [3,4,1]

jeromedockes: exactly. so that if we inadvertently change the way they are computed, and do 2 full independent evaluations one for X and one for y, the shuffling will be done once while we are collecting X, and then it will be done again when we are collecting y, and the orders won't be the same and the test will fail. some other tests would fail as well but it's important and easy/fast to check so I thought we might as well have one dedicated to just that

rcap107: makes sense, thanks for the explanation

PR #1931, opened 2026-02-26
expose tags, classes_, _estimator_type in Learner classes

jeromedockes:
The SkrubLearner, ParamSearch and OptunaSearch dataop wrapper classes now expose some attributes used by scikit-learn machinery:

- the scikit-learn estimator tags
- `classes_` in fitted classifiers
- the legacy `_estimator_type` for old scikit-learn versions

the code for doing this is already in skrub, but at the moment it is only used in private internal classes that allow us to use scikit-learn tools like `cross_validate` and `GridSearchCV`. it was originally not added to `SkrubLearner` because, due to accepting a dict rather than X, y pair it is not interchangeable with a scikit-learn estimator anyway. However, having those attributes increases the partial compatibility, making it easier for client code or third-party libraries that want to handle both skrub learners and scikit-learn estimators (eg skore)

### DIFF
--- a/skrub/_data_ops/_estimator.py
+++ b/skrub/_data_ops/_estimator.py
@@ -51,20 +51,6 @@ _SEARCH_FITTED_ATTRIBUTES = _SKLEARN_SEARCH_FITTED_ATTRIBUTES_TO_COPY + [
 ]
 
 
-def _get_default_sklearn_tags():
-    class _DummyTransformer(TransformerMixin, BaseEstimator):
-        pass
-
-    try:
-        return _DummyTransformer().__sklearn_tags__()
-    except AttributeError:
-        # sklearn < 1.6
-        return None
-
-
-_DEFAULT_SKLEARN_TAGS = _get_default_sklearn_tags()
-
-
 class _SharedDict(dict):
     """A dict that does not get copied during deepcopy/sklearn clone.
 
@@ -87,6 +73,20 @@ def _copy_attr(source, target, attributes):
             setattr(target, a, getattr(source, a))
         except AttributeError:
             pass
+
+
+def _get_default_sklearn_tags():
+    class _DummyTransformer(TransformerMixin, BaseEstimator):
+        pass
+
+    try:
+        return _DummyTransformer().__sklearn_tags__()
+    except AttributeError:
+        # sklearn < 1.6
+        return None
+
+
+_DEFAULT_SKLEARN_TAGS = _get_default_sklearn_tags()
 
 
 class _DataOpWrapperMixin(_CloudPickle):


jeromedockes:
> * Why do we have "_default_sklearn_tags" as a callable rather than a module-level constant that is built by a callable at module import?

It was to have the check on the existence of tags (only exist in recent scikit-learn) in only one place. I removed the function and inlined the computation of default tags.

> * We should probably move forward and apply sklearn common tests to the objects concerned by this PR

Almost all the estimator checks will fail because the learners do not follow the scikit-learn API (they do not take X, y inputs but a dictionary)

jeromedockes:
> I removed the function and inlined the computation of default tags.

actually getting the default tags does take a few dozen microseconds so in the end I did what you suggest and compute only once to put in a global variable

jeromedockes:
> actually getting the default tags does take a few dozen microseconds so in the end I did what you suggest and compute only once to put in a global variable

although it was only called in the improbable case that we are getting the tags on a dataop that contains no estimator (apply), so probably 0 times in most programs... anyway not super important

GaelVaroquaux:
> > * Why do we have "_default_sklearn_tags" as a callable rather than a module-level constant that is built by a callable at module import?

> It was to have the check on the existence of tags (only exist in recent scikit-learn) in only one place. I removed the function and inlined the computation of default tags.

I agree, but my gut feeling was that we could call the function that you wrote at module-loading time and store the result in a constant.

> > * We should probably move forward and apply sklearn common tests to the objects concerned by this PR

> Almost all the estimator checks will fail because the learners do not follow the scikit-learn API

Point taken.
Still, I think that we should, when possible, try to call the common tests, because we run the risk of not following the evolutions of the conventions.

GaelVaroquaux:
> actually getting the default tags does take a few dozen microseconds so in the end I did what you suggest and compute only once to put in a global variable

Sorry, I just saw that, which makes my above comment irrelevant :)

jeromedockes:
>> actually getting the default tags does take a few dozen microseconds so in the end I did what you suggest and compute only once to put in a global variable
>
> Sorry, I just saw that, which makes my above comment irrelevant :)

I don't think that duration is completely irrelevant for functions that might get called often (slowness spread over many small inefficiencies is hard to come back from). Here the situations where we fall back to default tags are very rare but that's not obvious. Anyway I kept the suggestion of having a global variable.

> Still, I think that we should, when possible, try to call the common tests, because we run the risk of not following the evolutions of the conventions.

Ok I think I will create a wrapper class that adapts to the scikit-learn interface so we can use it for the estimator checks. (we also discussed making such an estimator available to users)

PR #1939, opened 2026-02-27
Wording details around data-ops

GaelVaroquaux:
Small wording changes to try to make data-ops purpose more clear

### DIFF
--- a/examples/data_ops/1110_data_ops_intro.py
+++ b/examples/data_ops/1110_data_ops_intro.py
@@ -8,10 +8,10 @@ This example shows data wrangling for machine learning using Skrub's
 The challenge of data-wrangling for machine learning is the need to
 apply the wrangling operations to new data, for prediction.
 
-Skrub's DataOps build pipelines blending data wrangling and machine
+Skrub's DataOps build pipelines that blend data wrangling and machine
 learning by recording all the operations involved in pre-processing data
-and training models. They result an a full *learner* that starts from the
-raw data. We will also how it can be saved, loaded it back, and then make
+and training models. They result in an a full *learner* that starts from the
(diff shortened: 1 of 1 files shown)


examples/data_ops/1110_data_ops_intro.py
```diff
@@ -1,16 +1,22 @@
...
-to build a machine learning pipeline that records all the operations involved in
-pre-processing data and training a model. We will also show how to save the model,
-load it back, and then use it to make predictions on new, unseen data.
+This example shows data wrangling for machine learning using Skrub's
+:ref:`DataOps <user_guide_data_ops_index>`.
+
+The challenge of data-wrangling for machine learning is the need to
+apply the wrangling operations to new data, for prediction.
+
+Skrub's DataOps build pipelines blending data wrangling and machine
```
rcap107:
```suggestion
Skrub's DataOps build pipelines that blend data wrangling and machine
```

examples/data_ops/1110_data_ops_intro.py
```diff
@@ -1,16 +1,22 @@
...
+learning by recording all the operations involved in pre-processing data
+and training models. They result an a full *learner* that starts from the
```
rcap107:
```suggestion
and training models. They result in an a full *learner* that starts from the
```

examples/data_ops/1110_data_ops_intro.py
```diff
@@ -1,16 +1,22 @@
...
+raw data. We will also how it can be saved, loaded it back, and then make
```
rcap107:
```suggestion
raw data. We will also how show it can be saved, loaded back, and then used to make
```

Issue #1934, opened 2026-02-27
DOC - Update the examples in the gallery so that they load datasets from their path


rcap107:
#1852 added dataset paths to the fetchers, which means that now it is possible to write 

```python
from skrub.datasets import fetch_employee_salaries
import pandas as pd

path = fetch_employee_salaries().path
df = pd.read_csv(path)
```
instead of having magical access to the dataframe from the bunch object. 

#1852 has updated some of the examples so that they use the path. All examples in the gallery that rely on skrub's fetchers should be updated so that they load from the  path rather than using the dataset loaded in the bunch. 

At some point in the future we are also going to remove loading from the fetchers entirely, so that it becomes necessary to load from the path: this is a more realistic situation anyway, and we want the examples to reflect it. 

So for this issue (or through multiple issues) we want the examples in the gallery and elsewhere in the documentation to be updated so that they all load from file rather. For some examples this is quite simple to implement, some others may require more attention/rewriting. 

The example changes in #1852 (like here https://github.com/skrub-data/skrub/pull/1852/changes#diff-050e3180cbff14a3de42224ccfae3ee7e9ae1a12eea4d38210e5535963e93e1c) can be used as templates

jeromedockes:
👍  . also, at least one of the DataOps examples (or upcoming howto sections) should show a DataOp that takes the path as an input, rather than an already-loaded dataset (the read_csv is done inside the pipeline, not outside at the start of the example)

MuditAtrey:
Hi, I'd like to work on this! I'll use the changes in #1852 as a template and start updating the gallery examples.

MuditAtrey:
my first PR (#1940) updated 4 gallery examples `0040_fuzzy_joining.py`, `0060_multiple_key_join.py`, `0070_join_aggregation.py`, `0080_interpolation_join.py` to load datasets from their file paths using `pd.read_csv()`.

Looking at the remaining examples like `examples/data_ops/1110_data_ops_intro.py` still accesses the dataframe directly from the bunch:

```python
training_data = fetch_employee_salaries(split="train").employee_salaries
```

This one also uses `split="train"`, which makes it more complex than the straightforward cases. Is this example in scope for this issue, and if so, what's the preferred approach for handling the train/test split?

rcap107:
[...]

PR #1940, opened 2026-02-28
DOC update gallery examples to load datasets from path

MuditAtrey:
Part of #1934

#1852 updated some of the gallery examples to load datasets from their
file paths. This PR updates the remaining straightforward cases that
were not covered by #1852.

The following files have been updated to use `pd.read_csv()` on the
fetcher's path field, instead of accessing dataframes directly from
the Bunch object:

- `examples/0040_fuzzy_joining.py`
- `examples/0060_multiple_key_join.py`
- `examples/0070_join_aggregation.py`
- `examples/0080_interpolation_join.py`


GaelVaroquaux:
This is an improvement. I'm merging

PR #1941, opened 2026-03-01
FIX - do not collect lazyframes in CheckInput

### LINKED ISSUE #1925: do not collect lazyframes in CheckInput
to be consistent with the rest of the library and follow the policy we chose not to collect lazyframes for the user to avoid performance surprises (in the same way we don't densify sparse arrays), CheckInput should not collect lazyframes but do no special handling and let the usual error be raised

MuditAtrey:
Closes #1925
Removes the automatic collection of Polars LazyFrames in `_check_is_dataframe.` Preivously, passing a LazyFrame would trigger a warning and call `collect()` automatically. This is now consistent with the rest of the library with no special handling, a `TypeError` is raised instead.
Updated `test_input_is_lazy` to reflect the new behaviour.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -52,6 +52,9 @@ Bug Fixes
   and :pr:`1900` by :user:`Jérôme Dockès <jeromedockes>`.
 - Errors raised when a polars LazyFrame is passed where an eager DataFrame is
   expected are now clearer. :pr:`1916` by :user:`Jérôme Dockès <jeromedockes>`.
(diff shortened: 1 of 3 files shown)


MuditAtrey:
all checks passing, does anything need changing?

rcap107:
Hello @MuditAtrey, thanks for the PR, and I'm sorry for the longer turnaround on the review for this one. It looks good to me, we can merge it.

PR #1943, opened 2026-03-05
Allow passing kwargs to cv splitter in dataops

jeromedockes:
When calling `mark_as_X()` to identify a node as the X table to be split for cross-validation, we can choose a CV splitter, and additional arguments to be passed to `split()`. Those are evaluated at the same time as X and y, then used to split X and y, then the rest of the pipeline is cross-validated on those splits.

This allows for example splitting while providing groups to ensure generalization across groups. Note that in the example below the groups (like X and y) are not known in advance but loaded / constructed inside of the pipeline

```python
import skrub

df = skrub.datasets.toy_products()
print(df)
```

```
   description  price            seller     category
0        mouse     10   supermarket.com  electronics
1       hammer     15  bestproducts.com        tools
2     keyboard     20   supermarket.com  electronics
3      usb key      9  bestproducts.com  electronics
4      charger     13  bestproducts.com  electronics
5  screwdriver     12   supermarket.com        tools
```

```python
from sklearn.dummy import DummyClassifier
from sklearn.model_selection import LeaveOneGroupOut

data = skrub.var("df", df)
groups = data["seller"]
X = data[["description", "price"]].skb.mark_as_X(
    cv=LeaveOneGroupOut(), split_kwargs={"groups": groups}
)
y = data["category"].skb.mark_as_y()
pred = X.skb.apply(DummyClassifier(), y=y)
split = pred.skb.train_test_split()

# The train set only contains data from the "supermarket.com" seller.

print(f"train:\n{split['X_train']}")

# The test set only contains data from the "bestproducts.com" seller.

print(f"test:\n{split['X_test']}")
```

```
train:
   description  price
0        mouse     10
2     keyboard     20
5  screwdriver     12
test:
  description  price
1      hammer     15
3     usb key      9
4     charger     13
```

graph

### DIFF
--- a/skrub/_data_ops/_estimator.py
+++ b/skrub/_data_ops/_estimator.py
@@ -705,10 +705,13 @@ def cross_validate(learner, environment, *, keep_subsampling=False, cv=None, **k
         :meth:`DataOp.skb.subsample`), use a subsample of the data. By
         default subsampling is not applied and all the data is used.
 
-    cv : int, cross-validation generator or iterable, default=None
-        The default is 5-fold without shuffling. Can be a cross-validation
-        splitter, an iterable yielding pairs of (train, test) indices, or an
-        int to specify the number of folds for KFold splitting.
+    cv : int, cross-validation iterator or iterable, default=None
+        Cross-validation splitting strategy. It can be:
+
+        - None: 5-fold (stratified) cross-validation
+        - integer: specify the number of folds
+        - sklearn `CV splitter <https://scikit-learn.org/stable/modules/cross_validation.html#cross-validation-iterators>`_
+        - iterable yielding (train, test) splits as arrays of indices.
 
     kwargs : dict
         All other named arguments are forwarded to


skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2587,30 +2588,71 @@ def cross_validate(self, environment=None, *, keep_subsampling=False, **kwargs):
...
+
+            - None: 5-fold (stratified) cross-validation
+            - integer: specify the number of folds
+            - sklearn `CV splitter <https://scikit-learn.org/stable/modules/cross_validation.html#cross-validation-iterators>`_
+            - iterable yielding (train, test) splits as arrays of indices.
+
+        split_kwargs : dict or None, default=None
+            Named arguments for the ``split()`` function (such as groups for a
+            :class:`~sklearn.model_selection.GroupKFold`). Note that
+            ``split_kwargs`` (and ``cv``) can themselves be DatOps.
```
```suggestion
            ``split_kwargs`` (and ``cv``) can themselves be DataOps.
```

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2587,30 +2588,71 @@ def cross_validate(self, environment=None, *, keep_subsampling=False, **kwargs):
...
+
         Returns
         -------
-        The input DataOp, which has been marked as being ``X``
+        A new DataOp, which has been marked as being the ``X`` table (which
```
rcap107, line 2629: maybe this should not be called a table, since data ops can be used with non-tabular data like images?

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2657,9 +2697,60 @@ def mark_as_X(self):
...
-        new = self._data_op._skrub_impl.__copy__()
+        if cv is None:
+            if split_kwargs is not None:
+                raise TypeError(
+                    "To pass split_kwargs you must also provide a splitter. "
+                    f"Got cv=None and split_kwargs={split_kwargs}."
+                )
+            new = self._data_op._skrub_impl.__copy__()
+        else:
+            new = SplitX(self._data_op, cv, split_kwargs)
```
rcap107, line 2753: Just to understand here: if the cv is none, then we just return the regular DataOp (or raise an error if needed), while if cv is not none then SplitX is created to hold the data op and the arguments, and acts as kind of a "sentinel" to the later parts of the evaluation so that the other machinery is aware of the custom cv and its kwargs?

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2587,30 +2588,71 @@ def cross_validate(self, environment=None, *, keep_subsampling=False, **kwargs):
...
         -------
-        The input DataOp, which has been marked as being ``X``
+        A new DataOp, which has been marked as being the ``X`` table (which
+        must be split for cross-validation).
 
         See also
         --------
+        :meth:`DataOp.skb.mark_as_y`
+            The equivalent of this function for the targets: mark a node as
+            being the ``y`` table.
```
rcap107, line 2636: same comment about the table

skrub/_data_ops/_inspection.py
```diff
@@ -297,11 +298,11 @@ def _node_kwargs(data_op, url=None):
         "fontname": "sans-serif",
         "color": "black",
     }
-    if data_op._skrub_impl.is_X:
-        label = f"X: {label}"
+    if impl.is_X:
+        label = "X" if isinstance(impl, SplitX) else f"X: {label}"
```
rcap107: why is the label removed here?

skrub/_data_ops/_estimator.py
```diff
@@ -582,25 +586,85 @@ def _find_Xy(data_op):
...
+            nodes["y"],
+            values["X"],
+            values["y"],
             "fit_transform",
             environment,
             msg=msg,
         )
-    return Xy_values["X"], Xy_values.get("y")
+    values.setdefault("y")
+    return values
```
rcap107, line 608: still trying to follow along 

the change from Xy_values["y"] to values.setdefault("y") is because now I want to return values, which contains both X and y, but it can also contain the cv and split_kwargs, right?

skrub/_data_ops/_estimator.py
```diff
@@ -694,15 +763,19 @@ def cross_validate(learner, environment, *, keep_subsampling=False, **kwargs):
...
         X,
         y,
+        cv=splitter,
         **kwargs,
     )
     if (fitted_learners := result.pop("estimator", None)) is not None:
         result["learner"] = [_to_env_learner(p) for p in fitted_learners]
+    if (indices := result.pop("indices", None)) is not None:
+        result["train_indices"] = indices["train"]
+        result["test_indices"] = indices["test"]
```
rcap107, line 781: wasn't this part also done in a different PR? or is it a separate change

skrub/_data_ops/tests/test_estimators.py
```diff
@@ -622,6 +628,91 @@ def split(X, shuffle):
...
+    assert split["X_test"].shape[0] == 30
+
+    # Check for unsupervised estimator (train/test split without y)
+
+    split = features_with_groups.skb.train_test_split()
+    assert split["X_train"].shape[0] == 50
+
+    split = features_no_groups.skb.train_test_split()
+    assert split["X_train"].shape[0] == 75
+
```
rcap107, line 724: why does this have to be tested? isn't this case covered by the assertions over `pred_with_groups` and `pred_no_groups`?

jeromedockes: yes those last 4 statements are for the unsupervised case (it works the same, it is mostly for test coverage of a "if y is None" line somewhere -- see the comment a few lines above

skrub/_data_ops/_estimator.py
jeromedockes: yes, before this function returned a tuple X, y. now there are also a splitter and kwargs to return so I changed the return type to a dict. so before we would grab the keys X and y (defaulting to None for y if it is missing) and return in a tuple, now we just fill the value for y as None if it is missing and return the full dict

skrub/_data_ops/_estimator.py
jeromedockes: yes you are right it is in #1953 . I just copied the patch here in the meanwhile because I want to use return_indices=True for the unit tests in this PR. once 1953 is merged this should disappear from the diff

skrub/_data_ops/_inspection.py
jeromedockes: I removed the wrong half of the string 🤣  what we want is to remove the "X:" part because it is redundant in this case but I did the opposite

skrub/_data_ops/_skrub_namespace.py
jeromedockes: makes sense, any suggestion for another term?

skrub/_data_ops/_skrub_namespace.py
jeromedockes: yes:

- when there are a splitter and kwargs, we need to create a new node that groups together X the splitter and the kwargs, so we create the SplitX node. when splitting data we can find this node and grab from there X, the splitter and kwargs (and y separately) and evaluate them so we have what we need to create the splitted data
- when there are no splitter and kwargs, we could do the same thing (and just have splitter and kwargs be None in the created SplitX node). however I thought it might be better to avoid adding a node to the graph and complexifying it in the very common case where the user does not specify a splitter here. so when splitter is None we keep the current approach from the main branch, which is to return a copy of the input node, in which we have set the `is_X` attribute so that it can be found when we want to split the data. we could challenge that choice and always create an additional node but I think it's good that less advanced users who haven't chosen a splitter don't have to wonder what is that SplitX node and can look at a slightly simpler graph. also it is consistent with the treatment of mark_as_y

skrub/_data_ops/_skrub_namespace.py
rcap107: I don't have a good suggestion, so I think we shouldn't worry about it in this particular PR. I opened #1960  to not forget about this.

skrub/_data_ops/_estimator.py
```diff
@@ -641,6 +705,11 @@ def cross_validate(learner, environment, *, keep_subsampling=False, **kwargs):
         :meth:`DataOp.skb.subsample`), use a subsample of the data. By
         default subsampling is not applied and all the data is used.
 
+    cv : int, cross-validation generator or iterable, default=None
+        The default is 5-fold without shuffling. Can be a cross-validation
+        splitter, an iterable yielding pairs of (train, test) indices, or an
+        int to specify the number of folds for KFold splitting.
```
glemaitre: `KFold` could point to the scikit-learn doc

skrub/_data_ops/_estimator.py
```diff
@@ -641,6 +705,11 @@ def cross_validate(learner, environment, *, keep_subsampling=False, **kwargs):
         :meth:`DataOp.skb.subsample`), use a subsample of the data. By
         default subsampling is not applied and all the data is used.
 
+    cv : int, cross-validation generator or iterable, default=None
+        The default is 5-fold without shuffling. Can be a cross-validation
```
glemaitre: I'm wondering if we should somehow link to either:

- https://scikit-learn.org/stable/modules/cross_validation.html#cross-validation
- https://scikit-learn.org/stable/glossary.html#term-CV-splitter

glemaitre: We could be consistent with the docstring that you use in the file below.

jeromedockes: thanks @glemaitre , I copied the other docstring here which is linking to the scikit-learn cv guide

PR #1952, opened 2026-03-09
Adding "None" to the list of strings recognized as null values in Cleaner

lisaleemcb:
# Bug Fix Pull Request

## Description

added the string 'none' and 'None' to the list of strings to be marked as null values, STR_NA_VALUES, and fixed a small typo

Closes #1932 

## Checklist
- [ ] I have added tests that verify the bug fix
- [ ] I have added an entry to CHANGES.rst describing the fix
- [ ] My code follows the code style of this project
- [ ] I have checked my code and corrected any misspellings

## AI Disclosure
- [ ] This PR contains AI-generated code
    - [ ] I have tested the code generated in my PR
    - [ ] I have read and understood every line that has been generated by the AI agent
    - [ ] I can explain what the AI-generated code does

PR #1954, opened 2026-03-10
exposed the list of values to be considered as null strings in the Cleaner

lisaleemcb:
Added the ability to input custom strings to be considered as nulls to the `Cleaner` class

## Description

Custom strings can now be inputted as nulls to be removed when using the `Cleaner`.

Example usage:

```
list_with_nulls = ["foo", "bar", "baz", "None", "none"]
df = pd.DataFrame(list_with_nulls, columns=["values"])
c = Cleaner(null_strings=["foo", "alpha"])
c.fit_transform(df)
```

Closes #1952 

## Checklist

## How Has This Been Tested?

Currently missing tests

## AI Disclosure

### DIFF
--- a/skrub/_clean_null_strings.py
+++ b/skrub/_clean_null_strings.py
@@ -250,20 +250,20 @@ class CleanNullStrings(SingleColumnTransformer):
     def fit_transform(self, column, y=None):
         del y
 
-        if self.null_strings is not None:
-            if isinstance(self.null_strings, str):
-                self.null_strings_ = STR_NA_VALUES + [self.null_strings]
-            elif isinstance(self.null_strings, Sequence) and all(
-                isinstance(item, str) for item in self.null_strings
-            ):
-                self.null_strings_ = STR_NA_VALUES + list(self.null_strings)
-            else:
-                raise TypeError(
-                    "Expected null_strings to be either a string or a sequence of strictly strings, "
-                    f"\n         got: {self.null_strings}."
-                )
-        else:
+        if self.null_strings is None:
             self.null_strings_ = STR_NA_VALUES.copy()
+        elif isinstance(self.null_strings, str):
+            self.null_strings_ = STR_NA_VALUES + [self.null_strings]
+        elif isinstance(self.null_strings, Sequence) and all(
+            isinstance(item, str) for item in self.null_strings
+        ):
+            self.null_strings_ = STR_NA_VALUES + list(self.null_strings)
+        else:
+            raise TypeError(
+                "Expected null_strings to be either a string or a sequence of strictly strings, "
+                f"\n         got: {self.null_strings}."
+            )
+
         if not (sbd.is_pandas_object(column) or sbd.is_string(column)):
             raise RejectColumn(f"Column {sbd.name(column)!r} does not contain strings.")
         return self.transform(column)


jeromedockes:
Thanks @lisaleemcb  for starting on this! just to be sure I understand: we decided that for now you can add to the default list but not override it altogether or remove anything from it right?

also rather than isinstance(list) we can check for something a bit more general like collections.abc.Sequence and then convert to a list ourselves so that for example a tuple would work as well

lisaleemcb:
Hi @jeromedockes , yes that is the current implementation but we can consider adding that feature as well if you would like!

And thanks for the suggestion, I'll work on implementing the change.

jeromedockes:
> Hi @jeromedockes , yes that is the current implementation but we can consider adding that feature as well if you would like!

I would keep it like this for now :)

skrub/_clean_null_strings.py
```diff
@@ -234,8 +236,28 @@ class CleanNullStrings(SingleColumnTransformer):
...
     def fit_transform(self, column, y=None):
         del y
+
+        if self.null_strings is not None:
+            if isinstance(self.null_strings, str):
+                self.null_strings_ = STR_NA_VALUES + [self.null_strings]
+            elif isinstance(self.null_strings, Sequence) and all(
+                isinstance(item, str) for item in self.null_strings
+            ):
+                self.null_strings_ = STR_NA_VALUES + self.null_strings
```
jeromedockes:
```suggestion
                self.null_strings_ = STR_NA_VALUES + list(self.null_strings)
```

or `[*STR_NA_VALUES, *self.null_strings]`

skrub/_clean_null_strings.py
```diff
@@ -234,8 +236,28 @@ class CleanNullStrings(SingleColumnTransformer):
...
+            else:
+                raise TypeError(
+                    "Expected either a string or a sequence of strictly strings."
```
jeromedockes: maybe something like `f"Expected null_strings to be a string or a list of strings, got: {self.null_strings}"`

skrub/_clean_null_strings.py
```diff
@@ -234,8 +236,28 @@ class CleanNullStrings(SingleColumnTransformer):
...
+                )
+        else:
+            self.null_strings_ = STR_NA_VALUES
```
jeromedockes:
```suggestion
            self.null_strings_ = STR_NA_VALUES.copy()
```

skrub/_table_vectorizer.py
```diff
@@ -509,6 +519,10 @@ class TableVectorizer(TransformerMixin, BaseEstimator):
     datetime_format : str, default=None
         The format to use when parsing dates. If None, the format is inferred.
 
+    null_strings : str or sequence of str, default=None
+        Additional strings to consider as null values, beyond the default list
+        in ``STR_NA_VALUES``.
```
jeromedockes: STR_NA_VALUES is not exposed to users so we should reword this sentence to just say the default list without saying what it is, or maybe actually show the list somewhere

skrub/_table_vectorizer.py
```diff
@@ -801,6 +816,7 @@ def __init__(
         self.drop_if_constant = drop_if_constant
         self.drop_if_unique = drop_if_unique
         self.datetime_format = datetime_format
+        self.null_strings = null_strings
```
jeromedockes, line 817: we also need to pass it to _get_preprocessors for it to have an effect, in `_make_pipeline()` defined below

skrub/_clean_null_strings.py
```diff
@@ -234,8 +241,29 @@ class CleanNullStrings(SingleColumnTransformer):
...
+    def __init__(
+        self,
+        null_strings=None,
+    ):
+        self.null_strings = null_strings
+
     def fit_transform(self, column, y=None):
         del y
+
+        if self.null_strings is not None:
```
jeromedockes: I think we can avoid one level of nesting and simplify a bit with something like

```python
if self.null_strings is None:
    self.null_strings_ = STR_NA_VALUES.copy()
elif:
    ...
else:
    raise
```

or otherwise we can make the default `()` (the empty tuple) and not have to deal with the `None` case at all, as you prefer

PR #1955, opened 2026-03-10
Adding cache support to the text encoder

emassoulie:
Added a variant to the `TextEncoder`'s `vectorize` method, called `vectorize_with_cache`, that stores the values it computes in a custom parquet file for fast reuse.


jeromedockes:
I guess this supersedes #1555 ?

PR #1962, opened 2026-03-13
Implementing various doc and refactoring changes for the new ApplyToCols

### LINKED ISSUE #1926: Leftover fixes for ApplyToCols
After merging #1919 and #1851 there are still some parts of the ApplyToCols code and documentation that need to be fixed. 

- [ ] Rename `_apply_on_each_col.py` and `_apply_sub_frame.py` to `_apply_to_each_column` and `_apply_to_subframe` for consistency
- [ ] The content of the docstrings of  `ApplyToEachCol` an `ApplyToSubFrame` (specifically the examples) need to be migrated to the docstring of `ApplyToCols`
- [ ]  Both Apply transformers have additional attributes that need to be copied in ApplyToCols (used_inputs, all_outputs etc). The docstring for these attributes has already been added by mistake in #1851, but we still need it so it's not a big deal. 
- [ ] There should be an informative error when a user tries to get `transformer_` when ApplyToEachCol was used. For example "you tried to access transformer_ but this is ApplyToEachCol so it only has `transformerS_`. And viceversa for ApplyToSubFrame. 
- [ ] Docstrings need to be updated again to reconcile SingleColumnTransformer and the new ApplyToCols

These are relatively minor changes, but we need to deal with them before releasing.

rcap107:
Closes #1926

### DIFF
--- a/skrub/_single_column_transformer.py
+++ b/skrub/_single_column_transformer.py
@@ -62,7 +62,7 @@ class RejectColumn(ValueError):
     passed a column that does not contain strings, or that contains strings but none
     of them look like dates.
 
-    :class:`ApplyToCols` relies on these exceptions to whether a column should be
+    :class:`ApplyToCols` relies on these exceptions to decide whether a column should be
     transformed or passed through unchanged.
 
     How these rejections are handled depends on the ``allow_reject`` parameter.


rcap107:
I discussed this PR IRL with @jeromedockes. There are some leftover points to address: 
- we want to make ApplyToEachCol and ApplyToSubFrame private, so that users only need to learn about ApplyToCols and not worry about the difference between the two. 
- this means that the .skb.apply method needs to be fixed so that it returns ApplyToCols as estimator rather than one of the other two (but this will be done in a bug fix release)
- the documentation should be simplified: the point of this change was to avoid that users would have to understand the difference between the two Apply* objects, so we should avoid having to explain that part for most examples (it would defeat the point)

CHANGES.rst
```diff
@@ -39,10 +39,10 @@ Changes
   or on all selected columns as a single dataframe. In most cases, this replaces
   the original ``ApplyToCols`` and ``ApplyToFrame``. As a result, ``ApplyToCols``
   and ``ApplyToFrame`` have been renamed :class:`ApplyToEachCol` and
-  :class:`ApplyToSubFrame` respectively.
+  :class:`ApplyToSubFrame` respectively, and are now private.
```
jeromedockes: ApplyToCols and ApplyToFrame have been merged into a single class ApplyToCols that covers the functionality of both the old classes by detecting automatically whether the provided transformer should be applied independently on each column,  or on all selected columns as a single dataframe.



to users, private internals don't exist so renamed + made private is the same as removed. also now you put all the parameters in applytocols it replaces the old ones in all cases

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,16 +1,23 @@
...
+
 Advanced columnwise operations
 ------------------------------
 
 
 Rejection handling with ``ApplyToCols``
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-:class:`ApplyToCols` allows flexible manipulation of dataframes by automatically
+|ApplyToCols| allows flexible manipulation of dataframes by automatically
```
jeromedockes: not introduced by this PR but I think it should be fixed before the release: the content is out of order here. AFAICT this is the first time we are hearing about ApplyToCols in the user guide and we start with column rejection before an introduction to the class in general and basic usage. We have a link to basic usage but it should come before the advanced topic not after

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,16 +1,23 @@
...
 applying any of Skrub's columnwise transformers to multiple columns in any given dataframe.
 More information can be found on this class's basic usage on :ref:`apply_to_each_col`.
 
 If the input columns are unable to be transformed, a specific exception exists to indicate
-this: :class:`RejectColumn`. It can be called with a custom user message, to explain the rejection.
+this: |RejectColumn|. It can be called with a custom user message, to explain the rejection.
```
jeromedockes:
```suggestion
this: |RejectColumn|.
```

this is the case for almost all exceptions

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -67,17 +74,17 @@ dtype: object
 The single column transformer
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-In cases where the user wants to apply a custom transformation to a series and needs the ApplyToCols
+In cases where the user wants to apply a custom transformation to a series and needs the |ApplyToCols|
```
jeromedockes: "you" or "we" rather than "the user"

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -37,33 +42,107 @@ All multi-column transformers provided by skrub can take skrub selectors as
...
+>>> pca = ApplyToCols(PCA(n_components=2), cols=["a", "b"])
+>>> pca.fit_transform(df).round(2)
+       c       d  pca0  pca1
+0    0.0     0.0 -2.52  0.67
+1    0.0     0.0  7.50  0.00
+2  100.0     0.0 -2.49 -0.33
+3    0.0  1000.0 -2.49 -0.33
+
+
+If |ApplyToCols| is used with a transformer that inherits from
```
jeromedockes, line 70: I get why it is in this order but "most skrub transformers belong to this category" here seems to contradict a bit the "by default" above, maybe we can tweak the wording a bit

doc/modules/multi_column_operations/multi_column_operations.rst
```diff
@@ -37,33 +42,107 @@ All multi-column transformers provided by skrub can take skrub selectors as
...
+>>> df_enc.dtypes
+date_1    datetime64[...]
+date_2    datetime64[...]
+dtype: ...
+
+
+Dealing with columns that cannot be handled by a transformer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+|ApplyToCols| can raise a |RejectColumn| exception if it cannot handle a specific
```
jeromedockes: applytocols does not raise the exception.

it provides a way to let the encoder decide which columns it can be applied to. for example, if we do not know in advance which columns can be transformed to datetime, we can use applytocols to map ToDatetime to all columns in a dataframe and pass allow_reject=True. in that case, non-datetime columns will be left unchanged. By default, all columns in `cols` must be transformed and if one of them cannot be transform it will result in an error

doc/modules/multi_column_operations/selectors.rst
```diff
@@ -49,7 +49,7 @@ Using selectors:
     1      420.0     297.0
 
 * **transformers**: various transformers in skrub use selectors to select and transform columns
-  in a scikit-learn pipeline: :class:`~skrub.ApplyToEachCol`, :class:`~skrub.ApplyToSubFrame`,
+  in a scikit-learn pipeline: :class:`~skrub.ApplyToCols`, :class:`~skrub.ApplyToSubFrame`,
```
jeromedockes: applytosubframe should be removed

skrub/_apply_to_cols.py
```diff
@@ -32,25 +38,22 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
-          transformer, which fits a separate clone of `transformer` to each
-          column in `cols`.
-        - "frame" means `transformer` is wrapped in a :class:`ApplyToSubFrame`
-          transformer, which fits a single clone of `transformer` to the
-          selected part of the input dataframe.
         - "auto" chooses the wrapping depending on the input and transformer.
           If the transformer has a ``__single_column_transformer__`` attribute,
           "cols" is chosen. Otherwise "frame" is chosen.
+        - "cols" means `transformer` is cloned and fitted separately to each
+          column in `cols`.
```
```suggestion
          column in `cols` (possibly in parallel).
```

skrub/_apply_to_cols.py
```diff
@@ -32,25 +38,22 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
+        - "frame" means `transformer` is fitted on all columns in `cols` together.
```
jeromedockes: , provided as a single dataframe

skrub/_apply_to_cols.py
```diff
@@ -71,17 +74,192 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
+        The names of columns that were transformed.
+
+    all_outputs_ : list of str
+        All column names in the output dataframe.
+
+    created_outputs_ : list of str
+        The names of columns in the output dataframe that were created by one
+        of the fitted transformers.
+
+    transformers_ : dict
```
jeromedockes, line 86: here we should try to help the user understand this super weird thing where 2 types have been mashed into one so the attributes that exist depend on its mood. I'm not sure what the sphinx template will let us do though, I don't think we can have sections 🤔

skrub/_apply_to_cols.py
```diff
@@ -71,17 +74,192 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
+    ...     A=[-10., 10.], B=[-10., 0.], C=["Paris", "Rome"],
+    ...     D=pd.to_datetime(["2024-05-13T12:05:36", "2024-05-15T13:46:02"]))
+    ... )
+    >>> df
+        A     B      C                   D
+    0 -10.0 -10.0  Paris 2024-05-13 12:05:36
+    1  10.0   0.0   Rome 2024-05-15 13:46:02
+
+
+    By default, the same transformer is applied to all selected columns. It is
```
jeromedockes: here also I don't think 'by default' is the right way to think about it

jeromedockes: also we are emphasizing too much this distinction, the first example should be just about 'look this transformer was applied to some columns and the others were kept unchanged' -- and propably it should be a single column transformer as that is the most common use case. next we can point out: if we have a transformer for single columns (most skrub encoders) like the DatetimeEncoder, a separate clone is applied to each column and you can find them in transformers_. for transformers that take 2d inputs (most scikt-learn transformers), all the selected columns are passed as a single dataframe to one transformer, and it is available as transformer_. the selection of the correct mode is done automatically by inspecting the transformer but you can force it with the how parameter

skrub/_apply_to_cols.py
```diff
@@ -228,3 +415,18 @@ def get_feature_names_out(self, input_features=None):
         check_is_fitted(self)
 
         return self._wrapped_transformer.get_feature_names_out(input_features)
+
+    def __getattr__(self, name):
+        if name == "transformers_":
```
```suggestion
        if name == "transformers_" and isinstance(getattr(self, "_wrapped_transformer", None), ApplyToSubFrame)
```

we need to check that this is actually the reason. for example, it could be that the applytocols is not fitted

skrub/_apply_to_cols.py
```diff
@@ -228,3 +415,18 @@ def get_feature_names_out(self, input_features=None):
...
+            raise AttributeError(
+                "transformers_ is only available for single-column transformers, "
+                "or if how='cols'. You may be looking for transformer_ instead."
+            )  # noqa
```
jeromedockes: why did you need the noqa here?

jeromedockes:
we could also try to come up with a shorter phrase than "if transformer is a single-column transformer or how='cols'" to refer to the 2 modes of operation, it would help if we have a short way of saying 'it depends on the effective how'. we can have a fitted attribute `how_` that is either 'cols' or 'frame', or maybe say something like 'in columnwise mode' vs 'in single-frame mode', ... finding some way to distinguish the parts that are in common from those that differ clearly, and to streamline the discussion of the parts that are different

skrub/_apply_to_cols.py
rcap107: I was trying to stop the formatter from complaining and forgot to remove it after adding the line break

skrub/_apply_to_cols.py
rcap107: can I check that with 
```python
check_is_fitted(self, "transformer_")
```
or should I do it in some other way

skrub/_apply_to_cols.py
rcap107: Here is an alternative, I don't think it's much better

image

It does not seem like numpydoc supports having split groups of attributes

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
rcap107: I was thinking of adding a subsection in the introduction of the main features of skrub that mentions the basic use case of ApplyToCols

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,19 +1,75 @@
...
+
+.. |ApplyToCols| replace:: :class:`ApplyToCols`
+.. |RejectColumn| replace:: :class:`core.RejectColumn`
+.. |SingleColumnTranformer| replace:: :class:`core.SingleColumnTranformer`
+
+
 Advanced columnwise operations
 ------------------------------
 
+The single column transformer
```
rcap107, line 12: I inverted the order of the sections here to have the single column transformer first

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
rcap107: I moved the file about ApplyToCols to the section on default wrangling so now it's at the start of the user guide

doc/modules/default_wrangling/apply_to_cols.rst
rcap107: I rewrote the paragraph above to address this

skrub/_apply_to_cols.py
jeromedockes: that will raise a notfitted exception that you would have to catch so maybe not the most convenient, just getattr should be easier I think

skrub/_apply_to_cols.py
```diff
@@ -32,25 +39,11 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
-        - "auto" chooses the wrapping depending on the input and transformer.
-          If the transformer has a ``__single_column_transformer__`` attribute,
-          "cols" is chosen. Otherwise "frame" is chosen.
-
     allow_reject : bool, default=False
         Whether to allow refusing to transform columns for which the provided
         transformer is not suited, for example rejecting non-datetime columns if
-        transformer is a DatetimeEncoder. See the documentation of
-        :class:`ApplyToEachCol` for details.
+        transformer is a DatetimeEncoder. Only relevant if the transformer is a
```
jeromedockes, line 45: can you remove the obsolete mention of the "how" parameter and state what happens to rejected columns (they are kept unchanged)?

skrub/_apply_to_cols.py
```diff
@@ -71,33 +64,226 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
+        The names of columns that were transformed.
+
+    all_outputs_ : list of str
+        All column names in the output dataframe.
+
+    created_outputs_ : list of str
+        The names of columns in the output dataframe that were created by one
+        of the fitted transformers.
+
+    Other Attributes
```
jeromedockes: I like the idea but it seems sphinx simply omits the whole section from the output. let's just remove the heading it's ok, already a bit better now that the condition is simplified without the how option

skrub/_apply_to_cols.py
```diff
@@ -71,33 +64,226 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
+    1  10.0   0.0  0.000000  0.000000 2024-05-15 13:46:02
+
+    Since we selected only column "C", the transformer was applied only to that column,
+    while the other columns were left unchanged.
+
+    Scikit-learn transformers that can be applied to multiple columns at once can also
+    be used with ``ApplyToCols``. For example, to apply a
+    :class:`~sklearn.preprocessing.StandardScaler` to the numeric columns:
+
+    >>> scaler = ApplyToCols(StandardScaler(), cols=["A", "B"])
```
jeromedockes: can we use a PCA instead? eg having 3 numeric columns in the input and doing a PCA(n_components=2). I think it makes the distinction clearer

skrub/_apply_to_cols.py
```diff
@@ -208,7 +388,9 @@ def transform(self, X):
             The transformed data.
         """
 
-        check_is_fitted(self)
+        check_is_fitted(
```
jeromedockes: why do we need this change? I think the old `check_is_fitted(self)` should be enough

skrub/_apply_to_cols.py
```diff
@@ -225,6 +407,27 @@ def get_feature_names_out(self, input_features=None):
         feature_names_out : ndarray of str objects
             Transformed feature names.
         """
-        check_is_fitted(self)
+        check_is_fitted(
```
jeromedockes: same here

skrub/_apply_to_cols.py
```diff
@@ -225,6 +407,27 @@ def get_feature_names_out(self, input_features=None):
...
 
         return self._wrapped_transformer.get_feature_names_out(input_features)
+
+    def __getattr__(self, name):
+        if name == "transformers_" and isinstance(
+            getattr(self, "_wrapped_transformer", None), ApplyToSubFrame
+        ):
+            raise AttributeError(
+                "transformers_ is only available for single-column transformers. "
+                "You may be looking for transformer_ instead."
```
jeromedockes: nitpick: let's mimick the standard python wording 'Did you mean: 'transformer_'?'

skrub/_single_column_transformer.py
```diff
@@ -30,7 +30,73 @@
...
+    skrub._single_column_transformer.RejectColumn: Could not find a datetime format...
+
+    ``RejectColumn`` exceptions can be used to indicate that a column cannot be
+    handled by the current transformer: this may mean that the data is invalid,
+    or that the transformer is simply not designed to handle that type of data.
+    For example, a ``ToDatetime`` transformer might raise ``RejectColumn`` when it is
+    passed a column that does not contain strings, or that contains strings but none
+    of them look like dates.
+
+    :class:`ApplyToCols` relies on these exceptions to whether a column should be
```
jeromedockes:
```suggestion
    :class:`ApplyToCols` relies on these exceptions to decide whether a column should be
```

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -0,0 +1,240 @@
...
+.. |StandardScaler| replace:: :class:`~sklearn.preprocessing.StandardScaler`
+.. |OneHotEncoder| replace:: :class:`~sklearn.preprocessing.OneHotEncoder`
+
+.. _user_guide_multiple_columns:
+
+Operating over multiple columns at once with |ApplyToCols|
+===========================================================
+
+Very often and for various reasons, transformers must be applied to multiple
+columns at the same time. For example, all numeric columns in a dataframe may need
```
jeromedockes: I'm not sure I get "at the same time" or "multiple columns at once" -- this is how they are usually applied. rather I would say "transforming only part of a dataframe", "only some of the columns", or something along those lines

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -0,0 +1,240 @@
...
+to be scaled at the same time.
+While the heuristics used by the :class:`TableVectorizer` are usually good enough
+to apply the proper transformers to different datatypes, using it may not be an
+option in all cases. In scikit-learn pipelines, the column selection operation can
+is done with the :class:`~sklearn.compose.ColumnTransformer`.
```
jeromedockes:
```suggestion
be done with the :class:`~sklearn.compose.ColumnTransformer`.
```

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -0,0 +1,240 @@
...
+
+Skrub provides the |ApplyToCols| transformer to achieve the same results with
+a larger degree of control over which columns are being transformed.
+|ApplyToCols| maps a transformer to columns in a dataframe, so that all
+columns that satisfy a certain condition are transformed, while the others are
+left untouched.
+
+.. tip::
+
+    All multi-column transformers provided by skrub can take skrub selectors as
```
jeromedockes: many rather than all? tablevectorizer and many others don't have that parameter
or maybe the tip could be when a skrub transformer has a cols parameter to specify a column list, that can be a selector as well

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -0,0 +1,240 @@
...
+>>> from sklearn.pipeline import make_pipeline
+>>> from skrub import ApplyToCols
+>>> from sklearn.preprocessing import OneHotEncoder, StandardScaler
+>>> import pandas as pd
+>>> df = pd.DataFrame({"text": ["foo", "bar", "baz"], "number": [1, 2, 3]})
+
+
+We use the |s.numeric| and |s.string| selectors to choose the respective columns:
+
+.. admonition:: Why ``sparse_output=False``?
```
jeromedockes: can we use a ordinalencoder to avoid this discussion? unless you think this is the best place to include it.

also doing the 2 seems redundant: I think we can remove the standardscaler part

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -0,0 +1,240 @@
...
+0  c1          0.0         1.0        0.0     2023.0         1.0      15.0        1.673741e+09
+1  c2          0.0         0.0        1.0     2023.0         2.0      20.0        1.676851e+09
+2  c3          1.0         0.0        0.0     2023.0         3.0      10.0        1.678406e+09
+
+Note that the column "id" was not encoded and was instead left as-is.
+
+Dealing with columns that cannot be handled by a transformer
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+|ApplyToCols| lets the underlying encoder decide which columns it can be applied to.
```
jeromedockes:
```suggestion
|ApplyToCols| can allow the underlying encoder to decide which columns it can be applied to.
```

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -0,0 +1,240 @@
...
+transformed dataframe. The ``rename_columns`` parameter can be used to avoid
+name collisions:
+
+>>> scaler = ApplyToCols(
+...     StandardScaler(), keep_original=True, rename_columns="{}_scaled"
+... )
+>>> scaler.fit_transform(df)
+      A      B  A_scaled  B_scaled
+0 -10.0    0.0      -1.0      -1.0
+1  10.0  100.0       1.0       1.0
```
jeromedockes: I have the impression this is very similar to the examples section of the docstring (but not exactly the same, so we don't spot at once that we don't need to read it if we already read the docstring). Do you think maybe we should have these explanations in one place only, and have one point to the other?

also I think the docstring one is a bit more concise so I think I prefer it a bit, if we replace the standardscaler with a PCA there

skrub/_apply_to_cols.py
rcap107: without this,`test_check_is_fitted_missing_fitted_attribute_transform` is failing because it's not raising the notfitted exception, and I'm not sure what is going on to break that

jeromedockes: i don't understand the test. we just need to test:

- don't fit and call transform -> not fitted error
- fit with single column transformer and access transformer_ -> custom message
- fit with regular transformer and access transformers_ -> custom message
- access some_other_attribute -> standard attributeerror message

jeromedockes: check_is_fitted will just look for the presence of the attribute so check_is_fitted(transformer if hasattr transformer) does not seem like it should be there

doc/modules/default_wrangling/apply_to_cols.rst
rcap107: I'll remove it

doc/modules/default_wrangling/apply_to_cols.rst
rcap107: it's probably not the best place to do this, we should still have it somewhere, not sure where though

rcap107: the reason for having the numeric encoder was to show that it's possible to put the transformers generated with applytocols in a pipeline, but I did not explain that well.

skrub/_apply_to_cols.py
rcap107: oops, a "how" slipped through

jeromedockes:
> +.. admonition:: Why ``sparse_output=False``?
>
> it's probably not the best place to do this, we should still have it somewhere, not sure where though

maybe one of the howto / faq / explain this error thingies when we have them?

jeromedockes:
> the reason for having the numeric encoder was to show that it's possible to put the transformers generated with applytocols in a pipeline, but I did not explain that well.

yes but all transformers can be used in a pipeline it's what the pipeline is for. so I don't think we need to repeat that here and we can keep it short instead WDYT

rcap107:
> the reason for having the numeric encoder was to show that it's possible to put the transformers generated with applytocols in a pipeline, but I did not explain that well.
> yes but all transformers can be used in a pipeline it's what the pipeline is for. so I don't think we need to repeat that here and we can keep it short instead WDYT

I think it's better to make it clear that combining transformers should be done like

```python
apply1  = ApplyToCols(transformer1())
apply2 = ApplyToCols(transformer2())
make_pipeline(apply1, apply2)
```
instead of 
```python
pipe = make_pipeline(transformer1(), transformer2())
ApplyToCols(pipe)
```

rcap107:
> +.. admonition:: Why ``sparse_output=False``? it's probably not the best place to do this, we should still have it somewhere, not sure where though
> maybe one of the howto / faq / explain this error thingies when we have them?

yeah I think it makes sense to have one entry that explains the sparse error when invariably someone will smack into that

skrub/_apply_to_cols.py
```diff
@@ -71,33 +65,229 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
...
+        transformer is a :class:`~core.SingleColumnTransformer`.
+
+    output_to_input_ : dict
+        Maps the name of each column in the transformed output to the name of
+        the input column from which it was derived. Only available when the
+        transformer is a :class:`~core.SingleColumnTransformer`.
+
+    transformer_ : Transformer
+        The fitted transformer. Only available when ``how="frame"`` and the
+        transformer is not a :class:`~core.SingleColumnTransformer`.
```
```suggestion
        The fitted transformer. Only available when the transformer is **not** a :class:`~core.SingleColumnTransformer`.
```

doc/modules/default_wrangling/apply_to_cols.rst
```diff
@@ -0,0 +1,157 @@
...
+.. |RejectColumn| replace:: :class:`core.RejectColumn`
+.. |ToDatetime| replace:: :class:`ToDatetime`
+.. |SingleColumnTransformer| replace:: :class:`~skrub.core.SingleColumnTransformer`
+.. |StandardScaler| replace:: :class:`~sklearn.preprocessing.StandardScaler`
+.. |OneHotEncoder| replace:: :class:`~sklearn.preprocessing.OneHotEncoder`
+.. |make_pipeline| replace:: :class:`~sklearn.pipeline.make_pipeline`
+
+.. _user_guide_multiple_columns:
+
+Operating over multiple columns at once with |ApplyToCols|
```
jeromedockes: here also I don't get the "at once" part, rather transforming only some of the columns or something like that

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,19 +1,76 @@
...
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+In cases where we want to apply a custom transformation to a series we need the |ApplyToCols|
+structure to handle multiple columns, or if this transformation needs to be able to reject certain
+columns and communicate this to |ApplyToCols|, we must to create a transformer from scratch
+that is capable of handling this exception.
```
jeromedockes:
```suggestion
structure to handle multiple columns, and if this transformation needs to be able to reject certain
columns and communicate this to |ApplyToCols|, we must to create a transformer from scratch
that raises this exception when appropriate.
```

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,19 +1,76 @@
...
+
+Hence the ``SingleColumnTransformer`` class. It is originally a base class from which many transformers are
+inherited, but it can also be used to create new transformers. As long we specify
```
jeromedockes: I don't understand this sentence

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,19 +1,76 @@
...
+corresponding ``fit`` and ``transform`` methods to our custom transformer, it can be passed on like any other.
+
+For instance, we might want to create a custom transformer specialized in parsing zip codes of a certain
+format, that returns |RejectColumn| with a custom warning when the length of the provided
```
jeromedockes:
```suggestion
format, that raises |RejectColumn| when the length of the provided
```

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,19 +1,76 @@
...
+zip code is incorrect:
```
jeromedockes: we can reiterate here that the goal is that in this case ApplyToCols will pass the column through without transformation

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -1,19 +1,76 @@
...
+2  HS014   WB9M88
+>>> ZipcodeParser().fit_transform(df["sent"])
+  letters  numbers
+0      AB      123
+1      BD      601
+2      HS       14
+>>> ZipcodeParser().fit_transform(df["received"])
+Traceback (most recent call last):
+    ...
+skrub._single_column_transformer.RejectColumn: This transformer only takes zip codes of length 5.
```
jeromedockes: we show the exception being raised here but later we should show it resulting in the column being passed through. in the example the datetimeencoder can be replaced with the zipcode one

doc/modules/default_wrangling/apply_to_cols.rst
rcap107: "Transforming selected columns", "Transforming specific columns"?

jeromedockes: yep either of those works!

Issue #1720, opened 2025-10-30
BUG - Fix doc generation command so that it can run on Windows


rcap107:
At the moment, it is not possible to generate the documentation on Windows because we are relying on Make and makefiles. 

Scikit-learn does this using a bat file:
https://github.com/scikit-learn/scikit-learn/blob/main/doc/make.bat

We should add the equivalent to skrub.

DCchoudhury15:
@rcap107 can i work on this?

jeromedockes:
please do, @DCchoudhury15 thanks a lot for offering!

PR #1963, opened 2026-03-13
DOC Add make.bat for building documentation on Windows

DCchoudhury15:
## Description

Addresses #1720

Adds a `make.bat` script to the `doc/` folder so that Windows users
can build the documentation without needing Make/Makefiles.

The script supports the same targets as the existing `Makefile`:
- `html` - full build including gallery
- `html-noplot` - fast build without running examples
- `linkcheck` - check external links
- `linkcheck-noplot` - check links without gallery
- `clean` - remove all build output

Also updates `CONTRIBUTING.rst` to document the Windows equivalents
alongside the existing Linux/macOS commands.

## Checklist

## How Has This Been Tested?

Tested the equivalent `make html-noplot` build successfully on Linux
using `pixi run build-doc-quick`. The `make.bat` mirrors the same
Sphinx commands as the `Makefile`.

## AI Disclosure
- [] This PR contains AI-generated code
    - [] I have tested the code generated in my PR
    - [] I have read and understood every line that has been generated by the AI agent
    - [] I can explain what the AI-generated code does

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -621,15 +621,15 @@ environments:
       - pypi: https://files.pythonhosted.org/packages/89/18/58c64cafcf8eb677a99ef593121f719e6dcbdb7d1c594ae5a10d4997ca8a/fonttools-4.61.1-cp314-cp314-manylinux1_x86_64.manylinux2014_x86_64.manylinux_2_17_x86_64.manylinux_2_5_x86_64.whl
       - pypi: https://files.pythonhosted.org/packages/7b/91/984aca2ec129e2757d1e4e3c81c3fcda9d0f85b74670a094cc443d9ee949/joblib-1.5.3-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/43/dc/51acc6791aa14e5cb6d8a2e28cefb0dc2886d8862795449d021334c0df20/kiwisolver-1.4.9-cp314-cp314-manylinux2014_x86_64.manylinux_2_17_x86_64.whl
-      - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/matplotlib/3.11.0.dev1945+gb4b0e3a6d/matplotlib-3.11.0.dev1945+gb4b0e3a6d-cp314-cp314-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
+      - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/matplotlib/3.11.0.dev1835+ge3d6dfc70/matplotlib-3.11.0.dev1835+ge3d6dfc70-cp314-cp314-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
       - pypi: https://files.pythonhosted.org/packages/4b/27/20770bd6bf8fbe1e16f848ba21da9df061f38d2e6483952c29d2bb5d1d8b/narwhals-2.17.0-py3-none-any.whl
       - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/numpy/2.5.0.dev0/numpy-2.5.0.dev0-cp314-cp314-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
-      - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/pandas/3.1.0.dev0+400.gc44750bf8c/pandas-3.1.0.dev0+400.gc44750bf8c-cp314-cp314-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl
+      - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/pandas/3.1.0.dev0+243.g5b19efbb47/pandas-3.1.0.dev0+243.g5b19efbb47-cp314-cp314-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl
       - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/pillow/12.2.0.dev0/pillow-12.2.0.dev0-cp314-cp314-manylinux_2_27_x86_64.manylinux_2_28_x86_64.whl
       - pypi: https://files.pythonhosted.org/packages/52/d2/c6e44dba74f17c6216ce1b56044a9b93a929f1c2d5bdaff892512b260f5e/plotly-6.6.0-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/0a/49/737c1a6273c585719858261753da0b688454d1b634438ccba8a9c4eb5aab/polars-1.38.1-py3-none-any.whl
       - pypi: https://files.pythonhosted.org/packages/f2/5a/61d60ec5cc0ab37cbd5a699edb2f9af2875b7fdfdfb2a4608ca3cc5f0448/polars_runtime_32-1.38.1-cp310-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
-      - pypi: https://pypi.anaconda.org/scientific-python-nightly-wheels/simple/pyarrow/24.0.0.dev231/pyarrow-24.0.0.dev231-cp314-cp314-manylinux_2_28_x86_64.whl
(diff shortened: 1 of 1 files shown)


DCchoudhury15:
I was looking into the issue with Windows documentation builds and tried putting together a `make.bat` script for the `doc/` folder. I attempted to map out the standard `Makefile` targets (like `html`, `html-noplot`, `clean`, etc.) and also added the Windows equivalents to `CONTRIBUTING.rst`. 

I do have a quick question about testing this, though. Since I'm developing strictly on Linux right now, I don't actually have a local Windows environment to run this batch script natively. I tried to mirror the existing logic as closely as possible, but I'm basically hoping the Windows CI pipeline will catch any batch syntax quirks I might have missed. Does this approach work for this issue? I'll keep a close eye on the action logs and push fixes if the CI complains, but let me know if I should be handling this differently or if I missed any specific cache folders in the clean target!

rcap107:
Hi @DCchoudhury15, thanks for the PR. The CI is unlikely to spot any issues with the file because the docs are built with the regular makefile and the test matrix is only checking for the actual test suite. 

I think there isn't an alternative to doing it the manual way and try to run it somehow on a  machine that runs Windows. I'll try to do that to review the file when I can.

DCchoudhury15:
Thanks for the clarification @rcap107! I'm going to set up a Windows VM 
locally to test this properly and will report back with the results. 
I'll push any fixes needed based on what I find.

rcap107:
Hi @DCchoudhury15, before merging this PR there should also be an update to the pyproject.toml file to add target specific commands. Pixi allows to do this using https://pixi.prefix.dev/latest/workspace/multi_platform_configuration/#activation 

There should be a specific entry for building the documentation with the bat file if the platform is win64.

DCchoudhury15:
Hi @rcap107!

I tested the `make.bat` script on a real Windows 11 machine with Python 3.13.4, Sphinx 7.4.7, and Graphviz 14.1.3, and the core targets are working great. The `help` command lists everything correctly, and running `html-noplot` successfully built the docs with just the pre-existing warnings. I opened the generated `index.html` in the browser and everything rendered properly. The `clean` command correctly wiped the `_build`, `auto_examples`, and generated reference folders. I also ran `linkcheck-noplot` which passed fine, only catching broken links that were already there. I skipped the full `html` build just because it takes a long time, but since `html-noplot` passed cleanly i presume it should work exactly the same hopefully.

I also added the Windows specific pixi tasks to `pyproject.toml` under `[tool.pixi.feature.doc.target.win-64.tasks]`. I made sure to verify on Linux that the existing pixi doc commands still work fine after the change. I did run into one weird issue on Windows though. When running the pixi `build-doc-quick` command there, pixi still seemed to try and execute `make` instead of `make.bat`. I am not sure if my syntax is slightly off or if the base feature-level tasks are overriding the target-level ones. Do you know the right way to configure this override in pixi?

doc/make.bat
```diff
@@ -0,0 +1,96 @@
...
+)
+if exist "generated_for_index\" (
+    rmdir /q /s "generated_for_index"
+    echo. Removed generated_for_index\
+)
+if exist "reference\" (
+    for %%i in ("reference\*.rst") do del /q "%%i" 2>nul
+    echo. Removed reference\*.rst
+)
+goto end
```
rcap107, line 65: I noticed that more files are being cleaned here than in the original makefile. That's good! Could you update the linux makefile to match?

pyproject.toml
```diff
@@ -227,6 +227,13 @@ clean-doc = { cmd = "rm -rf _build/ && rm -rf auto_examples/ && rm -rf reference
 linkcheck = { cmd = "make linkcheck", cwd = "doc" }
 linkcheck-quick = { cmd = "make linkcheck-noplot", cwd = "doc" }
 
+[tool.pixi.feature.doc.target.win-64.tasks]
+build-doc = { cmd = "make.bat html", cwd = "doc" }
+build-doc-quick = { cmd = "make.bat html-noplot", cwd = "doc" }
+clean-doc = { cmd = "make.bat clean", cwd = "doc" }
+linkcheck = { cmd = "make.bat linkcheck", cwd = "doc" }
+linkcheck-quick = { cmd = "make.bat linkcheck-noplot", cwd = "doc" }
```
rcap107: This should fix the windows targets
```suggestion
build-doc = { cmd = ".\\make.bat html", cwd = "doc" }
build-doc-quick = { cmd = ".\\make.bat html-noplot", cwd = "doc" }
clean-doc = { cmd = ".\\make.bat clean", cwd = "doc" }
linkcheck = { cmd = ".\\make.bat linkcheck", cwd = "doc" }
linkcheck-quick = { cmd = ".\\make.bat linkcheck-noplot", cwd = "doc" }
```

DCchoudhury15:
Hi @rcap107!

I just pushed the updates for both of your suggestions. I changed the `win-64` tasks in `pyproject.toml` to use `.\\make.bat` instead of just `make.bat`, and after testing on Windows, this completely fixed the issue. `pixi run --environment doc build-doc-quick` now correctly fires off `.\make.bat html-noplot`. I also updated the `clean` target in the Linux `Makefile` so it matches the Windows one perfectly, wiping out `auto_examples/`, `generated/`, `generated_for_index/`, and `reference/generated/`. I verified this on my Linux setup and everything clears out properly when running the pixi clean command now.

I ran the full `make.bat html` build to test that optuna example. Almost everything ran perfectly (17 out of 18 examples passed), but `1131_optuna_choices.py` did fail with a `WinError 1314` regarding missing privileges. 

From looking at the traceback, it seems `optuna` tries to create a symlink for a file lock when `search.fit(env)` is called. I guess Windows blocks symlink creation unless you have admin rights or Developer Mode enabled? Since this seems to be a Windows-specific quirk with how `optuna` handles its storage backend and probably not an issue with our `make.bat` script, do you think we are good to just proceed as is? I imagine the Linux CI ran it without any issues since symlinks work out of the box there, but I would love to hear what you think!

pyproject.toml
```diff
@@ -223,10 +223,17 @@ python = "~=3.14.0"
 [tool.pixi.feature.doc.tasks]
 build-doc = { cmd = "make html", cwd = "doc" }
 build-doc-quick = { cmd = "make html-noplot", cwd = "doc" }
-clean-doc = { cmd = "rm -rf _build/ && rm -rf auto_examples/ && rm -rf reference/generated/", cwd = "doc" }
+clean-doc = { cmd = "rm -rf _build/ && rm -rf auto_examples/ && rm -rf generated/ && rm -rf generated_for_index/ && rm -rf reference/generated/", cwd = "doc" }
```
rcap107: Simplifying a bit since the command is the same in the makefile
```suggestion
clean-doc = { cmd = "make clean", cwd = "doc" }
```

rcap107:
Hi @DCchoudhury15, thanks a lot for the fixes.

> I ran the full `make.bat html` build to test that optuna example. Almost everything ran perfectly (17 out of 18 examples passed), but `1131_optuna_choices.py` did fail with a `WinError 1314` regarding missing privileges.
> 
> From looking at the traceback, it seems `optuna` tries to create a symlink for a file lock when `search.fit(env)` is called. I guess Windows blocks symlink creation unless you have admin rights or Developer Mode enabled? Since this seems to be a Windows-specific quirk with how `optuna` handles its storage backend and probably not an issue with our `make.bat` script, do you think we are good to just proceed as is? I imagine the Linux CI ran it without any issues since symlinks work out of the box there, but I would love to hear what you think!

Yes, that's the same problem that I saw. I think we can proceed as is, but I think there should be a mention of this problem in the contribution doc where the Windows command is explained, so that people that try to use it don't get surprising results. Something like 

> If you are working on Windows, building the example 1131 on Optuna may fail with a permission error, but the rest of the documentation build should run without problem. In that case, it's fine to ignore the problem if your contribution does not touch that particular example: the CI will build the example as normal in the PR. 

That should help with avoiding some confusion.

I think the PR is pretty much done and is only missing a message like the one I wrote to avoid potential confusion for new contributors. I might leave it open for a few days to let other maintainers leave comments if needed, but I'll likely merge this at the start of next week.

DCchoudhury15:
Hi @rcap107, I've added the note to CONTRIBUTING.rst right after the make.bat html section. Let me know if the wording looks good or if you'd like any changes!

rcap107:
> Hi @rcap107, I've added the note to CONTRIBUTING.rst right after the make.bat html section. Let me know if the wording looks good or if you'd like any changes!

Looks good to me, I think we can merge this after all. Thanks a lot for the help!

PR #1970, opened 2026-03-16
setting the `__module__` of types to the public module in which they are exposed

jeromedockes:
@lisaleemcb @rcap107  pointed out that the repr of some types looks a bit weird because it includes the path to a private submodule. Also, as currently the RejectColumn is defined in a private module a user seeing it in a traceback and wanting to catch it may not know where to import it from. We can set the `__module__` attribute of those types to `skrub` or `skrub.core`, where they are imported and exposed as part of the public API

before

```
>>> import polars as pl
>>> import skrub
>>> from skrub import DatetimeEncoder

>>> type(skrub.var('a'))
<class 'skrub._data_ops._data_ops.DataOp'>
>>> skrub.TableVectorizer
<class 'skrub._table_vectorizer.TableVectorizer'>
>>> DatetimeEncoder().fit_transform(pl.Series([1, 2, 3]))
    ...
skrub._single_column_transformer.RejectColumn: Column '' does not have Date or Datetime dtype.
```

after

```
>>> import polars as pl
>>> import skrub
>>> from skrub import DatetimeEncoder

>>> type(skrub.var('a'))
<class 'skrub.DataOp'>
>>> skrub.TableVectorizer
<class 'skrub.TableVectorizer'>
>>> DatetimeEncoder().fit_transform(pl.Series([1, 2, 3]))
    ...
skrub.core.RejectColumn: Column '' does not have Date or Datetime dtype.
```

I made the change just for those 3 examples for now, if we decide to make this change we would do it for all classes. Also I haven't updated any doctests so many will probably fail

### DIFF
--- a/skrub/_utils.py
+++ b/skrub/_utils.py
@@ -291,7 +291,7 @@ def set_module(module):
     >>> from skrub._utils import set_module
     >>> class A: pass
     >>> A
-    <class '__main__.A'>
+    <class 'skrub._utils.A'>
     >>> set_module('skrub')(A)
     <class 'skrub.A'>
     """


lisaleemcb:
Ah, I see! Thanks for the clarification.

glemaitre:
Nice :)

PR #1973, opened 2026-03-17
Chore - remove "use_table_report" config parameter

### LINKED ISSUE #1971: CHORE - remove the configuration entry that is setting the patch display
At the moment, we allow users to permanently replace the repr of pandas and polars dataframes so that it uses the TableReport instead of the respective default repr. While this was intended to be a replacement for the patch_display, after further discussion we realized that this is not a good solution: we are patching something in a different library by default in a non-obvious way: this is not good and we should remove this functionality.

rcap107:
## Description
This PR removes the "use_table_report" configuration parameter. This config parameter is patching over the default behavior of pandas and polars dataframes in a non-transparent way. 

Given it is a big change, it's better if the user changes this using patch_display and unpatch_display instead. 

For the moment I have kept the data ops version of this parameter, but maybe we should remove both.

### DIFF
--- a/skrub/_reporting/tests/test_patch_display.py
+++ b/skrub/_reporting/tests/test_patch_display.py
@@ -2,7 +2,7 @@ import pickle
 
 import pytest
 
-from skrub import config_context, patch_display, set_config, unpatch_display
+from skrub import patch_display, unpatch_display
 from skrub.conftest import skip_polars_installed_without_pyarrow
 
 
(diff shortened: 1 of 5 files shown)


jeromedockes:
> For the moment I have kept the data ops version of this parameter, but maybe we should remove both.

they are of a different nature:

 - `use_table_report_data_ops` decides if skrub itself should use the table report when it generates a display for a skrub dataop. I think it's ok to keep.
 - `use_table_report` when set to true will monkey-patch methods on pandas.DataFrame and polars.DataFrame, which goes beyond configuring skrub itself and we decided to remove, as you did in this PR :+1:

skrub/_reporting/tests/test_patch_display.py
```diff
@@ -54,93 +60,51 @@ def test_patch_display(df_module, repeat_patch, repeat_unpatch, capsys):
...
     assert "data-test-associations-skipped" not in df5._repr_html_()
+    unpatch_display()
 
     df6 = pd_module.make_dataframe(
         {f"col_{i}": [i + j for j in range(3)] for i in range(5)}
     )
     patch_display(max_plot_columns=None, max_association_columns=None)
     assert "data-test-plots-skipped" not in df6._repr_html_()
     assert "data-test-associations-skipped" not in df6._repr_html_()
+    unpatch_display()
```
jeromedockes, line 110: can you add something like

```python
@pytest.fixture(autouse=True)
def teardown_unpatch_display():
    yield
    unpatch_display()
```

in this module? so that the display is always unpatched in the teardown of each test of this module

(in addition to the unpatch that may be needed inside of tests as those you have here)

rcap107: it's already here, isn't this what you're asking for?

https://github.com/skrub-data/skrub/pull/1973/changes#diff-88442ded8eb580e553d008675669c794fbb2b9e4b4317e659267567508c2cdbeR10

jeromedockes: yep, sorry I missed it :)

jeromedockes (approved):

PR #1974, opened 2026-03-17
CHORE - Silencing a deprecation warning in nightlies

rcap107:
Nightlies are failing because of a pandas 4.0 deprecation warning in the joiners.

Example: https://github.com/skrub-data/skrub/actions/runs/23178427828

Given this is a long term warning and we might be done with the joiners by the time pandas 4.0 gets released, we can silence this in the meanitme.


jeromedockes:
> Given this is a long term warning and we might be done with the joiners by the time pandas 4.0 gets released

also we are fine with both the old and the new behavior

jeromedockes (approved):

PR #1975, opened 2026-03-18
Add advanced tabular pipeline guide examples

### LINKED ISSUE #1966: DOC - Improving the user guide section about the tabular pipeline
The user guide section on the tabular pipeline should be expanded with some additional examples on how to use the tabular pipeline for situations that are slightly more complex than what is currently explained.

- Show that the resulting pipeline can be modified by adding more steps
- Show how to use more specific transformers/steps
- Show that the model can be a pipeline that adds the steps needed by the user

kudos07:
## Description

Expand the user guide section on `tabular_pipeline` with examples for slightly more advanced usage.

This adds examples showing:
- how to extend the generated pipeline with additional scikit-learn steps via the `.steps` attribute and `make_pipeline`
- how to pass a scikit-learn `Pipeline` as the estimator to `tabular_pipeline`
- how to move from `tabular_pipeline` to a more explicit custom pipeline by reusing and modifying its pieces
- how to customize the `TableVectorizer`, including changing the high-cardinality encoder
- how to combine `ApplyToCols` with `TableVectorizer` to transform only a subset of columns
- that the logic inside `tabular_pipeline` is intentionally simple, so users do not lose much by writing their own pipeline when they need more control

Addresses #1966 

## Checklist

## How Has This Been Tested?

- Ran the updated example snippets locally as Python checks against the current codebase
- Verified that the documented pipeline step names and object representations match the current implementation


rcap107:
Hi @kudos07, thanks a lot for the PR. I'll review it in the next days, but at a first glance I think it should be shortened a bit. 

I'm not sure how to format this, but I am wondering if we should turn this into an example in the gallery instead: we'd start from the regular tabular pipeline and then gradually increase the complexity which is what is shown here. That would also allow showing the estimator representation of the pipeline, which isn't available in the user guide because doctests do not support that functionality.

I'd like to discuss this with the other maintainers a bit more, so you may want to put this particular PR on hold for some time.

kudos07:
Thanks for the note, that sounds fair.

I can see why this would work better as a gallery example, especially since it follows more of a step by step progression than a short guide section.

No problem keeping this on hold while you discuss it with the other maintainers. If you decide the gallery is the better place for it, I’m happy to rework it in that format and cut it down as needed.

If you want, I can also make it a bit more casual and GitHub-native.

doc/modules/default_wrangling/tabular_pipeline.rst
```diff
@@ -81,3 +84,114 @@ problems, but may not beat properly tuned ad-hoc pipelines.
...
+You can use the ``.steps`` attribute of the resulting pipeline together with
+:func:`~sklearn.pipeline.make_pipeline` to build a new pipeline that has more
+than just the steps in the tabular pipeline. This is useful when you want to
+keep skrub's default preprocessing, while inserting an extra transformation
+before the final estimator:
+
+>>> from sklearn.feature_selection import SelectPercentile, f_regression
+>>> from sklearn.pipeline import make_pipeline
+>>> from skrub import tabular_pipeline
+>>> base_pipeline = tabular_pipeline("regressor")  # doctest: +SKLEARN_VERSION >= "1.4"
```
rcap107:
```suggestion
>>> base_pipeline = tabular_pipeline("regressor")
```

sklearn 1.4 is a min req so this is unnecessary

doc/modules/default_wrangling/tabular_pipeline.rst
```diff
@@ -81,3 +84,114 @@ problems, but may not beat properly tuned ad-hoc pipelines.
...
+>>> extended_pipeline = make_pipeline(
+...     *[step[1] for step in base_pipeline.steps[:-1]],
+...     SelectPercentile(score_func=f_regression, percentile=50),
+...     base_pipeline.steps[-1][1],
+... )
+>>> [name for name, _ in extended_pipeline.steps]  # doctest: +SKLEARN_VERSION >= "1.4"
```
rcap107:
```suggestion
>>> [name for name, _ in extended_pipeline.steps]
```

doc/modules/default_wrangling/tabular_pipeline.rst
```diff
@@ -81,3 +84,114 @@ problems, but may not beat properly tuned ad-hoc pipelines.
...
+than just the steps in the tabular pipeline. This is useful when you want to
+keep skrub's default preprocessing, while inserting an extra transformation
+before the final estimator:
+
+>>> from sklearn.feature_selection import SelectPercentile, f_regression
+>>> from sklearn.pipeline import make_pipeline
+>>> from skrub import tabular_pipeline
+>>> base_pipeline = tabular_pipeline("regressor")  # doctest: +SKLEARN_VERSION >= "1.4"
+>>> extended_pipeline = make_pipeline(
+...     *[step[1] for step in base_pipeline.steps[:-1]],
```
rcap107, line 113: I don't think the list expansion is obvious to all users, so it's better to explain this in a sentence above, along with an explanation of what a pipeline is (a list of pairs (name, estimator))

doc/modules/default_wrangling/tabular_pipeline.rst
```diff
@@ -81,3 +84,114 @@ problems, but may not beat properly tuned ad-hoc pipelines.
...
+Because the result is an ordinary scikit-learn pipeline, you can take some of
+its steps, change the parameters that matter for your data, and assemble a new
+pipeline yourself.
+
+For example, you can start from the default regressor pipeline, then replace the
+high-cardinality encoder in the |TableVectorizer| with a |TextEncoder|:
+
+>>> from sklearn.base import clone
+>>> from skrub import TextEncoder, tabular_pipeline
+>>> base = tabular_pipeline("regressor")  # doctest: +SKLEARN_VERSION >= "1.4"
```
rcap107: same as above, and same below
```suggestion
>>> base = tabular_pipeline("regressor")
```

doc/modules/default_wrangling/tabular_pipeline.rst
```diff
@@ -81,3 +84,114 @@ problems, but may not beat properly tuned ad-hoc pipelines.
...
+>>> vectorizer = clone(base.named_steps["tablevectorizer"]).set_params(
+...     high_cardinality=TextEncoder()
+... )  # doctest: +SKLEARN_VERSION >= "1.4"
+>>> vectorizer  # doctest: +ELLIPSIS +SKLEARN_VERSION >= "1.4"
```
rcap107:
```suggestion
>>> vectorizer
```

ellipsis is enabled by default in the configuration

doc/modules/default_wrangling/tabular_pipeline.rst
```diff
@@ -81,3 +84,114 @@ problems, but may not beat properly tuned ad-hoc pipelines.
...
+... )
+>>> [name for name, _ in custom_pipeline.steps]
+['applytocols', 'tablevectorizer', 'ridge']
+
+This kind of composition is often enough when only a few columns need special
+treatment and the rest of the dataframe can still go through the default
+|TableVectorizer| behavior.
+
+The logic used by the tabular pipeline is quite simple
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
rcap107: this section should be moved at the start of this entire section, almost as a warning that the following steps are usually not necessary

doc/modules/default_wrangling/tabular_pipeline.rst
rcap107: re-reading this example, I _really_ don't want anything like this, it's defeating the point of using both the tabular_pipeline and the sklearn pipeline with ApplyToCols

rcap107 (commented):
Thanks for the PR @kudos07, I think it's a good start. Overall, this is mostly what I would like to have in the documentation. I think part of it could be simplified and trimmed. 

About turning this into an example, in the end we might go with a different solution instead. In #1901 we are going to add a new section to the documentation that collects self-contained material that can answer the question "how do I do this?". I think this content would fit that section more than the user guide, or the example gallery. 

So, once #1901 gets merged you should be able to merge from upstream and get the new section before moving the material you added to that part of the documentation. 

Does that make sense to you?

doc/modules/default_wrangling/tabular_pipeline.rst
kudos07: Removed the redundant doctest directive.

doc/modules/default_wrangling/tabular_pipeline.rst
kudos07: Removed the version guard here as well, thanks.

doc/modules/default_wrangling/tabular_pipeline.rst
kudos07: Removed here too. This is just for me to indicate I have removed them.

doc/modules/default_wrangling/tabular_pipeline.rst
kudos07: This one done too

doc/modules/default_wrangling/tabular_pipeline.rst
kudos07: Moved this section up so it introduces the advanced customization material, thanks.

doc/modules/default_wrangling/tabular_pipeline.rst
kudos07: @rcap107 
What I have in mind is:

- keep the .steps example as the main illustration of extending a tabular_pipeline
- keep the “pipeline as estimator” example, since it still fits the mental model of tabular_pipeline
- remove the examples that combine tabular_pipeline with ApplyToCols or manual reassembly of several pieces, since those are more confusing than helpful in this section
- if we still want to document those more advanced compositions, move them later to the new “how do I do this?” section once that exists

That way this page stays focused on how to extend tabular_pipeline without undermining the point of using it in the first place.

kudos07:
Implemented the changes in the latest update. I removed the redundant doctest directives, clarified the **.steps** example, moved the “logic is quite simple” section earlier, and removed the more confusing custom-composition examples so the page stays focused on extending **tabular_pipeline**.

doc/modules/default_wrangling/tabular_pipeline.rst
```diff
@@ -73,11 +74,72 @@ problems, but may not beat properly tuned ad-hoc pipelines.
      - Native support
      - :class:`~sklearn.impute.SimpleImputer`
 
-.. note::
```
rcap107, line 76: This note should not be removed: it's a footnote for the table above

doc/modules/default_wrangling/tabular_pipeline.rst
rcap107: it doesn't look like the section was moved, maybe you forgot to commit the change?

rcap107 (commented):
A couple leftover comments, but this PR is almost done. Thanks @kudos07

kudos07:
Thanks @rcap107 , I addressed both points in the latest update. I restored the note under the table since it is referenced by the footnotes above, and I also moved the “The logic used by the tabular pipeline is quite simple” section so it now appears before the advanced examples.

rcap107:
Hi @kudos07, I realized after re-reading with some more attention that the notes on sklearn < 1.4 were removed for a good reason, and we just forgot to remove the footnote markers in the table. I fixed that part and added a cross-reference in the same table. The other changes you made look good!

rcap107 (approved):
Looks good to me, thanks @kudos07

rcap107:
It seems that the CI is being annoying again, I'll merge this when I can

kudos07:
@rcap107 Why does the CI behave like this?

rcap107:
> @rcap107 Why does the CI behave like this?

I'm tempted to blame GitHub here. It's not the first time we've had issues with running actions in the past months, likely due to vibe coded commits clogging up. 

It's not a big deal, just annoying.

kudos07:
@rcap107 got it!

Issue #1969, opened 2026-03-16
Extending the TableReport filters so that they can use custom selectors to pick columns


rcap107:
At the moment, the TableReport allows the creation of custom filters which are defined as a dictionary where the key is the name of the filter to show in the report, and the value is the list of columns that should be selected by the given filter.

It would be good to extend this functionality so that it allows users to pass skrub selectors in addition to list of columns. Then, the TableReport would expand the list of columns under the hood and add the filter to the list of filters.

rcap107:
Closed by #1976

PR #1976, opened 2026-03-18
updated how custom filters are implemented for the Table Report

lisaleemcb:
## Description

Previously, the `custom_filters` kwarg passed to the `TableReport` required a nested dict such as

CUSTOM_FILTERS = {'FILTER ID': {filter1: [0,1,2]}}.

Additionally, the only custom filter type that could be passed was a list of ints, indicating the indices of the desired columns.

My proposed change is to 

a) streamline the user input such that only one dictionary is required, for example,
`CUSTOM_FILTERS = {filter1: selector}`

and

b) changed the required input to be a skrub Selector object.

Addresses #1969 

## TBC

## How Has This Been Tested?


skrub/_reporting/_table_report.py
```diff
@@ -45,6 +48,58 @@ def _check_max_cols(max_plot_columns, max_association_columns):
...
+    if column_filters is None:
+        return column_filters
+    for i, (display_name, filter) in enumerate(column_filters.items()):
+        if isinstance(filter, s.Selector):
+            filter_index = filter.expand_index(df)
+            custom_filters[f"custom_{i}"] = {
+                "display_name": display_name,
+                "columns": filter_index,
+            }
+            continue
```
jeromedockes: ~this will break out of the loop before processing the other filters~. also we should use `sbd` to access the columns and the shape. regarding the name, in scikit-learn functions that validate input and return a potentially transformed or sanitized version of it are often called `check_something` and skrub tends to follow that habit, too

maybe we can refactor this part slightly along the lines of

```python
def _check_col_filter(name, cols, df):
    err_msg = (
        "Custom column filters should be either a Selector or a list of column names"
        f"\n  or a list of column indices. Got a bad filter for key {name!r}: {cols!r}"
    )
    if isinstance(cols, s.Selector):
        return cols.expand_index(df)
    if not isinstance(cols, Sequence):
        raise TypeError(err_msg)
    if all(isinstance(c, str) for c in cols):
        all_col_names = set(sbd.column_names(df))
        bad_col_names = [c for c in cols if c not in all_col_names]
        if bad_col_names:
            raise ValueError(
                "The following column names passed for "
                f"filter {name!r} are not in the dataframe: {bad_col_names}"
            )
        return s.make_selector(cols).expand_index(df)
    if all(isinstance(c, numbers.Integral) for c in cols):
        bad_idx = [c for c in cols if not 0 <= c < sbd.shape(df)[1]]
        if bad_idx:
            raise ValueError(
                "The following column indices passed for "
                f"filter {name!r} are out of range: {bad_idx}"
            )
        return list(cols)
    raise TypeError(err_msg)


def _check_column_filters(column_filters, df):
    if column_filters is None:
        return None
    if not isinstance(column_filters, Mapping):
        raise TypeError(
            "column_filters should be a dict mapping names to column lists, "
            f"got object of type: {type(column_filters)}"
        )
    return {
        str(name): {
            "display_name": str(name),
            "columns": _check_col_filter(name, cols, df),
        }
        for name, cols in column_filters.items()
    }

```

skrub/_reporting/_html.py
```diff
@@ -77,9 +77,10 @@ def _get_high_association_columns(summary):
     return list(columns)
 
 
-def _get_column_filters(summary):
+def _get_column_filters(summary, column_filters):
```
jeromedockes: the column_filters parameter does not seem to be used

skrub/_reporting/_table_report.py
lisaleemcb: Okay, thanks for the feedback! I'll take a look at your suggestions.

skrub/_reporting/_html.py
lisaleemcb: Sorry, remnant from previous implementation. Thanks for the catch!

skrub/_reporting/_table_report.py
lisaleemcb: So your solution looks great, thanks for the help - I've updated the tests accordingly.

rcap107: Not necessarily related to this PR, but why would the continue there break out early? Is it doing something specific beyond skipping to the next iteration of the loop? 🤔

jeromedockes: yes you are right -- i have no idea why i said that, I should not do late reviews :D

PR #1977, opened 2026-03-19
Update docs to add link to quickstart example 

lisaleemcb:
This PR just adds a link to the first example in the Examples section of the documentation, to the User Guide section. This just boosts the visibility of the Getting Started example for new users.


rcap107:
Simple change, but very useful. Thanks a lot @lisaleemcb

PR #1984, opened 2026-03-23
Removed overplotted counts in Table Report

### LINKED ISSUE #1929: Overlap between strings in the TableReport plots for large numbers
This is a screenshot taken from the TableReport of this dataset: https://www.kaggle.com/datasets/mylesoneill/warcraft-avatar-history?select=wowah_data.csv

Image

There is a slight overlap between the number written on the first bar, and the corresponding percentage. 

This should be easy to fix by moving the number on the bar slightly to the left.

lisaleemcb:
We have decided to remove the overplotting of the value counts for the bar histograms in the TableReport. This is due to tricky formatting issues that are highly specific to the data and therefore difficult to address cleanly.

Closes #1929

### DIFF
--- a/skrub/_reporting/_plotting.py
+++ b/skrub/_reporting/_plotting.py
@@ -312,7 +312,7 @@ def value_counts(value_counts, n_unique, n_rows, color=COLOR_0):
     ax_percentage = ax.twinx()
     ax_percentage.set_ylim(ax.get_ylim())
     ax_percentage.set_yticks(ticks=range(len(counts)))
-    ax_percentage.tick_params(axis="both", length=0, pad=0)
+    ax_percentage.tick_params(axis="both", length=0)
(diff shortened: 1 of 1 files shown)


skrub/_reporting/_utils.py
```diff
@@ -87,7 +87,7 @@ def format_number(number):
     if isinstance(number, numbers.Integral):
         return f"{number:,}"
     if isinstance(number, numbers.Real):
-        # Import placed here to avoid circular import related to dispatch
+        #    Import placed here to avoid circular import related to dispatch
```
jeromedockes: not related to your pr but now the patch option has been removed this should not be the case anymore, we can move the import to the top

lisaleemcb: Updated!

PR #1985, opened 2026-03-23
fix: suppress matplotlib interactive display when generating report plots

### LINKED ISSUE #1968: ENH - Avoid showing histogram plot in TableReport in interactive mode
When matplotlib is used in interactive mode, creating a `TableReport` will create plots while we are not interested to show them outside of the report itself.

It could be handy to use the following context manager when creating the figure to avoid this issue: https://matplotlib.org/stable/api/_as_gen/matplotlib.pyplot.ioff.html

satishkc7:
## What this fixes

Closes #1968.

When matplotlib is in interactive mode (e.g. a Jupyter notebook with `%matplotlib inline` or after calling `plt.ion()`), calling `plt.subplots()` inside any of the `_plotting` functions causes figures to be rendered inline as a side effect — even though the plots are only meant to be serialized to SVG and embedded in the `TableReport` HTML.

## Fix

Wrap the plotting function call inside the `_plot` decorator with `plt.ioff()` as a context manager. This disables interactive mode for the duration of figure creation and serialization, without affecting the caller's interactive mode setting after the call returns.

```python
# before
with warnings.catch_warnings():
    ...
    return plotting_fun(*args, **kwargs)

# after
with warnings.catch_warnings(), plt.ioff():
    ...
    return plotting_fun(*args, **kwargs)
```

`plt.ioff()` as a context manager is available since matplotlib 3.4, which is already the minimum required version.

## Test

Added `test_plots_do_not_show_in_interactive_mode` in `test_plotting.py` which enables interactive mode via `plt.ion()`, calls `histogram()`, and asserts no new figures are left open afterward.

### DIFF
--- a/skrub/_reporting/tests/test_plotting.py
+++ b/skrub/_reporting/tests/test_plotting.py
@@ -28,4 +28,3 @@ def test_histogram():
     data = pd.Series([0.0])
     _, n_low, n_high = _plotting.histogram(data)
     assert (n_low, n_high) == (0, 0)
-


skrub/_reporting/tests/test_plotting.py
```diff
@@ -28,3 +29,19 @@ def test_histogram():
     data = pd.Series([0.0])
     _, n_low, n_high = _plotting.histogram(data)
     assert (n_low, n_high) == (0, 0)
+
+
+def test_plots_do_not_show_in_interactive_mode():
+    """Plotting functions must not display figures even when matplotlib is in
+    interactive mode (e.g. inside a Jupyter notebook with %matplotlib inline).
+    All figures are serialized to SVG and embedded in the HTML report."""
```
jeromedockes: the last sentence seems unrelated to this test?

skrub/_reporting/tests/test_plotting.py
```diff
@@ -28,3 +29,19 @@ def test_histogram():
...
+    rng = np.random.default_rng(0)
+    data = pd.Series(rng.normal(size=50))
+    plt.ion()
```
jeromedockes: can we use the context manager here as well so we restore the matplotib state instead of always turning ioff() in the finally clause

skrub/_reporting/tests/test_plotting.py
```diff
@@ -28,3 +29,19 @@ def test_histogram():
...
+    try:
+        before = plt.get_fignums()
+        _plotting.histogram(data)
+        after = plt.get_fignums()
```
jeromedockes: this will not work because the plotting function closes the figure it creates so by the time we reach this line the figure has flashed and been closed. I guess we would need to patch one of the matplotlib functions to really test this. I think it might be ok to do without a test for this

skrub/_reporting/_plotting.py
```diff
@@ -102,7 +102,7 @@ def plot_with_config(*args, **kwargs):
             for k, v in params.items():
                 plt.rcParams[k] = v
 
-            with warnings.catch_warnings():
+            with warnings.catch_warnings(), plt.ioff():
```
jeromedockes, line 105: 👍

skrub/_reporting/tests/test_plotting.py
satishkc7: Good catch, removed that sentence. I've since dropped the test entirely (see below).

skrub/_reporting/tests/test_plotting.py
satishkc7: Agreed. I've removed the test entirely (addressed in the next comment), which eliminates the try/finally altogether.

skrub/_reporting/tests/test_plotting.py
satishkc7: You're right, the figure is already closed by the time we compare fignums, so the assertion never catches anything. Since patching the backend would be overly complex for this, I've removed the test per your suggestion. The fix in _plotting.py is straightforward enough to stand without it.

jeromedockes (approved):
LGTM, thank you!

Issue #1425, opened 2025-06-02
DOC - Move the note on single-column transformers to the Notes section of the docstring


rcap107:
At the moment, 
- MinHashEncoder
- GapEncoder
- StringEncoder
- TextEncoder
- DatetimeEncoder

all have the note on single column transformers at the very top of the docstring.
![Image](https://github.com/user-attachments/assets/16728c9c-04b2-401e-a27a-48cd6e31ec92)

The docstring should be modified so that: 
- [ ] It is stored in a drop down that is hidden by default (see https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html and the `:collapsible:` option
- [ ] It should include cross-references to the SingleColumnTransformer now that it's public

Vincent-Maladiere:
> it does not bring any information after the first time the user has read it.

I think the point is to ensure new users reading the doc see this disclaimer, as it's a pretty important paradigm shift from scikit-learn. I'm hesitant on this issue, because we might have an expert bias.

rcap107:
Discussing with @glemaitre, we could split the note so that at the top there is only a short mention like "such and such is a column transformer, so it works differently from typical scikit-learn transformers, click here for more detail" and have the rest of the paragraph in the note section of the docstring.

GaelVaroquaux:
> Discussing with ***@***.***, we could split the note so that at the top there is only a short mention like "such and such is a column transformer, so it works differently from typical scikit-learn transformers, click here for more detail" and have the rest of the paragraph in the note section of the docstring.

OK but not as a "note".

Also, this is a bit long: it's going to show up on all the docstrings that appear when people open a parenthesis in an IDE, diminishing their usefulness.

rcap107:
> Discussing with ***@***.***, we could split the note so that at the top there is only a short mention like "such and such is a column transformer, so it works differently from typical scikit-learn transformers, click here for more detail" and have the rest of the paragraph in the note section of the docstring.
> OK but not as a "note".
> 
> Also, this is a bit long: it's going to show up on all the docstrings that appear when people open a parenthesis in an IDE, diminishing their usefulness.

Not sure if I follow here. What I mean is that I want to replace the current paragraph with the sentence from above, and move the rest of the paragraph where the notes section of the docstring is. The current note does not appear in Code, so the new version of it shouldn't appear either

rcap107:
From IRL discussion:

Good to have, but should not be a blocker for 0.6.0. We can release a fix in 0.6.1

[...]

PR #1986, opened 2026-03-24
updated docstring for SingleColumnTransformer note

lisaleemcb:
I've updated the docstring for classes that inherit from the `SingleColumnTransformer` class. @rcap107 and I discussed the right tone to make the dropdown bar compelling enough so that users will click on, but not so alarming that it causes concern. @GaelVaroquaux  @jeromedockes Please let me know your feedback on the change, thank you!

Address #1425

### DIFF
--- a/skrub/_reporting/_table_report.py
+++ b/skrub/_reporting/_table_report.py
@@ -145,7 +145,7 @@ class TableReport:
 
             export SKB_MAX_PLOT_COLUMNS=30
 
-      : int, default=30
+    max_association_columns : int, default=30
         Maximum number of columns for which associations should be computed.
         If the number of columns in the dataframe is greater than this value,
         the associations will not be computed. If "all", the associations


doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
```diff
@@ -128,3 +128,51 @@ We can see that the only column that has a transformer is "birthday":
 
 >>> to_datetime.transformers_
 {'birthday': ToDatetime()}
+
+.. _single_column_transformer:
+
+The single column transformer
```
jeromedockes: this section seems duplicated with the first one in this file?

skrub/_single_column_transformer.py
```diff
@@ -14,18 +14,22 @@
...
+    " parameter in a ``skrub.ApplyToEachCol`` or a ``skrub.TableVectorizer``.\n\n"
     "To apply to all columns::\n\n"
     "   ApplyToCol({class_name}())\n\n"
     "To apply to selected columns::\n\n"
     "   ApplyToEachCol({class_name}(), cols=['col_name_1', 'col_name_2'])"
 )
 _SINGLE_COL_PARAGRAPH = textwrap.indent(_SINGLE_COL_LINE, prefix=" " * 4)
-_SINGLE_COL_NOTE = f".. note::\n\n{_SINGLE_COL_PARAGRAPH}\n"
+_SINGLE_COL_NOTE = (
+    f".. admonition:: A note on using single column transformations \n"
```
jeromedockes, line 32: maybe we can make it a bit shorter and be explicit that it directly applies to this transformer, eg something like "{class_name} transforms a single column"?

can you edit the mentions of ApplyToEachCol,ApplyToCol, which are outdated, to refer to ApplyToCols instead?

the rendering seems to parse it as a quotation, I guess the indentation needs to be adjusted:

Image

doc/modules/column_level_featurizing/advanced_columnwise_operations.rst
rcap107: I think this was added by mistake

lisaleemcb: So it was a solution I found for doing the cross reference. Line 132 adds a label to the section defined by line 134 so that you can link to it. However, I am new to Sphinx so feel free to change!

jeromedockes: sorry I didn't mean just this line, the cross-reference is done correctly. I meant the whole section that starts [here](https://github.com/lisaleemcb/skrub/blob/2d3799acc55d633d105546d54decbe4c660584fe/doc/modules/column_level_featurizing/advanced_columnwise_operations.rst?plain=1#L12) and the one that starts [here](https://github.com/lisaleemcb/skrub/blob/2d3799acc55d633d105546d54decbe4c660584fe/doc/modules/column_level_featurizing/advanced_columnwise_operations.rst?plain=1#L134) seem to have almost the same content

lisaleemcb: Aaah, sorry, I see it now! Deleted.

skrub/_single_column_transformer.py
lisaleemcb: fixed the block quote and tightened the message, but we can of course continue to refine the wording

skrub/_single_column_transformer.py
```diff
@@ -14,18 +14,24 @@
...
-    " estimators, its ``fit``, ``transform`` and ``fit_transform`` methods expect a"
-    " single column (a pandas or polars Series) rather than a full dataframe. To apply"
-    " this transformer to one or more columns in a dataframe, use it as a parameter in"
-    " a ``skrub.ApplyToCols`` or a ``skrub.TableVectorizer``.\n\n"
+    "``{class_name}`` is a type of "
+    ":ref:`single column transformation <single_column_transformer>` . Unlike most "
+    "scikit-learn estimators, its ``fit``, ``transform`` and ``fit_transform`` methods"
+    " expect a single column (e.g. Series) not a full dataframe."
+    " To apply this transformer to one or more columns in a dataframe, use it as a"
+    " parameter in a ``skrub.ApplyToCols`` or a ``skrub.TableVectorizer``.\n\n"
```
jeromedockes: while we are at it we can turn those into links

```suggestion
    " To apply this transformer to one or more columns in a dataframe, use it "
    "in a :class:`ApplyToCols` or a :class:`TableVectorizer`.\n\n"
```

skrub/_reporting/_table_report.py
```diff
@@ -145,7 +145,7 @@ class TableReport:
 
             export SKB_MAX_PLOT_COLUMNS=30
 
-    max_association_columns : int, default=30
+      : int, default=30
```
rcap107: Why was this removed?

lisaleemcb: Thanks for the catch, I think it was deleted by accident. I've added it back

Issue #1883, opened 2026-02-03
DISC -  Clarify the behavior of the associations tab in the TableReport


rcap107:
I'm opening this issue to discuss two possible UX improvements for the Associations tab of the TableReport.

I had someone ask me why the TableReport wasn't printing out all the associations between columns for a dataframe with 8 columns: that should result in 28 rows, and it was printing only 20.

This is because there is a hardcoded limit on the number of rows that are printed by the TableReport, which is set to 20. This limit isn't explained clearly anywhere. It is still possible to get the associations for all the pairs of columns, but this requires using the `column_associations` function, which isn't reported well either. 

So for this issue I'd like to know:
- how can we make this behavior more obvious? clearly, we should add a message in the Associations tab saying that we're displaying only the top 20 associations, but we should also add an entry in the user guide explaining how to use `column_associations`. 
- do we want to add a config parameter that lets the user specify the number of rows to be printed? I don't think this is really needed, given that the TableReport is supposed to give an indicative view of the table, rather than being very accurate. I think that if `column_associations` is explained properly, users that need that functionality would just use that instead.

So, the first point should definitely be addressed, while the second is more of a discussion point. 

I'd rather have a clear plan for this issue before anyone takes it up.

GaelVaroquaux:
On Tue, Feb 03, 2026 at 07:40:53AM -0800, Riccardo Cappuzzo wrote:
>   • how can we make this behavior more obvious? clearly, we should add a message in the Associations tab saying that we're displaying only the top 20 associations, but we should also add an entry in the user guide explaining how to use column_associations.

Let's start by this, not only in the user guide but also as a small text on the TableReport

jeromedockes:
I think we should just have a "show more" button and a scroll bar. we compute all of them to find the first 20 and they don't take much space to store so we can put the first 5,000 in the report (and a message saying "use column_associations()" for users who click "show more" 249 times)

PR #1992, opened 2026-03-26
Increased the number of top associations to display

lisaleemcb:
Additionally made the associations columns scrollable to compensate for the increased limit.

This addresses some, but not all of #1883


jeromedockes:
thank you @lisaleemcb ! I noticed that the header moves and the vertical borders in the header disappear when we scroll. Usually the default setting of border-collapse: collapse which merges together the borders of table cells doesn't play super well with sticky position; we can make it separate and then do a couple of adjustments to the table's borders. I opened a PR on your branch to show those suggestions:

https://github.com/lisaleemcb/skrub/pull/1

PR #1995, opened 2026-03-26
allow passing additional arguments to scorers in dataops

jeromedockes:
We can now call `.skb.with_scoring` to define which scorer should be used and pass additional arguments to it (such as sample weights). those can be dataops (ie sample weights are computed dynamically from the inputs)

```
>>> import skrub
>>> from sklearn.dummy import DummyClassifier

>>> df = skrub.datasets.toy_products()
>>> data = skrub.var("df", df)
>>> X = data[["description", "price"]].skb.mark_as_X(cv=2)
>>> y = data["category"].skb.mark_as_y()
>>> sample_weight = X["price"]

>>> pred = X.skb.apply(DummyClassifier(), y=y).skb.with_scoring(
...     "accuracy", kwargs={"sample_weight": sample_weight}
... )

>>> pred.skb.cross_validate()
   fit_time  score_time  test_score
0  0.002466    0.003201    0.888889
1  0.002033    0.002878    0.647059
```

### DIFF
--- a/skrub/_data_ops/_skrub_namespace.py
+++ b/skrub/_data_ops/_skrub_namespace.py
@@ -2862,7 +2862,7 @@ class SkrubNamespace:
 
             - a single string;
             - a callable that returns a single value.
-            - ``None``, the ``esti`mator`'s default evaluation criterion is used.
+            - ``None``, the estimator's default evaluation criterion is used.
 
             If ``scoring`` represents multiple scores, one can use:
 


doc/modules/data_ops/validation/tuning_validating_data_ops.rst
```diff
@@ -193,3 +193,74 @@ The test set only contains data from the "bestproducts.com" seller.
...
+==========================================
+
+Sometimes we have additional information to pass to the scorer such as sample
+weights, group information etc.
+
+We can control how scoring is performed by using
+:meth:`DataOp.skb.with_scoring`. It has a ``scoring`` parameter, which can be
+anything :func:`sklearn.model_selection.cross_validate` accepts for ``scoring``
+such as a metric name, callable scorer, or dict mapping metric names to scorers
+(see the reference documentation of :meth:`DataOp.skb.with_scoring` for
```
rcap107: crossreference seems to be broken here

doc/modules/data_ops/validation/tuning_validating_data_ops.rst
```diff
@@ -193,3 +193,74 @@ The test set only contains data from the "bestproducts.com" seller.
...
+
+Passing additional arguments to the scorer
+==========================================
+
+Sometimes we have additional information to pass to the scorer such as sample
+weights, group information etc.
+
+We can control how scoring is performed by using
+:meth:`DataOp.skb.with_scoring`. It has a ``scoring`` parameter, which can be
+anything :func:`sklearn.model_selection.cross_validate` accepts for ``scoring``
```
rcap107: not limited to this PR, I was wondering if here (and elsewhere) we should use the `~` to have the short name of the module rather than linking all of `sklearn.model_selection.cross_validate`

doc/modules/data_ops/validation/tuning_validating_data_ops.rst
```diff
@@ -193,3 +193,74 @@ The test set only contains data from the "bestproducts.com" seller.
...
+
+>>> sample_weight = X["price"]
+>>> pred.skb.with_scoring(
+...     "accuracy", kwargs={"sample_weight": sample_weight}
+... ).skb.cross_validate() # doctest: +SKIP
+   fit_time  score_time  test_accuracy
+0  0.003045    0.003275       0.888889
+1  0.002659    0.003026       0.647059
+
+Besides passing extra arguments, :meth:`DataOp.skb.with_scoring` can also be
```
rcap107, line 241: also broken here

doc/modules/data_ops/validation/tuning_validating_data_ops.rst
```diff
@@ -193,3 +193,74 @@ The test set only contains data from the "bestproducts.com" seller.
...
+-0.6365141682948128
+
+(The score above is the negative log loss, not the default accuracy which would
+be positive.)
+
+Note that :meth:`DataOp.skb.with_scoring` only changes how scoring is performed
+(the outputs of :meth:`DataOp.skb.cross_validate`,
+:meth:`DataOp.skb.make_randomized_search`, :meth:`SkrubLearner.score` etc.),
+**not** the actual outputs of the learner (it does _not_ affect the outputs of
+:meth:`DataOp.skb.eval`, :meth:`SkrubLearner.predict`, etc.)
```
rcap107: `SkrubLearner.score` and `SkrubLearner.predict` xrefs are also broken

doc/modules/data_ops/validation/tuning_validating_data_ops.rst
```diff
@@ -193,3 +193,74 @@ The test set only contains data from the "bestproducts.com" seller.
...
+>>> learner.fit(split['train'])
+SkrubLearner(data_op=<Scoring <Apply DummyClassifier> (1 scorers)>
+    This DataOp will be scored with:
+      - 'neg_log_loss'
+    Use .skb.cross_validate(…) or .skb.make_learner(…).score(…) to compute scores.)
+>>> learner.score(split['test']) # doctest: +SKIP
+-0.6365141682948128
+
+(The score above is the negative log loss, not the default accuracy which would
+be positive.)
```
rcap107: Maybe a  rewrite to make it more evident that we're not using the default scoring metric? 
```suggestion
(The score above is `negative` because it is the `negative log loss`, rather than the default accuracy which would
be positive.)
```

skrub/_data_ops/tests/test_estimators.py
```diff
@@ -173,6 +181,152 @@ def test_cross_validate(data_op, data, n_jobs):
     assert score.mean() == pytest.approx(0.84, abs=0.05)
 
 
+def test_cross_validate_with_scoring():
+    X_a, y_a = make_classification(random_state=0)
+    X = skrub.X(X_a)
+
+    def get_sw(X):
```
rcap107: I was wondering for a while what "sw" meant here
```suggestion
    def get_sample_weights(X):
```

skrub/_data_ops/tests/test_interactive_features.py
```diff
@@ -276,6 +289,32 @@ def test_repr():
     >>> print(e.skb.describe_param_grid())
     - choose_float(10, 100): choose_float(10, 100)
       choose_float(1, 100, log=True, n_...): choose_float(1, 100, log=True, n_steps=100, default=10)
+    >>> from sklearn.dummy import DummyClassifier
```
rcap107, line 292: Unrelated from the current PR, asking here because I just noticed this test. Is this needed because the regular doctests are not sufficient?

skrub/_data_ops/_data_ops.py
```diff
@@ -699,7 +706,7 @@ def __repr__(self):
         if preview is NULL:
             return result
         subsample_msg = " (on a subsample)" if uses_subsampling(self) else ""
-        header = f"Result{subsample_msg}:"
+        header = f"{self._skrub_impl.__skrub_preview_heading__()}{subsample_msg}:"
```
rcap107, line 709: Was the preview_heading function added for possible changes in the future? It doesn't look like it's simplifying things for now

skrub/_data_ops/_estimator.py
```diff
@@ -245,11 +259,22 @@ def report(self, *, environment, mode, **full_report_kwargs):
...
+    def _score(self, environment):
+        score_node = _find_scoring_node(self.data_op)
+        if score_node is None:
+            return self._eval_in_mode("score", environment)
+        estimator = self.__skrub_to_Xy_pipeline__(environment)
+        cv_data = _compute_X_y_and_cv(self.data_op, environment)
+        X, y = cv_data["X"], cv_data.get("y")
+        return estimator._score(X, y, cast_to_float=False)
+
     def __getattr__(self, name):
```
rcap107, line 271: Following along the code:

This function is used when the learner is called with predict/predict_proba/score etc. to trigger the evaluation of all the steps in the given mode 

Then return the result of the evaluated learner

If it's asking for the scorer, then more stuff needs to be done which is why it's calling _score instead

skrub/_data_ops/_estimator.py
```diff
@@ -566,6 +601,69 @@ def _eval_in_mode(self, mode, X, y=None):
...
+        scorer = check_scoring(self, scoring)
+        kwargs = kwargs or {}
+        if not hasattr(scorer, "get_metadata_routing"):
+            return partial(scorer, **kwargs)
+        scorer = copy.deepcopy(scorer)
+        if hasattr(scorer, "_kwargs"):
+            scorer._kwargs = {**scorer._kwargs, **kwargs}
+        elif hasattr(scorer, "_scorers"):
+            scorer._scorers = {
+                k: self._prepare_scorer(v, kwargs) for k, v in scorer._scorers.items()
```
rcap107, line 623: The recursive call here is to evaluate all the possible nested estimators in case they are dicts that contain dicts, correct? 

The loop ends when the code reaches a scorer, which doesn't have a _scorers attribute (meaning it's not nested anymore), or when it does have _kwargs, in which case the kwargs are set and the scorer is returned

skrub/_data_ops/_estimator.py
```diff
@@ -566,6 +601,69 @@ def _eval_in_mode(self, mode, X, y=None):
...
+        score_node = _find_scoring_node(self.data_op)
+        if score_node is None:
+            return self._eval_in_mode("score", X, y)
+        env = self._get_env(X, y)
+        scorers = evaluate(
+            score_node._skrub_impl.scorers, mode="fit_transform", environment=env
+        )
+        all_scores = []
+        for scorer_info in scorers:
+            scorer = self._prepare_scorer(scorer_info["scoring"], scorer_info["kwargs"])
```
rcap107, line 656: I lost track of where is the code that lets me specify a custom name for the scorer

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2833,6 +2834,141 @@ def is_y(self):
...
+
+        This records a scikit-learn
+        `scorer <https://scikit-learn.org/stable/modules/model_evaluation.html#the-scoring-parameter-defining-model-evaluation-rules>`_
+        to be used when measuring the quality of predictions.
+
+        It allows configuring the default scoring metrics, and most importantly
+        to compute any additional arguments needed for scoring (such as sample
+        weights) as part of the DataOp evaluation.
+
+        This method can be called several times to add multiple scorers that
```
rcap107: maybe "chained" to make it clearer that all the calls should come one after the other ?

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2833,6 +2834,141 @@ def is_y(self):
...
+        use different kwargs. Several scoring strategies can be passed to a
+        single calls if they do not need different kwargs (see the description
+        of the `scoring` parameter).
```
rcap107: here and in the other places, but I saw you didn't put the double backticks anywhere
was it intentional?
```suggestion
        of the ``scoring`` parameter).
```

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2833,6 +2834,141 @@ def is_y(self):
         """Whether this DataOp has been marked with :meth:`.skb.mark_as_y()`."""
         return self._data_op._skrub_impl.is_y
 
+    @checked_data_op_constructor
+    def with_scoring(self, scoring, kwargs=None, name=None):
```
rcap107, line 2838: this wasn't added to the api reference so it doesn't appear in the docs

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2833,6 +2834,141 @@ def is_y(self):
...
+            with scoring information attached to it, which will be used for
+            example in cross-validation.
+
+        Notes
+        -----
+        If this method is used several times, all calls to it must be grouped
+        -- there can be no other nodes in-between. For example
+        `pred.skb.score_with('accuracy').skb.score_with('roc_auc')` is allowed
+        but not
+        `pred.skb.score_with('accuracy').skb.apply_func(a_function).skb.score_with('roc_auc')`.
```
```suggestion
        `pred.skb.score_with('accuracy').skb.score_with('roc_auc')` is allowed,
        while
        `pred.skb.score_with('accuracy').skb.apply_func(a_function).skb.score_with('roc_auc')`
        is not.
```

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2833,6 +2834,141 @@ def is_y(self):
...
+        >>> pred.skb.with_scoring("accuracy").skb.with_scoring(
+        ...     "accuracy",
+        ...     kwargs={"sample_weight": sample_weight},
+        ...     name="weighted_accuracy",
+        ... ).skb.cross_validate() # doctest: +SKIP
+           fit_time  score_time  test_accuracy  test_weighted_accuracy
+        0  0.002738    0.005733       0.666667                0.888889
+        1  0.002845    0.005705       0.666667                0.647059
+
+        When we have several scorers but they use the same kwargs we can pass a list
```
```suggestion
        When we have several scorers that use the same kwargs, we can pass a list
```

doc/modules/data_ops/validation/tuning_validating_data_ops.rst
rcap107: I think it's broken because with_scoring was not added to the api reference

CHANGES.rst
```diff
@@ -11,6 +11,11 @@ Ongoing Development
 
 New Features
 ------------
+- It is now possible to pass additional (dynamically computed) arguments to the
+  scorers used by :class:`DataOp` objects for validation, hyperparameter search
+  etc. For example, sample weights. This is achieved by passing the scorers and
+  their arguments to :meth:`DataOp.skb.with_scoring`. :pr:`1995` by
```
rcap107, line 17: xref broken here, should be fixed once the api template is updated

skrub/_data_ops/tests/test_interactive_features.py
jeromedockes: yeah it is here to check specific aspects of the object's repr that are not necessarily covered by the main doctests

skrub/_data_ops/_data_ops.py
jeromedockes: it is to allow the node type to override the default which is just "Results". the Scoring node uses it to provide more detail

skrub/_data_ops/_estimator.py
jeromedockes: yes 👍

skrub/_data_ops/_estimator.py
jeromedockes: yes there will be at most one level of recursion, for MultiMetric scorers

skrub/_data_ops/_estimator.py
jeromedockes: `.skb.with_scoring(..., name='the name')`

doc/modules/data_ops/validation/tuning_validating_data_ops.rst
jeromedockes: ok I did it here, we can edit other places in other pr

jeromedockes:
thanks for the review @rcap107 , I checked the links and they should be good now

jeromedockes:
thanks for the reviews :)

Issue #1844, opened 2026-01-15
DOC - add an example that combines Aeon and the skrub Data Ops


rcap107:
We should have an example about timeseries that uses aeon (https://www.aeon-toolkit.org/en/stable/) and the skrub data ops to show how the two libraries interface with each other.

wenebob:
Hey @rcap107, can I work on this issue?

rcap107:
Hi @Faith-Nchifor, thanks for offering to help. At the moment, this issue is not very well defined, and would benefit from prototyping in a sprint, rather than working completely offline. 

I think it would be better to work on some of the other issues marked as "help wanted" instead.

lisaleemcb:
Hi @wenebob ! After discussions, we think we have a good timeseries dataset for a forecasting example. It was used for a previous EuroSciPy event and involves estimating the electricity usage across France in relationship to seasonal weather patterns. Does this sound like an issue you would be interested in working on? As the example has been use previously, we have some resources we could provide to get you started.

https://github.com/skrub-data/EuroSciPy2025

wenebob:
Hi @lisaleemcb,
Yes, I'm interested in working this issue

lisaleemcb:
Hi @wenebob, great! We're sprucing up the dataset for you a little bit, and will have it prepped and ready to go by next Friday.

rcap107:
Hi @wenebob, we didn't have the time to set up the full dataset fetcher for the electricity dataset, but it has been prepared and can be downloaded from here: https://github.com/skrub-data/skrub-data-files/raw/refs/heads/main/electricity_usage.zip

If you want to start working on the example, you can download it from the link and start writing the text. Once the fetcher is ready, we can work it in your code without trouble. 

You can find additional info on how to put together the table in the repo that was linked before (https://github.com/skrub-data/EuroSciPy2025)

wenebob:
Got it.
Thanks @rcap107!

PR #2013, opened 2026-04-02
FEAT - Add a fetcher for the electricity forecasting dataset

lisaleemcb:
Added a timeseries dataset for a forecasting example.

To do: upload to osf.io
Addresses #1844


rcap107:
The dataset has been updated on osf at https://osf.io/download/d8ykq

skrub/datasets/_fetching.py
```diff
@@ -606,3 +606,40 @@ def fetch_california_housing(data_home=None):
...
+    data_home: str or path, default=None
+        The directory where to download and unzip the files.
+
+    Returns
+    -------
+    Path : PosixPath
+         The path to the electricity usage CSV files
+
+    """
+    return download_dataset("electricity_usage", data_home=None)
```
rcap107:
```suggestion
    return download_dataset("electricity_usage", data_home=data_home)
```

skrub/datasets/_fetching.py
```diff
@@ -606,3 +606,40 @@ def fetch_california_housing(data_home=None):
...
+    Parameters
+    ----------
+    data_home: str or path, default=None
+        The directory where to download and unzip the files.
+
+    Returns
+    -------
+    Path : PosixPath
+         The path to the electricity usage CSV files
+
```
rcap107, line 643: I think it would be good to have a See Also section that links the full tutorial (https://github.com/skrub-data/EuroSciPy2025), and the Notes section that explains in more detail the content of each dataset. As it is, it's quite hard to figure out how to use the dataset from the docstring.

rcap107 (commented):
Thanks for the PR @lisaleemcb! I left a few comments

jeromedockes:
one thing we need to check is the license(s) of the original data, to make sure we can redistribute it like this.

rcap107 (approved):
Looks good to me, thanks a lot @lisaleemcb

doc/api_reference.py
```diff
@@ -322,6 +322,7 @@
                     "datasets.fetch_country_happiness",
                     "datasets.fetch_credit_fraud",
                     "datasets.fetch_drug_directory",
+                    "datasets.fetch_electricity_usage",
```
jeromedockes: maybe "electricity_consumption" ? or whatever "load" means, I think more details are provided in the original data source, eg

https://transparencyplatform.zendesk.com/hc/en-us/articles/16647979768084-Actual-Total-Load-Day-ahead-Per-Bidding-Zone-6-1-A-6-1-B

skrub/datasets/tests/test_fetching.py
```diff
@@ -174,3 +175,51 @@ def _error_on_get(*args, **kwargs):
...
+
+@xfail_with_download_error
+@pytest.mark.parametrize(
+    "dataset_name, dataset_path",
+    [
+        ("electricity_usage", ("electricity_usage", "electricity_usage")),
+    ],
+)
+def test_dataset_paths(dataset_name, dataset_path):
+    "Test datasets whose fetcher returns a PosixPath."
```
jeromedockes: nitpick: on windows it will not be a posixpath

jeromedockes: for testing a single dataset we do not need the parametrize, you can just use the values directly in the test

skrub/datasets/tests/test_fetching.py
```diff
@@ -174,3 +175,51 @@ def _error_on_get(*args, **kwargs):
...
+                "Total Load - Day Ahead _ Actual_202501010000-202601010000.csv",
+                "Total Load - Day Ahead _ Actual_202401010000-202501010000.csv",
+                "Total Load - Day Ahead _ Actual_202301010000-202401010000.csv",
+                "Total Load - Day Ahead _ Actual_202201010000-202301010000.csv",
+                "Total Load - Day Ahead _ Actual_202101010000-202201010000.csv",
+            ],
+        ),
+    ],
+)
+def test_dataset_files(dataset_name, files):
```
jeromedockes: here also, could you remove the parametrize? also could you use pathlib utilities rather than `os` for manipulating file and directory paths

skrub/datasets/tests/test_fetching.py
```diff
@@ -174,3 +175,51 @@ def _error_on_get(*args, **kwargs):
...
+    "Test datasets whose fetcher returns a PosixPath."
+    path = getattr(_fetching, f"fetch_{dataset_name}")()
+    for file in files:
+        print(f"Checking for {file}")
+        print(file in os.listdir(path))
```
jeromedockes:
```suggestion
```

jeromedockes (commented):
thanks @lisaleemcb ! :)

doc/api_reference.py
lisaleemcb: Per discussion I'll change to electricity_forecasting, as the dataset contains data on both the electricity load and weather data.

doc/api_reference.py
```diff
@@ -326,6 +326,7 @@
...
```
jeromedockes: I see you changed the name of the function; it needs to be updated here too

skrub/datasets/tests/test_fetching.py
```diff
@@ -174,3 +175,45 @@ def _error_on_get(*args, **kwargs):
     with TemporaryDirectory() as temp_dir:
         with pytest.raises(OSError, match="Can't download"):
             _ = _fetching.fetch_employee_salaries(data_home=temp_dir)
+
+
+@xfail_with_download_error
+def test_dataset_paths(dataset="electricity_forecasting"):
```
jeromedockes: i think we can remove this test

skrub/datasets/tests/test_fetching.py
```diff
@@ -174,3 +175,45 @@ def _error_on_get(*args, **kwargs):
...
+                "Total Load - Day Ahead _ Actual_202501010000-202601010000.csv",
+                "Total Load - Day Ahead _ Actual_202401010000-202501010000.csv",
+                "Total Load - Day Ahead _ Actual_202301010000-202401010000.csv",
+                "Total Load - Day Ahead _ Actual_202201010000-202301010000.csv",
+                "Total Load - Day Ahead _ Actual_202101010000-202201010000.csv",
+            ],
+        ),
+    ],
+)
+def test_dataset_files(dataset_name, files):
```
jeromedockes: We don't need to make this generic for now:

```python
def test_electricity_forecasting():
    path = _fetching.fetch_electricity_forecasting()
    downloaded = [f.name for f in Path(path).iterdir() if f.is_file()]
    assert {'weather_bayonne.csv', 'total load.csv', ...}.issubset(downloaded)
```

lisaleemcb: updated!

skrub/datasets/tests/test_fetching.py
```diff
@@ -174,3 +175,29 @@ def _error_on_get(*args, **kwargs):
...
+            "Total Load - Day Ahead _ Actual_202501010000-202601010000.csv",
+            "Total Load - Day Ahead _ Actual_202401010000-202501010000.csv",
+            "Total Load - Day Ahead _ Actual_202301010000-202401010000.csv",
+            "Total Load - Day Ahead _ Actual_202201010000-202301010000.csv",
+            "Total Load - Day Ahead _ Actual_202101010000-202201010000.csv",
+        ]
+    )
+    path = _fetching.fetch_electricity_forecasting()
+    downloaded = [f.name for f in Path(path).iterdir() if f.is_file()]
+    assert files == set(downloaded)
```
jeromedockes: it seems the weather ones are parquet not csv? also it is more common to write `assert actual == expected` than `assert expected == actual` so writing it in that order makes the pytest error a little easier to read for people used to that convention

jeromedockes (commented):
LGTM besides the little fix needed on the test! thank you @lisaleemcb

skrub/datasets/tests/test_fetching.py
lisaleemcb: Hi @jeromedockes , thanks for the tip about the boolean check!  

Regarding the files, the folder should have both but I'm not checking for the parquet files. Do you want me to change that?

rcap107:
Discussed IRL: there should be only csv files. We also need to upload the new zip version (with csv only) on osf.

rcap107:
The new version is now on osf, so we can merge this after the tests pass. Thanks a lot @lisaleemcb!

jeromedockes (approved):

PR #2014, opened 2026-04-02
Regrouping pages on model tuning in the DataOps User Guide

emassoulie:
The five pages on validation and tuning at the end of the DataOps section of the User Guide have been fused into three main parts:
- Validating a DataOps model
- Hyperparameter tuning
- Going further with Optuna

These parts have examples that should be shortened (or made into "example" pages), and subsection titles that have been made more explicit.


emassoulie:
Marked as draft because, in addition to the restructuring, the pages also need to be trimmed a little (with the larger examples perhaps moved to their own section).

doc/modules/data_ops/validation/hyperparameter_tuning.rst
```diff
@@ -52,8 +56,8 @@ for ``alpha``.
    in :ref:`this example <example_optuna_choices>`.
 
 
-Here are the different kinds of choices, along with their default outcome when
-we are not using hyperparameter search:
+Skrub provides over 10 different ``choose`` methods for tuning use cases, all detailed
```
rcap107, line 59: I wouldn't say there are 10 "different" choose methods: the choose methods are 4 (from, int, float and bool), but they can be used in different ways

doc/modules/data_ops/validation/tuning_validating_data_ops.rst
```diff
@@ -93,8 +96,8 @@ on several splits. This is done with :meth:`.skb.cross_validate()
 
 .. _user_guide_data_ops_splitting_data:
 
-Splitting the data in train and test sets
-=========================================
+More advanced train/test splitting
```
rcap107, line 99: I think the original title should be kept here: this section isn't about a more advanced way of defining train and test splits, we really are just splitting the data in two

rcap107 (commented):
Thanks for the PR @emassoulie, overall I think it's an improvement over the current documentation. I left a couple of comments where I was not convinced by the wording, but aside from that I think we can merge this soon.

PR #2017, opened 2026-04-02
add caching to .skb.apply

jeromedockes:
adds caching to estimators added with `.skb.apply()` and function calls added with `deferred()` or `.skb.apply_func()`. Caching is off by default and can be enabled with `set_config(cache=True)` or `set_config(cache='/path/to/cache_dir')`. For a given estimator or call, caching can be forbidden entirely in the pipeline definition e.g. `X.skb.apply_func(f, no_cache=True)` in which case the call is never cached regardless of configuration.


jeromedockes:
<details>
<summary>example script</summary>

```python
import time

import skrub
from sklearn.ensemble import HistGradientBoostingClassifier

skrub.set_config(cache=True)

X_a, y_a = make_classification(n_samples=50_000, n_features=50)
pred = skrub.X(X_a).skb.apply(HistGradientBoostingClassifier(), y=skrub.y(y_a))
split = pred.skb.train_test_split()
learner = pred.skb.make_learner()

tic = time.perf_counter()
learner.fit(split["train"])
toc = time.perf_counter()
print("first fit", toc - tic)

tic = time.perf_counter()
learner.fit(split["train"])
toc = time.perf_counter()
print("second fit", toc - tic)

tic = time.perf_counter()
result = learner.predict_proba(split["test"])
toc = time.perf_counter()
print("first predict", toc - tic)

tic = time.perf_counter()
result = learner.predict_proba(split["test"])
toc = time.perf_counter()
print("second predict", toc - tic)
```

</details>

```
first fit 0.28028534399345517
second fit 0.022888949999469332
first predict 0.014127899004961364
second predict 0.005811317008920014
```

skrub/_data_ops/_caching.py
```diff
@@ -0,0 +1,124 @@
...
+import joblib
+
+from .. import _config
+
+# The functions meant to be cached. They are at the top because joblib
+# invalidates the cache when the line number changes.
+
+
+def _call_fitting_method(estimator, method_name, args, kwargs, estimator_id):
+    result = getattr(estimator, method_name)(*args, **kwargs)
```
rcap107, line 14: this is basically saying

```python
result = getattr(estimator, method_name)(*args, **kwargs)

# method_name = fit_transform
# estimator has fit_transform -> return the function fit_transform of estimator
result = fit_transform(*args, **kwargs)
```

correct?

skrub/_data_ops/_caching.py
```diff
@@ -0,0 +1,124 @@
...
+
+
+def _call_deferred_func(func, args, kwargs, globals, closure, defaults, kwdefaults):
+    if globals or closure or defaults:
+        # The deferred function has skrub DataOps (that need to be
+        # evaluated) in its global variables, free variables or default
+        # arguments. In this case after those are evaluated, we recompile a
+        # new function in which the DataOps have been replaced by their
+        # computed value. More details in the docstring of
+        # `skrub.deferred`.
```
rcap107, line 29: here we are "unpacking" the data op if it contains a function, so that the result is just the function we started with, including all the additional arguments

skrub/_data_ops/_caching.py
```diff
@@ -0,0 +1,124 @@
...
+
+import pickle
+import types
+
+import joblib
+
+from .. import _config
+
+# The functions meant to be cached. They are at the top because joblib
+# invalidates the cache when the line number changes.
```
rcap107: I don't understand this part, when would the line number change?

skrub/_data_ops/_caching.py
```diff
@@ -0,0 +1,124 @@
...
+            return
+        self.cached_func = {}
+        self.memory = joblib.Memory(cache_dir, verbose=0)
+        self.cache_dir = cache_dir
+
+    def has_memory(self):
+        self._check_cache_dir()
+        return self.memory is not None
+
+    def cache(self, func, ignore=()):
```
rcap107, line 87: what does the ignore do?

skrub/_data_ops/_caching.py
```diff
@@ -0,0 +1,124 @@
...
+            pass
+        result = self.memory.cache(func, ignore=ignore)
+        self.cached_func[key] = result
+        return result
+
+    def call_deferred_func(
+        self, func, args, kwargs, globals, closure, defaults, kwdefaults, *, no_cache
+    ):
+        all_args = (func, args, kwargs, globals, closure, defaults, kwdefaults)
+        if no_cache or not self.has_memory():
```
rcap107, line 106: could you add a comment here to explain that `self.has_memory()` tries to set the memory? I think it's not obvious here where the memory is set the first time the function is called

skrub/_data_ops/_data_ops.py
```diff
@@ -462,7 +465,7 @@ def _get_preview(obj):
     return obj
 
 
-def _checked_deferred_call_constructor(f):
```
rcap107, line 465: curiosity, why did you remove the _?

skrub/_data_ops/_caching.py
rcap107: looking at the code below, I'm guessing it's to avoid issues with unsupervised transformers that don't take y

could you add a comment about that?

skrub/_data_ops/_data_ops.py
```diff
@@ -1752,7 +1791,64 @@ def pretty_repr(self):
...
+        for i in dis.get_instructions(func.__code__)
+        if i.opname == "LOAD_GLOBAL"
+    ]
+    f_globals = {
+        name: func.__globals__[name]
+        for name in globals_names
+        if name in func.__globals__
+        and not isinstance(
+            func.__globals__[name],
+            (
```
rcap107, line 1828: I'm not following what is happening here: what is it that is being removed by the filter?

skrub/_data_ops/_data_ops.py
```diff
@@ -1505,21 +1520,42 @@ def eval(self, *, mode, environment):
...
         if "fit" in method_name:
-            y_arg = () if self.unsupervised else (y,)
+            args = (X,) if self.unsupervised else (X, y)
         elif method_name == "score":
-            y_arg = (y,)
+            args = (X, y)
         else:
-            y_arg = ()
-        method_kwargs = yield from self._eval_kwargs(method_name)
-        return getattr(self.estimator_, method_name)(X, *y_arg, **method_kwargs)
```
rcap107, line 1522: This is more of a general comment than something about this specific PR, but I noticed that all these blocks where we're checking whether the name of the method is such or such make it quite hard to follow what the code is doing. I don't have a suggestion on how to improve this, it's just something I noticed while reviewing data ops PRs over time.

skrub/_data_ops/_data_ops.py
```diff
@@ -1671,26 +1707,29 @@ class Call(DataOpImpl):
...
-                argdefs=e.defaults,
-                closure=tuple(types.CellType(c) for c in e.closure),
+        if getattr(e.func, "_skrub_is_deferred", False):
+            raise ValueError(
+                "A deferred function was wrapped in a DataOp, "
+                "probably by passing it (indirectly) to .skb.apply_func():\n"
+                f"{e.func!r}.\n"
+                "This results in deferring the function twice.\n"
+                "Please pass the original, undecorated function instead.\n"
+                "(Note: it can be accessed from the deferred function as f.func)."
```
rcap107, line 1721: this is a great candidate for the error page

skrub/_data_ops/_data_ops.py
```diff
@@ -1825,6 +1928,27 @@ def deferred(func):
     INFO x = 3
     3
 
+    It is possible to pass only the ``no_cache`` argument, in which case a
```
rcap107: is this related to the next line? I would have expected something like 

```python
>>> no_cache = skrub.deferred(no_cache=True)
>>> @no_cache
... def f(x) ...
```

unless I'm misunderstanding

skrub/_data_ops/_data_ops.py
```diff
@@ -1890,75 +2014,30 @@ def deferred(func):
...
-            Call(
-                func,
-                args,
-                kwargs,
-                globals={},
-                closure=(),
-                defaults=(),
-                kwdefaults={},
-            )
+    if not isinstance(func, DataOp) and getattr(func, "_skrub_is_deferred", False):
```
rcap107, line 2022: if func isn't a data op, how does it get the _skrub_is_deferred attr?

skrub/_docs/modules/data_ops/ml_pipeline/caching.rst
```diff
@@ -0,0 +1,58 @@
...
+.. code:: python
+
+          skrub.X().skb.apply(skrub.TableVectorizer()).skb.apply(
+              HistGradientBoostingRegressor(), y=skrub.y()
+          )
+
+The transformations applied by the TableVectorizer() can be reused.
+
+For this to happen, we need to enable caching in the configuration, either
+``skrub.set_config(cache=True)`` to store cached results in a default location
```
rcap107, line 35: Could you be more explicit about where it's stored by default?

skrub/tests/test_config.py
```diff
@@ -283,3 +289,29 @@ def test_get_deprecated_int_env_default(monkeypatch):
...
+    assert _load_cache_env_var() is False
+    monkeypatch.setenv("SKB_CACHE", "1")
+    assert _load_cache_env_var() is True
+    monkeypatch.setenv("SKB_CACHE", "/tmp/cache")
+    assert _load_cache_env_var() == "/tmp/cache"
+
+
+def test_unchanged():
+    # make codecov happy
+    assert repr(skrub._config.UNCHANGED) == "unchanged"
```
rcap107, line 317: what an important test :D

skrub/_config.py
```diff
@@ -386,6 +428,9 @@ def set_config(
         data_dir = Path(data_dir).expanduser().resolve()
         local_config["data_dir"] = str(data_dir)
 
+    if cache is not UNCHANGED:
```
rcap107, line 431: I understand this is a sentinel value, but I'm not sure why it's needed

The id check triggers if the cache param has been changed at some point during runtime? Why is that important?

skrub/_data_ops/tests/test_caching.py
```diff
@@ -0,0 +1,147 @@
...
+
+
+@pytest.fixture(autouse=True)
+def reset_counts():
+    f.n_calls = 0
+    DummyTransformer.reset()
+
+
+@pytest.mark.parametrize("with_cache", (False, True))
+def test_caching(with_cache, tmp_path):
```
rcap107, line 48: TIL tmp_path is a pytest fixture

I couldn't figure out where it was coming from

skrub/_data_ops/tests/test_caching.py
```diff
@@ -0,0 +1,147 @@
...
+    skrub.set_config(cache=tmp_path)
+    a = skrub.var("a", 0, becomes_default=True)
+    data_op = (
+        skrub.as_data_op(3)
+        .skb.apply(
+            DummyTransformer(), fit_transform_kwargs={"a": a}, transform_kwargs={"a": a}
+        )
+        .skb.apply_func(lambda x: x)
+    )
+    assert data_op.skb.eval() == 4
```
rcap107, line 118: here the default value of "a" (0) is tested to be triggered by lambda functions that return None -> the parameter "a" is None -> pick up the default value

and this is triggered with and without a lambda function, and after defining a learner

and the idea is that none of these settings should fail even though the parameters are lambda functions that can't be pickled 

so this is more of a test that no error is raised, rather than checking that an error is raised

if this is the right read on the test, could you add a comment to clarify the intent a bit?

skrub/_data_ops/tests/test_caching.py
```diff
@@ -0,0 +1,147 @@
...
+    assert data_op.skb.eval({"a": lambda: None}) == 4
+    assert data_op.skb.make_learner(fitted=True).transform({"a": lambda: None}) == 4
+
+
+def test_cache_reuse_across_data_ops(tmp_path):
+    # cache reused with different estimator objects in different graphs
```
rcap107, line 126: at high level, what is ensuring that the cache is reused? is it joblib here, or is it in the cache object somewhere I didn't notice?

rcap107 (changes requested):
Thanks for the PR @jeromedockes ! This will be very useful indeed. 

I left my usual comments asking for clarification and a few more comments where I found something unclear. 

I think that the next thing to worry about is the max size of the cache, though that can come in a follow-up PR at some point.

skrub/_data_ops/tests/test_caching.py
jeromedockes: > so this is more of a test that no error is raised, rather than checking that an error is raised

yes, the lambda are just there to have either the function or one of its arguments to be something that cannot be serialized to triggers a picklingerror when joblib tries to hash it. we check that in this case we fall back to calling the function without caching and there is no error. i'll add a comment

skrub/_data_ops/tests/test_caching.py
jeromedockes: the cache is reused because we have set the config to use the cache, and both dataops use the same class DummyTransformer initialized with the same arguments so it has the same hash and there is a cache hit.

we check that the cache is reused by looking at DummyTransformer.n_calls

skrub/_data_ops/_caching.py
jeromedockes: i think I'll just remove it if it's unclear. joblib invalidates the cache when the line in the source file where the function is defined, so we put those functions in a separate module and near the top of it so that there is a smaller chance that they get moved around due to unrelated modifications of the rest of skrub code

skrub/_data_ops/_caching.py
jeromedockes: yeah it's just a wrapper for caching

`_call_fitting_method(tv, "fit_transform", X, 123)` will give the same result as `tv.fit_transform(X)`. the wrapping + different signature is to make it easier to cache

skrub/_data_ops/_caching.py
jeromedockes: when the function has default parameter values, global variables or closure variables that contain dataops or choices, we first evaluate them, and compile a new function that contains their values (rather than dataops) and then call it. this function gets all the already-evaluated pieces and takes care of compiling the function if necessary and calling it. and it gets cached

skrub/_data_ops/_caching.py
jeromedockes: ignore tells joblib that the ignored parameters should not be taken into account for deciding if a cached result is valid or not. for example if you do `deduplicate(X, n_jobs=3)` and you cache it you may want to ignore n_jobs because it does not change the result so if someone calls `deduplicate(X, n_jobs=5)` the result that has been computed for n_jobs=3 can be reused even though the value for that parameter has changed.

we use it to rely on an estimator id and avoid hashing the fitted estimator, which may have spurious variations and potentially be expensive

skrub/_data_ops/_data_ops.py
jeromedockes: because now _skrub_namespace also needs to use this function

skrub/_data_ops/_data_ops.py
jeromedockes: we check if any of the global variables referenced in the function need to be evaluated (they contain a dataop or a choice). this is done by `needs_eval`. for some types we know we can skip `needs_eval`, for example very often global variables are functions, types (classes) or modules, and for those we know they don't need to be evaluated so we don't need to call `needs_eval`. this filter is for performance

skrub/_data_ops/_data_ops.py
jeromedockes: I'll remove the comment. the point is you can use any of these

```
@deferred()
def f(): ...

@deferred(no_cache=True)
def f(): ...

# and also

@deferred  # note no parentheses here
def f(): ...
```

skrub/_data_ops/_data_ops.py
jeromedockes: that gets added by `deferred` to the wrapper it constructs

skrub/tests/test_config.py
jeromedockes: yeah I try to make sure the patch coverage is always 100% even if there are some bits that don't strictly require being tested. here UNCHANGED has a `__repr__` because it gets displayed as the default value in the html ref doc

skrub/_config.py
jeromedockes: other config options use `None` as the sentinel: `set_params(cardinality_threshold=10, float_precision=None)` means "don't modify the configuration for `float_precision`", not "set the `float_precision` to `None`"

for the `cache` we can't use `None` as a sentinel because cache=None means "disable caching"

I guess we could just remove the None option, it made more sense when the option was called `cache_dir`

but I think anyway even for other options here it might be clearer to have a more explicit sentinel than None. for example it's not immediately obvious that `table_report_plots_threshold=None` means keep the old threshold, and not set `table_report_plots_threshold` to None, which could mean always plot regardless of the number of columns. so it could make sense to use UNCHANGED everywhere

jeromedockes (commented):
thanks for the review @rcap107 . I put a few replies and I'll work on addressing it soon

rcap107 (dismissed):
Looks good to me! Thanks a lot @jeromedockes 

Leaving this open in case someone else wants to weigh in before merging

jeromedockes:
I tried a bit the caching on a realistic pipeline which I took from here:

https://github.com/probabl-ai/skore-example-electricity-load

running the cross-validation with a warm cache we see 10x speedup on the HGB pipeline and 42x on the TabICL one (that is re-running the exact same pipeline though). disclaimer: I didn't check the code that ran the experiments. on that pipeline only the final estimator is costly so caching doesn't help much for hyperparam search for example. with caching disabled the runtime was the same as on main, and on a cold cache the biggest overhead due to writing results was +13% (for the fastest pipeline, HGB + no quantile regression). the outputs were the same as those obtained from the main branch. the cache size did grow quickly, up to 2G, so providing guidance on how to handle that, and later tooling, will be important.

it would be interesting to do some similar experiments on a pipeline that does have expensive intermediate steps, e.g. a tablevectorizer with text

jeromedockes:
irl we decided we need limiting of the cache size in this PR. I added one possible option but reverting to draft until we have settled on the api and I add the tests

rcap107:
Do you think this PR will make it in the next release? 🤔

jeromedockes:
> Do you think this PR will make it in the next release? 🤔

i think so :+1:

rcap107 (approved):
Looks good to me, the cache pruning change looks like a fairly light addition overall

jeromedockes:
cool, thanks for reviewing!

GaelVaroquaux:
> Merged [1]#2017 into main.

Wohoo! Exciting

PR #2023, opened 2026-04-07
Issue 1249 add DropSimilar transformer

emassoulie:
Creation of the `DropSimilar` transformer, that removes columns for which there exist an association score above a certain threshold.


emassoulie:
Documentation still pending!
At the moment, the class seems to work but does not react to the "correct" threshold: columns with association 0.8 and 0.9 will not be dropped by a DropSimilar with parameter 0.8, but will if the parameter is 0.5.

skrub/_drop_similar.py
```diff
@@ -0,0 +1,81 @@
...
+        pairs_to_drop = _filter_associations(association_df, self.threshold)
+
+        self.to_drop_.extend(pairs_to_drop["right_column_name"].unique())
+
+        self._dropper = DropCols(self.to_drop_)
+
+        return self._dropper.fit_transform(X, y)
+
+    def transform(self, X):
+        check_is_fitted()
```
jeromedockes:
```suggestion
        check_is_fitted(self)
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,81 @@
...
+        if not (0 <= self.threshold <= 1):
+            raise ValueError("Threshold must be between 0 and 1")
+
+        self.to_drop_ = []
+
+        association_df = _column_associations.column_associations(X)
+
+        pairs_to_drop = _filter_associations(association_df, self.threshold)
+
+        self.to_drop_.extend(pairs_to_drop["right_column_name"].unique())
```
jeromedockes: use `sbd.unique`. also I think it would be good to convert to built-in types ie a python list of python strings with `sbd.to_list`

skrub/_drop_similar.py
```diff
@@ -0,0 +1,77 @@
+import polars as pl
```
rcap107:
```suggestion
try:
	import polars as pl
except ImportError:
	pass
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,77 @@
...
+    See Also
+    --------
+    DropUninformative :
+        Drops columns for which various other criteria indicate that they contain
+        little to no information (amount of nulls, of distinct values...)
+
+    Cleaner :
+        Runs several checks to sanitize a dataframe, including converting columns
+        to standard formats or dropping certain columns.
+    """
```
rcap107: There should be an example here

skrub/_drop_similar.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@dispatch
+def _filter_associations(obj):
+    raise_dispatch_unregistered_type(obj, kind="Series")
+
+
+@_filter_associations.specialize("pandas")
+def _filter_associations_pandas(obj, threshold):
+    return obj[obj["cramer_v"] > threshold]
```
rcap107:
```suggestion
    return obj[obj["cramer_v"] >= threshold]
```

If we're using the boundaries of the threshold [0, 1], having strictly larger here means that columns that have an association of exactly 1 would not be selected

skrub/_drop_similar.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+@_filter_associations.specialize("polars")
+def _filter_associations_polars(obj, threshold):
+    return obj.filter(pl.col("cramer_v") > threshold)
```
rcap107: same here

skrub/_drop_similar.py
```diff
@@ -0,0 +1,77 @@
...
+
+
+class DropSimilar(TransformerMixin):
+    """Drop columns found too redundant to the rest of the dataframe,
+    according to association defined by Cramér's V.
+
+    This is done by computing Cramér's V between every possible two columns,
+    and sorting these couples in descending order. Then, for every association above
+    a preestablished threshold, one of the two columns is dropped.
```
rcap107:
```suggestion
    the given threshold, one of the two columns is dropped.
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,77 @@
...
+
+    Parameters
+    ----------
+    threshold : float, default=0.8
+        The Cramér association score value above which to start dropping columns.
+
+    Attributes
+    ----------
+    to_drop : list
```
rcap107:
```suggestion
    to_drop_ : list
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,77 @@
...
+        to standard formats or dropping certain columns.
+    """
+
+    def __init__(self, threshold=0.8):
+        self.threshold = threshold
+
+    def fit_transform(self, X, y=None):
+        # check that the threshold is correct
+        if not (0 <= self.threshold <= 1):
+            raise ValueError("Threshold must be between 0 and 1")
```
rcap107:
```suggestion
            raise ValueError(f"Threshold must be between 0 and 1, got {self.threshold!r}")
```

skrub/tests/test_drop_similar.py
```diff
@@ -0,0 +1,74 @@
...
+def test_drop_similar(df_module, table_with_associations, threshold, result):
+    ds = DropSimilar(threshold=threshold)
+    res = ds.fit_transform(table_with_associations)
+    resulting_columns = list(res.columns)
+    assert resulting_columns == result
+
+
+def test_wrong_threshold(df_module, table_with_associations):
+    ds = DropSimilar(threshold=-0.5)
+    with pytest.raises(ValueError):
```
rcap107: Here and in the other places please add a match clause to make sure the correct error is being caught

skrub/tests/test_drop_similar.py
```diff
@@ -0,0 +1,74 @@
...
+    ],
+)
+def test_drop_similar(df_module, table_with_associations, threshold, result):
+    ds = DropSimilar(threshold=threshold)
+    res = ds.fit_transform(table_with_associations)
+    resulting_columns = list(res.columns)
+    assert resulting_columns == result
+
+
+def test_wrong_threshold(df_module, table_with_associations):
```
rcap107: This test could use a parametrization, also it may be simplified a bit: `table_with_associations` is not needed here 
```suggestion
def test_wrong_threshold(df_module):
```

skrub/tests/test_drop_similar.py
```diff
@@ -0,0 +1,74 @@
...
+        (0.5, ["letters"]),
+        (0, ["letters"]),
+    ],
+)
+def test_drop_similar(df_module, table_with_associations, threshold, result):
+    ds = DropSimilar(threshold=threshold)
+    res = ds.fit_transform(table_with_associations)
+    resulting_columns = list(res.columns)
+    assert resulting_columns == result
+
```
rcap107, line 86: Can you add a test that checks for a threshold of 1.0, and with a dataframe that has two perfectly correlated columns?

skrub/tests/test_drop_similar.py
rcap107: also there should be a test with a string parameter

rcap107 (commented):
Thanks for the PR @emassoulie, I left a few comments to improve and iterate over, but I think it's in a good shape so far!

rcap107:
Something to not forget: there should be a get_feature_names_out method that returns the list of columns that are not dropped by this object (that is, the surviving columns)

emassoulie:
It turns out that the base `SingleColumnTransformer` has a default `fit` method, whereas the dataframe transformer `TransformerMixin` doesn't! This means `DropSimilar` doesn't register as an estimator. I'll add in a `fit` in this specific class.

skrub/_drop_similar.py
```diff
@@ -0,0 +1,194 @@
+try:
```
GaelVaroquaux, line 1: Can we have a module level docstring. This is important for people exploring the code, but also these days for coding agents who try to understand what a code does and how to use it

skrub/_drop_similar.py
```diff
@@ -0,0 +1,194 @@
...
+    'employee_position_title',
+    'date_first_hired',
+    'year_first_hired']
+    >>> ds = DropSimilar(threshold=0.4)
+    >>> clean_df = ds.fit_transform(df)
+
+    `ds` has now removed a column for each pair with association above 0.6.
+    These associations are stored in the `table_associations` attribute:
+
+    >>> ds.table_associations
```
GaelVaroquaux, line 88: By scikit-learn standards, "table_associations" should end with a trailing "_" (and thus be called "table_associations_", because it is derived from data

skrub/_drop_similar.py
```diff
@@ -0,0 +1,194 @@
...
+    all_outputs_ : list
+        The names of columns that the transformer keeps
+
+    table_associations_ : dataframe
+        A dataframe with columns `left_column_name', 'right_column_name' and 'cramer_v'
+        listing association scores between every pair of columns.
+
+    See Also
+    --------
+    DropUninformative :
```
GaelVaroquaux, line 60: Great!!
Keep in mind to add this object (DropSimilar) to the see also of DropUninformative

GaelVaroquaux (commented):
That's a great start!

Two tasks to keep in mind:
* Adding the new object to the API documents
* Adding it to the docs and possibly an example, so that people discover it

skrub/_drop_similar.py
rcap107: I opened an issue about this, because in general we are not very systematic about doing it. https://github.com/skrub-data/skrub/issues/2061

skrub/_drop_similar.py
```diff
@@ -0,0 +1,194 @@
...
+    'employee_position_title',
+    'date_first_hired',
+    'year_first_hired']
+    >>> ds = DropSimilar(threshold=0.4)
+    >>> clean_df = ds.fit_transform(df)
+
+    `ds` has now removed a column for each pair with association above 0.6.
+    These associations are stored in the `table_associations_` attribute:
+
+    >>> ds.table_associations_
```
rcap107: this should be shortened to avoid cluttering the docstring

```suggestion
    >>> ds.table_associations_.head()
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,194 @@
...
+        check_is_fitted(self)
+        return self.all_outputs_
+
+    def fit(self, X, y=None):
+        self.fit_transform(X, y=y)
+        return self
+
+    def fit_transform(self, X, y=None):
+        # check that the threshold is correct
+        if isinstance(self.threshold, bool) or not (
```
rcap107, line 132: This check may be improved by splitting it in two: a check for the dtype (has to be number, and not a bool) that raises TypeError if it fails, and a ValueError that checks that the threshold is in the proper range

skrub/tests/test_drop_similar.py
```diff
@@ -0,0 +1,120 @@
...
+    ds = DropSimilar(threshold=threshold)
+    res = ds.fit_transform(table_with_associations)
+    resulting_columns = list(res.columns)
+    kept_cols = list(ds.get_feature_names_out())
+    assert kept_cols == resulting_columns
+    assert resulting_columns == result
+
+
+@skip_polars_installed_without_pyarrow
+def test_wrong_threshold(df_module):
```
rcap107: This test could be simplified with parametrize, though it should account for changing the condition and the different errors

rcap107 (commented):
A few more comments to deal with, but the PR is progressing well. Thanks @emassoulie

skrub/_drop_similar.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Cleaner :
+        Runs several checks to sanitize a dataframe, including converting columns
+        to standard formats or dropping certain columns.
+
+    Examples
+    --------
+    >>> from skrub import DropSimilar
+    >>> from skrub.datasets import fetch_employee_salaries
+    >>> df = fetch_employee_salaries().X
```
rcap107: This example should be updated so that it uses the `toy_cities` generator that is added by #2042 once that PR is merged

skrub/_drop_similar.py
jeromedockes: or maybe `column_associations_` because they are associations between columns and that is already the name of the (public) function that computes them. but also as it is potentially largeish i wonder if we should store it, or only the list of columns to drop

emassoulie:
The suggested `toy_cities` function has been added to the docstring, and should be functional as soon as https://github.com/skrub-data/skrub/pull/2042 is merged!

emassoulie:
The CI doctest fails in terms of the randomly generated dataframe, but only on certain columns! This error does not occur locally for me, but @rcap107 has confirmed it does for him. I will add a `# doctest: +skip` for this PR, but this may underline a bigger issue that could come back elsewhere.

rcap107:
The changes in the _generating file are bug fixes for an inconsistency in the datetimes that was added with #2042

skrub/datasets/_generating.py
```diff
@@ -278,14 +278,18 @@ def toy_cities(seed=0, size=1000, nulls=0.1, n_metrics=4):
 
     # Next, the "start" and "end" datetime columns are constructed.
     s = rng.integers(0, int(now), size=size)
-    e = rng.integers(s, np.ones(size) * now)
+    e = rng.integers(s, int(now), size=size)
     v = np.vstack([s, e])
 
     df_dates = pd.DataFrame(v.T, columns=["start", "end"])
     if hasattr(df_dates, "map"):
```
rcap107, line 288: can you add a comment here explaining that the if is due to the fact that old versions of pandas don't have map?

skrub/tests/test_drop_similar.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+@skip_polars_installed_without_pyarrow
+def test_fit_transform(table_with_associations):
+    ds = DropSimilar()
+    res_fit_transform = ds.fit_transform(table_with_associations)
+    ds.fit(table_with_associations)
+    res_transform = ds.transform(table_with_associations)
+    fit_transform_columns = list(res_fit_transform.columns)
+    transform_columns = list(res_transform.columns)
```
rcap107:
```suggestion
    fit_transform_columns = sbd.column_names(res_fit_transform)
    transform_columns = sbd.column_names(res_transform)
```

sbd needs to be imported for this

skrub/tests/test_drop_similar.py
```diff
@@ -0,0 +1,137 @@
...
+        ("lower", _TYPE_ERROR),
+    ],
+)
+def test_wrong_threshold(df_module, threshold, error):
+    ds = DropSimilar(threshold=threshold)
+    with pytest.raises(ValueError, match=error):
+        ds.fit_transform(df_module.make_dataframe({}))
+
+
+def test_without_pyarrow(monkeypatch):
```
rcap107, line 122: can you add a comment that explains what this test is doing?

skrub/_drop_similar.py
```diff
@@ -0,0 +1,172 @@
...
+        The names of columns that the transformer keeps
+
+    table_associations_ : dataframe
+        A dataframe with columns `left_column_name', 'right_column_name' and 'cramer_v'
+        listing association scores between every pair of columns.
+
+    See Also
+    --------
+    DropUninformative :
+        Drops columns for which various other criteria indicate that they contain
```
rcap107:
```suggestion
        Drops a single column if certain criteria indicate that it contains
```
making it more explicit that DropUninformative is a SingleColumnTransformer

skrub/_drop_similar.py
```diff
@@ -0,0 +1,172 @@
...
+    0  SHAoqcdajQ  Vilnius            17.0 2004-09-02 03:22:56 2014-06-26 11:36:43
+    1  HVAFYLGCDW      NaN             NaN 1979-10-22 01:43:56 1987-10-15 06:30:05
+    2  oQIauSCbNL     Rome            13.0 1986-08-09 19:01:10 2002-04-06 03:56:09
+    3  SjeSbCepzv  Vilnius            17.0 2008-11-26 15:57:13 2021-11-16 23:16:13
+    4  ubagaIBHnG   London             8.0 1982-09-13 20:54:54 2000-12-02 07:36:41
+
+    >>> ds = DropSimilar(threshold=0.8)
+    >>> clean_df = ds.fit_transform(df)
+
+    `ds` has now removed a column for each pair with association above 0.6.
```
rcap107:
```suggestion
    `ds` has now removed a column for each pair with association above 0.8.
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,172 @@
...
+    >>> clean_df.head() # doctest: +SKIP
+              uid   cities               start                 end
+    0  SHAoqcdajQ  Vilnius 2004-09-02 03:22:56 2014-06-26 11:36:43
+    1  HVAFYLGCDW      NaN 1979-10-22 01:43:56 1987-10-15 06:30:05
+    2  oQIauSCbNL     Rome 1986-08-09 19:01:10 2002-04-06 03:56:09
+    3  SjeSbCepzv  Vilnius 2008-11-26 15:57:13 2021-11-16 23:16:13
+    4  ubagaIBHnG   London 1982-09-13 20:54:54 2000-12-02 07:36:41
+    """  # noqa: E501
+
+    def __init__(self, threshold=0.8):
```
rcap107:
```suggestion
    def __init__(self, threshold=1):
```

I think we should be more conservative and only drop columns that have perfect correlation, but this may need some discussion

skrub/_drop_similar.py
```diff
@@ -0,0 +1,172 @@
...
+    """Drop columns found too redundant to the rest of the dataframe,
+    according to association defined by Cramér's V.
+
+    This is done by computing Cramér's V between every possible two columns,
+    and sorting these couples in descending order. Then, for every association above
+    the given threshold, one of the two columns is dropped.
+
+    Parameters
+    ----------
+    threshold : float, default=0.8
```
rcap107:
```suggestion
    threshold : float, default=1
```

rcap107 (commented):
A few minor changes left (mostly documentation), but this is close to being ready as far as I'm concerned, thanks a lot @emassoulie 

I also suggested changing the threshold to 1, so that only columns that are perfectly correlated with each other are dropped by default.

emassoulie:
Thanks for the help and the feedback @rcap107! I'll address these straight away.

skrub/_drop_similar.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Attributes
+    ----------
+    to_drop_ : list
+        The names of columns evaluated for removal
+
+    all_outputs_ : list
+        The names of columns that the transformer keeps
+
+    table_associations_ : dataframe
```
jeromedockes: column_associations?

skrub/_drop_similar.py
jeromedockes: or at least something much higher like .98

skrub/_drop_similar.py
```diff
@@ -0,0 +1,172 @@
...
+        if sbd.is_polars(X):
+            try:
+                import pyarrow  # noqa F401
+            except ImportError:
+                raise ImportError(
+                    "DropSimilar requires the Pyarrow package to run on Polars"
+                    " dataframes."
+                )
+
+        self.to_drop_ = []
```
jeromedockes: i don't think you need to initialize it here and extend later you can just do to_drop = sbd.to_list(sbd.unique...

jeromedockes (commented):
thanks @emassoulie !

skrub/_drop_similar.py
emassoulie: It does seem better, it makes the default approach a "bare minimum" one. Sounds good to me!

skrub/_drop_similar.py
emassoulie: I think the name was meant to convey that it is a "table of associations". If it isn't clear enough I'll update it.

jeromedockes: yep, please update it! thanks :)

skrub/datasets/_generating.py
emassoulie: Is it OK to put this comment in the DropSimilar PR? Or should it be on its own separate merge?

rcap107: it's fine to have it here, it's not a big change in this case

emassoulie:
All done! Hope the PR is satisfactory.

skrub/_drop_similar.py
```diff
@@ -0,0 +1,168 @@
...
+    0  SHAoqcdajQ  Vilnius            17.0 2004-09-02 03:22:56 2014-06-26 11:36:43
+    1  HVAFYLGCDW      NaN             NaN 1979-10-22 01:43:56 1987-10-15 06:30:05
+    2  oQIauSCbNL     Rome            13.0 1986-08-09 19:01:10 2002-04-06 03:56:09
+    3  SjeSbCepzv  Vilnius            17.0 2008-11-26 15:57:13 2021-11-16 23:16:13
+    4  ubagaIBHnG   London             8.0 1982-09-13 20:54:54 2000-12-02 07:36:41
+
+    >>> ds = DropSimilar(threshold=0.8)
+    >>> clean_df = ds.fit_transform(df)
+
+    `ds` has now removed a column for each pair with association above 0.8.
```
rcap107:
```suggestion
    ``ds`` has now removed a column for each pair with association above 0.8.
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,168 @@
...
+    0           cities    encoded_cities  1.000000
+    1              uid            cities  0.052979
+    2              uid    encoded_cities  0.052979
+    3   encoded_cities               end  0.046455
+    4           cities               end  0.046455
+
+    A single pair is above the threshold, `cities` and `encoded_cities`,
+    with an association score of 1. Since one is an encoding of the other,
+    this is to be expected.
+    Therefore, one of these two has been marked as dropped by `ds`:
```
rcap107: rst needs double backticks to format as code, single backticks are formatted as italics

```suggestion
    A single pair is above the threshold, ``cities`` and ``encoded_cities``,
    with an association score of 1. Since one is an encoding of the other,
    this is to be expected.
    Therefore, one of these two has been marked as dropped by ``ds``:
```

skrub/_drop_similar.py
```diff
@@ -0,0 +1,168 @@
...
+    1  HVAFYLGCDW      NaN             NaN 1979-10-22 01:43:56 1987-10-15 06:30:05
+    2  oQIauSCbNL     Rome            13.0 1986-08-09 19:01:10 2002-04-06 03:56:09
+    3  SjeSbCepzv  Vilnius            17.0 2008-11-26 15:57:13 2021-11-16 23:16:13
+    4  ubagaIBHnG   London             8.0 1982-09-13 20:54:54 2000-12-02 07:36:41
+
+    >>> ds = DropSimilar(threshold=0.8)
+    >>> clean_df = ds.fit_transform(df)
+
+    ``ds`` has now removed a column for each pair with association above 0.8.
+    These associations are stored in the `table_associations_` attribute:
```
rcap107:
```suggestion
    These associations are stored in the ``table_associations_`` attribute:
```

rcap107 (approved):
Looks good to me, thanks a lot for this PR @emassoulie !

PR #2027, opened 2026-04-12
ENH add has_dtype selector

### LINKED ISSUE #1681: FEA - Add a `has_dtype` selector

kudos07:
## Description

Add a new `has_dtype(*dtypes)` selector to let users select columns by passing the exact dtype objects they want to match.

The behavior is intentionally simple and backend-specific: a column is selected when `sbd.dtype(col)` exactly matches one of the provided dtypes. This follows the issue discussion and keeps the selector hands-off, so users can pass dtypes such as `pl.List` or package-specific dtype objects reported by their dataframe library.

Addresses #1681

## Checklist

## How Has This Been Tested?

- `python -m pytest skrub\selectors\tests\test_selectors.py -q`
- `python -m pytest skrub\_dataframe\tests\test_common.py -q -k dtype`

The selector tests cover:
- exact dtype match
- multiple accepted dtypes
- no-match behavior
- behavior across pandas nullable dtypes, pandas numpy dtypes, and polars
- pickling of the new selector


skrub/selectors/_selectors.py
```diff
@@ -292,6 +293,50 @@ def float():
     return Filter(sbd.is_float, name="float")
 
 
+def _has_dtype(column, *dtypes):
+    return sbd.dtype(column) in dtypes
+
+
+def has_dtype(*dtypes):
+    """
+    Select columns whose dtype exactly matches one of the provided dtypes.
```
jeromedockes, line 302: maybe 'is equal to' rather than 'exactly matches'? numpy and pandas dtypes define equality so for example `np.dtype('int64') == int` or `pd.Int32Dtype() == 'Int32'` are True. polars is more strict. also could you show in the Examples section one example like this (eg using 'int' or `int` to select a pandas integer column), as these shorthands are commonly use to check pandas dtypes without going through the trouble of building an actual dtype object

jeromedockes (commented):
LGTM besides a small comment on the docstring, thank you very much @kudos07 !

rcap107:
Small note, this PR needs a changelog entry since it's a new feature

kudos07:
@jeromedockes @rcap107 I pushed both follow-ups to the PR:

- added the changelog entry, including the :pr: / :user: metadata required by the changelog check
- adjusted the has_dtype docstring again to avoid a backend-specific doctest failure while keeping the wording aligned with the review discussion

skrub/selectors/_selectors.py
kudos07: changed the wording to equality-based, and dropped the shorthand example because it was not stable in CI.

rcap107 (approved):
Looks good to me, thanks a lot @kudos07! This is very useful

I'll leave the PR open for a few more days in case @jeromedockes wants to add something, and I'll merge it afterwards

rcap107:
No further comments, merging this.

Thanks a lot @kudos07!

PR #2030, opened 2026-04-13
config to control if dataops repr dropdown is open

### LINKED ISSUE #2029: ENH - add a configuration parameter to always show the graph in data ops
Data ops generate a graph that records all the nodes defined up to the current data op. By default, the graph is generated but hidden in a drop down, like so: 
Image

We should add an entry in the configuration that forces the graph to show by default (so that the drop-down is always active): 
Image

This would be very useful for the data ops examples: at the moment they're just walls of text, and having the graphs would help break them in section.

Note that the configuration is already reset to the default configuration after each example is run, so that's not a problem.

jeromedockes:
closes #2029

### DIFF
--- a/examples/data_ops/1120_multiple_tables.py
+++ b/examples/data_ops/1120_multiple_tables.py
@@ -44,6 +44,10 @@ import pandas as pd
 import skrub
 import skrub.datasets
 
+# Small display detail: open the graphs by default in the visualizations shown
+# in the rest of this notebook.
+skrub.set_config(data_ops_open_graph_dropdown=True)
+
 dataset = skrub.datasets.fetch_credit_fraud(split="train")
 
 # %%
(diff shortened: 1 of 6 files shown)


rcap107 (commented):
Thanks for the PR @jeromedockes, this will be very useful for the documentation! Speaking of which, would you mind adding the config parameter to one of the introductory examples? Maybe we could do it in a separate PR, but it's a small change and I think it makes sense to have it here.

examples/data_ops/1120_multiple_tables.py
```diff
@@ -44,6 +44,10 @@
 import skrub
 import skrub.datasets
 
+# Small display detail: open the graphs by default in the visualizations shown
```
rcap107, line 47: I did not realize how large the final graph was going to be 🤔 I wonder if it might be a bit too big. I don't think it's a problem though, I think it conveys well how the pipeline is growing over time. 

Also, if we're going to replace some of those operations with deferred functions, the graph should shrink anyway.

rcap107 (approved):

PR #2032, opened 2026-04-15
DOC - Adding crossreferences and improving DataOps docstrings

rcap107:
In this PR I'm adding crossreferences to various data ops docstrings, expanding some of the examples and in general doing a cleaning pass.

### DIFF
--- a/skrub/_data_ops/_choosing.py
+++ b/skrub/_data_ops/_choosing.py
@@ -539,7 +539,9 @@ def optional(value, *, name=None, default=OPTIONAL_VALUE):
     >>> print(reduced.skb.describe_param_grid())
     - pca: [PCA(n_components=2), None]
 
-If we perform a hyperparameter search (for example with :meth:`DataOp.skb.make_grid_search`), both pipelines (with and without a PCA) will be considered and the one giving the best predictions will be selected.
+    If we perform a hyperparameter search (for example with
+    :meth:`DataOp.skb.make_grid_search`), both pipelines (with and without a PCA)
+    will be considered and the one giving the best predictions will be selected.
     See also
     --------
     choose_bool :
--- a/skrub/_data_ops/_choosing.py
+++ b/skrub/_data_ops/_choosing.py
@@ -457,8 +457,10 @@ class Optional(Choice):
 
 
 def optional(value, *, name=None, default=OPTIONAL_VALUE):
-    """A choice between ``value`` and ``None``. ``value`` can be an estimator, or \
-    a DataOp that evaluates to an estimator.
+    """A choice between ``value`` and ``None``.
+
+    Typically, ``value`` is an estimator and this choice is passed to
+    :meth:`DataOp.skb.apply`.
 
     ``optional`` allows to build a branch in the search space where the ``value``
     is a component of the pipeline
--- a/skrub/_data_ops/_choosing.py
+++ b/skrub/_data_ops/_choosing.py
@@ -526,7 +526,7 @@ def optional(value, *, name=None, default=OPTIONAL_VALUE):
     None
 
     In practice, ``optional`` is used with :func:`DataOp.skb.apply` to make the
-    application of a DataOp conditional.
+    application of a transformer optional.
     For example, if we want to make the application of PCA optional, we can do:
 
     >>> import skrub
@@ -539,6 +539,7 @@ def optional(value, *, name=None, default=OPTIONAL_VALUE):
     >>> print(reduced.skb.describe_param_grid())
     - pca: [PCA(n_components=2), None]
 
+If we perform a hyperparameter search (for example with :meth:`DataOp.skb.make_grid_search`), both pipelines (with and without a PCA) will be considered and the one giving the best predictions will be selected.
     See also
     --------
     choose_bool :
--- a/skrub/_data_ops/_choosing.py
+++ b/skrub/_data_ops/_choosing.py
@@ -457,12 +457,14 @@ class Optional(Choice):
 
 
 def optional(value, *, name=None, default=OPTIONAL_VALUE):
-    """A choice between ``value`` and ``None``. ``value`` can be a DataOp,
-    an estimator, or any other object.
+    """A choice between ``value`` and ``None``. ``value`` can be an estimator, or \
+    a DataOp that evaluates to an estimator.
 
-    ``optional`` allows to build a branch in the search space where a given component
-    (eg a dimensionality reduction step, a feature selection step, etc) is either
-    present or not. When the component is not present, it is represented by ``None``.
+    ``optional`` allows to build a branch in the search space where the ``value``
+    is a component of the pipeline
+    (e.g., a dimensionality reduction step, a feature selection step, etc) which may
+    be present or not. When the component is not present, it is represented by ``None``.
+    This is equivalent to ``skrub.choose_from([value, None], name=name)``.
 
     When a learner is fitted *without hyperparameter tuning*, the outcome of
     this choice is ``value``. Pass ``default=None`` to make ``None`` the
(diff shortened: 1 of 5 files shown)


rcap107 (commented):
For some reason the crossreferences in the RejectColumn docstring don't work

skrub/_data_ops/_choosing.py
```diff
@@ -454,7 +457,12 @@ def __repr__(self):
 
 
 def optional(value, *, name=None, default=OPTIONAL_VALUE):
-    """A choice between ``value`` and ``None``.
+    """A choice between ``value`` and ``None``. ``value`` can be a DataOp,
+    an estimator, or any other object.
```
MarieSacksick: `any other object` is not very clear to me; could it be a list for instance? From what I understand, it can only be an object that would be a pipeline step, is that correct?

skrub/_data_ops/_choosing.py
```diff
@@ -454,7 +457,12 @@ def __repr__(self):
...
+
+    ``optional`` allows to build a branch in the search space where a given component
+    (eg a dimensionality reduction step, a feature selection step, etc) is either
+    present or not. When the component is not present, it is represented by ``None``.
```
MarieSacksick: What do you mean by "component"? Does it mean the "value"? 
I don't get it, because it would mean that someone adds an optional step... that is empty?  
Maybe it's because it's not clear to me what optional does, the example gives code where it's used alone, while I would expect it, from the first paragraph, to be part of a dataops plan.

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -185,7 +185,7 @@ def apply(
         score_kwargs=None,
     ):
         """
-        Apply a scikit-learn estimator to a dataframe or numpy array.
+        Apply an estimator with a scikit-learn interface to a dataframe or numpy array.
```
MarieSacksick: I can understand this two ways: 
- same scikit-learn API (but then I wonder if it's just fit/predict or other?)
- inherits from BaseEstimator
so it might worth to precise?

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1076,10 +1078,46 @@ def eval(self, environment=None, *, keep_subsampling=False):
         Result:
         ―――――――
         15
+
+        Note that in this case ``c`` has a value of ``15``, which has been computed
+        by evaluating the DataOp on the values of the variables.
+
+        To actually evaluate the DataOp and obtain the result, we can use
```
MarieSacksick: I find this sentence surprising, because it says above that the dataOp already has been evaluated. Two possible changes:  
removing the word "evaluate" above, and replace it by "pre-compute" maybe? By the way, is it always pre-computed, sometimes it may take a long time?
and/or:  
```suggestion
        To actually obtain the result, we can use
```

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1846,41 +1897,80 @@ def train_test_split(
...
         ...     DummyClassifier(), y=y
         ... )
 
-        >>> split = delayed.skb.train_test_split(random_state=0)
+        We can split the data into a training and test set with
+        ``.skb.train_test_split()``. It it also possible to specify parameters
+        for the splitting function, such as the random state, or the size of the
+        test set:
+
+        >>> split = delayed.skb.train_test_split(random_state=0, test_size=0.2)
```
MarieSacksick, line 1909: an API remark, and not a remark related to this PR, but I find it very surprising to apply train test split _after_ applying the model

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1846,41 +1897,80 @@ def train_test_split(
...
         >>> y = orders["delayed"].skb.mark_as_y()
         >>> delayed = X.skb.apply(skrub.TableVectorizer()).skb.apply(
         ...     DummyClassifier(), y=y
         ... )
 
-        >>> split = delayed.skb.train_test_split(random_state=0)
+        We can split the data into a training and test set with
+        ``.skb.train_test_split()``. It it also possible to specify parameters
+        for the splitting function, such as the random state, or the size of the
+        test set:
```
MarieSacksick: these parameters are not listed at the beginning of the docstring. I assume there are the same as the scikit-learn TTS accept; could it be written clearly somewhere? (I would say that above would make it clearer than here in the middle of the example)

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1846,41 +1897,80 @@ def train_test_split(
...
+        >>> from sklearn.metrics import accuracy_score
         >>> predictions = learner.predict(split["test"])
         >>> accuracy_score(split["y_test"], predictions)
         0.0
+
+        If :func:`~skrub.DataOp.skb.mark_as_X` was defined to use a specific
+        splitting function, the splitter will be used by ``train_test_split``:
+
+        >>> from sklearn.model_selection import LeaveOneOut
+        >>> X = orders.skb.drop("delayed").skb.mark_as_X(cv=LeaveOneOut())
```
MarieSacksick, line 1934: what happens if a cv is given, but also `split_func` in TTS? cv has priority?

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -3088,8 +3188,22 @@ def set_description(self, description):
     def description(self):
         """A user-defined description or comment about the DataOp.
 
-        This can be set with ``.skb.set_description()`` and is displayed in the
-        execution report.
+        This can be set with :func:`DataOp.skb.set_description` and is displayed
+        in the execution report.
```
MarieSacksick: where is it displayed? I understand that it should appear with the _repr_ of c, and I also tried to display the graph, but it appears in none, so I don't understand `is displayed in the execution report.`

MarieSacksick (commented):
A couple of elements were not clear to me, or seemed ambiguous :)

skrub/_data_ops/_choosing.py
rcap107: In principle it can be any other object, but in most cases users will use estimators or dataops that evaluate to estimators

skrub/_data_ops/_choosing.py
rcap107: `optional` allows to add a step that may be turned off while executing hyperparameter search

if I have 
```python
data_op.skb.apply(optional(PCA()))
```

then it means that adding the PCA is itself a hyperparameter

skrub/_data_ops/_skrub_namespace.py
rcap107: I'll remove this section from the PR, explaining eval and preview is hard to do and we should discuss it in a separate issue

skrub/_data_ops/_skrub_namespace.py
rcap107: the parameters here are already in the docstring as `split_func_kwargs`, they're forwarded directly to the underlying scikit-learn object, but it could be clarified

skrub/_data_ops/_skrub_namespace.py
rcap107: yes, if cv is given in the final step, what is defined in the mark_as_X is overwritten

skrub/_data_ops/_skrub_namespace.py
rcap107: added a mention of the function

skrub/_data_ops/_choosing.py
```diff
@@ -454,7 +457,14 @@ def __repr__(self):
 
 
 def optional(value, *, name=None, default=OPTIONAL_VALUE):
-    """A choice between ``value`` and ``None``.
+    """A choice between ``value`` and ``None``. ``value`` can be an estimator, or \
+    a DataOp that evaluates to an estimator.
```
jeromedockes:
```suggestion
    """A choice between ``value`` and ``None``.

    Typically, ``value`` is an estimator and this choice is passed to
    :meth:`DataOp.skb.apply`.
```

skrub/_data_ops/_choosing.py
```diff
@@ -512,6 +524,34 @@ def optional(value, *, name=None, default=OPTIONAL_VALUE):
 
     >>> print(optional(PCA(), default=None).default())
     None
+
+    In practice, ``optional`` is used with :func:`DataOp.skb.apply` to make the
+    application of a DataOp conditional.
```
jeromedockes:
```suggestion
    application of a transformer optional.
```

skrub/_data_ops/_choosing.py
```diff
@@ -512,6 +524,34 @@ def optional(value, *, name=None, default=OPTIONAL_VALUE):
...
+    >>> import skrub
+    >>> from skrub.datasets import toy_products
+    >>> from sklearn.decomposition import PCA
+
+    >>> products = skrub.var("products", toy_products())
+    >>> vectorized = products.skb.apply(skrub.TableVectorizer())
+    >>> reduced = vectorized.skb.apply(skrub.optional(PCA(n_components=2), name="pca"))
+    >>> print(reduced.skb.describe_param_grid())
+    - pca: [PCA(n_components=2), None]
+
```
jeromedockes, line 541:
```suggestion

If we perform a hyperparameter search (for example with :meth:`DataOp.skb.make_grid_search`), both pipelines (with and without a PCA) will be considered and the one giving the best predictions will be selected.
```

jeromedockes (approved):

PR #2033, opened 2026-04-16
DOC - improving selector docstrings

rcap107:
In this PR I am adding more crossreferences between selectors and improving the documentation and examples of some of them

### DIFF
--- a/skrub/selectors/_base.py
+++ b/skrub/selectors/_base.py
@@ -173,7 +173,7 @@ def _select_col_names_polars(df, col_names):
 
 
 def select(df, selector):
-    """Apply a selector to a dataframe and return a dataframe with the selected
+    """Apply a selector to a dataframe and return a dataframe with the selected \
     columns.
 
 
--- a/skrub/selectors/_base.py
+++ b/skrub/selectors/_base.py
@@ -173,7 +173,8 @@ def _select_col_names_polars(df, col_names):
 
 
 def select(df, selector):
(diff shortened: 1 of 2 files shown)


skrub/selectors/_base.py
```diff
@@ -173,7 +173,8 @@ def _select_col_names_polars(df, col_names):
 
 
 def select(df, selector):
-    """Apply a selector to a dataframe and return the resulting dataframe.
+    """Apply a selector to a dataframe and return a dataframe with the selected
+    columns.
```
jeromedockes, line 177: nice improvement, thanks @rcap107 ! small formatting detail: I think the summary (first line) of docstring has to fit on one single line

rcap107: It seems to be rendering alright though? I can revert the change but I'd rather keep it explicit

jeromedockes: ok good, I'm surprised numpydoc or whatever didn't yell at you. you can always break the line with 

```
the selected \
columns
```

if you need to make both numpydoc (1 line) and ruff (line length) happy

rcap107: maybe numpydoc is skipping this for some reason 🤔 I added the line break anyway to be sure

jeromedockes: ok thanks. maybe we can do that for all the docstrings in this pr where the summary spans several lines (eg expand_index). apart from that it looks great!

jeromedockes (approved):

PR #2036, opened 2026-04-21
ENH - adding doc link to html repr of estimators

### LINKED ISSUE #1247: Introduce a SkrubEstimator to have docs in HTML display
Scikit-learn estimators, and subclasses, have a mechanism to link to the help page (the question mark on the display):
Image

When we subclass in skrub, we must indicate the logic to link to the docs (there is a subclassing mechanism, explained here: https://scikit-learn.org/dev/developers/develop.html#developer-api-for-html-representation).

We should introduce a SkrubEstimator that does this, and systematically inherit from it.

rcap107:
This PR improves the html repr of the skrub estimators so that now they show the (?) symbol that leads to the documentation page. 

At the moment, this is what a user sees in a jupyter notebook:
image
Note that only the OneHotEncoder has a (?)

With this PR, it looks like this:
image

I updated:
- ApplyToCols
- All the transformers that inherit from  SingleColumnTransformers by adding the required methods to the base class
- SquashingScaler
- TableVectorizer and Cleaner
- SelectCols and DropCols
- SkrubLearner

I did not touch any of the joiners. 

For the change itself I had to override the _doc_link_template. I also had to modify the VisualBlock in ApplyToCols because otherwise cases like `ApplyToCols(TableVectorizer())` would not show the link for the `TableVectorizer`; I'm not entirely sure this is the best way of doing it, and maybe we could abstract more the way I added `_doc_link_template`.


rcap107:
Coverage is failing because I'm not testing every line. I'll wait for someone to review the current version of the code because simplifying it may make it easier to test as well.

glemaitre:
Compare to my PR in https://github.com/skrub-data/skrub/pull/1051, here you are not backporting the feature. But I think it is fine. Basically, you would benefit from the feature if you have a recent enough scikit-learn version, otherwise you will not see the help which I think it is fine.

One thing that I find strange is to have to redefine the property everywhere. I think that it would be best to define it in a mixin or base class that would be inherited from each skrub components.

rcap107:
> Compare to my PR in #1051, here you are not backporting the feature. But I think it is fine. Basically, you would benefit from the feature if you have a recent enough scikit-learn version, otherwise you will not see the help which I think it is fine.

I forgot about #1051 🙈 

> One thing that I find strange is to have to redefine the property everywhere. I think that it would be best to define it in a mixin or base class that would be inherited from each skrub components.

Yes, that's what I was not convinced by. I'll remove the duplication using a base class.

skrub/_base.py
```diff
@@ -0,0 +1,17 @@
+from sklearn.base import BaseEstimator
+
+
+class BaseTransformer(BaseEstimator):
```
jeromedockes: Maybe we can call it SkrubBaseEstimator instead because the point is not transformer vs estimator, but that it should point to the skrub documentation.
also it applies to skrublearners which are not (always) transformers

it could also be SkrubEstimator but that is too similar to SkrubLearner

jeromedockes: also could you add a docstring to this class to say it is a base class for stuff shared by all estimators defined in skrub, which at the moment is only the documentation url

skrub/_data_ops/_estimator.py
```diff
@@ -787,6 +787,17 @@ def describe_params(self):
         """
         return describe_params(eval_choices(self.data_op), choice_graph(self.data_op))
 
+    _doc_link_module = "skrub"
```
jeromedockes: here also we could remove this and replace BaseEstimator by SkrubBaseEstimator as the base class right? also as the base class of _BaseParamSearch

jeromedockes (commented):
i like the approach of replacing BaseEstimator with a skrub base class in all MROs. we should make that explicit in the class docstring and use it everywhere we inherit from baseestimator and it makes sense ( i think there are a few more like DropSimilar, and the joiners if we care to update those)

skrub/_apply_to_cols.py
```diff
@@ -431,6 +433,18 @@ def get_feature_names_out(self, input_features=None):
 
         return self._wrapped_transformer.get_feature_names_out(input_features)
 
+    def _sk_visual_block_(self):
+        # This is needed because when ApplyToCols is used with a transformer like
```
jeromedockes: i'm not sure i understood this comment, but also the scikit-learn diagram machinery is quite complicated so maybe it's not easy to explain in a short comment

rcap107: When I added this block I had the problem that for some reason the (?) signs did not appear properly for the TableVectorizer, but now I can't replicate the problem anymore so I think it can be removed

rcap107: I also tested sklearn 1.5, but it seems to work on the old version so not sure what happened

rcap107: Nevermind, I found the problem. Without that block, the docstring for the TableVectorizer itself is not added.

rcap107: Without this function, the transformer inside applytocols doesn't get the doc link

Image

with the change, it shows up properly

Image

skrub/_apply_to_cols.py
```diff
@@ -431,6 +433,20 @@ def get_feature_names_out(self, input_features=None):
 
         return self._wrapped_transformer.get_feature_names_out(input_features)
 
+    def _sk_visual_block_(self):
+        # This is needed because cases like ApplyToCols(TableVectorizer())
```
jeromedockes, line 437: @glemaitre could you give a bit of context on why this is needed here?

skrub/_apply_to_each_col.py
```diff
@@ -15,7 +16,7 @@
 _SELECT_ALL_COLUMNS = selectors.all()
 
 
-class ApplyToEachCol(BaseEstimator, TransformerMixin):
+class ApplyToEachCol(SkrubBaseTransformer, TransformerMixin):
```
jeromedockes: while we're at it let's put the mixin before the base class

skrub/_base.py
```diff
@@ -0,0 +1,24 @@
+from sklearn.base import BaseEstimator
+
+
+class SkrubBaseTransformer(BaseEstimator):
+    """Base class for all skrub transformers.
+
+    This is a class that all skrub transformers inherit from.
+    For the moment, it's only used for the documentation url, but eventually
+    it will be used for other things as well.
+    """
```
jeromedockes:
```suggestion
    """Base class for all skrub estimators.

    This is a class that all skrub transformers inherit from.
    For the moment, it's only used to set the documentation url for estimator diagrams.
    
    Think twice before adding anything to this class: it is a base class of *all* skrub estimators, including meta-estimators like ApplyToCols, the SingleColumnTransformer base class, and the SkrubLearners created by DataOps.
    """
```

skrub/_data_ops/tests/test_estimators.py
```diff
@@ -1460,3 +1460,13 @@ def load_data():
     pred = X.skb.apply(DummyClassifier(), y=y)
     search = pred.skb.make_grid_search(scoring="roc_auc").fit({})
     assert search.results_.shape[0] == 1
+
+
+def test_learner_docstring():
+    data_op, data = get_data_op_and_data("simple")
+    split = data_op.skb.train_test_split(data)
+    learner = data_op.skb.make_learner().fit(split["train"])
```
jeromedockes, line 1468: do we need it to be fitted? otherwise maybe we can slightly simplify & speedup


```suggestion
    learner = skrub.var('a').skb.make_learner()
```

jeromedockes (approved):
thank you @rcap107 !

skrub/_data_ops/tests/test_estimators.py
rcap107: whoops didn't see the comment, I'll add fix this when I'm cleaning up for the release

PR #2037, opened 2026-04-23
FIX - fixing behavior of DropUninformative for datetime columns

rcap107:
This PR is slightly altering the behavior of DropUninformative when `drop_if_unique=True` so that it ignores datetime columns the same way it ignores numeric columns.

In the stable branch, if the parameter is set to True then the transformer will drop columns that contain no null values and all other values are unique, except in the case the columns are numbers. This means that unique datetimes are also dropped, which I think is unintended behavior.

The PR is addressing this by not dropping datetime columns, however as I was working on this I started wondering whether we should alter the behavior in the other direction, so that if `drop_if_unique=True` then _all columns_ are dropped regardless of dtype (and possibly presence of nulls?). I wonder if this behavior would be simpler to explain, rather than having to specify the conditions: if the flag is on, any column where everything is unique is fair game. Then, users can always use ApplyToCols to exclude specific columns.

### DIFF
--- a/skrub/_drop_uninformative.py
+++ b/skrub/_drop_uninformative.py
@@ -22,8 +22,9 @@ class DropUninformative(SingleColumnTransformer):
 
     drop_if_unique : bool, default=False
         If True, drop the column if all values are distinct. Missing values count as
-        one additional distinct value. Numeric columns are never dropped. This may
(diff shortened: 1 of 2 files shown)


rcap107:
Closing this, we might be deprecating the parameter altogether instead

PR #2039, opened 2026-04-27
ENH: add exclude_cols to ApplyToCols (#1873)

### LINKED ISSUE #1873: DISC - Add explicit `exclude_cols` to ApplyToCols
When one is using one of the Apply transformers, I think that a very common pattern is one where the transformation is to be done over all columns, except for a set of columns specified by the user. At the moment,  this pattern can be handled easily by using the skrub selectors like so 

```python
from skrub import ApplyToCols
import skrub.selectors as s

ApplyToCols(transformer, cols=s.all() - "excluded_col")
```

However, this pattern still requires an additional import, and writing explicitly `s.all()`. I am wondering if it may be easier to have a parameter that is called `exclude` and that works as a shortcut to select "all columns except the excluded one". In this way, users wouldn't have to import the selectors, and this would simplify many common situations. 

```python
ApplyToCols(transformer, exclude="excluded_col")

```

I think this would be a useful parameter, but I can see it could be redundant, which is why I am tagging this as a discussion.

sabasiddique1:
# Bug Fix Pull Request

## Description

Add support for an `exclude_cols` parameter in `ApplyToCols` to simplify applying a transformer to all columns except a given subset.

This avoids requiring users to explicitly use selector expressions like `s.all() - "col"` and aligns the behavior with the existing `.skb.apply` API.

Internally, this is implemented by computing:

```
cols = make_selector(cols) - exclude_cols
```

before passing the result to `wrap_transformer`.

Addresses #1873

---

## Checklist

* [x] I have read the contributing guidelines
* [x] I have added tests that verify the change
* [x] I have added an entry to CHANGES.rst describing the change
* [x] My code follows the code style of this project
* [x] I have checked my code and corrected any misspellings

---

## How Has This Been Tested?

Added tests in `skrub/tests/test_apply_to_cols.py` to verify:

* excluding a single column
* excluding multiple columns
* correct behavior when combining `cols` selector with `exclude_cols`

Tests were executed using:

```
python -m pytest skrub/tests/test_apply_to_cols.py
```

All tests passed successfully.

---

## AI Disclosure

* [x] This PR contains AI-generated code

  * [x] I have tested the code generated in my PR
  * [x] I have read and understood every line that has been generated by the AI agent
  * [x] I can explain what the AI-generated code does

### DIFF
--- a/skrub/_apply_to_cols.py
+++ b/skrub/_apply_to_cols.py
@@ -184,10 +184,7 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
     transform all numeric columns except ``"C"``:
 
     >>> scaler = ApplyToCols(StandardScaler(), cols=s.numeric(), exclude_cols="C")
-    >>> scaler.fit_transform(df)                                     # doctest: +SKIP
-        C   city                   D    A    B
-    0  19  Paris 2024-05-13 12:05:36 -1.0 -1.0
-    1  20   Rome 2024-05-15 13:46:02  1.0  1.0
+    >>> scaler.fit_transform(df)  # doctest: +SKIP
 
 
     It is possible to set ``allow_reject=True`` to allow the transformer to reject
--- a/skrub/_apply_to_cols.py
+++ b/skrub/_apply_to_cols.py
@@ -33,11 +33,15 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
         The transformer to apply to the selected columns.
 
     cols : str, sequence of str, or skrub selector, optional
-        The columns to attempt to transform. Only the selected columns will have
-        the transformer applied. Columns outside this selection are passed
-        through unchanged (``fit_transform`` is not called on them) and remain
-        unmodified in the output. The default is to attempt transforming all
-        columns.
+        The columns to attempt to transform. The transformer is applied to the
+        columns matched by ``cols`` and not matched by ``exclude_cols``.
+        Columns outside this selection are passed through unchanged
+        (``fit_transform`` is not called on them) and remain unmodified in the
+        output. The default is to attempt transforming all columns.
+
+    exclude_cols : str, sequence of str, or skrub selector, optional
+        Columns to exclude from transformation. The transformed columns are the
+        ones matched by ``cols`` and not matched by ``exclude_cols``.
 
     allow_reject : bool, default=False
         Whether to allow refusing to transform columns for which the provided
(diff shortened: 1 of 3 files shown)


rcap107:
Hi @sabasiddique1, a quick comment on the changelog: the check is failing because the changelog entry is missing the PR number and github handle

skrub/_apply_to_cols.py
```diff
@@ -175,6 +180,12 @@ class ApplyToCols(TransformerMixin, BaseEstimator):
     0  Paris 2024-05-13 12:05:36 -1.0 -1.0 -1.0
     1   Rome 2024-05-15 13:46:02  1.0  1.0  1.0
 
+    We can also exclude columns from a broader selection. For example, to
+    transform all numeric columns except ``"C"``:
+
+    >>> scaler = ApplyToCols(StandardScaler(), cols=s.numeric(), exclude_cols="C")
```
rcap107: This line is breaking the doctest below because it's redefining the scaler without fitting it, so the scaler doesn't have a `transformer_` argument anymore

Something like this would fix the problem:

```python
    >>> exc_scaler = ApplyToCols(StandardScaler(), cols=s.numeric(), exclude_cols="C")
    >>> df_enc = exc_scaler.fit_transform(df)  # doctest: +SKIP
    >>> df_enc
        C   city                   D    A    B
    0  19  Paris 2024-05-13 12:05:36 -1.0 -1.0
    1  20   Rome 2024-05-15 13:46:02  1.0  1.0
```

rcap107 (commented):
Hi @sabasiddique1, I think the PR is already in a good shape, thanks a lot! 

Other than fixing the doctests, I think we can move on and merge this pretty quickly

rcap107:
Hi @sabasiddique1, we would like to have this PR in the next release which we're planning to do next week. Do you want to fix the problem with the docstring? Otherwise we will do that on our side and merge afterwards

sabasiddique1:
Hey @rcap107, thanks for the update! I can take care of fixing the docstring and will push an update shortly so it can be included in the next release.

sabasiddique1:
Hi @rcap107,

I updated the example to use a new variable (`exc_scaler`) and added `# doctest: +SKIP` to avoid failures. All tests, doctests, and pre-commit checks are passing locally.

Does this look good to merge?

rcap107 (approved):
Looks good to me, thanks a lot @sabasiddique1, this will be useful!

jeromedockes (approved):
thank you very much @sabasiddique1 !

PR #2040, opened 2026-04-28
MAINT - deprecating drop_if_unique

rcap107:
Follow-up of #2037. After IRL discussions, we reached the conclusion that `drop_if_unique` is unlikely to bring value and may actually cause harm as it drops informative columns. This PR deprecates the functionality.

### DIFF
--- a/skrub/_drop_uninformative.py
+++ b/skrub/_drop_uninformative.py
@@ -1,4 +1,5 @@
 import numbers
+import warnings
 
 from sklearn.utils.validation import check_is_fitted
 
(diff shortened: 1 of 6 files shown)


skrub/_drop_uninformative.py
```diff
@@ -20,11 +21,6 @@ class DropUninformative(SingleColumnTransformer):
         If True, drop the column if it contains only one unique value. Missing values
         count as one additional distinct value.
 
-    drop_if_unique : bool, default=False
```
jeromedockes, line 23: maybe keep the entry to have a place to say it is deprecated? something like

```
drop_if_unique : deprecated
   only kept for backwards compat will be removed in 0.10
```

jeromedockes (commented):
we should also probably say it is deprecated in the Cleaner docstring

jeromedockes (approved):

PR #2041, opened 2026-04-28
add methods to find a node in a dataop

jeromedockes:
- `find_X_y` returns the nodes marked as X and y and if they exist the cv and split_kwargs
- find looks for a node by name or by a callable predicate

### DIFF
--- a/skrub/_data_ops/tests/test_data_ops.py
+++ b/skrub/_data_ops/tests/test_data_ops.py
@@ -10,7 +10,7 @@ from sklearn.base import BaseEstimator
 from sklearn.datasets import make_classification, make_regression
 from sklearn.dummy import DummyRegressor
 from sklearn.linear_model import LogisticRegression, Ridge
-from sklearn.model_selection import train_test_split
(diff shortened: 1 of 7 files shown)


skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -907,3 +907,49 @@ def test_copy_attrs():
         .skb.set_name("transform")
     )
     assert isinstance(out.skb.applied_estimator.skb.eval().transformer_, PassThrough)
+
+
+def test_find():
```
glemaitre, line 912: They are personally the type of test that I like to parametrize with two tests: one with an assert and one to check the error.

glemaitre (approved):
LGTM

skrub/_data_ops/tests/test_data_ops.py
glemaitre: Going more in details, providing the expected require to move the code outside of the test function and it is not nice. So it looks already good. Merging.

PR #2042, opened 2026-04-28
Added `toy_cities` function to `datasets._generating`

emassoulie:
Addition of a `toy_random` function for random generation of all-purpose test datasets (with null proportions, random values for metrics, closely associated columns, etc.)


GaelVaroquaux:
How about calling it "toy_dataset", or "toy_dataframe". I don't think that putting the focus on the random aspect is crucial

emassoulie:
> How about calling it "toy_dataset", or "toy_dataframe". I don't think that putting the focus on the random aspect is crucial

Don't `toy_orders` and `toy_products` also create toy dataframes? I feel the random generation is what makes this function's specificity. Perhaps something like `toy_custom` would move the focus elsewhere?

GaelVaroquaux:
> Don't toy_orders and toy_products also create toy dataframes? I feel the random generation is what makes this function's specificity. Perhaps something like toy_custom would move the focus elsewhere?

OK, fair enough. But "random" is an information on the process of creation, not something very descriptive about the dataset. Can we find a better name?

jeromedockes:
> OK, fair enough. But "random" is an information on the process of creation, not something very descriptive about the dataset. Can we find a better name?

apparently it returns something with a 'capitals' column containing city names so maybe toy_countries, toy_cities.

emassoulie:
The name has been changed! Is this modification worth adding to the changelog, or can it be bundled with `DropSimilar` for instance?

rcap107:
> The name has been changed! Is this modification worth adding to the changelog, or can it be bundled with `DropSimilar` for instance?

The two PRs are quite distinct from each other, so the entries should be separate. In this case it just happened that we needed the new dataset for the other PR, but that doesn't mean they are strongly related.

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
             ],
         }
     )
+
+
+def toy_cities(seed=0, size=1000, nulls=0.1, n_metrics=4):
+    np.random.seed(seed)
```
rcap107: setting the seed inside the function actually changes the seed even outside the function, which is something that should be avoided

it should be set using something like this instead

```python
    rng = np.random.default_rng(seed=seed)
```

and the other instances of np.random should be replaced by this `rng`

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
...
+        "Stockholm",
+        "Vienna",
+        "Vilnius",
+        "Warsaw",
+        "Zagreb",
+    ]
+
+    d = {}
+
+    d["uid"] = [random_string() for i in range(size)]
```
rcap107: `random_string()` is using token_hex and isn't reproducible, so this should be replaced with something like 

```python
strings = ["".join(rng.choice(list(string.ascii_letters), 10)) for _ in range(size)]
```

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
...
+    d["cities"] = np.random.choice(capitals, size=size)
```
rcap107: `np.random` should be replaced by `rng`

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
...
+    for i in range(size):
```
rcap107: this part could be simplified using masks, something like 

```python
p = rng.uniform(0, 1, size=size)
df["cities"] = df["cities"].where(p<nulls)
```

in general, for loops should always be avoided when working with dataframes because they're very inefficient

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
...
+        p = np.random.random()
+        if p < nulls:
+            d["cities"][i] = None
+
+    cities_array = np.array(d["cities"]).reshape(-1, 1)
+    d["encoded_cities"] = list(OrdinalEncoder().fit_transform(cities_array))
```
rcap107: simplifying a bit:

```python
d["encoded_cities"] = OrdinalEncoder().fit_transform(d[["cities"]]))
```

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
...
+
+    start_times, end_times = [], []
+    for _ in range(size):
```
rcap107: this could be simplified with something like 

```python
import pandas as pd
import numpy as np
import time
from datetime import datetime

# reuse the same rng generator from outside
rng = np.random.default_rng(seed=19)

t = time.time()
size = 1_000

s = rng.integers(0, int(t), size=size)
e = rng.integers(s, np.ones(size)*t)

v = np.vstack([s,e])

df = pd.DataFrame(v.T, columns=["start", "end"])
# here I'm converting directly to timestamp since we're looking for that dtype
df = df.map(datetime.fromtimestamp)
```

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
...
+        s = np.random.randint(0, int(t))
+        e = np.random.randint(s, int(t))
+        p = np.random.random()
+        start_times.append(time.ctime(s))
+
+        if p >= nulls:
```
rcap107: nulls here should be added using masks like in the other comment

skrub/datasets/_generating.py
```diff
@@ -175,3 +179,66 @@ def toy_products():
...
+            end_times.append(time.ctime(e))
+        else:
+            end_times.append(None)
+
+    d["start_times"] = pd.to_datetime(start_times)
+    d["end_times"] = pd.to_datetime(end_times)
+
+    for k in range(n_metrics):
```
rcap107: this can be generated with something like 

```python
import numpy as np
rng = np.random.default_rng(seed=0)
n_rows = 15
n_metrics = 2
m = rng.random(size=(n_rows, n_metrics))
metrics = pd.DataFrame(m, columns=[f"metric_{_}" for _ in range(n_metrics)])
```

and then concatenated

rcap107 (commented):
Thanks for the PR @emassoulie, I think it can be improved a bit before merging it, I left a few comments about it.

skrub/datasets/_generating.py
emassoulie: I'd avoided that because of a type error it raised, I've now changed it back but the error is still there (`unhashable type: 'list'` on `d[['cities']]`. Is there a workaround that I haven't seen?

skrub/datasets/_generating.py
emassoulie: `fromtimestamp` isn't recognized, I'll clean that line up tomorrow! Thanks for all the suggestions

skrub/datasets/_generating.py
```diff
@@ -175,3 +178,103 @@ def toy_products():
...
+    Returns
+    -------
+    pandas dataframe
+        The randomly-generated dataframe, with `size` rows and
+        `5 + n_metrics` columns.
+
+    Examples
+    --------
+    >>> from skrub.datasets import toy_cities
+    >>> toy_cities(seed=5, size=3, n_metrics=2)
```
rcap107:
```suggestion
    >>> df = toy_cities(seed=5, size=3, n_metrics=2)
    >>> df # doctest: +SKIP
```

skrub/datasets/_generating.py
```diff
@@ -175,3 +178,103 @@ def toy_products():
...
+            uid     cities  encoded_cities               start                 end  metric_0  metric_1
+    0  IPbQyAGoYc  Stockholm               1 1992-02-15 04:59:22                 NaT  0.227319  0.895448
+    1  otDvgcachZ     Vienna               2 2024-01-27 06:03:38 2024-11-26 17:22:08  0.872195  0.018517
+    2  jHNmownYjU        NaN               0 2001-04-28 14:12:32 2023-05-13 04:40:58  0.707496  0.001200
+    """  # noqa: E501
+    rng = np.random.default_rng(seed=seed)
+    t = time.time()
```
rcap107: Thinking about it, it's probably better to have a fixed timestamp to ensure reproducibility. Currently, datetime columns change every time the function is called.

skrub/datasets/tests/test_generating.py
```diff
@@ -25,3 +26,26 @@ def test_make_deduplication_data():
         "test1",
         "test1",
     ]
+
+
+def test_toy_cities():
```
rcap107, line 32: Could you also add a test that checks that setting p=0 adds no nulls and p=1 has only nulls in the two columns?

skrub/datasets/_generating.py
```diff
@@ -175,3 +178,103 @@ def toy_products():
             ],
         }
     )
+
+
+def toy_cities(seed=0, size=1000, nulls=0.1, n_metrics=4):
+    """Custom dataframe example with a variety of column types.
```
rcap107:
```suggestion
    """Generate a synthetic dataframe example with a variety of column types.
    
```

The first line of the docstring should be a very short sentence, and then there should be a newline (numpy formatting things)

rcap107 (commented):
Thanks @emassoulie! I left a few more comments that shouldn't take too much effort to implement

skrub/datasets/_generating.py
emassoulie: Good idea! I've given t an arbitrary fixed value.

skrub/datasets/tests/test_generating.py
emassoulie: I'm on it! Come to think of it, shouldn't the function run its own check that `nulls` is between 0 and 1?

rcap107: yes, in fact there should be error checking for all the parameters (and the relative tests)

skrub/datasets/_generating.py
```diff
@@ -175,3 +177,105 @@ def toy_products():
...
+    df_cities["cities"] = df_cities["cities"].where(p >= nulls)
+    df_cities["encoded_cities"] = OrdinalEncoder().fit_transform(df_cities[["cities"]])
+
+    # Next, the "start" and "end" datetime columns are constructed.
+    s = rng.integers(0, int(now), size=size)
+    e = rng.integers(s, np.ones(size) * now)
+    v = np.vstack([s, e])
+
+    df_dates = pd.DataFrame(v.T, columns=["start", "end"])
+    df_dates = df_dates.map(datetime.datetime.fromtimestamp)
```
rcap107: DataFrame.map was called applymap in pandas 1.5 and has been deprecated in pandas 3.0, so we can check if the attribute "map" is present and use applymap if not 

```suggestion
    if hasattr(df_dates, "map"):
        df_dates = df_dates.map(datetime.datetime.fromtimestamp)
    else:
        df_dates = df_dates.applymap(datetime.datetime.fromtimestamp)
```

rcap107:
Coverage is failing because the error checks are not being tested in

emassoulie:
> Coverage is failing because the error checks are not being tested in

Ohhh right, this probably warrants a separate test function. I'll add this in.

skrub/datasets/_generating.py
```diff
@@ -175,3 +178,123 @@ def toy_products():
...
+    # Check that the other variables are integers
+    if not isinstance(seed, int) or seed < 0:
+        raise ValueError(f"seed must be a positive integer, got {seed}.")
+    if not isinstance(size, int) or size < 0:
+        raise ValueError(f"size must be a positive integer, got {size}.")
+    if not isinstance(n_metrics, int) or n_metrics < 0:
+        raise ValueError(f"n_metrics must be a positive integer, got {n_metrics}.")
+
+    rng = np.random.default_rng(seed=seed)
+    now = 1770000000
```
rcap107: it's better to not use magic numbers (especially without comments), something like this is a bit more wordy but it shows immediately what is the time range we are working with 
```suggestion
    now = datetime.fromisoformat("2024-01-01").timestamp()
```

skrub/datasets/_generating.py
```diff
@@ -5,10 +5,13 @@
 
 from __future__ import annotations
 
+import datetime
```
rcap107: this saves from having to write `datetime.datetime`

```suggestion
from datetime import datetime
```

skrub/datasets/_generating.py
```diff
@@ -175,3 +178,123 @@ def toy_products():
...
+    # Next, the "start" and "end" datetime columns are constructed.
+    s = rng.integers(0, int(now), size=size)
+    e = rng.integers(s, np.ones(size) * now)
+    v = np.vstack([s, e])
+
+    df_dates = pd.DataFrame(v.T, columns=["start", "end"])
+    if hasattr(df_dates, "map"):
+        df_dates = df_dates.map(datetime.datetime.fromtimestamp)
+    else:
+        df_dates = df_dates.applymap(datetime.datetime.fromtimestamp)
```
rcap107: after changing the import, this skips some redundancy 

```suggestion
    if hasattr(df_dates, "map"):
        df_dates = df_dates.map(datetime.fromtimestamp)
    else:
        df_dates = df_dates.applymap(datetime.fromtimestamp)
```

skrub/datasets/_generating.py
rcap107: an alternative could be 

```suggestion
	# timestamp corresponding to '2026-02-02T03:40:00'
    now = 1770000000
```

rcap107 (commented):
a couple more cosmetic comments, but aside from that the PR is almost ready

rcap107 (approved):
LGTM, thanks a lot @emassoulie!

PR #2048, opened 2026-05-04
ENH - Adding a function that exports the TableReport in Markdown format

rcap107:
This PR is a work in progress study of how we could improve the way the table report is exported in plain text or json, both for readable human output, and possibly for agents to read. 

Some sticking points:
- if the plaintext is used as is, it exposes LLMs to prompt injection from the columns and the most frequent values. This needs to be accounted for and we should not take responsibility for it.
- We probably don't need strict markdown formatting checks
- How can we avoid providing users with a gun to shoot themselves in the foot with? This PR opens the door of "hey now I can just generate a summary and pass it to an LLM, that's convenient", and then all sorts of problems can come up. 

The PR is still in draft mode for the time being, but the high level structure is ready.


skrub/_reporting/_table_report.py
```diff
@@ -417,6 +417,19 @@ def json(self):
...
+    def markdown(self):
+        """Get the report as a Markdown string.
+
+        Returns
+        -------
+        str :
+            The Markdown report.
+        """
+
+        if self._summary["dataframe_is_empty"]:
```
jeromedockes: kind of a detail but I guess that handling of empty dataframe could be moved into `to_markdown`, to have it all in 1 place?

skrub/_reporting/_data/templates/report.md
```diff
@@ -0,0 +1,63 @@
...
+{% if col.value_counts or (col.mean is defined and col.mean is not none) %}
+
+### Column `{{ col.name }}` — Details
+{% if col.value_counts %}
+
+**Most frequent values:**
+
+| Value | Count |
+|---|---|
+{% for value, count in col.value_counts %}| {{ value }} | {{ count }} |
```
jeromedockes: it looks good but there is in some places  a lot of whitespace in the output

jinja lets you [control](https://jinja.palletsprojects.com/en/stable/templates/#whitespace-control) how whitespace around statements is handled, either with configuration options or by adding `-` or `+` next to the `%` in the tags to indicate if it should be swallowed or not.

 could you try to play a bit with that to see if the output can be made a bit more compact, and possibly the source a bit more readable by having control tags in separate lines? for example for this section I think something like this improves it slightly:

```
{% for col in summary.columns %}
{% if col.value_counts or (col.mean is defined and col.mean is not none) %}
### Column `{{ col.name }}` — Details

{% if col.value_counts -%}

**Most frequent values:**

| Value | Count |
|---|---|
{%- for value, count in col.value_counts %}
| {{ value }} | {{ count }} |
{%- endfor %}
{%- endif %}
{%- if col.mean is defined and col.mean is not none -%}
```

skrub/_reporting/_data/templates/report.md
```diff
@@ -0,0 +1,63 @@
...
+{% endfor %}
+
+---
+
+## Associations (Cramér's V)
+
+{% if summary.top_associations %}
+| Left column | Right column | Cramér's V |
+|---|---|---|
+{% for assoc in summary.top_associations %}| {{ assoc.left_column_name }} | {{ assoc.right_column_name }} | {% if assoc.cramer_v > 0.9 %}**{{ "%.4f" | format(assoc.cramer_v) }}**{% else %}{{ "%.4f" | format(assoc.cramer_v) }}{% endif %} |
```
jeromedockes: IIRC the max number of associations is rather high now, maybe we want to truncate it here?

jeromedockes (commented):
thanks @rcap107 ! LGTM besides a couple of nitpicks

skrub/_reporting/_table_report.py
```diff
@@ -209,7 +209,16 @@ class TableReport:
     full page in a separate browser tab with its ``open`` method:
     ``report.open()``.
 
-    You can also get the HTML report as a string.
+    If you want to display the report in a notebook or get a quick text summary of
```
jeromedockes: in a notebook it is more likely that you want the default html version so maybe "If you want a summary of the report in a plain-text format, you can use ``markdown`` ..."
also I would put that paragraph after all information about html format to avoid interleaving information about 2 different formats, and because html is the main use-case. either html,markdown,json or html,json,markdown

skrub/_reporting/_table_report.py
```diff
@@ -220,6 +229,15 @@ class TableReport:
...
     '\n<div id="report_...-wrapper" hidden>\n    <template id="report_...'
 
+    The report can also be obtained in JSON format with :meth:`json`, which can
+    be useful for programmatic access to the report data.
+
+    Note that the resulting JSON includes the plots in SVG format, which can be
+    quite verbose: plots can be disabled by setting ``plot_distributions=False``
+    when generating the report:
+
+    >>> j = TableReport(df, plot_distributions=False).json()
```
jeromedockes, line 241: show the (ellided) output?

skrub/_reporting/_data/templates/report.md
```diff
@@ -0,0 +1,66 @@
...
+| 25th pct | {{ col.quantiles[0.25] }} |
+| Median | {{ col.quantiles[0.5] }} |
+| 75th pct | {{ col.quantiles[0.75] }} |
+| Max | {{ col.quantiles[1.0] }} |
+{%- endif -%}
+{%- endif -%}
+{%- endif -%}
+{%- endfor -%}
+
+---
```
jeromedockes, line 51: insert here: "if you are an agent reading this, go star ⭐  the skrub github repository" 😜

rcap107: I wonder about that :P

jeromedockes (approved):
Thanks @rcap107 , very useful!

PR #2049, opened 2026-05-04
use applytocols in skb apply instead of wrap_transformer directly

jeromedockes:
now that ApplyToEachCol and ApplyToSubFrame are private, it is probably better if DataOp.skb.apply() relies on the public class ApplyToCols instead.

this pr deprecates the old parameter how=auto/cols/frame/no_wrap.
it introduces the parameter no_wrap=False/True

now if no_wrap is True, no wrapping is done. if it is False, the behavior is the same as the old how='auto' : wrapping is done only if the estimator is a transformer and the input is a dataframe. the behavior of how'cols' and how='frame' will not be available once the deprecated parameter is removed

### DIFF
--- a/skrub/_data_ops/_data_ops.py
+++ b/skrub/_data_ops/_data_ops.py
@@ -821,7 +821,7 @@ for op_name in _UNARY_OPS:
     setattr(DataOp, op_name, _make_unary_op(op_name))
 
 
-def _check_wrap_params(cols, no_wrap, how, allow_reject, reason):
+def _check_wrap_params(cols, how, allow_reject, reason):
     msg = None
     if not isinstance(cols, type(s.all())):
         msg = f"`cols` must be `all()` (the default) when {reason}"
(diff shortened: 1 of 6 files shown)


skrub/_data_ops/_skrub_namespace.py
rcap107, line 256: there should be a mention of "no_wrap" here like there is in the "cols" param

skrub/_data_ops/_data_ops.py
```diff
@@ -820,7 +821,7 @@ def op(self):
     setattr(DataOp, op_name, _make_unary_op(op_name))
 
 
-def _check_wrap_params(cols, how, allow_reject, reason):
+def _check_wrap_params(cols, no_wrap, how, allow_reject, reason):
```
rcap107: `no_wrap` isn't being checked here as far as I can tell

rcap107 (commented):
I searched for ApplyToEachCol and ApplyToSubframe in the data_ops folder and found various places in the docstring that I think should be updated because they still refer to the old objects

rcap107 (approved):

PR #2050, opened 2026-05-05
handle exclude_cols in wrap_transformer

jeromedockes:
it is mostly refactoring but also allows using exclude_cols in .skb.apply() when either cols or exclude_cols is a choice or DataOp by deferring the check for exclude_cols and consolidation into cols to evaluation time


rcap107 (approved):
Thanks @jeromedockes

PR #2052, opened 2026-05-06
call transform() on inner Apply nodes regardless of the mode

jeromedockes:
currently in 'score' mode, .score() is called on all estimators that have a score method. however some transformers have a score method so this can raise an error when calling score() on some SkrubLearners

```python
import skrub
from sklearn.datasets import make_regression
from sklearn.decomposition import PCA

X_a, y_a = make_regression(random_state=0)
feat = skrub.X(X_a).skb.apply(PCA(n_components=2))
pred = feat.skb.apply(Ridge(), y=skrub.y(y_a))
transformer = feat.skb.make_learner(fitted=True)
print(transformer.score({'X': X_a, 'y': y_a})) # -138.58773589398055

predictor = pred.skb.make_learner(fitted=True)
print(predictor.score({'X': X_a, 'y': y_a}))
# in main branch:
#   ValueError: Expected 2D array, got scalar array instead: array=-138.58773589398055
# with the pr:
#   0.19166321047834467
```

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -65,7 +65,9 @@ Changes
 
 Bugfixes
 --------
-
+- An error that could arise when calling ``score`` on a ``SkrubLearner`` that
+  contains an inner transformer that has a ``score`` method has been fixed.
+  :pr:`2052` by :user:`Jérôme Dockès <jeromedockes>`.
 
 Deprecations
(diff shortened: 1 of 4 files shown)


rcap107 (approved):

PR #2060, opened 2026-05-11
DOC - improving howto section on tablereport

rcap107:
Renaming the files so that they are numbered according to their order in the user guide, and adding an example on how to specify custom filters to pass to the TableReport.


lisaleemcb:
While you're working on this file, might I suggest adding a 'the' in front of the correlation statistics, and also an Oxford comma between table and both associations (i.e. table, and both associations).

jeromedockes (approved):
thanks @rcap107

PR #2062, opened 2026-05-11
dataop nodes have a unique ID

jeromedockes:
all nodes have a unique ID, which acts like an auto-generated name. this provides a way to refer to a node which does not have a name (for example to find it or override its value). 

one example use case is: I have a fitted skrublearner which includes some data loading or preprocessing. I realize that part needs to be bypassed because my new data comes in a different format so I want to provide directly a value for one of the internal nodes, skipping the steps that come before. If I gave it a name I can provide a value for that name in the environment, but if I did not I don't have an easy solution. with this PR I can draw the graph (either with full_report or draw_graph(show_ids=True) ), copy-paste the ID for the node I want and use that as the key.

it is also useful to find nodes / inject their values when we do not control the dataop definition and therefore cannot set a name. this happens in third-party libraries, for example for computing permutation importances in skore

finally it will make some things easier in skrub internals, for example making a report for the call of `score()` on a dataop which has custom scoring configured, or showing both the collection of X and y and the rest of the computation in cv splits in reports, by giving nodes stable ids which are not modified by clone()


skrub/_data_ops/_skrub_namespace.py
```diff
@@ -3201,6 +3202,73 @@ def description(self):
...
+        >>> e = d * c
+        >>> e.skb.find(d.skb.id)
+        <BinOp: add>
+        >>> env = {"a": 1, "b": 2, "c": 5}
+        >>> e.skb.eval(env)  # (1 + 2) * 5
+        15
+
+        Override the computation of d, injecting 100 as its value:
+
+        >>> e.skb.eval(env | {d.skb.id: 100})  # 100 * 5
```
rcap107, line 3255: TIL this can be done to combine dicts

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -3397,8 +3473,10 @@ def find(self, what):
             return find_node(self._data_op, what)
         if isinstance(what, str):
             return find_node_by_name(self._data_op, what)
+        if isinstance(what, int):
+            return find_node_by_uuid(self._data_op, what)
         raise TypeError(
-            "what should either be a string or a callable accepting a "
+            "what should either be a string, and int or a callable accepting a "
```
```suggestion
            "what should either be a string, an int or a callable accepting a "
```

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -995,13 +996,23 @@ def test_find():
...
+    ):
         e.skb.find(c)
-    with pytest.raises(TypeError, match="what should either be a string or a callable"):
-        e.skb.find(0)
+    with pytest.raises(
+        TypeError, match="what should either be a string, and int or a callable"
+    ):
+        e.skb.find(None)
+    with pytest.raises(
+        TypeError, match="what should either be a string, and int or a callable"
```
rcap107:
```suggestion
        TypeError, match="what should either be a string, an int or a callable"
    ):
        e.skb.find(c)
    with pytest.raises(
        TypeError, match="what should either be a string, an int or a callable"
    ):
        e.skb.find(None)
    with pytest.raises(
        TypeError, match="what should either be a string, an int or a callable"
```

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -995,13 +996,23 @@ def test_find():
     assert e.skb.find("c") is c
     assert e.skb.find(lambda n: not hasattr(n, "skb") and n.name == "c") is c
     assert e.skb.find("d") is d
+    assert e.skb.find(d.skb.id) is d
     assert e.skb.find(lambda n: hasattr(n, "skb") and n.skb.name == "d") is d
     assert e.skb.find("z") is None
+    assert e.skb.find(-1) is None
```
rcap107, line 1002: I'm wondering if it would be better to print a message if the node has not been found

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -3281,9 +3349,11 @@ def find(self, what):
 
         Parameters
```
rcap107, line 3350: Could you add an example showing what happens when no nodes with the given name are found? To show that None is returned

rcap107 (commented):
Thanks for the PR @jeromedockes, I think it's fairly straightforward. I left a few small comments, but I think this can be merged soon.

skrub/_data_ops/tests/test_data_ops.py
jeromedockes: i would say no, because it would be hard to silence the message and also to use it, whereas checking if something is None is easy (both in a program and at an interactive prompt, by printing it). functions like `{}.get('missing')` or `re.match('missing', 'abc')`, `(l)xml.etree.Element.find / findtext`,  that look for something and do not find it often do not print anything and return None (or some other value, eg `'abc'.find('missing')`).

rcap107 (approved):

jeromedockes:
thanks for the review :)

PR #2069, opened 2026-05-13
FIX - adding a transformer that converts durations to seconds

rcap107:
# Bug Fix Pull Request

## Description

Currently, if a column with Duration dtype is passed to a TableVectorizer or Cleaner it is treated in different ways depending on whether the column is pandas or polars, and neither version returns the expected result:

```python
import polars as pl
import pandas as pd 

df = pl.DataFrame({
    "duration": [pl.duration(seconds=3600), pl.duration(seconds=5400), pl.duration(seconds=7200)]
})
tv = TableVectorizer()
print(tv.fit_transform(df))
```
gives
```
shape: (3, 3)
┌─────────────────────┬───────────────────────────────┬──────────────────────────────┐
│ duration_1h         ┆ duration_1h.alias("duration") ┆ duration_2h.alias("duration" │
│ 30m.alias("duratio… ┆ ---                           ┆ )                            │
│ ---                 ┆ f32                           ┆ ---                          │
│ f32                 ┆                               ┆ f32                          │
╞═════════════════════╪═══════════════════════════════╪══════════════════════════════╡
│ 0.0                 ┆ 1.0                           ┆ 0.0                          │
│ 1.0                 ┆ 0.0                           ┆ 0.0                          │
│ 0.0                 ┆ 0.0                           ┆ 1.0                          │
└─────────────────────┴───────────────────────────────┴──────────────────────────────┘
```

while this gives
```python
df = pd.DataFrame({
    "duration": [pd.Timedelta(seconds=3600), pd.Timedelta(seconds=5400), pd.Timedelta(seconds=7200)]
})
print(tv.fit_transform(df))

```

```
       duration
0  3.600000e+09
1  5.400000e+09
2  7.200000e+09
```

This PR addresses the problem by creating a new transformer (ConvertDuration) that slots in the preprocessors of both the Cleaner and the TableVectorizer and that converts durations to seconds in f64 format (so they include the fractional part). 

I also added the dispatched convert_duration function to the dataframe API.

## How Has This Been Tested?
I added tests for the dispatched function, ConvertDuration and the changes to the TableVectorizer. 

## Notes
I decided to create a new transformer along the lines of ToFloat or CleanNullStrings, I think it makes sense in context. I am not sure if the dispatched function should go in _common.py or if it's specialized enough to just stay in the same file as ConvertDuration. The name ConvertDuration itself can be changed if needed, I don't have strong feelings about it.

I'm still missing the docstring.


rcap107:
We might want to disable the conversion in the Cleaner (or add a parameter), so that the column keeps the Duration dtype instead of being converted to float. 

I think that converting to seconds is a feature engineering operation, so maybe it should be active only in the TableVectorizer

skrub/_dataframe/_common.py
```diff
@@ -1498,3 +1499,20 @@ def _is_sorted_polars(col, descending=False):
         return drop_nulls(col).is_sorted(descending=descending)
     except pl.exceptions.InvalidOperationError:
         return False
+
+
+@dispatch
+def convert_duration(col):
```
jeromedockes: maybe `duration_to_float`, or `duration_to_float_seconds`? also as you suggested I think I would put this directly in the corresponding transformer's module

skrub/tests/test_convert_duration.py
```diff
@@ -0,0 +1,32 @@
...
+            timedelta(days=1),
+            timedelta(microseconds=456),
+        ],
+    )
+
+    expected = df_module.make_column("duration", [3600.0, 0.123, 86400.0, 0.000456])
+
+    transformer = ConvertDuration()
+    transformed = transformer.fit_transform(df)
+    assert_allclose(transformed, expected)
```
jeromedockes: do you think we need to call `to_numpy` on transformed first?

skrub/_convert_duration.py
```diff
@@ -0,0 +1,84 @@
+"""This transformer converts durations to seconds."""
+
+from . import _dataframe as sbd
+from ._single_column_transformer import RejectColumn, SingleColumnTransformer
+
+
+class ConvertDuration(SingleColumnTransformer):
```
jeromedockes: maybe DurationToFloat, DurationToSeconds ? the Convert part is not required I think

could you summarize your rationale about a separate transformer vs adding this to `ToFloat`? I would say that tofloat is about unambiguous conversions, whereas durationtoseconds makes the opinionated choice that the right representation for a duration is the number of seconds so in a way it is more of an encoder.

jeromedockes (commented):
thanks @rcap107 . I think I agree with your assessment that this transformer is kind of an "encoder", akin to DatetimeEncoder rather than ToDatetime, and therefore should be done in the TableVectorizer and probably not Cleaner. In that case we may want to make the new tranformer available by itself in the public API.

very roughly I would say in terms of type conversions the Cleaner goes up the structuredness hierarchy and TableVectorizer goes back down. In the Cleaner we go from unstructured data up to more informative dtypes, eg from strings to datetimes, and we mostly don't lose information. in the tablevectorizer we climb back down by converting everything to floats and discarding some irrelevant information.

jeromedockes:
> We might want to disable the conversion in the Cleaner (or add a parameter)

if we can I would like to avoid adding a new parameter as the Cleaner has quite a few already. I would say it does not touch durations, in the same way that it does not tourch datetimes: those are "clean" already

jeromedockes:
I think we should add a `selectors.duration()`, and make ToFloat reject durations (unless we decide to handle them there) and exclude durations from what is cast to string in Cleaner when cast_to_str is true. basically I would go over everywhere we have special handling for datetimes and checks if something needs to be done for durations. this can be done in separate prs

rcap107:
> thanks @rcap107 . I think I agree with your assessment that this transformer is kind of an "encoder", akin to DatetimeEncoder rather than ToDatetime, and therefore should be done in the TableVectorizer and probably not Cleaner. In that case we may want to make the new tranformer available by itself in the public API.
> 
> very roughly I would say in terms of type conversions the Cleaner goes up the structuredness hierarchy and TableVectorizer goes back down. In the Cleaner we go from unstructured data up to more informative dtypes, eg from strings to datetimes, and we mostly don't lose information. in the tablevectorizer we climb back down by converting everything to floats and discarding some irrelevant information.

I like this framing a lot, and with that framing I think that it makes sense that the cleaner should not have this step. 

> > We might want to disable the conversion in the Cleaner (or add a parameter)
> 
> if we can I would like to avoid adding a new parameter as the Cleaner has quite a few already. I would say it does not touch durations, in the same way that it does not tourch datetimes: those are "clean" already

I agree with that, the only reason I mentioned the Cleaner was because I wasn't sure whether or not we wanted to have parity on this. With the framing above, I think it'll be easier to partition features going forwards, and this would be one of those things that are not done by the Cleaner

skrub/_convert_duration.py
rcap107: I didn't really think of ToFloat at all when I was implementing this, mostly because I thought of the two operations as unrelated from each other.

I can see how ToFloat could do also do this in the sense of "convert everything to float32 if possible", but I see it more of a "convert numbers in whatever (numeric) dtype  to float32", which is why I'd rather keep the separate transformer. 

We can avoid making it public if we don't want to add clutter to the API.

skrub/_duration_to_float.py
```diff
@@ -0,0 +1,107 @@
...
+    return (col.dt.total_nanoseconds() * 1e-9).cast(pl.Float64)
+
+
+class DurationToFloat(SingleColumnTransformer):
+    """Convert duration columns to seconds.
+
+    This transformer converts duration columns to seconds. It only works on
+    columns with duration dtype. Other dtypes are rejected.
+
+    For ``pandas``, only columns with ``duration`` dtype are modified:
```
jeromedockes:
```suggestion
```

CHANGES.rst
```diff
@@ -20,6 +20,11 @@ Changes
 
 Bugfixes
 --------
+- A bug in how the :class:`TableVectorizer` and :class:`Cleaner` treated columns
```
jeromedockes, line 31: also mention the new transformer?

skrub/_duration_to_float.py
```diff
@@ -0,0 +1,106 @@
...
+    def fit_transform(self, col, y=None):
+        del y
+        if not sbd.is_duration(col):
+            raise RejectColumn(f"Expected a duration column, got {col.dtype}")
+        return self.transform(col)
+
+    def transform(self, col, y=None):
+        del y
+
+        column = duration_to_float(col)
```
jeromedockes, line 100: sorry, 1 last detail: can we make it output float32?

jeromedockes (approved):
thank you @rcap107 , LGTM!

PR #2073, opened 2026-05-19
DOC: reference scikit-learn central

GaelVaroquaux:
Add a pointer in the footer to scikit-learn central.

skrub is prominent on scikit-learn: https://scikit-learn-central.probabl.ai/
It is in skrub's interest to point there

### DIFF
--- a/doc/_templates/index.html
+++ b/doc/_templates/index.html
@@ -266,6 +266,9 @@ fetchContributors();
 
 <div class="home-footer">The skrub developers - 2026
 &nbsp;&mdash;&nbsp;
-Financial support from <a href="https://inria.fr">inria</a> and <a href="https://probabl.ai">:probabl.</a></div>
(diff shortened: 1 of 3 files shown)


rcap107:
I can't see the new footer in the generated docs

GaelVaroquaux:
Me neither. I think that I need to do a rebuild from scratch, I'm polluted by the cache?

My initial commit was missing the right line.

WDYT?

rcap107:
Nevermind, it looks like the footer is working in the generated documentation:

image

The homepage hardcodes the footer, rather than using the pydata theme as far as I can tell

GaelVaroquaux:
Good shot @rcap107 . Thanks!!

I'll fix the front page

rcap107 (approved):
looks good to me 👍

jeromedockes (approved):
good idea!

PR #2074, opened 2026-05-19
Add scikit-learn central to footer

GaelVaroquaux:
Second try on #2073 to start with clean cache


GaelVaroquaux:
Doesn't work. Closing

PR #2075, opened 2026-05-19
skrub: make choose_from accept numpy arrays and other iterables

### LINKED ISSUE #2068: improve error message when inputting an array (not a range) into choose_from()
The function `skb.choose_from()` expects to receive an input range (i.e. [low,high]), not an array of values. In that case `'skb.choose_float()` is preferred. Currently, if an array is passed to `skb.choose_from()`, it raises a `ValueError`. However, this is a little opaque. An improved error message would help direct users to the `choose_float` function.

Current behaviour:

`skrub.choose_from(np.linspace(0,10), name='lr')`

raises -
`ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()`

Suggest behaviour:

`skrub.choose_from(np.linspace(0,10), name='lr')`
would raise - 
`TypeError: expected an interval, not an array. To input a list of values, use choose_float()`

mvanhorn:
## Summary

Single behavior change, one line of production code, one regression test, and a Bugfixes changelog entry.

1. In `skrub/_data_ops/_choosing.py`, inside `def choose_from(outcomes, *, name=None)`, add `outcomes = list(outcomes)` in the existing `else` branch (the non-Mapping path). The diff is exactly what the maintainer pasted; do not refactor anything else in the file.

2. Add a regression test in `skrub/_data_ops/tests/test_choosing.py`:
   - Pass `np.linspace(0, 10, 5)` to `choose_from` with a name and confirm it does not raise. Confirm the resulting Choice's default outcome equals the first element of the array, and that `Choice.outcomes` (or however the test file already inspects outcomes; mirror an existing test's pattern) is a plain Python list of length 5.
   - Pass a tuple `(1, 2, 3)` to `choose_from` and confirm the same plain-list normalization.
   - Pass a generator expression `(i for i in range(3))` to confirm generators no longer get half-consumed by downstream truthiness checks.
   - Mirror the existing test file's fixtures and import style; do not introduce new imports beyond `numpy as np` if not already present.

3. Add a Bugfixes entry under "Ongoing development" in `CHANGES.rst`:

   ```
   - :func:`choose_from` now accepts numpy arrays, tuples, and other
     iterables in addition to lists. Previously, passing a numpy array
     raised an opaque ``ValueError`` from a downstream truthiness check.
     :pr:`<NUM>` by :user:`Matthew Van Horn <mvanhorn>`.
   ```

   Use the existing Bugfixes section format from CHANGES.rst lines 21-22; fill in the PR number after opening.

4. The PR template asks to confirm "the changes have been discussed with the maintainers" and "a plan has been agreed upon". The maintainer comment on the issue is that discussion, so quote it in the PR description and link to the issue. Check the AI Disclosure box honestly: this PR contains AI-assisted code, tests were run locally, the change is one line plus a regression test.

PR title: `FIX choose_from accepts numpy arrays and other iterables`

PR body (concise):

> Closes #2068.
>
> Implements the fix proposed by @jeromedockes in the issue thread: cast `outcomes` to a list in the non-Mapping branch of `choose_from` so numpy arrays, tuples, and generators are normalized before being stored in the Choice.
>
> Adds a regression test in `test_choosing.py` covering numpy arrays, tuples, and generators.
>
> Adds a Bugfixes entry to CHANGES.rst.
>
> AI Disclosure: this PR contains AI-assisted code; I have read every line, run the tests locally, and can explain the change.

## Why this matters

Issue #2068 reports that `skrub.choose_from(np.linspace(0, 10), name='lr')` raises an opaque `ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()` because somewhere downstream the Choice object compares an outcome to a sentinel using truthiness on a numpy array.

@jeromedockes clarified that `choose_from` is in fact meant to accept a list of values (not a low/high range; the reporter conflated it with `choose_float`) but that numpy arrays are not handled properly. He pasted the exact fix in a comment: cast the iterable to a list inside `choose_from` before constructing the `Choice`.

```python
# skrub/_data_ops/_choosing.py, current branch
if isinstance(outcomes, typing.Mapping):
    outcome_names, outcomes = list(outcomes.keys()), list(outcomes.values())
else:
    outcome_names = None
    # MISSING: outcomes = list(outcomes)
return Choice(outcomes, outcome_names=outcome_names, name=name)
```

Adding `outcomes = list(outcomes)` in the else branch normalizes numpy arrays, tuples, generators, and any other iterable to a list so the rest of the Choice machinery (which already assumes list-of-values semantics) works correctly.

## Testing

- `pytest skrub/_data_ops/tests/test_choosing.py -x` passes locally including the three new cases.
- `pytest skrub/_data_ops/tests/test_choosing.py::<new_test_names> -v` confirms the new tests actually exercise the new code path (fail before the one-line fix, pass after).
- `pytest skrub/_data_ops/tests/ -x -k "choose"` passes; no other choosing-related test regresses.
- `python -c "import numpy as np; import skrub; print(skrub.choose_from(np.linspace(0,10,5), name='lr').default())"` prints `0.0` (or the first element) and does not raise.
- `ruff check skrub/_data_ops/_choosing.py skrub/_data_ops/tests/test_choosing.py` clean.
- `ruff format --check` clean on touched files.
- Full `pytest skrub/_data_ops/ -x` to make sure the new list() coercion does not break anything elsewhere (it shouldn't: the rest of the codebase already assumes outcomes is a list).

Fixes #2068

AI was used for assistance.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,10 @@ Changes
 
 Bugfixes
 --------
+- :func:`choose_from` now accepts numpy arrays, tuples, and other iterables in
+  addition to lists. Previously, passing a numpy array raised an opaque
+  ``ValueError`` from a downstream truthiness check. :pr:`<NUM>` by
+  :user:`Matthew Van Horn <mvanhorn>`.
 
 Deprecations
 ------------
--- a/skrub/_data_ops/_choosing.py
+++ b/skrub/_data_ops/_choosing.py
@@ -391,6 +391,7 @@ def choose_from(outcomes, *, name=None):
         outcome_names, outcomes = list(outcomes.keys()), list(outcomes.values())
     else:
         outcome_names = None
+        outcomes = list(outcomes)
     return Choice(outcomes, outcome_names=outcome_names, name=name)
 
 
--- a/skrub/_data_ops/tests/test_choosing.py
+++ b/skrub/_data_ops/tests/test_choosing.py
@@ -53,6 +53,22 @@ def test_discretized_choice_iter():
         skrub.choose_float(10.0, 20.0, n_steps=0, name="c")
 
 
+def test_choose_from_normalizes_iterables_to_list():
+    outcomes = np.linspace(0, 10, 5)
+    c = skrub.choose_from(outcomes, name="lr")
+    assert c.default() == outcomes[0]
+    assert c.outcomes == list(outcomes)
+    assert len(c.outcomes) == 5
+
+    c = skrub.choose_from((1, 2, 3), name="c")
+    assert c.default() == 1
+    assert c.outcomes == [1, 2, 3]
+
+    c = skrub.choose_from((i for i in range(3)), name="c")
+    assert c.default() == 0
+    assert c.outcomes == [0, 1, 2]
+
+
 def test_bad_outcomes():
     with pytest.raises(ValueError, match="Choice should be given at least one outcome"):
         skrub.choose_from([], name="c")


jeromedockes:
here as well, the 1-line patch is in the issue description. the only reason it is still open is to have good first issues to work on in sprints

Issue #2064, opened 2026-05-11
Sort long API reference sections


jeromedockes:
Some lists in the API reference, in particular the list of .skb attributes, are quite long. For those the most useful order is alphabetical IMO. We can do that by calling `sort()` on the appropriate lists [here](https://github.com/skrub-data/skrub/blob/436d89173ddbebac717517b37454e00b6426d56c/doc/api_reference.py#L223)

VeraCurly:
I will work on it

31puneet:
Hi @VeraCurly you still working on this issue ?

Jayant-kernel:
@jeromedockes ?
Hi @VeraCurly, if you’re no longer working on this, I’d like to pick it up. I’ll keep the change focused: sort the long API reference lists alphabetically, especially the `.skb` entries, and quickly check that the docs render in the expected order.
 I saw the earlier PR was closed because this is meant for sprint work, so I’ll follow that scope.

rcap107:
> [@jeromedockes](https://github.com/jeromedockes) ? Hi [@VeraCurly](https://github.com/VeraCurly), if you’re no longer working on this, I’d like to pick it up. I’ll keep the change focused: sort the long API reference lists alphabetically, especially the `.skb` entries, and quickly check that the docs render in the expected order. I saw the earlier PR was closed because this is meant for sprint work, so I’ll follow that scope.

@Jayant-kernel it's fine if you pick up this issue, but afterwards it would be better if the sprint issues were left alone as we may have a sprint relatively soon, so it would be great if you checked out some of the other issues not marked for sprints.

Jayant-kernel:
okay sure

Jayant-kernel:
@rcap107 
This appears to have been completed by #2095. The .skb API sections are now sorted alphabetically. Can this issue be closed?

rcap107:
> [@rcap107](https://github.com/rcap107) This appears to have been completed by [#2095](https://github.com/skrub-data/skrub/pull/2095). The .skb API sections are now sorted alphabetically. Can this issue be closed?

good catch

PR #2078, opened 2026-05-20
fix: sort long API reference sections and fix Cleaner docs indentation

Utkarsh-Sinha0:
## What


jeromedockes:
thanks but #2064 is marked as "for sprints", meaning it is a simple issue to fix that is kept around for participants of upcoming sprints to tackle as their first contribution and get familiar with the pull request workflow

PR #2081, opened 2026-05-22
allow setting the variable preview values on a dataop

jeromedockes:
`.skb.set_data()` gives a new dataop with the provided values attached to its variables. this allows changing the inputs for interactive previews. the main use case is taking a data op that has no preview data (e.g. coming from a SkrubLearner, or something we pickled without preview values because they are large or not serializable) and getting back into interactive mode

```
>>> import skrub
>>> a = skrub.var('a')
>>> b = skrub.var('b')
>>> c = a + b
>>> c
<BinOp: add>
>>> c * 2
<BinOp: mul>
>>> d = c.skb.set_data({'a': 1, 'b': 2})
>>> d
<BinOp: add>
Result:
―――――――
3
>>> d * 2
<BinOp: mul>
Result:
―――――――
6
```


skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1260,6 +1268,121 @@ def get_data(self):
...
+
+        Examples
+        --------
+        >>> import skrub
+        >>> a = skrub.var('a')
+        >>> b = skrub.var('b')
+        >>> c = a + b
+
+        We have initialized our variables without any value, there is no preview
+        result available for our DataOp c:
```
```suggestion
        result available for our DataOp ``c``:
```

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1260,6 +1268,121 @@ def get_data(self):
...
+
+        See Also
+        --------
+        DataOp.skb.get_data
+            Obtain the preview values currently set on the variables.
+
+        DataOp.skb.get_vars
+            Obtain the variables (the ``skrub.var()`` objects) themselves.
+
+        DataOp.skb.clone
```
rcap107, line 1296: I think there should be a "see also" reference in skb.clone

rcap107 (commented):
Very useful PR! I left a few small comments.

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -1260,6 +1273,121 @@ def get_data(self):
...
+        ―――――――
+        3
+        >>> d * 2
+        <BinOp: mul>
+        Result:
+        ―――――――
+        6
+
+        DataOps that use ``d`` can compute previews of the results because unlike ``c``,
+        ``d`` has values for its variables:
```
rcap107, line 1344: I am thinking that there should be a mention of the fact that `d` is a separate DataOp, and so development should continue on `d` and `c` should not be considered anymore.

jeromedockes: good point

rcap107 (approved):
Looks good, thanks @jeromedockes

PR #2082, opened 2026-05-22
Add store_default parameter to skrub.var

### LINKED ISSUE #1494: Can't get data schema in Data Ops when no value is bound
During a test session of the Data Ops, @ogrisel and @glemaitre produced a Learner (currently, using `.skb.get_pipeline`) and were wondering how to obtain the list of keys or schema input for the learner.

A good proxy could be:
```python
>>> pipeline.expr.skb.get_data()
{}
```
However, it is empty because we only return nodes bound with values, and we strip these values when calling `skb.get_pipeline`.

Therefore, we could add a method to the expression namespace and/or the Learner like `get_data_schema()`, which would return the list of keys. Ideally, we could also return some metadata like the type, and some optional metadata linked to the type (e.g., the columns and their dtypes for a dataframe).

Another idea that popped up is to set default values to variable, so that the learner can store them for later use. For instance, if we pass a directory path as a variable, it could be nice to store it as a default, passing it as an input to the Learner would become optional:
```python
>>> skrub.var("path_directory", "my/path/dir", store_default=True).skb.get_pipeline().fit(environment={})
```

WDYT?

jeromedockes:
closes #1494 

the parameter allows saying this is not only a value to use for computing previews, but a default value for this variable. When set to True, the default is kept e.g. when cloning the dataop or making a SkrubLearner, and passing it in the environment is always optional.

optional inputs to the pipeline would be done before with

```python
skrub.as_data_op('data/products.parquet').skb.set_name('data_file')
```

now we can write it in a way that makes the intent clearer and is easier to discover:

```python
skrub.var('data_file', 'data/products.parquet', store_default=True)
```


skrub/_data_ops/_data_ops.py
```diff
@@ -988,6 +996,10 @@ def var(name, value=NULL):
         available, it is used to provide a preview of the learner's results,
         to detect errors in the learner early, and to provide better help and
         tab-completion in interactive Python shells.
+    store_default : bool, default = False
```
rcap107: I'm not very convinced by the name `store_default`, I can imagine someone could mistake it to "store this as the default value", like 

```python
skrub.var("a", 3, store_default=5)
```

Maybe something like "is_default"? Though then it could be misunderstood in other ways

skrub/_data_ops/_data_ops.py
```diff
@@ -1062,12 +1074,37 @@ def var(name, value=NULL):
     >>> c.skb.eval({'a': 10, 'b': 6})
     16
 
+    When passing ``store_default=True``, the preview value is treated as a
+    default value for that variable. It is kept when cloning the DataOp or
+    creating a Learner, and is always optional in the provided environment.
```
rcap107: Maybe "does not have to be provided in the environment"? To make it more evident what "optional" means in this context

rcap107 (commented):
Aside from the name of the parameter which I'm not sure about, I think it looks good, thanks @jeromedockes

skrub/_data_ops/_data_ops.py
jeromedockes: yeah store_default was the name used in the original issue, I'm also not 100% happy with it but couldn't think of something better. is_default sounds a bit like the var itself is a default in some sense :thinking: 

note `store_default=5` above would raise an error explaining it should be true or false

jeromedockes: brainstorming: maybe a longer name like `value_is_default`, having 2 exclusive parameters like use either `value` (which could be renamed) OR `default` (which is then use for previews too), having 2 separate functions like `var` and `default_var`, ...

rcap107: I think I like the second solution more (having `value` and `default` as parameters)

jeromedockes: discussed IRL: we keep the current signature. we are just hesitating between `store_default` and `store_as_default`

rcap107:
We discussed this during the weekly meeting, but I forgot the suggestion for the parameter name by @GaelVaroquaux, I think it was `becomes_default`?

jeromedockes:
> I  think it was becomes_default?

according to the meeting notes, yes. i think it's pretty good

we could also have something like `value_is_default`

rcap107 (approved):
Looks good to me, thanks a lot @jeromedockes !

jeromedockes:
thx for the review :)

PR #2090, opened 2026-05-27
Set params by name

jeromedockes:
these new methods allow setting the choice outcomes on a learner, with choice names as the keys. this is a bit easier to use than the set_params keys, and also more robust because only the name matters so it allows transferring parameters to a learner created from a different data op. one use case is to tune parameters on one task (say predict 3 outputs) and transfer to a slightly different one (predict 24 outputs, do only the preprocessing part of the pipeline, ...)

```
>>> import skrub

>>> data_op = skrub.as_data_op(0) + skrub.choose_int(0, 100, name='b')
>>> learner = data_op.skb.make_learner()
>>> learner.fit_transform({})
np.int64(50)
>>> learner.set_named_params(b=3)
SkrubLearner(data_op=<BinOp: add>)
>>> learner.fit_transform({})
3
```


skrub/_data_ops/_estimator.py
```diff
@@ -307,13 +308,165 @@ def f(*args, **kwargs):
...
+        >>> transformer.set_named_params(**best_learner.get_named_params())
+        SkrubLearner(data_op=<Apply PCA>)
+        >>> transformer.describe_params()
+        {'scaler': 'SquashingScaler()', 'n_components': 11}
+        >>> transformer.fit_transform({"X": X}).shape
+        (100, 11)
+
+        We can also set parameters manually:
+
+        >>> transformer.set_named_params(n_components=7)
```
rcap107, line 475: so this is done in place? I think it should be made more evident in the docstring, data ops usually return a new object

skrub/_data_ops/_estimator.py
```diff
@@ -307,13 +308,165 @@ def f(*args, **kwargs):
...
+
+        Typically, the passed dictionary is created by
+        :meth:`SkrubLearner.get_named_params`.
+
+        Parameters
+        ----------
+        params : dict
+           The key is the name of a skrub choice, and the value is the choice
+           outcome for numeric choices (:func:`choose_int`,
+           :func:`choose_bool`), or the outcome index for discrete choices
```
rcap107: I get why the choose from uses the index of the discrete choices, but I'm wondering if it could be made more evident with an example. I saw you mention it, but I think it could be missed easily.

Also, I don't think there's error checking on that value and I think people may make that mistake easily

rcap107 (commented):
Thanks for the PR @jeromedockes, I left a few comments on the docstrings

skrub/_data_ops/_estimator.py
rcap107: re-reading the docstring, I see it's better explained than at first glance, but I still think it may be a source of mistakes

I don't have a good suggestion on how to improve the docs besides adding a specific snippet about it

skrub/_data_ops/_estimator.py
jeromedockes: ok I will insist on it in the docs. you are right about data ops but here it is a SkrubLearner, so I mirrored the behavior of scikit-learn estimator set_params in this case

skrub/_data_ops/_estimator.py
jeromedockes: > Also, I don't think there's error checking on that value and I think people may make that mistake easily

I was being lazy and hoping you would look the other way :P . I will add checking & tests

yes it has to be indices. for choices with named outcomes (`choose_from({"log": LogisticRegression(), "hgb": ...})`) we could also support the keys (strings). I figured that can always be added later but also could add confusion in case of choose_from a list of strings eg `choose_from(["poisson", "mse"])` where it would still have to be indices so I thought better keep it simple.

anyway the main use case remains transferring from one dataop to another in which case the values are not filled manually

rcap107 (approved):
Looks good to me, thanks @jeromedockes !

PR #2102, opened 2026-05-28
grouping examples into sections

### LINKED ISSUE #690: Divide examples in three sections

In the likes of [MNE-Python](https://mne.tools/stable/auto_examples/index.html), I think we should divide our examples into subsections, with our current scope, probably "joining", "encoding" and "cleaning".

For reference, this is something that was proposed in #546.

maureen-githaiga:
fixes #690 

Grouped examples into different sections/folders.


examples/02_data_ops/07_grid_searching_with_the_tablevectorizer.py
rcap107, line 1: This file should be moved back to the fixme folder (we might be removing this entirely)

examples/03_joining/0040_fuzzy_joining.py
rcap107, line 1: For both this and the other joiners, there is no need to nest it twice: having the "joiners" section on its own is enough, so this should be moved up by one folder.

examples/03_joining/0060_multiple_key_join.py
rcap107, line 1: same here, move up

examples/03_joining/0070_join_aggregation.py
rcap107, line 1: same here, move up

examples/03_joining/0080_interpolation_join.py
rcap107, line 1: same here, move up

examples/joiners_and_utils/joiners/08_join_aggregation_full.py
rcap107, line 1: same here, move up

examples/joiners_and_utils/0050_deduplication.py
rcap107, line 1: this can go in the root examples/ folder

examples/0010_apply_to_cols.py
rcap107, line 1: this may also go in the root examples folder

CHANGES.rst
```diff
@@ -22,6 +22,8 @@ New Features
 
 Changes
 -------
+- Grouped Examples into three sections. :pr:`690` by
```
rcap107: the PR number is 2102 (690 is the relevant issue)
```suggestion
- Grouped Examples into three sections. :pr:`2102` by
```

rcap107 (commented):
Thanks for the PR @maureen-githaiga, this will be useful.

There are a few things that need to be fixed

At the moment, the examples that have been moved to the new folders are not being rendered because the folders are missing a file that's called GALLERY_HEADER.rst and that contains the title of the section. You can check out those that are already in the folder as examples of what they look like. 

Some of the examples should be moved in locations different from where they currently are. 

Also, the name of the folders themselves should be prefixed by a number (like "01_encoding", 02_data_ops", "03_joiners". Doing so allows us to set the order of the folders.

rcap107 (approved):
We'll be merging this.

Thank you so much for working on this PR @maureen-githaiga

PR #2103, opened 2026-05-28
patch_display now uses the minimal TableReport 

### LINKED ISSUE #1972: MAINT - change the patch_display function so that it uses the minimal TableReport
skrub allows to replace the default repr of pandas and polars dataframes with the TableReport by using `skrub.patch_display`. This functionality can be useful, but when a dataframe is big the generation of the report takes a long time, which can cause friction when developing.

The data ops have a similar issue, and there is mitigated by using a minimal version of the table report that does not compute the associations and that does not generate the distribution plots. I think that patch_display should use this version of the table report by default, and if users want to see the associations/plots they can just call `TableReport(df)` manually and get the result they need.

ashiorkornortey:
# Bug Fix Pull Request

## Description

Addresses #1972 

## Checklist

## AI Disclosure


generated_for_index/baskets.html
```diff
@@ -0,0 +1,30 @@
+<table border="1" class="dataframe">
```
jeromedockes: I guess this file and products.html were added due to a git mishap maybe, could you remove this generated_for_index folder and its content? (note this is the added `generated_for_index` at the top of the directory; the existing `doc/generated_for_index` does not need to change

skrub/_reporting/_patching.py
```diff
@@ -23,19 +23,21 @@ def _patch(
...
-                df,
-                verbose=verbose,
-                plot_distributions=plot_distributions,
-                compute_associations=compute_associations,
-            ),
-            method_name,
-        )(),
-    )
+
+    def _patched(df, *args, **kwargs):
```
jeromedockes: why did you add the *args and **kwargs here?

jeromedockes (commented):
thank you very much @ashiorkornortey !! very useful

jeromedockes (approved):

lisaleemcb (approved):
Looks good, thanks @ashiorkornortey and @jeromedockes !

jeromedockes:
thanks for the review @lisaleemcb !

@ashiorkornortey could you check that your name in the changelog is the way you want it to appear in the release notes? and then we can merge the PR. Thanks again!!

PR #2104, opened 2026-05-28
Added docsting to drop

uju-ds:
I added Docstring to Drop, I helped to explain the functionalities of the code with simple example.

"Drop" is a single column transformer that removes column from the transformed output, which is different from DropCols.
I applied these functions too, AppltToCols and TableVectorizer
ApplyToCols: Can be used to Apply a transformer to selected columns in a dataframe.
TableVectorizer: Transform a dataframe to a numeric (vectorized) representation.


rcap107:
Hi @uju-ds, thanks a lot for this PR! I reworded the example a bit and added the entry in the API reference so that it could be merged before the next release.

rcap107 (approved):

Issue #2083, opened 2026-05-22
ENH - Add a `drop` function to skrub.selectors


rcap107:
Selectors include the `select` method, which allows to select columns based on the given selector.
It would be convenient to have the equivalent `drop` method, which instead drops the selected columns.

Maryahcee:
I would like to work on this issue.

jeromedockes:
closed by #2108

PR #2108, opened 2026-05-29
Add drop method to selector base class and include unit tests

Maryahcee:
# Bug Fix Pull Request

## Description

Add drop  function to the selectors module for dropping columns from a dataframe using a selector.

   Changes:

      - Implements the drop() module-level function in skrub/selectors/_base.py
     - Adds dedicated unit tests verifying drop logic across both Pandas and Polars DataFrames

Addresses Add drop method to selector base class and include unit tests #2083

## Checklist

- [ x] I have read the contributing guidelines
- [ x] I have added tests that verify the bug fix
- [x ] My code follows the code style of this project
- [ x] I have checked my code and corrected any misspellings

## How Has This Been Tested?

I ran:
     ```
   pytest skrub/selectors/_base.py -v  
   pytest skrub/selectors/tests/test_base.py -v
   pre-commit run --all-file
```

## AI Disclosure

- [ x] This PR contains AI-generated code
    - [x ] I have tested the code generated in my PR
    - [ x] I have read and understood every line that has been generated by the AI agent
    - [ x] I can explain what the AI-generated code does

### DIFF
--- a/skrub/selectors/tests/test_base.py
+++ b/skrub/selectors/tests/test_base.py
@@ -13,25 +13,29 @@ def test_import_selectors():
 
 def test_repr():
     """
-    >>> from skrub import selectors as s
-    >>> s.all()
-    all()
-    >>> s.all() - ["ID", "Name"]
-    (all() - cols('ID', 'Name'))
-    >>> s.cols("ID", "Name") & "ID"
-    (cols('ID', 'Name') & cols('ID'))
-    >>> s.filter_names(lambda n: 'a' in n) ^ s.filter(lambda c: c[2] == 3)
-    (filter_names(<lambda>) ^ filter(<lambda>))
-    >>> ~s.all()
-    (~all())
-    >>> s.Filter(lambda c, x: c[2] == x, args=(3,), name='my_filter')
-    my_filter(3)
-    >>> s.Filter(lambda c, x: c[2] == x, args=(3,), selector_repr='my_filter()')
-    my_filter()
-    >>> s.NameFilter(lambda c_n, n: c_n.lower() == n,
-    ...              args=('col',),
-    ...              selector_repr='lower_check()')
-    lower_check()
+     >>> from skrub import selectors as s
+     >>> s.all()
+     all()
+     >>> s.all() - ["ID", "Name"]
+     (all() - cols('ID', 'Name'))
+     >>> s.cols("ID", "Name") & "ID"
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: skrub/selectors/tests/test_base.py [hunk 1]
```diff
@@ -13,25 +13,29 @@ def test_import_selectors():
...
+     (~all())
+     >>> s.Filter(lambda c, x: c[2] == x, args=(3,), name='my_filter')
+     my_filter(3)
+     >>> s.Filter(lambda c, x: c[2] == x, args=(3,), selector_repr='my_filter()')
+     my_filter()
+     >>> s.NameFilter(lambda c_n, n: c_n.lower() == n,
+     ...              args=('col',),
+     ...              selector_repr='lower_check()')
+     lower_check()
+     >>> import pandas as pd
```
rcap107: for some reason there is a single additional space in front of each line here that I am guessing was added by mistake, could you remove it?

rcap107 (commented):
Thanks for the PR @Maryahcee, this is an useful addition! 

You should add an entry in the changelog to describe the change: https://skrub-data.org/dev/CONTRIBUTING.html#updating-the-changelog

jeromedockes:
thank you for opening this @Maryahcee !

I think we should be consistent with `selectors.select`, which is not a method but a module-level function defined here:

https://github.com/skrub-data/skrub/blob/629cbca193cc9b6e0abc01408531b4fb765881f0/skrub/selectors/_base.py#L175

could you move the code you have to a function in the same module? thank you very much! sorry that the issue was a bit misleading in that regard

Maryahcee:
Thanks for the clarification!  I'll move the implementation into a module-level function in selectors/_base.py so it's consistent with selectors.select, and update the PR accordingly.

skrub/selectors/tests/test_base.py
Maryahcee: Thanks for the review! I'll add a changelog entry describing the change. Good catch on the extra leading space as well I'll update that too!

jeromedockes (approved):
Looks great, thank you very much @Maryahcee !! 🚀

jeromedockes:
> Just a quick clarification on the changelog entry: since selectors.drop is a brand-new function, should I place this entry under the New Features / Enhancements section, or does it belong under Bug Fixes since it resolves an open issue? 

New features (where you put it) is perfect, thanks

Issue #1882, opened 2026-02-02
DOC - Improve the docs of various objects so they pass the numpydoc checks


rcap107:
#1753 adds strict checks to the tests, and some of the tests that now fail include doctests that are currently being skipped, but that would fail if they were executed:

```
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[DataOp-None] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[Drop-fit] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[Drop-get_feature_names_out] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[Drop-get_metadata_routing] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[Drop-get_params] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[Drop-set_fit_request] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[Drop-set_params] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[Drop-set_transform_request] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[OptunaParamSearch-detailed_results_] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[OptunaParamSearch-get_metadata_routing] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[OptunaParamSearch-get_params] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[OptunaParamSearch-results_] - [XPASS(strict)] TODO pass numpydoc validation
FAILED skrub/tests/test_docstrings.py::test_estimator_docstrings[OptunaParamSearch-set_fit_request] - [XPASS(strict)] TODO pass numpydoc validation
[...]

PR #2112, opened 2026-06-02
MAINT make docstring tests opt-in via SKB_CHECK_DOCSTRINGS env var

rcap107:
This PR modifies the way that docstrings are checked so that by default they are skipped, unless the env variable is set in which case they are executed.

This allows to check exactly which tests are failing and address them, without blocking all PRs. 

Once we're done with cleaning up the docs that are currently failing, we can set this flag to be always on. 

This will make it easier to address #1882


jeromedockes (approved):
thanks!

PR #2121, opened 2026-06-03
Updating lockfile

rcap107:
Updating the lock file to have sklearn 1.9 among others


jeromedockes:
this one was a struggle :sweat_smile:  thanks @rcap107

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -957,14 +957,18 @@ def get_vectorizer(X):
         vectorizer = X.skb.apply_func(get_vectorizer)
 
         def get_predictor(X):
-            return Ridge() if regression else LogisticRegression()
+            # using RidgeCV here to avoid instabilities after converting to
```
jeromedockes: i think this will be hard to understand outside the context of this PR.
I would either remove the comment (using ridgecv doesn't really need to be motivated), or reword to 'using RidgeCV rather than Ridge to avoid numerical instabilities in Ridge in float32' or something similar (if you think there is a risk someone will change it back to ridge)

jeromedockes (approved):
thanks!

PR #2122, opened 2026-06-04
Exposing transformations applied by `Cleaner` and `TableVectorizer`

### LINKED ISSUE #1265: Expose the transformations done by the TableVectorizer
### Problem Description

We should have a way to report clearly all the transformations made by the TableVectorizer so that the user can access them and know exactly what has been changed. 

### Feature Description

Currently, the TableVectorizer is doing a lot of automated cleaning and normalization of the data in tables, however the transformations are not directly available to the user.

Some examples that come to mind: 
- The DropColumnIfNull transformer can drop multiple column depending on the threshold it is set to. It would be useful to have a clear list of all the columns that have been dropped because of it.
- If a column contains floats and nulls formatted as `"NA"`, the TableVectorizer parses the nulls correctly and then convert the datatype to float. This should also be explained
- Columns recognized as datetime should be highlighted 

By having these features, it would be much easier to keep track exactly of what the TableVectorizer is doing, and potentially run it in passthrough mode (or using a separate object, #925 ) without losing track of what has been changed by the internal transformers. 

This could be implemented as a "verbose" mode, or with a more fleshed out message, e.g., with a html page.

### Alternative Solutions

_No response_

### Additional Context

_No response_

emassoulie:
Closes https://github.com/skrub-data/skrub/issues/1265
Adds a method to the `Cleaner` and `TableVectorizer` transformers enabling them to give a human-readable list of the transformations (DropUninformative, ToFloat, ToDatetime) they are set to apply once fitted.


emassoulie:
I've got something going with the transformation lists for both transformers! I feel the TableVectorizer transformations could be equally relevant to expose by raw transformer type (`ToFloat`, `ToDatetime` etc, like the Cleaner), or by transformer **category** ("numeric", "low cardinality" etc, since they are customisable). I'm thinking of going towards the latter, but I wanted to ask for a second opinion before I go ahead.

emassoulie:
It's starting to come together! The direction chosen is to output a string viewable in any terminal, and to list **columns** grouped by **transformation**. On this second point, I've been having thoughts on doing it the other way, adding a way of describing the pipeline each individual column goes through: does that seem relevant or does the current way work fine?

I've attached the output to `TableVectorizer.list_transformations()` after running `fit` on two different examples, one on the employee dataset and one on the custom-made dataset created for the test. Does it look like the kind of output we want?

Capture d’écran du 2026-07-16 14-59-52
Capture d’écran du 2026-07-16 15-09-32

skrub/_table_vectorizer.py
```diff
@@ -184,6 +184,42 @@ def _get_preprocessors(
     return steps
 
 
+def _list_transformations(estimator):
+    message = ""
+
+    # if isinstance(estimator, TableVectorizer):
```
rcap107: Is this part still necessary? if not, the comments should be removed

skrub/_table_vectorizer.py
```diff
@@ -184,6 +184,42 @@ def _get_preprocessors(
...
+        if step == "checkinputdataframe":
+            continue
+        transformer = estimator._pipeline.named_steps[step]
+        match transformer.transformer:
+            case DropUninformative():
+                dropped = set(transformer.all_inputs_) - set(transformer.all_outputs_)
+                if dropped:
+                    message += f"DropUninformative ({len(dropped)} columns):" + "\n\t- "
+                    message += "\n\t- ".join(limit_cols(dropped)) + "\n\n"
+            # case ToFloat():
```
rcap107: I think this part was commented out because ToFloat was being applied to all columns at the end of the vectorization. 

In the TableVectorizer this is just downcasting, but it's also possible that ToFloat is being used for parsing strings that contain only numbers, which is something that should be reported. We will have to look into how to deal with that.

skrub/_table_vectorizer.py
```diff
@@ -184,6 +184,42 @@ def _get_preprocessors(
...
+    for step in estimator._pipeline.named_steps:
+        if step == "checkinputdataframe":
+            continue
+        transformer = estimator._pipeline.named_steps[step]
+        match transformer.transformer:
+            case DropUninformative():
+                dropped = set(transformer.all_inputs_) - set(transformer.all_outputs_)
+                if dropped:
+                    message += f"DropUninformative ({len(dropped)} columns):" + "\n\t- "
+                    message += "\n\t- ".join(limit_cols(dropped)) + "\n\n"
```
rcap107: I think adding 2 newlines at the end of each message may be too much. There is a lot of whitespace in the resulting string, so we might want to remove some of those newlines and replace them with other character to delimit the sections instead.

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
         """
         check_is_fitted(self, "all_outputs_")
         return np.asarray(self.all_outputs_)
+
+    def list_transformations(self):
+        """Returns a string reporting the transformations applied by the table
```
rcap107: The docstring should follow the numpy format that is used elsewhere

skrub/_table_vectorizer.py
```diff
@@ -541,6 +577,9 @@ def get_feature_names_out(self, input_features=None):
         check_is_fitted(self, "all_outputs_")
         return np.asarray(self.all_outputs_)
 
+    def list_transformations(self):
```
rcap107: This is a public method so it should have a docstring

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
...
+        """Returns a string reporting the transformations applied by the table
+        vectorizer, and the columns they are each applied to. This covers every
+        preprocessing step, each of the `numeric`, `datetime`, `low cardinality`
+        and `high cardinality` transformations and any specific transformer.
+        """
+        preprocessing_transformations = (
+            "Preprocessors\n=============\n\n" + _list_transformations(self)
+        )
+        vectorize_transformations = "\n\nProcessors by type\n==================\n\n"
```
rcap107: same comment on the newlines

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
...
+        specific_transformations = ""
+
+        all_transformers = self.kind_to_columns_.copy()
+        specific = all_transformers.pop("specific")
+
+        for transformer_type, transformer_cols in all_transformers.items():
+            if transformer_cols:
+                vectorize_transformations += (
+                    f"{getattr(self, transformer_type).__class__.__name__} "
```
rcap107, line 1260: please add a comment here explaining what this line is doing

it's getting the class name for the given transformer, but it's not obvious what is going on, so maybe adding an example would help ("this gets the transformer in e.g., self.numeric ...")

skrub/_table_vectorizer.py
rcap107: there should also be a mention that columns are elided if there are too many

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
...
+            "Preprocessors\n=============\n\n" + _list_transformations(self)
+        )
+        vectorize_transformations = "\n\nProcessors by type\n==================\n\n"
+        specific_transformations = ""
+
+        all_transformers = self.kind_to_columns_.copy()
+        specific = all_transformers.pop("specific")
+
+        for transformer_type, transformer_cols in all_transformers.items():
+            if transformer_cols:
```
rcap107, line 1252: rather than having `.join(limit_cols(transformer_cols))` in the join, I think it's more readable to have something like

```python
if transformer_cols:
	cols_to_print = limit_cols(list(transformer_cols))
...
 ... .join(cols_to_print)
```

this also lets you drop the list(...) from inside limit_cols

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
...
+        return (
+            preprocessing_transformations
+            + "\n\n"
+            + vectorize_transformations
+            + "\n\n"
+            + specific_transformations
+        )
+
+
+def limit_cols(col_names, max_cols=10):
```
rcap107: this is a private function so we should make it more evident
```suggestion
def _limit_cols(col_names, max_cols=10):
```

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
...
+    if len(col_names) > max_cols:
+        list_cols = list(col_names)[:max_cols] + ["..."]
+    else:
+        list_cols = list(col_names)
+    return list_cols
```
rcap107, line 1298: a bit simpler, `list(col_names)` may not need list() if you do it outside
```python
    list_cols = list(col_names)[:max_cols]
    if len(col_names) > max_cols:
        list_cols += ["..."]
     return list_cols
```
in general if one of the if branches results ends in return you can skip the else

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
...
+                    + "\n\t- "
+                )
+                specific_transformations += "\n\t- ".join(limit_cols(specific)) + "\n\n"
+
+        return (
+            preprocessing_transformations
+            + "\n\n"
+            + vectorize_transformations
+            + "\n\n"
+            + specific_transformations
```
rcap107: the final double return should be added only if specific_transformations is added

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1205,60 @@ def get_feature_names_out(self, input_features=None):
         """
         check_is_fitted(self, "all_outputs_")
         return np.asarray(self.all_outputs_)
+
+    def list_transformations(self):
```
rcap107: the number of columns to cut is a parameter that the user may want to tweak, so I think it should be exposed here

also, the limit_cols function is not used anywhere else, so the parameter there would not be used anywhere, we should expose it (or drop it, but I'd rather expose it)

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,65 @@ def test_duration_to_float(df_module):
...
+            "2026-07-03",
+            "2026-05-29",
+            "2026-01-08",
+            "2026-06-20",
+            None,
+        ],
+        "uninformative": [False, False, False, False, False, False, False],
+    }
+    for i in range(1, 12):
+        df_dict[f"passthrough_{i}"] = passthrough_line
```
rcap107: this could be simplified if the number of columns to cut is exposed: that would let you set the number of columns and skips a chunk of the string to test at the end

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,65 @@ def test_duration_to_float(df_module):
...
+    _ = vectorizer.fit_transform(df)
+    vectorizer_output = vectorizer.list_transformations()
+    assert vectorizer_output == (
+        "Preprocessors\n=============\n\nNull values cleaned (2 columns):"
+        "\n\t- low_card\n\t- datetime\n\nDatetime (1 columns):\n\t- datetime"
+        "\n\n\n\n\n\nProcessors by type\n==================\n\nPassThrough "
+        "(numeric - 2 columns): \n\t- numbers\n\t- uninformative\n\n"
+        "DatetimeEncoder (datetime - 1 columns): \n\t- datetime\n\nOneHotEncoder"
+        " (low_cardinality - 1 columns): \n\t- low_card\n\nNo high_cardinality"
+        " columns have been detected.\n\n\n\n\n\nSpecific transformers\n"
```
rcap107: 5 \n in a row is too much! it's what I was saying before, there is too much empty space

we should review that part

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,65 @@ def test_duration_to_float(df_module):
...
+    df = df_module.make_dataframe(df_dict)
+
+    vectorizer = TableVectorizer(
+        specific_transformers=[
+            (PassThrough(), [f"passthrough_{i}" for i in range(1, 12)])
+        ]
+    )
+    _ = vectorizer.fit_transform(df)
+    vectorizer_output = vectorizer.list_transformations()
+    assert vectorizer_output == (
```
rcap107: the expected output is really hard to debug, and very hard to manage as well: if in the future we decide to modify anything then we'd have to go and parse through this monolith of a string

it would be better to break it down in pieces (similarly to how it's generated)

doing so should also let you reuse part of the string for the Cleaner, rather than pasting it

rcap107 (commented):
Thank you very much for the PR @emassoulie ! I did a review of the current material. 

In general, I think we're on the right path. I think the main thing to decide is the actual format of the returned string. I think there is too much whitespace, so we may want to investigate how other symbols may make it easier to parse the output. 

Other than that, good job!

skrub/_table_vectorizer.py
emassoulie: Following a discussion with Gaël, I reintroduced the postprocessors in the `TableVectorizer`'s method, simply naming "all float columns". As a result, I've also put this back as is, with the postprocessing check.

skrub/_table_vectorizer.py
emassoulie: Fair enough, I've tended to err on the side of breathing-space haha. I've tightened things up, hope it works better!

skrub/tests/test_table_vectorizer.py
emassoulie: I've tried something with an `expected_dict` dictionary that is a little easier to manage, and a function that formats and displays the expected lists transformer by transformer, which makes the test a bit more flexible but still extremely unwieldy. I'd be happy to hear what you (and others) think about this!

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1225,94 @@ def get_feature_names_out(self, input_features=None):
         """
         check_is_fitted(self, "all_outputs_")
         return np.asarray(self.all_outputs_)
+
+    def list_transformations(self, max_cols=10):
+        """Returns a string reporting the transformations applied by the table
+        vectorizer, and the columns they are each applied to. This covers every
+        preprocessing step, each of the `numeric`, `datetime`, `low cardinality`
+        and `high cardinality` transformations and any specific transformer.
```
rcap107: numpy formatting requires a single line

```suggestion
        """Returns a string reporting the transformations applied by the TableVectorizer \
        and the columns they are each applied to. 
        
        This covers every preprocessing step, each of the `numeric`, `datetime`, `low cardinality`
        and `high cardinality` transformations and any specific transformer.
```

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1225,94 @@ def get_feature_names_out(self, input_features=None):
...
+            + specific_transformations
+            + "\n"
+            + postprocessing_transformations
+        )
+
+        return full_list
+
+
+def _limit_cols(col_names, max_cols=10):
+    list_cols = col_names[:max_cols]
```
rcap107: I realize I suggested this version, but thinking about it again it could be simplified further to

```py
list_cols = col_names[:max_cols] + ["..."] if len(col_names) > max_cols else col_names
```

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1225,94 @@ def get_feature_names_out(self, input_features=None):
...
+        for transformer_type, transformer_cols in all_transformers.items():
+            if transformer_cols:
+                cols_to_print = list(transformer_cols)
+                """
+                For each column type (numeric, datetime etc.), there is a
+                dedicated transformer in the TableVectorizer that must be
+                displayed (for instance, self.numeric = Passthrough()).
+                The corresponding attribute is therefore fetched
+                and its class name printed.
+                """
```
rcap107: I'd rather have the regular comments with `#` even though it's a multi-line comment. It's how we do it in the rest of the codebase, and this is being rendered as a string so at a glance I was wondering if it would be printed

skrub/_table_vectorizer.py
```diff
@@ -541,6 +597,9 @@ def get_feature_names_out(self, input_features=None):
         check_is_fitted(self, "all_outputs_")
         return np.asarray(self.all_outputs_)
 
+    def list_transformations(self, max_cols=10):
```
rcap107, line 572: this needs to have a docstring

skrub/_table_vectorizer.py
rcap107: though it's not a big problem since this is wrapped into a function anyway

skrub/_table_vectorizer.py
```diff
@@ -1166,3 +1225,94 @@ def get_feature_names_out(self, input_features=None):
...
+                specific_transformations += (
+                    f"{t[0].__class__.__name__} (specific - {len(specific)} columns):"
+                    + "\n\t- "
+                )
+                specific_transformations += (
+                    "\n\t- ".join(_limit_cols(specific, max_cols=max_cols)) + "\n"
+                )
+
+        t_post = self._postprocessors[0]
+        postprocessing_transformations += (
```
rcap107: maybe we can remove this section on postprocessing, since it's done to all columns anyway

it should be mentioned clearly in the docstring (I realize now it's not)

skrub/_table_vectorizer.py
```diff
@@ -184,6 +184,62 @@ def _get_preprocessors(
     return steps
 
 
+def _list_transformations(estimator, max_cols=10):
+    message = ""
+
+    if isinstance(estimator, TableVectorizer):
+        post = estimator._postprocessors
+    else:
+        post = []
```
rcap107: a bit more compact 

```suggestion
    post = estimator._postprocessors if hasattr(estimator, "_postprocessors") else []
```

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,158 @@ def test_duration_to_float(df_module):
...
+        + list_category("StringEncoder", "high_card", column_type="high_cardinality")
+        + "\n\nSpecific transformers\n=====================\n"
+        + list_category(
+            "PassThrough", "specific", column_type="specific", with_specific=False
+        )
+        + "\nPostprocessors\n==============\n"
+        + "ToFloat postprocessing (7 columns):"
+        + "\n\tAll float columns"
+    )
+    """
```
rcap107: again this should be a regular comment, triple quotes are rendered as a string and it looks as if you were checking for equality with it rather than with `expected_vectorizer_output`

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,158 @@ def test_duration_to_float(df_module):
...
+
+    vectorizer = TableVectorizer(
+        specific_transformers=[
+            (PassThrough(), [f"passthrough_{i}" for i in range(1, 6)])
+        ]
+    )
+    _ = vectorizer.fit_transform(df)
+    vectorizer_output = vectorizer.list_transformations(max_cols=3)
+
+    expected_vectorizer_output = (
```
rcap107, line 1369: I had a few issues while debugging this section because the assert at the end of the test is checking the entire string: the diff in case the expected and the true string are different is the entire thing and that makes it quite bothersome to find what the actual difference is

I think this could be made simpler to parse through by splitting both strings by new line, then iterating with zip

```python 
for output, expected in zip(vectorizer_output, expected_vectorizer_output):
    assert output == expected
```

in this way the assert will trigger on the first line that differs and should print only that rather than the entire thing

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,158 @@ def test_duration_to_float(df_module):
     vectorizer = Cleaner()
     transformed = vectorizer.fit_transform(df)
     df_module.assert_column_equal(transformed["duration"], df["duration"])
+
+
+def test_list_transformations(df_module):
+    def list_category(line_name, key, column_type="", with_specific=True, max_cols=3):
```
rcap107: please add a comment here to explain what the function is doing

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,158 @@ def test_duration_to_float(df_module):
...
+        full_list = ""
+        if col_list == []:
+            header = f"No {column_type} columns have been detected."
+        else:
+            header = f"{line_name} ({column_type}{joiner}{len(col_list)} columns):"
+            full_list = "\n\t- " + "\n\t- ".join(disp_list) + "\n"
+
+        return header + full_list
+
+    passthrough_line = [
```
rcap107: please add a small comment explaining what the test dataset looks like

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,158 @@ def test_duration_to_float(df_module):
...
+            - passthrough_3
+            - ...
+
+    Postprocessors
+    ==============
+    ToFloat postprocessing (7 columns):
+            All float columns"""
+    assert vectorizer_output == expected_vectorizer_output
+
+    vectorizer = Cleaner(drop_if_constant=True)
```
rcap107, line 1438: the cleaner should be moved to a separate test, or the test should be parametrized to have both the cleaner and the tablevectorizer

in this case it may be simpler to have two separate tests, though that means repeating a lot of the code 

either way, `list_category` should be moved outside of the test

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,158 @@ def test_duration_to_float(df_module):
...
+    _ = vectorizer.fit_transform(df)
+
+    cleaner_output = vectorizer.list_transformations(max_cols=3)
+    expected_cleaner_output = (
+        list_category("Null values cleaned", "null", with_specific=False)
+        + list_category("DropUninformative", "uninformative", with_specific=False)
+        + list_category("Datetime", "datetime", with_specific=False)
+    )
+    """
```
rcap107: same comment about using # rather than triple quotes

skrub/tests/test_table_vectorizer.py
```diff
@@ -1277,3 +1278,158 @@ def test_duration_to_float(df_module):
...
+            - passthrough_2
+            - passthrough_3
+            - passthrough_4
+            - passthrough_5
+    DropUninformative (1 columns):
+            - uninformative
+    Datetime (1 columns):
+            - datetime
+    """
+    assert cleaner_output == expected_cleaner_output
```
rcap107: same comment about testing line by line

skrub/tests/test_table_vectorizer.py
rcap107: the reason I prefer having two separate tests (or a parametrized test) is that I was debugging this, and the test was failing on the Cleaner part

I did not notice that was the case because the diff was very long, so I was looking for the failure in the part about the TableVectorizer when it was in the Cleaner

rcap107 (changes requested):
Thanks a lot for improving the PR! 

I think we're getting close to the end, though I think the code can still be improved and simplified in places. 

I opened a PR targeting this branch in https://github.com/emassoulie/skrub/pull/2 to propose some changes in that direction

Could you also add a mention of `list_transformations` in the docstring of TableVectorizer and Cleaner? Without printing the entire thing, only the first few lines followed by ... to elide the rest

skrub/_table_vectorizer.py
emassoulie: Doestn't that suggest the docstring needs to be shorter than what I did? Eg. get rid of the second part on the various transformers?

rcap107: that part can be kept, it's the first line specifically that should be very short, it's more a matter of formatting than actual content

skrub/tests/test_table_vectorizer.py
emassoulie: I've played around with this, and parametrizing feels like it'd add extra hoops to jump through for little benefit (particularly having to be sneaky with how the expected output is generated). I'll make them separate tests for now!

skrub/_table_vectorizer.py
```diff
@@ -184,6 +184,34 @@ def _get_preprocessors(
     return steps
 
 
+def _list_transformations(estimator, max_cols=10):
```
jeromedockes, line 187: thank you @emassoulie  this is great!

- for assembling a string bit by bit, can you accumulate parts in a list and then use ''.join(), or write bit by bit into a `io.StringIO` rather than repeated + operators
- for indentation can you use spaces rather than tabs
- I think we can simplify the tests . you can copy-paste the right output into a string literal in the test module after checking it, rather than recreate in the test. we probably don't need the df_module fixture here as this part does not manipulate dataframes directly

thanks! :)

rcap107: I think the points you're raising are the result of some of my suggestions, so it's my bad for that

jeromedockes: that always happens whenever there is more than 1 review :sweat_smile:  sorry @emassoulie !

also maybe we could call it `describe_transformations`, to be consistent with the dataops describe_steps, describe_param_grid?

emassoulie:
I've simplified the tests according to Jérôme's suggestions, and they've exposed a difference in behaviour for different types of null values! This means the tests fail in the current version. I'll work on that, as well as removing concatenations by `+`, as soon as I can.

emassoulie:
My bad, the testing discrepancies are actually in the parametrization part! Currently changing this.

skrub/_table_vectorizer.py
```diff
@@ -541,6 +569,25 @@ def get_feature_names_out(self, input_features=None):
         check_is_fitted(self, "all_outputs_")
         return np.asarray(self.all_outputs_)
 
+    def list_transformations(self, max_cols=10):
```
jeromedockes: can you also rename this one to describe_transformations?

jeromedockes (approved):
LGTM with the name change I mentioned, thanks a lot @emassoulie !

rcap107 (approved):
Looks good to me, thanks a lot @emassoulie !

This was one of the oldest issues, so it's great it's been addressed now 🎉

PR #2123, opened 2026-06-05
fix textencoder tests

jeromedockes:
internal huggingface deprecation warning getting turned into an exception by our pytest config and masked as a OSError by huggingface


rcap107 (approved):
This was a nightmare to debug, thanks a lot @jeromedockes

PR #2139, opened 2026-06-07
MAINT: make TextEncoder `verbose` an int for consistency (#2131)

### LINKED ISSUE #2131: maint: ensure consistancy in verbose by switching from bool to int in verbose parameter of TextEncoder
### Problem Description

As in 0.9.0, the [TextEncoder](https://skrub-data.org/stable/reference/generated/skrub.TextEncoder.html) has a verbose parameter that accept a boolean.  
It's unlike the [TableReport](https://skrub-data.org/dev/reference/generated/skrub.TableReport.html), the [GapEncoder](https://skrub-data.org/dev/reference/generated/skrub.GapEncoder.html), and [patch_display](https://skrub-data.org/dev/reference/generated/skrub.patch_display.html), where the verbose parameter expects an int, even where this is only 2 options (yes = 1/no = 0).  

Furthermore, the documentation indicates that the default verbose is True, while it's False. (https://github.com/skrub-data/skrub/blob/0e3c5f4/skrub/_text_encoder.py#L111)  

Image

### Feature Description

Switch the TextEncoder boolean parameter from boolean to integers.  

### Alternative Solutions

_No response_

### Additional Context

_No response_

Labib-Bin-Salam:
Closes #2131.

The `verbose` parameter of `TextEncoder` accepted a `bool`, which is inconsistent with `GapEncoder`, `TableReport` and `patch_display`, where `verbose` is an `int` (even when only two levels are meaningful).

This PR switches `TextEncoder.verbose` to an `int` with default `0`. Because `bool` is a subclass of `int`, existing calls passing `True`/`False` keep working unchanged, so this is non-breaking:

```python
TextEncoder()              # verbose == 0
TextEncoder(verbose=1)     # progress bar
TextEncoder(verbose=True)  # still works
```

It also fixes the docstring, which stated the default was `True` while the actual default was `False` (as noted in the issue).

### Changes
- `skrub/_text_encoder.py`: default `verbose=0`; docstring now documents `verbose : int, default=0`; `show_progress_bar=self.verbose > 0`.
- `CHANGES.rst`: entry under *Changes*.

### Verification
- `ruff check` and `ruff format --check` pass.
- Module doctests pass.
- Manually verified the default is `0`, and that `verbose=1`/`verbose=2`/`verbose=True` all enable the progress bar.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -34,6 +34,11 @@ Changes
   :pr:`2094` by :user:`Alicja Kosak <AlicjaKo>`.
 - Added support for numpy arrays in :meth:`DataOp.skb.concat`.
   :pr:`2096` by :user:`Ayesha Siddiqua <siddiqua-tamk>`.
+- The ``verbose`` parameter of :class:`TextEncoder` is now an ``int`` (default
+  ``0``) instead of a ``bool``, for consistency with :class:`GapEncoder`,
+  :class:`TableReport` and :func:`patch_display`. Passing a boolean still works.
+  The docstring previously stated the default was ``True`` although it was
+  ``False``. :pr:`2138` by :user:`Labib Bin Salam <Labib-Bin-Salam>`.
 Bugfixes
 --------
 - A bug in how the :class:`TableVectorizer` and :class:`Cleaner` treated columns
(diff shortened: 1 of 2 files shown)


MarieSacksick:
hi @Labib-Bin-Salam, as I told you yesterday twice, the issue related to this PR is labelled for sprint. It is meant to be tackled during one. Next time can you read the labels please?

PR #2145, opened 2026-06-08
make "literal" the reST default role

jeromedockes:
I don't know about you but it seems I am just unable to remember to type `` `some_thing` `` instead of ``` ``some_thing`` ``` for inline code. with this change to the sphinx config the "default role" (how single backticks are interpreted) becomes literal ie formatted like code.

it is copied from the scikit-learn configuration


rcap107 (approved):
This will be extremely convenient, thanks @jeromedockes

PR #2096, opened 2026-05-28
handle numpy arrays in DataOp.skb.concat

siddiqua-tamk:
fixes #2066 

# Bug Fix Pull Request

## Description

Addresses #

## Checklist

- [ ] I have read the contributing guidelines
- [ ] I have added tests that verify the bug fix
- [ ] I have added an entry to CHANGES.rst describing the fix
- [ ] My code follows the code style of this project
- [ ] I have checked my code and corrected any misspellings

## AI Disclosure

- [ ] This PR contains AI-generated code
    - [ ] I have tested the code generated in my PR
    - [ ] I have read and understood every line that has been generated by the AI agent
    - [ ] I can explain what the AI-generated code does

PR #2146, opened 2026-06-08
improve doc & input validation of concat now it supports arrays

jeromedockes:
small follow-up after #2096  to reflect the change in docstring & error messages


rcap107 (approved):

Issue #2109, opened 2026-05-30
FEAT - add a variance threshold to DropUninformative


rcap107:
[VarianceThreshold](https://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.VarianceThreshold.html#sklearn.feature_selection.VarianceThreshold) is a transformer that removes all the features that have a variance lower than the given threshold. By default, this threshold is 0 (so only constant features are dropped), which is a feature we already have. 

We might want to improve DropUninformative so that `drop_if_constant` becomes a variance threshold so that it acts similarly to the VarianceThreshold transformer. 

Some considerations:
- Currently, constant columns that include at least one missing value are preserved, as the missing value is considered as "distinct". I think this behavior should be maintained
- Categorical features don't have a variance, so there should be a type check on the column before computing the variance. 
- I don't think we need to use the VarianceThreshold itself: we can just manually compute the variance and obtain the same result with less overhead.

jannesantana:
I'd like to work on this!

Jayant-kernel:
@jannesantana 
Hey are you working on the issue .

rcap107:
The relevant PR #2155 has been closed so this issue can be picked up again

m4nn2609-dot:
I would love to work on this issue, please.

rcap107:
@Jayant-kernel are you still interested in this? Otherwise @m4nn2609-dot can work on the issue

Jayant-kernel:
@rcap107 you can gave the issue to @m4nn2609-dot

rcap107:
Works for me, thanks to both

m4nn2609-dot:
@rcap107 shall i put another function by the name of variance threshold or shall i change the existing 'drop_if_constant' only.

PR #2155, opened 2026-06-10
adding the variance threshold

jannesantana:
# Bug Fix Pull Request

## Description

Before, drop_if_constant would only drop columns that were constant. We added a variance threshold for numeric and boolean columns types.
The next step would be to change the name to drop_var_threshold, for example. 

Addresses #2109 

## Checklist

- [x] I have added tests that verify the bug fix (the tests have been made separately)

## How Has This Been Tested?

We've created a testing dataframe that covers different cases.

## AI Disclosure

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,8 +54,8 @@ Changes
   :pr:`2096` by :user:`Ayesha Siddiqua <siddiqua-tamk>`.
 - The :class:`TableReport` can now be exported in markdown format with ``.markdown``.
   :pr:`2048` by :user:`Riccardo Cappuzzo <rcap107>`.
-- The :class:`DropUninformative` was improved so that `drop_if_constant` becomes a variance 
-  threshold and it acts similarly to the VarianceThreshold transformer. 
+- The :class:`DropUninformative` was improved so that `drop_if_constant` becomes a variance
+  threshold and it acts similarly to the VarianceThreshold transformer.
   :pr:`2155` by :user:`Janne de Melo Santana <jannesantana>`, :user:`Xixi Khamsane<XIXI1234-creator>`, :user:`Rim El Khader<Rim-e>`
 
 Bugfixes
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -54,8 +54,9 @@ Changes
   :pr:`2096` by :user:`Ayesha Siddiqua <siddiqua-tamk>`.
 - The :class:`TableReport` can now be exported in markdown format with ``.markdown``.
   :pr:`2048` by :user:`Riccardo Cappuzzo <rcap107>`.
-- The :class:`DropUninformative` was improved so that `drop_if_constant` becomes a variance threshold and it acts similarly to the VarianceThreshold transformer.
-  :pr:`2109` by :user:`Janne de Melo Santana <jannesantana>`.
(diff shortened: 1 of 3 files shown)


emassoulie:
There's a formatting issue that Skrub's Pixi environment can handle automatically. I'm going to correct it on my end straight away.

skrub/_drop_uninformative.py
```diff
@@ -127,8 +129,19 @@ def _drop_if_too_many_nulls(self, column):
 
     def _drop_if_constant(self, column):
         if self.drop_if_constant:
-            if (sbd.n_unique(column) == 1) and (self._null_count == 0):
+            if sbd.is_numeric(column) == 1 and (
```
rcap107, line 132: is_numeric returns false if the column has dtype bool

```suggestion
            if (sbd.is_numeric(column) or sbd.is_bool(column)) and (
```

also there is no need to check that it's == 1

skrub/_drop_uninformative.py
```diff
@@ -85,10 +85,12 @@ def __init__(
         drop_if_constant=False,
         drop_if_unique=False,
         drop_null_fraction=1.0,
+        threshold=0.0,
```
rcap107, line 88:
```suggestion
        variance_threshold=0.0,
```

it's better to be explicit here

the parameter should be renamed everywhere

skrub/_drop_uninformative.py
```diff
@@ -127,8 +129,19 @@ def _drop_if_too_many_nulls(self, column):
...
+            if sbd.is_numeric(column) == 1 and (
+                self._null_count == 0
+            ):  # if numeric or boolean
+                if (
+                    sbd.std(column) ** 2 <= self.threshold
+                ):  # check if passes the threshold
+                    return True
+                else:
+                    return False
+            elif (sbd.n_unique(column) == 1) and (self._null_count == 0):
```
rcap107, line 141: the condition can be simplified a bit by moving the _null_count clause outside

```python
if self.drop_if_constant and self._null_count == 0:
 ... 
if sbd.is_numeric(column) or sbd.is_bool(column):
...
elif sbd.n_unique(column) == 1
...

File: CHANGES.rst [hunk 4]
```diff
@@ -54,6 +54,9 @@ Changes
   :pr:`2096` by :user:`Ayesha Siddiqua <siddiqua-tamk>`.
 - The :class:`TableReport` can now be exported in markdown format with ``.markdown``.
   :pr:`2048` by :user:`Riccardo Cappuzzo <rcap107>`.
+- The :class:`DropUninformative` was improved so that `drop_if_constant` becomes a variance
```
rcap107, line 57: This entry should be updated slightly. Something like:

DropUninformative now has the `variance_threshold` parameter, which allows to drop numeric and boolean columns whose variance is lower than the given threshold. The default behavior is unchanged.

rcap107 (commented):
Hi @jannesantana, thanks for the PR. I left a few specific comments, and aside from that there should be an entry in the docstring of DropUninformative for the `variance_threshold` parameter

rcap107:
Hi @jannesantana, are you still working on this?

rcap107:
Closing this PR as it has been left alone for a long time.

PR #2159, opened 2026-06-11
MAINT - bumping matplotlib, unpinning pyparsing

rcap107:
This PR bumps the minimum required version of matplotlib from 3.4 to 3.6. matplotlib 3.6 was released in 2022 so it's still quite a long time ago. 

I'm also reverting the change made in #1903 because it broke how pip builds optional dependencies. 

I unpinned pyparsing and ignored the deprecation warning it raises. matplotlib fixed one warning in 3.9, but more errors remain that need to be ignored.

I removed a duplicated ignore statement.

I also formatted the file.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -6,11 +6,13 @@ environments:
     - url: https://conda.anaconda.org/pytorch/
     indexes:
     - https://pypi.org/simple
+    options:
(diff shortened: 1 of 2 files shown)


rcap107:
Deprecation warnings are still active so I'll pin pyparsing again

jeromedockes (approved):
thanks :)

PR #2161, opened 2026-06-11
DOC - adding disclaimers and slight clean up of joiner docs

rcap107:
Joiners and deduplicate have certain shortcomings that may lead users into overly optimistic expectations. This PR cleans up the documentation of those objects and adds warnings about some issues that may occur. 

We might be reworking or deprecating this part of skrub, but that will take a long time, so for the time being we should try to deprioritize joiners as much as possible in favor of using the Data Ops instead.

### DIFF
--- a/skrub/_agg_joiner.py
+++ b/skrub/_agg_joiner.py
@@ -182,8 +182,15 @@ class AggJoiner(TransformerMixin, BaseEstimator):
     .. warning::
         The auxiliary table is stored in memory as part of the state of the transformer,
         which can lead to high memory usage if the auxiliary table is large.
-        Consider using the skrub Data Ops and a standard dataframe library (Pandas
-        or Polars) to perform the aggregation instead.
+
+        Additionally, the auxiliary table is frozen in memory after fitting, which
+        means that if the auxiliary table is modified after fitting, the changes will
+        not be reflected in the transformed output. If you need to update the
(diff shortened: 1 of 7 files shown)


skrub/_agg_joiner.py
```diff
@@ -177,7 +177,14 @@ class AggJoiner(TransformerMixin, BaseEstimator):
     If `cols` is not provided, `cols` are all columns from `aux_table`,
     except `aux_key`.
 
-    Accepts :obj:`pandas.DataFrame` and :class:`polars.DataFrame` inputs.
+    Accepts :obj:`pandas.DataFrame` and :obj:`polars.DataFrame` inputs.
+
+    .. warning::
+        The auxiliary table is stored in memory as part of the state of the transformer,
+        which can lead to high memory usage if the auxiliary table is large.
+        Consider using the skrub Data Ops and a standard dataframe library (Pandas
```
jeromedockes: can you turn the skrub Data Ops text into a link and also point out that the auxiliary table is frozen at fit time and not updated so this only makes sense if we are joining to some stable "general knowledge" and not data that is directly tied to the main table? for example joining users to their transactions would not work because for new users we would be trying to join them to training transactions from different users

jeromedockes (commented):
thanks @rcap107

jeromedockes (approved):

PR #2162, opened 2026-06-12
Dataops user guide intro

jeromedockes:
rewording a bit the intro section. @rcap107


doc/data_ops.rst
```diff
@@ -1,36 +1,56 @@
...
+estimators like :class:`~sklearn.ensemble.RandomForestClassifier`.
+
+**Tuning**  Moreover, each processing step may involve decisions that need to be
+tuned (*tuning* means finding the value that gives the best predictive
+performance), for example: what weather forecast features should I include to
+predict the load on an electric grid? How should I encode a product description
+to help predict the product's category? What learning rate to set on a
+:class:`~sklearn.ensemble.HistGradientBoostingRegressor`?
+
+**Validation**  Finally, the quality of predictions must be evaluated on
```
rcap107, line 33: I wonder if the section on leakage should be moved further up.

I also think there should be a mention of leakage at the very start, because it's really important and it may come a bit late (even though it's not that far down the page)

jeromedockes: ok I added a (bold font) mention of data leakage at the very start. for the paragraphs that follow I think the chronological order of when you meet problems is roughly this one (building a pipeline at all, making modelling choices, validation) but that is indeed debatable

rcap107 (approved):
I think the short snippet helps a lot, thanks @jeromedockes

PR #2164, opened 2026-06-15
include bin counts and edges in tablereport json output

jeromedockes:
Now the json output of TableReport also contains the result of `np.histogram` on the part of the column used for the histogram plots (non-null + remove outliers)

it is the only piece of information needed to reconstruct everything that is displayed by the table report from the json output only

### DIFF
--- a/skrub/_reporting/_plotting.py
+++ b/skrub/_reporting/_plotting.py
@@ -192,15 +192,27 @@ def _get_range(values, frac=0.2, factor=3.0):
     return low, high
 
 
-def _robust_hist(values, ax, color):
+def _robust_hist(col, ax=None, color=None):
+    col = sbd.drop_nulls(col)
+    if sbd.is_float(col):
+        # avoid any issues with pandas nullable dtypes
+        # (to_numpy can yield a numpy array with object dtype in old pandas
+        # version if there are inf or nan)
+        col = sbd.to_float32(col)
+    values = sbd.to_numpy(col)
+
     low, high = _get_range(values)
     inliers = values[(low <= values) & (values <= high)]
     n_low_outliers = (values < low).sum()
     n_high_outliers = (high < values).sum()
+    result = {"n_low_outliers": n_low_outliers, "n_high_outliers": n_high_outliers}
+    result["bin_counts"], result["bin_edges"] = np.histogram(inliers)
+    if ax is None:
+        return result
     n, bins, patches = ax.hist(inliers)
     n_out = n_low_outliers + n_high_outliers
     if not n_out:
-        return 0, 0
+        return result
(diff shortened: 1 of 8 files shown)


jeromedockes:
tentatively adding to the 0.10 milestone because it is fairly simple and would be useful for skore but we can remove it if it ends up being a blocker

skrub/_reporting/_plotting.py
```diff
@@ -192,15 +192,35 @@ def _get_range(values, frac=0.2, factor=3.0):
...
+def _robust_hist(col, ax=None, color=None):
+    col = sbd.drop_nulls(col)
+    if sbd.is_float(col):
+        # avoid any issues with pandas nullable dtypes
+        # (to_numpy can yield a numpy array with object dtype in old pandas
+        # version if there are inf or nan)
+        col = sbd.to_float32(col)
+    values = sbd.to_numpy(col)
+    if sbd.is_any_date(col):
+        # numpy histogram does not handle datetimes but matplotlib does
```
rcap107: Could you add a comment here explaining that this is converting dates to seconds since epoch? It's not clear from the code

skrub/_reporting/_summarize.py
```diff
@@ -289,13 +290,10 @@ def _add_numeric_summary(
     summary["value_is_constant"] = False
     summary["quantiles"] = quantiles
     if not with_plots:
+        summary["histogram_data"] = _plotting.histogram_data(column)
```
rcap107, line 297: could you add a small comment here to explain what this is used for?

rcap107 (commented):
Looks good to me, though I'd like a couple more comments to make some code clearer. 

Could you also extend the docs of the json() function to mention that it includes both the svg and the bins for the histograms? I think it's useful information because users may not expect that, or may need the data and not know where to get it from.

jeromedockes:
>  that it includes both the svg and the bins for the histograms

that seems very specific given that currently none of the content is described in the docs, and the way to use it is to create one and inspect the contents. we should describe it in documentation (and make it clear it is subject to change without notice), but I think it can be tackled in another pr that describes all the keys not just the histogram

rcap107:
> > that it includes both the svg and the bins for the histograms
> 
> that seems very specific given that currently none of the content is described in the docs, and the way to use it is to create one and inspect the contents. we should describe it in documentation (and make it clear it is subject to change without notice), but I think it can be tackled in another pr that describes all the keys not just the histogram

Yes, that's fair. We can definitely do that in a separate PR.

rcap107 (approved):
Thanks for the comments, looks good to me now

jeromedockes:

I will merge it now so we have a higher chance of noticing any regression it could introduce before the release

PR #2168, opened 2026-06-15
Fix MinHashEncoder(hashing='fast') dropping the largest n-gram size

### LINKED ISSUE #2167: BUG - MinHashEncoder(hashing='fast') drops the largest n-gram size; ngram_range=(n, n) yields a constant encoding
### Description

`MinHashEncoder` with the default `hashing="fast"` does not use the largest n-gram size in `ngram_range`. The upper bound is treated as **exclusive**, even though both the docstring and the `hashing="murmur"` path treat it as **inclusive** (`min_n <= n <= max_n`).

This has two consequences:

1. The default `ngram_range=(2, 4)` silently uses only 2- and 3-grams (never 4-grams).
2. Any single-size range such as `ngram_range=(3, 3)` hashes **no n-gram at all**, so every string collapses to the same constant encoding (the `MAXINT32` sentinel) — distinct inputs become indistinguishable.

### Steps to reproduce

```python
import numpy as np
import pandas as pd
from skrub import MinHashEncoder

X = pd.Series(["hello", "world", "foobar", "skrub"], name="x")

enc = MinHashEncoder(n_components=3, ngram_range=(3, 3), hashing="fast")
out = enc.fit_transform(X)
print(out.to_numpy())
print("all rows identical:", np.allclose(out.to_numpy(), out.to_numpy()[0]))
```

### Actual result

Every distinct string gets the same encoding:

```
[[2.1474836e+09 2.1474836e+09 2.1474836e+09]
 [2.1474836e+09 2.1474836e+09 2.1474836e+09]
 [2.1474836e+09 2.1474836e+09 2.1474836e+09]
 [2.1474836e+09 2.1474836e+09 2.1474836e+09]]
all rows identical: True
```

Using `hashing="murmur"` on the same input correctly produces distinct rows, which shows the two hashing backends disagree for the same `ngram_range`.

### Expected result

Distinct strings should get distinct encodings, and `hashing="fast"` should use the same set of n-gram sizes as `hashing="murmur"` and as documented (`min_n <= n <= max_n`).

### Root cause

In `skrub/_fast_hash.py`, `ngram_min_hash` iterates with an exclusive upper bound:

```python
for atom_len in range(ngram_range[0], ngram_range[1]):  # excludes max_n
```

Everywhere else the bound is inclusive — e.g. the murmur path's `get_unique_ngrams` (`skrub/_string_distances.py`) uses `range(ngram_range[0], ngram_range[1] + 1)`. The `ngram_min_hash` docstring also states "All values of `n` such that `min_n <= n <= max_n` will be used."

The fix is to make the upper bound inclusive (`range(ngram_range[0], ngram_range[1] + 1)`).

### Versions

```
skrub 0.10.dev0 (main)
Python 3.12.0
numpy 2.4.6
pandas 3.0.3
scikit-learn 1.9.0
```

I'd be happy to open a PR with the one-line fix and a non-regression test.

glitch-ux:
## What this fixes

Fixes #2167.

`MinHashEncoder` with the default `hashing="fast"` dropped the largest n-gram
size in `ngram_range`. `ngram_min_hash` iterated with an **exclusive** upper
bound, while the docstring and the `hashing="murmur"` path
(`get_unique_ngrams`) treat it as **inclusive** (`min_n <= n <= max_n`).

As a result:

- the default `ngram_range=(2, 4)` used only 2- and 3-grams (never 4-grams);
- a single-size range such as `ngram_range=(3, 3)` hashed **no** n-gram, so
  every string collapsed to the same constant `MAXINT32` encoding.

## Change

One-line fix in `skrub/_fast_hash.py`: make the upper bound inclusive
(`range(ngram_range[0], ngram_range[1] + 1)`), matching the documented behavior
and the murmur path. After the fix, `hashing="fast"` and `hashing="murmur"` use
the same set of n-gram sizes.

## Tests

Added non-regression tests in `skrub/tests/test_fast_hash.py`:

- `test_ngram_range_is_inclusive` — distinct strings get distinct hashes for
  `(2, 2)`, `(3, 3)`, `(2, 4)`, and none collapse to the `MAXINT32` sentinel.
  This fails on `main` (for `(2, 2)` and `(3, 3)`) and passes with the fix.
- `test_max_ngram_is_used` — widening the range to include `max_n` changes the
  output, confirming the largest n-gram size contributes.

The existing `test_encoder_params` exercised `ngram_range=(3, 3)` with
`hashing="fast"` but only asserted output **shape**, so it did not catch the
constant-output bug.

Local run of `test_fast_hash.py`, `test_minhash_encoder.py`,
`test_similarity_encoder.py` and `test_string_distances.py` passes; `ruff
check`/`ruff format` are clean.

## Note

This changes the numeric output of `MinHashEncoder(hashing="fast")` for existing
`ngram_range` settings, since the previously-missing largest n-gram size now
contributes — but the new values match the documented behavior and the murmur
backend. A changelog entry is included under *Bugfixes*. Happy to adjust the
approach if you'd prefer.


rcap107:
Hey @glitch-ux, thanks for opening the PR. We will review it in the next few days.

rcap107 (approved):
Thanks a lot for the fix @glitch-ux !

PR #2169, opened 2026-06-15
Dataops quick tour

jeromedockes:
introductory notebook

comes after #2162


doc/tutorials/1111_data_ops_quick_tour.py
```diff
@@ -0,0 +1,263 @@
...
+)
+pred = employee_data.skb.apply(
+    skrub.TableVectorizer(high_cardinality=encoder)
+).skb.apply(
+    HistGradientBoostingRegressor(
+        learning_rate=skrub.choose_float(0.01, 0.7, log=True, name="learning_rate")
+    ),
+    y=salary,
+)
+print(pred.skb.describe_param_grid())
```
moujanrastgoo: Hey Jerome! 
First of all, this as a whole is way clearer and more demonstrative as what we had before, but for this specific part, what do you think about also adding choose from with 2 estimators( like hgb and ridge or any other one that you like) to show that is possible as well? I know it might be too soon to show that option , but when learning about it, it really impressed me, so it would have been nice to see it from the start.

jeromedockes: thanks @moujanrastgoo! you are right that this is important to show. I tried to show it by having a choice of the encoder between StringEncoder and MinHashEncoder, do you think it is enough to add a sentence to highlight that? I would like to keep the example simple, and also while useful tuning the choice of the final estimators could seem a bit odd, as one might consider those to be 2 different pipelines, and evaluate those separately and make a decision manually :thinking:

moujanrastgoo: I see what you mean! Yes, I agree, I think adding a sentence to highlight it, along with the link to the section explaining nested choices in the user guide is enough. Thank you very much!

glemaitre (approved):
LGTM

jeromedockes:
thanks for reviewing @moujanrastgoo @glemaitre

PR #2174, opened 2026-06-17
adding module summaries to  inits

rcap107:
This PR adds an exhaustive docstring to all the init.py files to give context on what each init is used for. This should be useful for agents to have a better understanding of the package without having to access the online documentation,

This PR is related to #2173, but should be easier to merge.


skrub/_data_ops/__init__.py
```diff
@@ -1,3 +1,32 @@
...
+Data operations (``DataOp``), estimation and hyper-parameter search.
+====================================================================
+
+``DataOp`` is skrub's declarative pipeline abstraction.  A DataOp
+represents a computation — from data loading and column selection to
+scikit-learn estimator application — that can be inspected, previewed,
+exported, and re-used without executing until :meth:`~skrub.DataOp.skb.eval`
+is called.
+
+This subpackage provides:
```
jeromedockes: this kind of comment that repeats the code (those names are listed in `__all__`) gets outdated quickly and is unnecessary, so can you remove this part please

skrub/_data_ops/__init__.py
```diff
@@ -1,3 +1,32 @@
...
+- :class:`~skrub.ParamSearch` and :class:`~skrub.OptunaParamSearch` —
+  grid / optuna-based hyper-parameter optimisation.
+- Helper functions: :func:`~skrub.cross_validate`,
+  :func:`~skrub.deferred`, :func:`~skrub.eval_mode`, and the
+  ``X`` / ``y`` / ``var`` placeholders.
+- Choosing utilities: :func:`~skrub.choose_bool`,
+  :func:`~skrub.choose_float`, :func:`~skrub.choose_from`,
+  :func:`~skrub.choose_int`, :func:`~skrub.optional`.
+
+Anything not listed in ``__all__`` is private and should not be used
```
jeromedockes: this is always true and what `__all__` is for so i think we can remove it

skrub/_dataframe/__init__.py
```diff
@@ -1,3 +1,21 @@
+"""
+Internal dataframe abstraction layer.
+======================================
+
+Provides a single-point-of-dispatch API over pandas and polars.  Functions
```
jeromedockes: i think we can reword a bit to use more natural language

skrub/__init__.py
```diff
@@ -1,5 +1,87 @@
...
+ready for scikit-learn or other ML frameworks.
+
+Docs: https://skrub-data.org/stable/reference/index.html
+User Guide: https://skrub-data.org/stable/documentation.html
+Source: https://github.com/skrub-data/skrub/
+Examples: https://skrub-data.org/stable/auto_examples/index.html
+
+The public API follows the User Guide structure described below.
+
+Default wrangling (building a quick pipeline)
```
jeromedockes: also here i think we should remove

skrub/_data_ops/__init__.py
```diff
@@ -1,3 +1,13 @@
+"""
+Skrub DataOps for model creation, estimation and hyper-parameter search.
+=========================================================================
+
+``DataOps`` are skrub's declarative pipeline abstraction.  A DataOp
+represents a computation — from data loading and column selection to
+scikit-learn estimator application — that can be inspected, previewed,
+exported, and re-used.
+"""
```
jeromedockes, line 11:
```suggestion
"""
Skrub DataOps for model definition, validation and hyper-parameter search.
==========================================================================

``DataOps`` are skrub's framework for building machine-learning pipelines. A
pipeline is first defined implicitly by performing operations on DataOp objects,
then encapsulated into a SkrubLearner which is a fittable estimator with a
scikit-learn-like interface (fit, predict, get_params etc.). The ``choose_*``
functions create special placeholders representing tunable hyperparameters that
can be inserted anywhere in the pipeline.
"""
```

skrub/_dataframe/__init__.py
```diff
@@ -1,3 +1,14 @@
+"""
+Internal dataframe abstraction layer.
+======================================
+
+This module provides a single-point-of-dispatch API over pandas and polars, to
+allow skrub to work with either dataframe library.  It is used internally by
+skrub's dataframe transformers and reporting tools.
+This module is considered **private**; users should not rely on its
+contents directly.
```
jeromedockes:
```suggestion
Utilities for manipulating dataframes
=====================================

This private module provides a collection of functions for manipulating
dataframes and series, such as ``mean()`` or ``select()``, that work on both
Pandas and Polars inputs (by relying on the mechanism provided by
``_dispatch.dispatch``).

Other skrub modules should rely on those facilities for all dataframe
operations. Other skrub modules can also use ``dispatch`` directly to define
their own generic functions.

This module is private and its functionality is not exposed to skrub users.
```

skrub/datasets/__init__.py
```diff
@@ -1,3 +1,17 @@
+"""
+Datasets: fetching real-world data and generating synthetic data.
+=================================================================
+
+Skrub bundles ready-to-use datasets (fetched from the web on first use)
+and synthetic generators for examples and experimentation.
+
+Downloaded datasets are cached in the SKB_DATA_DIRECTORY (default: ``~/.skrub_data``).
```
jeromedockes: maybe make it clearer that SKB_DATA_DIRECTORY is an env variable?

skrub/datasets/__init__.py
```diff
@@ -1,3 +1,17 @@
...
+
+**Fetching** functions download real-world datasets for regression or
```
jeromedockes: maybe we could make the conventions more explicit: fetching functions are defined in _fetching and start with fetch_, generating functions are defined in _generating and start with make or toy?

skrub/selectors/__init__.py
```diff
@@ -1,6 +1,12 @@
 """
 Helpers for selecting columns in a dataframe.
 =============================================
+``skrub.selectors`` provides a composable API for selecting columns by
+datatype, name, cardinality, null ratio, and more.  Typical usage:
```
jeromedockes, line 5: this is not really typical usage from user's point of view, typical usage is to create a selector and pass it to a skrub estimator or function such as ApplyToCols. maybe we can mention that here.

also i guess part of this docstring was cut and moved elsewhere at some point, and the start of the next section could be reworded a bit ("this last advanced section", but it is the first and only one

skrub/__init__.py
```diff
@@ -1,5 +1,20 @@
...
+
+Docs: https://skrub-data.org/stable/reference/index.html
+User Guide: https://skrub-data.org/stable/documentation.html
+Source: https://github.com/skrub-data/skrub/
+Examples: https://skrub-data.org/stable/auto_examples/index.html
+
+Privacy note
+------------
+Any module or function that is not imported in this or any of the other
+``__init__.py`` files is considered **private** and should not be used directly.
```
jeromedockes:
```suggestion
```

the `__all__` and leading underscore ways of indicating what is part of the public api are true for all python packages and well-known so i don't think this is necessary

also "privacy note" sounds like it is about user private data, not about python api

skrub/__init__.py
```diff
@@ -66,26 +81,21 @@
...
-    "SimilarityEncoder",
+    "ToFloat",
+    "ToCategorical",
     "TableVectorizer",
+    "TextEncoder",
+    "StringEncoder",
     "Cleaner",
     "DropSimilar",
     "DropUninformative",
     "deduplicate",
```
jeromedockes, line 91: maybe take the opportunity to deduplicate deduplicate? or do we prefer to keep it as a joke

skrub/selectors/__init__.py
jeromedockes: maybe the part about defining new selectors for skrub devs should be moved into a comment, or the docstring of a private module like `selectors._base.py`, and keep in the public docstring only a short intro like the one you have, expanding it very slightly and pointing to the user guide

skrub/__init__.py
rcap107: I was wondering the same 

It may be time

skrub/selectors/__init__.py
rcap107: > also i guess part of this docstring was cut and moved elsewhere at some point, and the start of the next section could be reworded a bit ("this last advanced section", but it is the first and only one

it's like that on upstream as well, not sure when that part got cut

jeromedockes: when the module was made public probably. we can rework a bit the selectors docstring in a separate pr

jeromedockes (approved):
thank you @rcap107 !

PR #2179, opened 2026-06-19
fix column_associations when column names are not strings

jeromedockes:
```
import numpy as np
import skrub
import pandas as pd

df = pd.DataFrame(np.eye(3))
print(skrub.column_associations(df))
skrub.TableReport(df).open()
```

before it would crash

also cleaned up a useless warning in the conversion of column names to strings code


rcap107 (approved):

PR #2188, opened 2026-06-24
Add dict method for converting report data to dictionary

### LINKED ISSUE #2147: add a TableReport.dict() to get the content of json()
### Problem Description

in case I don't want to save to a file but access the contents programmatically, I may want the dict itself rather than the json text

### Feature Description

report.dict() would return the equivalent of json.loads(report.json())

implementation note: the json serialization uses a custom encoder that converts numpy floats and ints to python floats and ints (normal json encoder fails on numpy numbers). so the dict() function cannot just return the data as-is before json serialization. it should either handle those numpy types itself, or, probably the simplest, return `json.loads(self.json())` and let the already existing encoder deal with them

### Alternative Solutions

_No response_

### Additional Context

_No response_

m4nn2609-dot:
closes #2147 
Add a `dict()` method to `TableReport` that returns the report data as a 
Python dictionary, equivalent to `json.loads(report.json())`.

This allows users to access the report contents programmatically without 
having to manually parse the JSON string.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,6 +11,8 @@ Ongoing development
 
 New Features
 ------------
+- :meth:`TableReport.dict` now allows exporting the report data as a Python 
+  -dictionary. :pr:`2188` by :user:`m4nn2609-dot <m4nn2609-dot>`.
 - New methods :meth:`SkrubLearner.get_named_params` and
   :meth:`SkrubLearner.set_named_params` allow getting and setting the outcomes for
   choices contained in the DataOp, keyed by choice name. It provides a more
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -11,8 +11,7 @@ Ongoing development
 
 New Features
 ------------
-- :meth:`TableReport.dict` now allows exporting the report data as a Python 
-  -dictionary. :pr:`2188` by :user:`m4nn2609-dot <m4nn2609-dot>`.
+- :meth:`TableReport.dict` now allows exporting the report data as a Python dictionary. :pr:`2188` by :user:`m4nn2609-dot <m4nn2609-dot>`.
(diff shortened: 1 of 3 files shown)


rcap107:
Hi @m4nn2609-dot, thanks for this PR. I saw you asked to work on this issue, but did not wait for a maintainer before opening a PR. For this time we'll let it slide, but if it happens again in the future we may close the PR without reviewing it. 

There are various reasons why we'd rather assign issues explicitly: there may already be an open PR dealing with the issue, the issue may not be detailed enough to work on, or we may have decided to not work on the issue after all. If you start working on the issue without asking, there is the risk that the PR gets closed anyway and you waste your time. 

It was not the case for this particular PR, but please keep it in mind for future contributions.

m4nn2609-dot:
Thanks for the heads up! I'll make sure to wait for explicit assignment before opening a PR next time. Appreciate the patience .Sorry for the unprofessional behavior from my side.

rcap107:
> Thanks for the heads up! I'll make sure to wait for explicit assignment before opening a PR next time. Appreciate the patience .Sorry for the unprofessional behavior from my side.

No worries! This kind of things happens all the time, but very often who opens the PR doesn't take the time to actually respond to our comments.

Regarding your PR, it's failing because it needs an entry in the changelog, the code was not formatted and it's missing a test for the new line.

For the test you can test that the output of the function is the same as the expected output (which basically means taking the code that generates the json in the first place and use that to prepare the expected)

You can find how to address the other two problems in the [contributing guide](https://skrub-data.org/stable/CONTRIBUTING.html)

m4nn2609-dot:
please review my pr
@rcap107

rcap107 (approved):
Looks good to me, thanks a lot @m4nn2609-dot

PR #2189, opened 2026-06-25
fix tablereport error in histogram when data range is very narrow

jeromedockes:
we can get an error in matplotlib / numpy histogram if the column is not constant but the data range does not contain enough representable floating numbers for the default number of bins

for example it would fail on a column like this

```
    low = np.float32(10.0)
    high = np.nextafter(low, 11.0)
    data = pd.Series([low, high])
```


skrub/_reporting/_plotting.py
```diff
@@ -193,7 +199,22 @@ def _get_range(values, frac=0.2, factor=3.0):
...
+def _get_safe_hist_range(values):
+    # make sure numpy can find bin edges between the range low and high bounds
+    if not len(values) or not (
+        np.issubdtype(values.dtype, np.floating)
+        or np.issubdtype(values.dtype, np.integer)
+    ):
+        return None
+    vmin, vmax = values.min(), values.max()
+    delta = max(np.spacing(vmin), np.spacing(vmax))
+    if vmax - vmin > 12 * delta:
```
rcap107, line 221: why 12 here and 6 below?

skrub/_reporting/_plotting.py
```diff
@@ -193,7 +199,22 @@ def _get_range(values, frac=0.2, factor=3.0):
     return low, high
 
 
+def _get_safe_hist_range(values):
```
rcap107, line 202: could you explain this function in a bit more detail? it's not clear what is happening in the code and what is being returned

skrub/_reporting/tests/test_plotting.py
```diff
@@ -28,3 +28,16 @@ def test_histogram():
     data = pd.Series([0.0])
     _, hist = _plotting.histogram(data)
     assert (hist["n_low_outliers"], hist["n_high_outliers"]) == (0, 0)
+
+    low = np.float32(10.0)
```
rcap107, line 37: can you add a comment here explaining what this line is for?

rcap107 (commented):
Thanks @jeromedockes 

The code looks mostly good to me, but I'd like a few more comments explaining what is happening before we merge this

skrub/_reporting/_plotting.py
jeromedockes: yep i added comments LMK if it is clearer now

skrub/_reporting/_plotting.py
jeromedockes: if the range is already wider than 12, we don't need to adjust it. otherwise we add 6 on each side so it is at least 12-wide afterwards.

probably 5 + 5 = 10 would be enough because there are 10 bins, but we don't care and have a little safety margin to make sure we don't crash

rcap107 (approved):
Thanks for the comments @jeromedockes, the PR looks good to me

PR #2194, opened 2026-06-26
support other kinds of predicate than name in find_fitted_estimator

jeromedockes:
support ids and callables in skrublearner.find_fitted_estimator, like truncated_after does


glemaitre (approved):
LGTM.

PR #2195, opened 2026-06-26
cache predict, predict_proba, etc. during SkrubLearner.score()

jeromedockes:
in skrublearner.skore if multiple scorers have been added with .skb.with_scoring several may call predict() on the same data. this PR caches the result. 

(it is a bit similar to what scikit-learn multimetricscorer, except multimetric cache only looks at the function name not the inputs (which is safe because it only applies caching for BaseScorers)

Moreover it is possible to pass `return_predictions=True` and score also returns the result of any response function like predict, predict_proba that has been called during scoring. indeed it is common to need both the scores and the predictions for further inspection like plotting

finally, for advanced cases where we happen to already have precomputed predictions it is possible to pass them in the environment under the special key _skrub_predictions . for example this is useful for skore which caches predictions separately

it is particularly relevant for models such as tabicl for which predict is expensive

```python
import skrub
from sklearn.datasets import make_regression


class VerboseRidge(Ridge):
    def predict(self, X):
        print("predict")
        return super().predict(X)


data_op = (
    skrub.X()
    .skb.apply(VerboseRidge(), y=skrub.y())
    .skb.with_scoring("r2")
    .skb.with_scoring("neg_mean_squared_error")
)

X, y = make_regression(random_state=0)
split = data_op.skb.train_test_split({"X": X, "y": y}, shuffle=False)
learner = data_op.skb.make_learner().fit(split["train"])
```
```
>>> learner.score(split["test"]) # predict only called once, not once for r2 and once for mse
predict
{'r2': 0.41038669368537584, 'neg_mean_squared_error': -8815.709381722563}

>>> learner.score(split['test'], return_predictions=True) # also get the predictions
predict
({'r2': 0.41038669368537584, 'neg_mean_squared_error': -8815.709381722563}, {'predict': array([ -62.7741653 ,   74.65355628,   12.91198187,  -79.56543396,
       -101.66458804, -149.8290312 ,   17.12525471,   89.08248582,
          7.17400477,  -78.93419072,  -72.9042891 ,  -78.00720039,
         70.52947871, -105.69623088,   71.93530794,  -27.72992668,
        139.66398704, -144.84887872, -171.30845333,   37.12898672,
         19.88109151, -102.65515896, -135.1107391 ,  -79.16918706,
       -118.83113416])})

# advanced use-case, when for some reason it is easier
# to compute the predictions ourselves than to get them with score(env, return_predictions=True):

>>> prediction = learner.predict(split['test'])
predict
>>> learner.score(split['test'] | {'_skrub_predictions': {'predict': prediction}}) # inject pre-computed predictions, predict() not called
{'r2': 0.41038669368537584, 'neg_mean_squared_error': -8815.709381722563}
```

Also, this PR adds support for passing to with_scoring a scorer that calls score() on the estimator. in this case this is interpreted as getting the default score that would be returned if with_scoring had not been used. this allows adding custom scorers while also retrieving the default score. Note as in scikit-learn cross_validate, using `None` as the scorer does just that.

```
>>> (
...     skrub.X()
...     .skb.apply(Ridge(), y=skrub.y())
...     .skb.with_scoring(None)
...     .skb.with_scoring("r2")
...     .skb.with_scoring(lambda e, X, y: e.score(X, y), name="default_score")
...     .skb.make_learner()
...     .fit(split["train"])
...     .score(split["test"])
... )
{'score': 0.41038669368537584, 'r2': 0.41038669368537584, 'default_score': 0.41038669368537584}
```


glemaitre (approved):
LGTM

skrub/_data_ops/_estimator.py
```diff
@@ -915,26 +918,58 @@ def _process_scores(scorer_info, scorer_output):
...
+        super().__init__(data_op, environment)
+        self.cache = cache
+
+    def _eval_in_mode(self, mode, X, y=None):
+        if y is not None:
+            # Only use caching for methods like predict, predict_proba etc.
+            # (They are the only ones to be called anyway unless a scorer does
+            # something very weird)
+            return super()._eval_in_mode(mode, X, y=y)
+        key = (mode, id(X))
```
glemaitre: Just a small remark here on how brittle it could be. Just waiting from the commit that we discussed IRL.

glemaitre (approved):
LGTM. I think it is a safer for the moment.

Issue #2192, opened 2026-06-25
FEAT - add `approx_n_unique` as selector


rcap107:
Polars has the [approx_n_unique](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.approx_n_unique.html#polars.approx_n_unique) method, which estimates the number of unique values in a column. This is much faster than the usual n_unique on large dataframes.

The `has_cardinality` selector relies on the `n_unique` function for both pandas and polars, and can be slow on larger dataframes.

We may want to have either a selector that uses explicitly approx_n_unique, or add the "approximate" flag to `has_cardinality`, to switch to the approximate function on demand.

The problem is that this function is only available for polars, so if the dataframe is a pandas dataframe the function should either fall back to the default n_unique, or raise an import error. If we go with the first option, I think it would be less disruptive if the change is done in has_cardinality, rather than having a separate function that sometimes doesn't work.

GaelVaroquaux:
Hi @rcap107

Thanks for the suggestion. Do you know if the approx method switches to an exact one when the data are small (and thus the cost/benefit of approx vs exact may vary)?

jeromedockes:
thanks @rcap107 . you mean the `cardinality_below` selector?  rather than approximating the number of unique values, the way to speed this up a lot would be to stop counting once the threshold has been reached, as it will typically be used with low thresholds (eg 30 to decide between encoders). in this case we don't need to know if n unique is approximately 10,000, or 11,000 if after looking at 100 values we already found 30 unique ones

also can you provide some timings on dataframes typical of those handled by skrub? i am seeing for example less than 2ms for n_unique on a column of 300,000 strings and 1000 unique values

all that being said, I also prefer adding a flag to the existing `cardinality_below` than introducing a new function

GaelVaroquaux:
> all that being said, I also prefer adding a flag to the existing cardinality_below than introducing a new function

I'm thinking in this direction too. Users often struggle finding the right functions to use

jeromedockes:
> the way to speed this up a lot would be to stop counting once the threshold has been reached

that should be relatively easy to do for both polars and pandas, by taking chunks of the series of an exponentially-growing size and counting n_unique for each. it should not be much slower than doing it at once for the whole column if we pick a common ratio large enough for the chunk size geometric sequence

PR #2196, opened 2026-06-29
DOC - selectors documentation rework

### LINKED ISSUE #2176: DOC - extend the documentation of the base Selector class with examples on composing selectors and general usage

I asked an agent to compose selectors to choose specific columns in a dataframe, and it immediately looked for the documentation of the base Selector class for information on this. The Selector docstring is very terse, and does not explain at all the basic features of the selectors. 

There should be short but general examples of how selectors in general work in the docstring of the base class.

rcap107:
This PR reworks and extends the documentation of the selectors to provide more context on when to use what, more cross-references, and additional examples.

This is in draft mode and needs to be trimmed down before merging. 

Not a blocker for the next release because we can push it in the doc branch if needed.


skrub/selectors/_base.py
```diff
@@ -79,7 +79,17 @@
 
 
 def all():
-    """Select all columns.
+    """Select all columns in a dataframe.
+
+    This is the most general selector, matching every column regardless of name,
+    type, or content. It is useful as a starting point for building complex
```
lisaleemcb:
```suggestion
    type, or content. It is a useful starting point for building complex
```

skrub/selectors/_base.py
```diff
@@ -272,48 +411,84 @@ def select(df, selector):
...
+    >>> s.select(df, s.all() & s.glob('*_mm'))
+       height_mm  width_mm
+    0      297.0     210.0
+    1      420.0     297.0
+
     """
     return _select_col_names(df, make_selector(selector).expand(df))
 
 
 def drop(df, selector):
```
lisaleemcb, line 403: Should this function be in the selectors section of the API documentation?

skrub/selectors/_base.py
```diff
@@ -272,48 +411,84 @@ def select(df, selector):
...
-    """Apply a selector to a dataframe and return a dataframe without the \
-    selected columns.
+    """Apply a selector to a dataframe and return it without the selected columns.
 
-    ``selector`` can be anything accepted by ``make_selector`` i.e. a selector,
-    column name or list of column names.
+    This is the complement of ``select()``: it returns a new dataframe containing
+    all columns except those matched by the selector. Useful for removing unwanted
```
lisaleemcb:
```suggestion
    all columns except those matched by the selector. 
```

skrub/selectors/_base.py
```diff
@@ -272,48 +411,84 @@ def select(df, selector):
...
+    columns (identifiers, metadata, noise) before modeling or analysis.
```
lisaleemcb:
```suggestion
```

skrub/selectors/_base.py
```diff
@@ -352,28 +549,122 @@ def drop(df, selector):
...
+    Selectors can be combined with operators to create complex selection rules:
+
+    - ``s.numeric() | s.boolean()`` - numeric OR boolean columns
+    - ``s.all() - s.glob('*_id')`` - all columns except ID-like ones
+    - ``s.string() & ~s.cardinality_below(10)`` - high-cardinality string columns
+
+    See Also
+    --------
+    skrub.ApplyToCols :
+        Apply a transformer to selected columns, use selectors for column selection.
```
lisaleemcb: Found this ref a little confusing

skrub/selectors/_base.py
```diff
@@ -352,28 +549,122 @@ def drop(df, selector):
...
+
+    >>> s.numeric().expand(df)
+    ['height_mm', 'width_mm', 'ID']
+
+    Combine selectors for complex rules:
+
+    >>> (s.numeric() & ~s.glob('*_ID')).expand(df)
+    ['height_mm', 'width_mm']
+
+    Use in data transformations: apply a scaler only to numeric columns:
```
lisaleemcb:
```suggestion
    Use in data transformations, for example to apply a scaler only to numeric columns:
```

skrub/selectors/_base.py
```diff
@@ -406,18 +698,27 @@ def expand(self, df):
...
         ----------
         df : dataframe
+            A pandas or polars dataframe to evaluate the selector against.
 
         Returns
         -------
-        list
-            The list of indices among ``df``'s columns that the item would select.
+        list of int
+            The indices (0-based positions) of columns from ``df`` that the
```
lisaleemcb:
```suggestion
            The indices (zero-indexed) of columns from ``df`` that the
```

skrub/selectors/_base.py
```diff
@@ -607,18 +911,70 @@ def __repr__(self):
...
 def filter(predicate, *args, **kwargs):
-    """Select columns for which ``predicate`` returns True.
+    """Select columns for which a custom predicate function returns ``True``.
+
+    This is the most flexible selector, allowing arbitrary column selection logic
+    based on column data (values, dtype, shape, etc.). The predicate receives the
+    actual column object (pandas/polars Series) for inspection.
+
+    This selector is useful in the following scenarios:
+    - When built-in selectors don't match your selection criterion
```
lisaleemcb: Formatting here, in the docs it's not rendering as a list

skrub/selectors/_base.py
```diff
@@ -607,18 +911,70 @@ def __repr__(self):
...
+    - To select columns based on column content properties (e.g., variance, range)
+    - To select columns by custom statistics or computed properties
+    - To combine multiple custom criteria in a single predicate
+
+    Parameters
+    ----------
+    predicate : callable
+        A function that takes a column and optional extra arguments, returning
+        ``True`` to select the column or ``False`` to exclude it.
+        Signature: ``predicate(col, *args, **kwargs) -> bool``
```
lisaleemcb, line 834: I think you want this to be rendered on a new line?

skrub/selectors/_selectors.py
```diff
@@ -157,15 +212,24 @@ def numeric():
     """
     Select columns that have a numeric data type.
 
+    Use this selector to find all numeric columns for scaling, normalization,
```
lisaleemcb:
```suggestion
    Use this selector to find all numeric columns.
```

skrub/selectors/_selectors.py
```diff
@@ -157,15 +212,24 @@ def numeric():
...
+    or statistical analysis. This includes both integer and floating-point types
```
lisaleemcb:
```suggestion
This includes both integer and floating-point types
```

skrub/selectors/_selectors.py
```diff
@@ -157,15 +212,24 @@ def numeric():
...
+    but excludes Boolean columns (which often need different handling).
 
-    This selects float and integer columns but not Boolean columns.
+    This selector matches both integer and floating-point columns, equivalent to
+    ``integer() | float()``.
+
+    Notes
```
lisaleemcb:
```suggestion
    Note
```

skrub/selectors/_selectors.py
```diff
@@ -201,11 +265,13 @@ def numeric():
     str_      ...
     dtype: object
 
+    Select all numeric columns (note: booleans are excluded):
```
lisaleemcb:
```suggestion
    Select all numeric columns:
```

skrub/selectors/_selectors.py
```diff
@@ -219,15 +285,25 @@ def integer():
     """
     Select columns that have an integer data type.
 
+    Use this selector when you specifically need integer-typed columns,
+    excluding floating-point and Boolean types. This is useful for selecting
+    discrete numeric features or ID-like columns.
 
-    This selects integer columns but not Boolean columns.
+    This selector selects only integer columns but not Boolean columns.
+    Note that ``integer() | float()`` is equivalent to :func:`numeric()`.
```
lisaleemcb: This is broken by a new line, might be useful to put on its own line

skrub/selectors/_selectors.py
```diff
@@ -262,11 +338,13 @@ def integer():
     str_      ...
     dtype: object
 
+    Select all integer columns (note: booleans are excluded):
```
lisaleemcb:
```suggestion
    Select all integer columns:
```

skrub/selectors/_selectors.py
```diff
@@ -280,12 +358,22 @@ def float():
     """
     Select columns that have a floating-point data type.
 
+    Use this selector when you specifically need floating-point-typed columns,
```
lisaleemcb:
```suggestion
    Use this selector when you specifically need columns of floating-point type,
```

skrub/selectors/_selectors.py
```diff
@@ -280,12 +358,22 @@ def float():
...
+    excluding integer and Boolean types. This is useful for selecting continuous
+    numeric features that have been measured or calculated as decimals.
+
+    This selector selects floating-point columns (float32, float64, etc.).
+    Note that ``integer() | float()`` is equivalent to :func:`numeric()`.
```
lisaleemcb: Again here, the expression here is split across lines

skrub/selectors/_selectors.py
```diff
@@ -336,13 +432,39 @@ def has_dtype(*dtypes):
...
     This selector takes a hands-off approach: skrub does not normalize or infer
     dtypes across dataframe libraries. A column is selected if
     ``sbd.dtype(column) == dtype`` for at least one of the provided ``dtypes``.
 
-    The most reliable way to use this selector is to get the dtype from an
-    existing column and pass that dtype object directly.
+    Parameters
+    ----------
+    *dtypes : dtype objects
+        One or more dtype objects to match. The most reliable approach is to get
```
lisaleemcb: Not sure I understand the second sentence

skrub/selectors/_selectors.py
```diff
@@ -381,6 +501,26 @@ def any_date():
     """
     Select columns that have a Date or Datetime data type.
 
+    Use this selector to find temporal columns that need date-specific
```
lisaleemcb:
```suggestion
    Use this selector to find temporal columns that require date-specific
```

skrub/selectors/_selectors.py
```diff
@@ -381,6 +501,26 @@ def any_date():
...
+    preprocessing, such as feature extraction (year, month, day) or
+    time-based aggregations.
+
+    This selector matches Datetime columns, including timezone-aware datetime
+    columns.
+
+    Notes
```
lisaleemcb, line 451: Should this be elevated to a warning or caution?

skrub/selectors/_selectors.py
```diff
@@ -381,6 +501,26 @@ def any_date():
...
+    -----
+    Behavior may differ between pandas and polars:
+    - Pandas: Selects datetime64 dtype columns
```
lisaleemcb: Not rendering as a list

skrub/selectors/_selectors.py
```diff
@@ -458,23 +647,31 @@ def string():
     """
     Select columns that have a String data type.
```
lisaleemcb:
```suggestion
    Select columns that have a string data type.
```

skrub/selectors/_selectors.py
```diff
@@ -458,23 +647,31 @@ def string():
...
 
+    Use this selector to find all text columns for encoding, NLP processing,
```
lisaleemcb:
```suggestion
    Use this selector to find all text columns.
```

skrub/selectors/_selectors.py
```diff
@@ -458,23 +647,31 @@ def string():
...
+    or text-based feature engineering. This includes both explicit string dtypes
```
lisaleemcb:
```suggestion
This includes both explicit string dtypes
```

skrub/selectors/_selectors.py
```diff
@@ -458,23 +647,31 @@ def string():
...
+    and object columns containing only strings.
 
     In pandas, object columns containing (only) strings are also selected.
```
lisaleemcb: Repeating information from previous line, or is this a separate point?

skrub/selectors/_selectors.py
```diff
@@ -458,23 +647,31 @@ def string():
...
 
     Notes
```
lisaleemcb, line 566: Again, should this be a caution since the behaviour can cause confusion?

skrub/selectors/_selectors.py
```diff
@@ -525,19 +713,30 @@ def object():
...
+    mixed types. For standard use cases, prefer more specific selectors like
+    :func:`string` or :func:`categorical`.
+
     Notes
     -----
-    Before pandas 3.0, columns containing only strings have the ``object``
-    dtype and are selected. From pandas 3.0 onwards they have the ``string``
-    dtype and are not. Use :func:`~skrub.selectors.string` for
-    selecting strings.
+    Before pandas 3.0, columns containing only strings can have the ``object``
```
lisaleemcb: If this is essentially the same point as in `string()`, I would homogenize the phrasing to make it easier to parse (i.e. put as a bullet point list).

skrub/selectors/_selectors.py
```diff
@@ -556,10 +755,19 @@ def object():
     string         ...
     dtype: object
 
+    Select object dtype columns (note: may contain mixed types):
```
lisaleemcb:
```suggestion
    Select object dtype columns (note: can contain mixed types):
```

skrub/selectors/_selectors.py
```diff
@@ -568,12 +776,30 @@ def boolean():
     """
     Select columns that have a Boolean data type.
 
+    **When to use:**
+    Use this selector to find Boolean columns that often need special handling
+    (e.g., encoding strategies different from numeric features). Boolean features
```
lisaleemcb:
```suggestion
    (e.g., encoding strategies different from numeric features).
```

skrub/selectors/_selectors.py
```diff
@@ -568,12 +776,30 @@ def boolean():
...
+    represent binary choices and may require different preprocessing than numeric
```
lisaleemcb:
```suggestion

```

skrub/selectors/_selectors.py
```diff
@@ -568,12 +776,30 @@ def boolean():
...
+    features.
```
lisaleemcb:
```suggestion

```

skrub/selectors/_selectors.py
```diff
@@ -591,17 +817,25 @@ def boolean():
...
-    Bool_    boolean
-    dtype: object
+
+    Select all Boolean columns:
 
     >>> s.select(df, s.boolean())
        bool_  Bool_
     0   True  False
 
+    Combine with numeric() to include both (note: boolean() is separate):
```
lisaleemcb:
```suggestion
    Combine with numeric() to include both:
```

skrub/selectors/_selectors.py
```diff
@@ -624,12 +858,29 @@ def cardinality_below(threshold):
...
+    ----------
+    threshold : int
+        Columns with fewer than this many unique values are selected.
+        Null values do not count in the cardinality.
+
+    Notes
+    -----
+    **Performance Consideration:** This selector requires computing the number
+    of unique values for each column. On large datasets (>1M rows), this may
+    be slow. Consider using on a sample if performance is critical.
```
lisaleemcb:
```suggestion
    be slow. Consider using on a subsample of data if performance is critical.
```

skrub/selectors/_selectors.py
```diff
@@ -624,12 +858,29 @@ def cardinality_below(threshold):
...
 
     See Also
     --------
     has_nulls :
```
lisaleemcb, line 779: We should specify that a null does not count as a unique value for the threshold

skrub/selectors/_selectors.py
```diff
@@ -689,27 +961,57 @@ def has_nulls(proportion=0.0):
     Select columns that contain at least one null value, or a proportion of null \
     values above a given threshold.
 
+    **When to use:**
+    Use this selector to identify columns needing imputation or to find columns
```
lisaleemcb:
```suggestion
    Use this selector to identify columns needing imputation or 
```

skrub/selectors/_selectors.py
```diff
@@ -689,27 +961,57 @@ def has_nulls(proportion=0.0):
...
+    with excessive missing data for exclusion. This is useful for data quality
```
lisaleemcb:
```suggestion
    with excessive missing data. This is useful for data quality
```

skrub/selectors/_base.py
rcap107: yes, it was missing but has been added now

skrub/selectors/_base.py
rcap107: What did you find confusing?

skrub/selectors/_base.py
rcap107: it doesn't look like that can be done in the parameters section

skrub/selectors/_selectors.py
rcap107: it renders as "Notes" with or without s

skrub/selectors/_selectors.py
rcap107: the point is that sometimes it's hard to figure out what is the dtype of a column without seeing it first, so the suggestion is to extract it explicitly 

the sentence probably should be removed from the parameter description here though

skrub/selectors/_selectors.py
rcap107: yeah maybe it's worth it

skrub/selectors/_selectors.py
rcap107: this is a separate point

skrub/selectors/_base.py
```diff
@@ -98,16 +103,38 @@ def all():
...
+       height_mm  width_mm kind
+    0      297.0     210.0   A4
+    1      420.0     297.0   A3
+
+    Combine with type selectors:
+
+    >>> s.select(df, s.all() & s.numeric())
+       height_mm  width_mm  ID
+    0      297.0     210.0   4
+    1      420.0     297.0   3
```
jeromedockes:
```suggestion
```

the all() has no effect here we can just do `s.select(df, s.numeric())`

skrub/selectors/_base.py
```diff
@@ -98,16 +103,38 @@ def all():
...
     """
     return All()
 
 
 def cols(*columns):
-    """Select columns by name.
+    """Select the columns whose names are explicitly listed.
```
jeromedockes: that is subjective but to me the old formulation was clearer

skrub/selectors/_base.py
```diff
@@ -248,12 +307,38 @@ def _select_col_names_polars(df, col_names):
 
 
 def select(df, selector):
-    """Apply a selector to a dataframe and return a dataframe with the selected \
-    columns.
+    """Apply a selector to a dataframe and return the selected columns.
+
+    It evaluates a selector against a specific dataframe and returns a new
+    dataframe containing only the columns matched by the selector.
```
jeromedockes:
```suggestion
    """Select the columns of a dataframe that are matched by the selector.

    This function returns a new dataframe containing only the columns of `df` matched by `selector`.
```

skrub/selectors/_base.py
```diff
@@ -248,12 +307,38 @@ def _select_col_names_polars(df, col_names):
...
+    df : dataframe
+        The dataframe to select columns from (pandas or polars).
+    selector : selector, str, or list
+        A selector object, single column name, or list of column names.
+
+    Returns
+    -------
+    dataframe
+        A new dataframe containing only the columns matched by the selector.
+        The column order matches the selector's expansion order.
```
jeromedockes:
```suggestion
```

i'm not sure that would help users figure out what will be the order in the output. for most selectors it will be the same order as in the input dataframe, except for `cols` it will be the order in which they are listed, and for user-defined selectors that override expand() it could be anything. we could either say that or just say nothing and make no guarantees about output column order wich is probably the simplest

skrub/selectors/_base.py
```diff
@@ -272,48 +357,73 @@ def select(df, selector):
...
+    1      420.0     297.0
+
     """
     return _select_col_names(df, make_selector(selector).expand(df))
 
 
 def drop(df, selector):
-    """Apply a selector to a dataframe and return a dataframe without the \
-    selected columns.
+    """Apply a selector to a dataframe and return it without the selected columns.
```
jeromedockes:
```suggestion
    """Select the columns of a dataframe that are NOT matched by the selector.
```

or "Drop the columns matched by the selector from a dataframe"

skrub/selectors/_base.py
```diff
@@ -332,18 +442,26 @@ def drop(df, selector):
...
     0   A5   5
     1   A4   4
 
-    We can also pass column names directly:
+    Drop specific columns by name (can pass names directly):
 
     >>> s.drop(df, ['height_mm', 'width_mm'])
       kind  ID
     0   A5   5
     1   A4   4
```
jeromedockes, line 465:
```suggestion
```

I think the double negation gets a bit weird: here s.select(df, s.string()) would be more natural. also with 2 examples it is probably enough to get the idea, we are not describing selectors in general just saying how drop works

skrub/selectors/_base.py
```diff
@@ -352,28 +480,117 @@ def drop(df, selector):
 
 
 class Selector:
-    """Generic selector type, that returns set columns when applied.
+    """Generic selector type that selects columns from a dataframe.
```
jeromedockes:
```suggestion
    """Pattern for matching columns in a dataframe.
```

skrub/selectors/_base.py
```diff
@@ -352,28 +480,117 @@ def drop(df, selector):
...
+    A ``Selector`` is a reusable rule for selecting columns based on various
+    criteria (data type, name pattern, content properties, etc.). Selectors
+    enable delayed selection: you can define a selection rule before the data
+    is available.
+
+    For each column in a dataframe, a selector evaluates the ``_matches(column)``
+    method:
+
+    - ``_matches`` returns ``True`` → column is selected
+    - ``_matches`` returns ``False`` → column is excluded
```
jeromedockes: this is for implementers not users

```suggestion
```

skrub/selectors/_base.py
```diff
@@ -352,28 +480,117 @@ def drop(df, selector):
...
+
+    **How to use selectors**
 
-    This class is not meant to be instantiated manually, ``Selector``
-    objects are created by calling one of the selector builders such
-    as :meth:`skrub.selectors.all()` or :meth:`skrub.selectors.make_selector()`.
+    1. **Direct selection:** ``s.select(df, selector)`` returns a filtered dataframe
```
jeromedockes: I think we can remove the numbering and bold formatting to have a simple bullet list

skrub/selectors/_base.py
```diff
@@ -352,28 +480,117 @@ def drop(df, selector):
...
+      applies a transformer to selected columns
+    3. **In DataOps:** ``skrub.X(df).skb.apply(transformer, cols=selector)``
+    4. **Manual expansion:** ``selector.expand(df)`` gets column names for manual use
+
+    **Combining Selectors**
+
+    Selectors can be combined with set operators to create complex selection rules:
+
+    - ``s.numeric() | s.boolean()`` - numeric OR boolean columns
+    - ``s.all() - s.glob('*_id')`` - all columns except ID-like ones
```
jeromedockes:
```suggestion
    - ``s.all() - s.glob('*_id')`` - all columns except those whose name ends with "_id"
```

skrub/selectors/_base.py
```diff
@@ -352,28 +480,117 @@ def drop(df, selector):
...
+    - ``s.string() & ~s.cardinality_below(10)`` - high-cardinality string columns
```
jeromedockes:
```suggestion
    - ``~s.cardinality_below(10)`` - high-cardinality columns
```

skrub/selectors/_base.py
```diff
@@ -352,28 +480,117 @@ def drop(df, selector):
...
+
+    Use a selector to get the names of matching columns:
+
+    >>> s.numeric().expand(df)
+    ['height_mm', 'width_mm', 'ID']
+
+    Selectors can be combined to create complex selection rules. For example,
+    to select numeric columns but exclude the 'ID' column:
+
+    >>> (s.numeric() & ~s.cols('ID')).expand(df)
```
jeromedockes:
```suggestion
    >>> (s.numeric() - 'ID').expand(df)
```

skrub/selectors/_base.py
```diff
@@ -352,28 +480,117 @@ def drop(df, selector):
...
+        -------
+        bool
+            ``True`` if the column should be selected, ``False`` otherwise.
+        """
         raise NotImplementedError()
 
     def expand(self, df):
-        """Lists the column names that the selector would retain if applied to \
-        the dataframe ``df``.
+        """Get the list of column names that the selector would select.
```
jeromedockes:
```suggestion
        """Get the names of columns matched by the selector in a list.
```

skrub/selectors/_base.py
```diff
@@ -406,18 +624,25 @@ def expand(self, df):
         return matching_col_names
 
     def expand_index(self, df):
-        """Lists the indices of dataframe ``df``'s columns that the selector \
-        would retain if applied to ``df``.
+        """Get the indices of columns that the selector would select.
```
jeromedockes:
```suggestion
        """Get the indices of columns matched by the selector in a list.
```

skrub/selectors/_base.py
```diff
@@ -607,18 +824,70 @@ def __repr__(self):
...
+    This is the most flexible selector, allowing arbitrary column selection logic
+    based on column data (values, dtype, shape, etc.). The predicate receives the
+    actual column object (pandas/polars Series) for inspection.
+
+    This selector is useful:
+
+    - When built-in selectors don't match your selection criterion
+    - To select columns based on column content properties (e.g., variance, range)
+    - To select columns by custom statistics or computed properties
+    - To combine multiple custom criteria in a single predicate
```
jeromedockes:
```suggestion
```

skrub/selectors/_base.py
```diff
@@ -607,18 +824,70 @@ def __repr__(self):
 
 
 def filter(predicate, *args, **kwargs):
-    """Select columns for which ``predicate`` returns True.
+    """Select columns for which a custom predicate function returns ``True``.
+
+    This is the most flexible selector, allowing arbitrary column selection logic
+    based on column data (values, dtype, shape, etc.). The predicate receives the
+    actual column object (pandas/polars Series) for inspection.
```
jeromedockes:
```suggestion
    This allows selecting columns based on an arbitrary criterion. The predicate takes as in put the column (a pandas or polars Series) and must return True if it should be selected and False otherwise.
```

skrub/selectors/_base.py
```diff
@@ -607,18 +824,70 @@ def __repr__(self):
...
+
+    Returns
+    -------
+    Selector
+        A ``Filter`` selector that matches columns where ``predicate`` returns ``True``.
+
+    See Also
+    --------
+    filter_names : Select columns based only on their name (not content)
+    Selector.expand : Evaluate a selector to get matching column names
```
jeromedockes:
```suggestion
```

skrub/selectors/_base.py
```diff
@@ -607,18 +824,70 @@ def __repr__(self):
...
+    instead of capturing them in a closure::
+
+        # Picklable (importable function + explicit args)
+        s.filter(str.startswith, 'prefix')
+
+        # Picklable alternative (with functools.partial)
+        import functools
+        s.filter(functools.partial(str.startswith, 'prefix'))
+
+        # NOT picklable (closure)
```
jeromedockes, line 867: here assuming this is at module level it is not a closure but a global variable and what prevents pickling is the lambda expression not the variable

skrub/selectors/_base.py
```diff
@@ -607,18 +824,70 @@ def __repr__(self):
...
+        prefix = 'test'
+        s.filter(lambda col: str(col.name).startswith(prefix))
+
+    For name-based selection, consider using :func:`filter_names`, :func:`glob`,
+    or :func:`regex` instead of ``filter`` for simpler selection.
+
+    **Performance**: The predicate is called once per column. Depending on the
+    size of the column and the specific operation, this may be expensive.
+    For large datasets, consider subsampling the data or using vectorized operations
+    if possible.
```
jeromedockes:
```suggestion
```

skrub/selectors/_base.py
```diff
@@ -655,6 +928,15 @@ def filter(predicate, *args, **kwargs):
     0   4
     1   3
 
+    Combine with type selectors:
+
+    >>> def has_high_range(col):
+    ...     return col.max() - col.min() > 100
+
+    >>> s.select(df, s.numeric() & s.filter(has_high_range))
```
jeromedockes, line 918: you can point out that due to short-circuit has_high_range will only get called on numeric columns here so there is no risk of it crashing on a string column

skrub/selectors/_base.py
```diff
@@ -669,20 +951,69 @@ def _default_name():
 
 
 def filter_names(predicate, *args, **kwargs):
-    """Select columns based on their name.
+    r"""Select columns based only on their name (not content or type).
```
jeromedockes:
```suggestion
    r"""Select columns based only on their name.

File: skrub/selectors/_base.py [hunk 58]
```diff
@@ -669,20 +951,69 @@ def _default_name():
...
+    (a string) to the predicate, rather than the column itself. Use this when
+    your selection logic depends only on naming patterns, not data values.
+
+    This selector is useful to:
 
+    - Select columns matching regex patterns (consider using ``glob()`` or ``regex()``
+      for simple patterns)
+    - Check column names for prefixes, suffixes, or substrings
+    - Filter by naming conventions (e.g., 'internal_*', 'tmp_*')
+    - Custom naming-based logic that doesn't involve column data
```
jeromedockes:
```suggestion
    The predicate takes as input the column name and must return True if it should be selected and False otherwise.
```

skrub/selectors/_base.py
```diff
@@ -669,20 +951,69 @@ def _default_name():
...
 
-    For a column whose name is ``col_name``, ``predicate`` is called as
-    ``predicate(col_name, *args, **kwargs)`` and the column is selected if
-    returns ``True``. Note this is different from ``filter``, because here the
-    predicate is passed the column name; with ``filter``, the predicate
-    is passed the actual column (pandas or polars Series).
+    Parameters
+    ----------
+    predicate : callable
+        A function that takes a column name (string) and optional extra arguments,
```
jeromedockes, line 949: here I would be consistent with the `filter` doc: either add the note about pickling in the entry for args and kwargs there or remove it here

skrub/selectors/_base.py
```diff
@@ -669,20 +951,69 @@ def _default_name():
...
+        Extra positional arguments passed to the predicate. Using explicit
+        arguments (instead of closures) helps with pickling the selector.
+    **kwargs : dict
+        Extra keyword arguments passed to the predicate. Using explicit
+        arguments (instead of closures) helps with pickling the selector.
 
+    Returns
+    -------
+    Selector
+        A ``NameFilter`` selector that matches columns where ``predicate(name)``
```
jeromedockes:
```suggestion
        A ``NameFilter`` selector that matches columns where ``predicate(name, *args, **kwargs)``
```

skrub/selectors/_base.py
```diff
@@ -669,20 +951,69 @@ def _default_name():
...
+        returns ``True``.
+
+    See Also
+    --------
+    filter : Select columns based on their content or properties
+    glob : Select columns by wildcard pattern matching
+    regex : Select columns by regular expression pattern matching
+    Selector.expand : Evaluate a selector to get matching column names
```
jeromedockes:
```suggestion
```

skrub/selectors/_base.py
```diff
@@ -669,20 +951,69 @@ def _default_name():
...
+    rather than lambdas or closures. Pass parameters via ``*args`` or ``**kwargs``::
+
+        # Picklable (importable function + explicit args)
+        s.filter_names(str.endswith, '_mm')
+
+        # Picklable alternative (with functools.partial)
+        import functools
+        s.filter_names(functools.partial(str.endswith, '_mm'))
+
+        # NOT picklable (closure)
```
jeromedockes, line 991: same comment as above

skrub/selectors/_base.py
```diff
@@ -701,14 +1032,15 @@ def filter_names(predicate, *args, **kwargs):
     0      297.0     210.0   A4   4
     1      420.0     297.0   A3   3
 
+    Select columns ending with a suffix (using lambda):
```
jeromedockes, line 1015:
```suggestion
    Prefer using *args and **kwargs to pass extra arguments to the predicate, rather than defining a dynamic function which may cause pickling errors.
    
    Select columns ending with a suffix (using lambda):
```

skrub/selectors/_selectors.py
```diff
@@ -78,16 +95,34 @@ def _regex(col_name, pattern, flags=0):
...
+        Select columns by name with Unix shell-style wildcard patterns.
+        Use this for simpler patterns.
+    filter_names :
+        Select columns based on custom name-based criteria.
+
+    Notes
+    -----
+    A column is selected if ``re.match(col_name, pattern, flags)`` returns a match.
+    Note that ``re.match`` only matches at the beginning of the string. Use '$' at
+    the end to require matching until the end of the column name.
```
jeromedockes: this is worded in a slightly confusing way but anyway it is just generic information about regular expressions so I think we can skip

(I realize it was already present in the existing docstring)


```suggestion
```

skrub/selectors/_selectors.py
```diff
@@ -78,16 +95,34 @@ def _regex(col_name, pattern, flags=0):
...
 
-    pattern can be a string pattern or a compiled regular expression, and flags
-    are regular expression flags as described in the ``re`` module
-    documentation:
-
-    https://docs.python.org/3/library/re.html#flags
+    Use this selector for complex name patterns that glob patterns cannot express.
+    This is useful for selecting columns with specific naming conventions or
+    patterns that glob patterns cannot express, so that regular expressions are
+    needed (e.g., columns matching '^feature_[0-9]+$').
```
jeromedockes:
```suggestion
    A column is selected if `re.match(pattern, column.name, flags) returns a match.
```

skrub/selectors/_selectors.py
```diff
@@ -78,16 +95,34 @@ def _regex(col_name, pattern, flags=0):
...
+    For simple wildcard patterns, prefer :func:`glob`.
```
jeromedockes:
```suggestion
    For simple wildcard patterns, consider :func:`glob`.
```

skrub/selectors/_selectors.py
```diff
@@ -101,18 +136,15 @@ def regex(pattern, flags=0):
...
+    Select columns matching a pattern:
 
     >>> s.select(df, s.regex('.*_mm'))
        height_mm  width_mm
     0      297.0     210.0
     1      420.0     297.0
 
-    A column is selected if ``re.match(col_name, pattern, flags)`` returns a
-    match. Note that it is enough to match at the beginning of the string:
+    Match at the beginning of the column name (no need for ^ prefix):
```
jeromedockes: i think we can remove this part, or we would need an example dataframe with columns where one is a prefix of the other to make the distinction clearer e.g "date", "date_of_birth" and we must use "date$" to not match date_of_birth

skrub/selectors/_selectors.py
```diff
@@ -126,23 +158,35 @@ def regex(pattern, flags=0):
     Columns: []
     Index: [0, 1]
 
-    Flags are passed to ``re.match``; the following are 3 equivalent ways of
-    setting re flags (re.IGNORECASE in this example):
+    Use regex flags for case-insensitive matching:
```
jeromedockes:
```suggestion
    using regex flags:
```

you can point to the doc

skrub/selectors/_selectors.py
```diff
@@ -219,15 +268,19 @@ def integer():
     """
     Select columns that have an integer data type.
 
+    Use this selector when you specifically need integer-typed columns,
+    excluding floating-point and Boolean types. This is useful for selecting
+    discrete numeric features or ID-like columns.
 
-    This selects integer columns but not Boolean columns.
+    Note that ``integer() | float()`` is equivalent to :func:`numeric()`.
```
jeromedockes:
```suggestion
    Boolean columns are not matched by this selector, only signed and unsigned ints.
```

skrub/selectors/_selectors.py
```diff
@@ -278,14 +333,19 @@ def integer():
 
 def float():
     """
-    Select columns that have a floating-point data type.
+    Select columns that have a floating-point data type (float32, float64, etc.)
+
+    Note that ``integer() | float()`` is equivalent to :func:`numeric()`.
```
jeromedockes:
```suggestion
```

skrub/selectors/_selectors.py
```diff
@@ -354,24 +447,40 @@ def has_dtype(*dtypes):
...
     >>> s.select(df, s.has_dtype(items_dtype, count_dtype))
            items  count
     0  [A4, A3]      2
     1      [A5]      1
 
-    Some dataframe libraries may also accept shorthand values that compare
-    equal to a dtype object. However, this is backend-specific, so the most
-    robust approach is still to pass the dtype obtained from the dataframe
-    library itself.
+    This also works with complex dtypes such as GeometryDtype in GeoPandas:
```
jeromedockes, line 426: I'm not sure we need to add this, it is a pretty simple concept "column is selected if column.dtype is in dtypes"

skrub/selectors/_selectors.py
```diff
@@ -381,6 +490,35 @@ def any_date():
     """
     Select columns that have a Date or Datetime data type.
 
+    Use this selector to select only date/datetime columns (including timezone-aware
+    datetime columns), for example to perform
+    time-based feature engineering, filtering, or analysis.
```
jeromedockes:
```suggestion
```

skrub/selectors/_selectors.py
```diff
@@ -381,6 +490,35 @@ def any_date():
...
+
+    Notes
+    -----
+
+    .. warning::
+
+      Behavior may differ between pandas and polars:
+
+      - Pandas: Selects datetime64 dtype columns
```
jeromedockes: I'm not sure what you mean. Date columns do not exist in pandas so how could they be selected

skrub/selectors/_selectors.py
```diff
@@ -381,6 +490,35 @@ def any_date():
...
+      - Polars: Selects both Date and Datetime dtypes
+
+    Only datetime columns are selected. Time-only, period, and duration types are
+    not selected.
+    String columns containing date-like values are not selected either, and
+    should be converted to datetime types first.
```
jeromedockes:
```suggestion
    Selection is based on the column's dtype: for example string columns containing date-like values are not selected.
```

i'm not sure this is necessary though, it just dilutes the simple description "select columns that have a Date or Datetime data type"

skrub/selectors/_selectors.py
```diff
@@ -419,10 +578,21 @@ def categorical():
     """
     Select columns that have a Categorical (or polars Enum) data type.
 
+    Use this selector to find columns with an explicitly defined
+    set of categories.
+
+    Note that categorical columns are different from string columns: categorical
+    columns have a fixed set of category values, while string columns contain
+    arbitrary text.
+
```
jeromedockes:
```suggestion
```

jeromedockes (commented):
_

skrub/selectors/_selectors.py
```diff
@@ -624,12 +840,34 @@ def cardinality_below(threshold):
...
+    Notes
+    -----
+    **Performance Consideration:** This selector requires computing the number
+    of unique values for each column. On large datasets (>1M rows), this may
+    be slow. Consider using on a subsample of data if performance is critical.
+
+    Missing values do not count as unique values for cardinality. For example,
+    a column with values `[1, 2, 2, None]` has a cardinality of 2.
+
```
jeromedockes:
```suggestion
```

skrub/selectors/_selectors.py
```diff
@@ -651,27 +889,36 @@ def cardinality_below(threshold):
...
+    >>> s.select(df, s.cardinality_below(10) & s.numeric())
+        a1    a2  a2_b    a3  a3_b  a4
+    0     1     1     1     1     1   1
+    1     1     1     1     2     2   2
+    2     1     2     2     3     3   3
+    3  <NA>  <NA>     2  <NA>     3   4
+
+    Note that numeric features are still treated as numeric even if they have low
+    cardinality (e.g., IDs).
```
jeromedockes:
```suggestion
```

not sure what the last sentence means in this context

skrub/selectors/_selectors.py
```diff
@@ -689,46 +936,104 @@ def has_nulls(proportion=0.0):
...
+        - 0.5: Selects columns with >50% missing values
+        - 1.0: Selects columns where all values are null
+
+    Notes
+    -----
+    Null values include NaN, None, NA, etc., depending on the dataframe library.
+    Behavior:
+
+    - pandas: Recognizes np.nan, None, pd.NA, pd.NaT
+    - polars: Recognizes null values
```
jeromedockes: and NaN are also treated as NULL

skrub/selectors/_base.py
rcap107: I'd rather keep this section, but maybe it could be moved to the notes part

skrub/selectors/_base.py
rcap107: same here, I'd like to keep this (at least in the notes)

skrub/selectors/_selectors.py
rcap107: I wanted to add an example with a library that's not pandas or polars to show that selectors are not limited to that

rcap107: unless the content was on a different line

skrub/selectors/_selectors.py
rcap107: that's the point, if someone is using pandas and polars on the same dataframe they'll get different results if they use any_date on the two versions 

not our fault there isn't a Date dtype in pandas

skrub/selectors/_selectors.py
rcap107: I think both parts are relevant

skrub/selectors/_base.py
```diff
@@ -352,28 +471,108 @@ def drop(df, selector):
 
 
 class Selector:
-    """Generic selector type, that returns set columns when applied.
+    """Pattern for matching columns in a dataframe.
+
+    A ``Selector`` is a reusable rule for selecting columns based on various
+    criteria (data type, name pattern, content properties, etc.). Selectors
+    enable delayed selection: you can define a selection rule before the data
+    is available.
```
jeromedockes, line 485: I think you need a new line if you want a paragraph break

skrub/selectors/_base.py
jeromedockes: ok maybe in this case change slightly the sentence above to something like " this selector can be useful for example: ..."

skrub/selectors/_selectors.py
jeromedockes: ok in this case we can keep it. I am just a bit hesitant about the sheer size of the documentation in general, here the selector does something very simple and pretty much all you need to know about it is contained in the first line of the docstring, and yet we end up with a full page in the reference docs. 

this is not specific to your change or even this particular docstring, and it is also partly due to the conventions imposed by numpydoc and the sphinx theme formatting, so maybe here is not the right place to discuss it. but I would really like us to come up with strategies to keep the documentation more concise, otherwise I am afraid users will refuse to read it

skrub/selectors/_selectors.py
jeromedockes: WDYM use pandas and polars on the same dataframe? it is either a pandas.DataFrame or a polars.DataFrame. 

in pandas dates are represented with datetime64 and also matched by this selector

actually I think listing explicitly the dtypes matched by this selector separately for pandas and polars is a good idea, I would just remove the "warning behaviour may differ" part

skrub/selectors/_selectors.py
jeromedockes: probably here it is fine btw, as it is in the ref doc and at the end of the page.

skrub/selectors/_selectors.py
jeromedockes: the second one is definitely relevant, it is already elsewhere in the docstring if I am not mistaken but it is important so we can keep it. for the first one,

-  I would have to see a realistic pipeline where counting unique values in a column is a bottleneck (related to #2192 btw)
- we would need to explain how to follow the suggestion (define your function that computes n unique quickly and use selectors.filter). but I don't think it's really worth it

skrub/selectors/_selectors.py
rcap107: I've had scuffed situations where for some reason I have to deal with both types at the same time (typically because I do my processing in polars but some other library really needs pandas). In more relatable situations, people may start working in pandas before switching with polars, and things would (mostly) work throughout in skrub except in this place where more columns would be selected than not 

I agree it's not a particularly grave danger, but it's still something that should be pointed out. Listing the types that are selected in pandas and polars should be enough

jeromedockes (approved):
very nice, thank you for the big overhaul @rcap107

PR #2202, opened 2026-07-02
allow filtering coords of the parallel coord plot

jeromedockes:
we can select which metrics and params to show in dataop parallel coordinate plots

if we have a huge number of scores it becomes unreadable

screenshot_2026-07-02T16:19:53+02:00

now we can do

```python
s.plot_results(show_times=[], show_scores=['neg_mape__12h__q_0.5', 'd2_pinball_score__12h__q_0.5'])
```

to select which to include

screenshot_2026-07-02T18:03:03+02:00

also fixes the fact that "std_" columns were only filtered out for "score" before

### DIFF
--- a/skrub/_data_ops/_estimator.py
+++ b/skrub/_data_ops/_estimator.py
@@ -1379,7 +1379,7 @@ class _BaseParamSearch(_DataOpWrapperMixin, SkrubBaseEstimator):
         min_score=None,
         show_scores=None,
         show_choices=None,
-        show_times=True,
+        show_times=None,
     ):
         """Create a parallel coordinate plot of the cross-validation results.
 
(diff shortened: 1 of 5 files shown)


skrub/_data_ops/_estimator.py
```diff
@@ -1303,6 +1303,20 @@ def detailed_results_(self):
         except NotFittedError:
             attribute_error(self, "results_")
 
+    def _get_score_names(self):
+        return [
+            k.removeprefix("mean_test_")
+            for k in self.cv_results_.keys()
+            if k.startswith("mean_test_")
```
rcap107, line 1310: is the if needed? isn't removeprefix a no op if the prefix isn't there anyway

skrub/_data_ops/tests/test_parallel_coord.py
```diff
@@ -56,7 +57,7 @@ def test_parallel_coord():
 
     pytest.importorskip("plotly")
 
-    fig = search.plot_results()
+    fig = search.plot_results(show_scores=["score"] if specify_show_scores else None)
```
rcap107: I don't see "score" defined anywhere, is that because the random search adds the "score" by itself? If there is more than one scoring method, is "score" still there or is it necessary to specify the metric?

Maybe this isn't related to this PR, but this line surprised me. I thought it should be `show_times=["score"]` instead.

skrub/_data_ops/_estimator.py
```diff
@@ -1371,29 +1397,91 @@ def plot_results(self, *, colorscale=DEFAULT_COLORSCALE, min_score=None):
             perform well, to make the plot less cluttered and to make better
             use of the colorscale's range.
 
+        show_scores : list of str, optional
+            List of score names to show
+            (e.g. "accuracy" in ``.skb.with_scoring(["roc_auc", "accuracy"])``).
+            By default all are shown.
+
+        show_choices : list of str, optional
+            List of choice names to show
```
rcap107, line 1406: maybe mention that only named choices can be selected?

rcap107 (commented):
Looks good to me, I left a few comments but it looks like everything is in order

Thanks @jeromedockes

skrub/_data_ops/tests/test_parallel_coord.py
jeromedockes: so this is kind of inherited from how scikit-learn cross_validate, gridearch etc work

https://scikit-learn.org/stable/modules/model_evaluation.html#the-scoring-parameter-defining-model-evaluation-rules

if you dont specify `scoring`, it uses the estimator's `score()` method and the result shows up as 'score' in the cv results

if you specify `scoring` (in skrub that can be done either by passing `scoring` to `.skb.cross_validate`, `.skb.make_randomized_search`, etc., or by using `.skb.with_scoring`), then that is used and the keys in cv_result will depend on the exact kind of scoring you used, for example if it is `scoring=['roc_auc', 'accuracy']` you will end up with `roc_auc` and `accuracy` keys

to summarize: `show_scores=['score']` is meaningless here, because 'score' is the only score anyway. I just added it to the test to make sure it doesn't crash in a case where `with_scoring` is not used. but I will reorganize the tests a bit then, and add a case with multiple scorers passed to `make_randomized_search` instead of `with_scoring`


show_times=["score"] is for showing the time it took to call score(), which is different from showing the value returned by score

skrub/_data_ops/_estimator.py
jeromedockes: the if does not control the application of removeprefix but th inclusion of k in the resulting list comprehension

if it starts with mean_test, remove the prefix and keep it
otherwise discard it

rcap107: oh right I misread it somehow, my bad

skrub/_data_ops/tests/test_parallel_coord.py
rcap107: yep, I thought it was something related to scikit-learn

thanks for reorganizing the tests, I think making this part more explicit would aid clarity

rcap107 (approved):
Looks good, thanks @jeromedockes

PR #2211, opened 2026-07-08
better error message when variables are missing from env

### LINKED ISSUE #2210: Better error message for no value after "make_learner"
The following code
```python
data_path = skrub.var("data_path", value=str(DATA_DIR / 'data.csv'))

data = data_path.skb.apply_func(pd.read_csv)

features = data.drop(columns=[*NON_FEATURE_COLS, GROUP_COL, TARGET_COL])
X = features.skb.mark_as_X(
    cv=splitter,
    split_kwargs={"groups": data[GROUP_COL]},
)
y = data[TARGET_COL].skb.mark_as_y()

predictions = X.skb.apply(skrub.tabular_pipeline(estimator), y=y)
learner = predictions.skb.make_learner()
report = skore.evaluate(learner, data={"data_dir": str(DATA_DIR)})
```

creates the following traceback:
```
UninitializedVariable: "No value has been provided for 'data_path'"
Evaluation of node <Var 'data_path'> failed. See above for full traceback. This node was defined here:
      File "/home/varoquau/tmp/Indays-July2026.mine/main.py", line 61, in evaluate_learner
        data_path = skrub.var("data_path", value=str(DATA_DIR / 'data.csv'))
```

The problem is that the "make_learner" call dropped the default value of `data_path`. But the error message is not helpful at all.

A solution would probably be to special-case it for "learners", to point out that "make_learner" drop the values.

jeromedockes:
closes #2210 

```python
import skrub

data_path = skrub.var("data_path", value="data")
learner = data_path.skb.make_learner()
learner.fit({"data_dir": "data"})
```

```
skrub._data_ops._data_ops.UninitializedVariable: No value has been provided for 'data_path'
Evaluation of node <Var 'data_path'> failed. See above for full traceback. This node was defined here:
      File "/home/jerome/workspace/scratch/2026/07/scratch.py", line 14, in <module>
        data_path = skrub.var("data_path", value="data")
- Note that preview values passed to initialize skrub variables
  are ignored by default when we pass an explicit 'environment' dictionary,
  for example when calling SkrubLearner.fit({'X': ..., 'y': ...}).
  Please pass a value for 'data_path' in the environment.
  You can also use skrub.var('data_path', value=..., becomes_default=True)
  to always retain the initialization value as a default.
- WARNING: the following keys were passed in the environment but have no corresponding variable in the DataOp:
  ['data_dir']
```

### DIFF
--- a/skrub/_data_ops/tests/test_errors.py
+++ b/skrub/_data_ops/tests/test_errors.py
@@ -662,6 +662,8 @@ def test_missing_var_message():
     )
     assert "skrub.var('a', value=..., becomes_default=True)" in full_msg
 
+    # Test case were we do use the preview values (no environment is passed)
+
     with pytest.raises(KeyError) as exc:
         (data_op + skrub.var("d")).skb.make_learner(fitted=True)
 
@@ -684,6 +686,8 @@ def test_missing_var_message_train_test_split():
     with pytest.raises(KeyError) as exc:
         X.skb.train_test_split()
     full_msg = "\n".join(traceback.format_exception(exc.value, exc.value, exc.tb))
+    print(full_msg)
+    assert "No value has been provided for 'a'" in full_msg
     assert (
         "ignored by default whenever we pass an explicit 'environment' dictionary"
         not in full_msg
--- a/skrub/_data_ops/tests/test_errors.py
+++ b/skrub/_data_ops/tests/test_errors.py
@@ -645,15 +645,46 @@ def test_mark_as_X_missing_cv():
 
 @pytest.mark.skipif(sys.version_info < (3, 11), reason="no add_note")
 def test_missing_var_message():
-    learner = (
+    data_op = (
         skrub.var("a", "a value")
         + skrub.var("var_b_name", "b value", becomes_default=True)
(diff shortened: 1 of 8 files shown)


GaelVaroquaux:
Thanks a lot!

Is it possible to detect whether a "make_learner" call has been called or not, to have a more precise error message?

skrub/_data_ops/tests/test_errors.py
```diff
@@ -640,3 +641,57 @@ def test_unhashable():
...
         skrub.var("a").skb.mark_as_X(split_kwargs={"groups": None})
+
+
+@pytest.mark.skipif(sys.version_info < (3, 11), reason="no add_note")
+def test_missing_var_message():
+    data_op = (
+        skrub.var("a", "a value")
+        + skrub.var("var_b_name", "b value", becomes_default=True)
+        + skrub.choose_from(["?", "!"], name="c")
+    )
```
GaelVaroquaux, line 652: This is such a fun application of DataOps. I think that we should have a example entitled "A 'formula' functionality for machine-learning", or something like this

GaelVaroquaux (approved):
LATM (looks awesome to me). Merging!

PR #2213, opened 2026-07-14
split returned by .skb.train_test_split: return last by default + allow choosing the split

jeromedockes:
Currently, when a splitter has been added with `.skb.mark_as_X(cv=the_splitter)`, if `.skb.train_test_split` is not passed an explicit `split_func` function, it defaults to the first split of the registered splitter.

With this PR,

- When no splitter has been added with mark_as_X or when we pass a split_func to .skb.train_test_split (ie when we are not using the mark_as_X splitter, either because there isn't one or we are overriding it), the behavior is unchanged (except the result contains new keys X and y, see below)
- When using the mark_as_X splitter, now by default the last split is returned rather than the first. it is usually what we want for timeseries splits (and for other splitters we typically don't care). Moreover it is possible to select a specific split by passing `split_index=3` to get the third split. Also row_indices_train and row_indices_test are added to the resulting dict, like they are in iter_cv_splits

Moreover, X and y (the full matrices before splitting) are now always added to the dicts returned by both train_test_split and iter_cv_splits . this can be useful for example to visualize the splits, makes the returned row indices more useful, and should usually come at no memory cost (X and y are already in memory anyway, for example iter_cv_splits keeps a reference to them until the generator is exhausted) -- and we can always select the subset of the dict we need and discard the rest

```python
import skrub
from sklearn import model_selection
from sklearn.dummy import DummyRegressor
import numpy as np

X = np.arange(30)[:, None]
y = X[:, 0]

pred = (
    skrub.var("X", X)
    .skb.mark_as_X(cv=model_selection.TimeSeriesSplit())
    .skb.apply(DummyRegressor(), y=skrub.var("y", y).skb.mark_as_y())
)
```

```
by default we now get the last split

>>> pred.skb.train_test_split()["y_test"]
array([25, 26, 27, 28, 29])

we can ask for a specific split. (split_index=0 gets the old behavior)

>>> pred.skb.train_test_split(split_index=0)["y_test"]
array([5, 6, 7, 8, 9])
>>> pred.skb.train_test_split(split_index=1)["y_test"]
array([10, 11, 12, 13, 14])

as before, we can still override the split function and pass the kwargs our
function expects.

>>> pred.skb.train_test_split(
...     split_func=model_selection.train_test_split,
...     test_size=12,
...     shuffle=True,
...     random_state=0,
... )["y_test"]
array([ 2, 28, 13, 10, 26, 24, 27, 11, 17, 22,  5, 16])
```

### DIFF
--- a/skrub/_data_ops/_estimator.py
+++ b/skrub/_data_ops/_estimator.py
@@ -1,5 +1,6 @@
 # Scikit-learn-ish interface to the skrub DataOps
 import copy
+import itertools
 import numbers
 from functools import partial
 
@@ -1027,15 +1028,27 @@ class _Splitter:
         del groups
         return self.splitter.get_n_splits(X, y, **self.split_kwargs)
 
-    def __call__(self, X, y=None):
-        # Can be used as a callable to get a single train/test split (the first
-        # CV split)
-        train, test = next(iter(self.split(X, y)))
-        X_train, X_test = _safe_indexing(X, train), _safe_indexing(X, test)
+    def get_single_split(self, X, y=None, split_index=-1):
+        if not isinstance(split_index, numbers.Integral):
+            raise TypeError(f"split_index should be an int, got: {split_index!r}")
+        raw_split_index = split_index
+        n_splits = self.get_n_splits(X, y)
+        if split_index < 0:
+            split_index = n_splits + split_index
+        if not 0 <= split_index < n_splits:
+            raise IndexError(
+                f"split_index {raw_split_index} out of range for splitter "
+                f"{self.splitter} with {n_splits} splits."
+            )
+        train, test = next(itertools.islice(self.split(X, y), split_index, None))
+        result = {"row_indices_train": train, "row_indices_test": test}
+        result["X_train"] = _safe_indexing(X, train)
+        result["X_test"] = _safe_indexing(X, test)
         if y is None:
-            return X_train, X_test
-        y_train, y_test = _safe_indexing(y, train), _safe_indexing(y, test)
-        return X_train, X_test, y_train, y_test
+            return result
+        result["y_train"] = _safe_indexing(y, train)
+        result["y_test"] = _safe_indexing(y, test)
+        return result
 
 
 def _compute_cv_data(data_op, environment, cv):
@@ -1049,24 +1062,6 @@ def _compute_cv_data(data_op, environment, cv):
             _Splitter(check_cv(data["cv"]), data["split_kwargs"]),
         )
     return data["X"], data["y"], None
-
-
-def _compute_train_test_split_data(data_op, environment, split_func, split_func_kwargs):
-    data = _compute_X_y_and_cv(data_op, environment)
-    if split_func is not None:
-        return data["X"], data["y"], split_func, split_func_kwargs
-    if "cv" in data:
-        return (
-            data["X"],
-            data["y"],
-            _Splitter(check_cv(data["cv"]), data["split_kwargs"]),
-            {},
-        )
-    # kwargs fall through to the default split_func when no splitter is set and
-    # no split_func is passed. This allows something like
-    # data_op.skb.train_test_split(shuffle=False), without having to import &
-    # pass sklearn.model_selection.train_test_split explicitly.
-    return data["X"], data["y"], model_selection.train_test_split, split_func_kwargs
 
 
 def _rename_cv_param_learner_to_estimator(kwargs):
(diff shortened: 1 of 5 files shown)


jeromedockes:
In fact I think `.skb.train_test_split` should be deprecated and replaced by `.skb.get_cv_split(environment=None, cv=None, split_index=-1)` i.e. instead of having a train_test_split that is completely different from the CV splitters like in scikit-learn, just have an easy way to get a single split out of a (any) splitter

however this pr improves the existing function without breaking backward compatibility, and the possible deprecation of train_test_split can be discussed separately

skrub/_data_ops/_estimator.py
```diff
@@ -1027,15 +1027,18 @@ def get_n_splits(self, X, y=None, groups=None):
...
+            raise TypeError(f"split_index should be an int, got: {split_index!r}")
+        n_splits = self.get_n_splits(X, y)
+        if not -n_splits <= split_index < n_splits:
+            raise IndexError(
+                f"split_index {split_index} out of range for splitter "
+                f"{self.splitter} with {n_splits} splits."
+            )
+        if split_index < 0:
+            split_index = n_splits + split_index
+        return next(itertools.islice(self.split(X, y), split_index, None))
```
rcap107, line 1045: So here `self.split` returns `n_splits` splits (train, test), and islice is selecting all the splits between `split_index` and the end, and finally `next(...)` is returning only the first one, which contains a single pair (train, test) where train, test are the indices of the specific samples

Is that correct?

Could you add a comment about this here?

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2009,11 +2009,24 @@ def train_test_split(
...
+            The function used to split X and y once they have been computed. If
+            ``split_func`` is not provided,
+
+            - If a cross-validation splitter has been set with
+              :meth:`~DataOp.skb.mark_as_X`, that splitter is used.
+              In this case the only accepted kwarg in ``split_func_kwargs`` is
+              ``'split_index'``, which indicates which split to use. By default
+              ``'split_index'`` is -1, i.e. the last split is used.
+            - Otherwise :func:`sklearn.model_selection.train_test_split` is
+              used (and ``split_func_kwargs`` are forwarded to it).
```
rcap107, line 2028: This part throws me off a bit. I get why it would be the final split when working with timeseries, but in general I thought it'd be more intuitive to return the first split rather than the last. Though, by default train_test_split shuffles so there is no actual actual meaning in the "first" split, so it shouldn't matter...

Maybe add a mention of the timeseries split here to make it obvious why the last split was chosen? It makes it read as less arbitrary

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2036,6 +2055,13 @@ def train_test_split(
               the test environment, if there is one (may not be the case for
               unsupervised learning).
 
+            If relying on the splitter passed to :meth:`~DataOp.skb.mark_as_X`,
+            the following keys are added:
+
+            - row_indices_train: the row indices (in X and y) of the training samples.
```
rcap107, line 2068: Is there a reason why these keys are added only when using the custom splitter rather than in all cases?

skrub/_data_ops/_estimator.py
```diff
@@ -1199,27 +1210,40 @@ def train_test_split(
...
-    X, y, split_func, split_func_kwargs = _compute_train_test_split_data(
-        data_op, environment, split_func, split_func_kwargs
-    )
+    data = _compute_X_y_and_cv(data_op, environment)
+    X, y = data["X"], data["y"]
+    if split_func is None and "cv" in data:
+        extra_kwargs = set(split_func_kwargs).difference({"split_index"})
+        if extra_kwargs:
+            raise ValueError(
+                "When using the splitter passed to .skb.mark_as_X(),\n"
```
rcap107, line 1223: I'd like to have here a mention of the fact that the kwargs should be passed directly to the splitter passed to mark_as_X, but the error message is already long enough

This would be a good place to add a link to the error page on data ops if we ever get around working on that.

rcap107 (commented):
Overall, it looks good to me. I have a couple of comments asking for clarification and I'm not entirely sure about choosing the last split (though that's not really asking for changes in the code). 

I also think that train_test_split is not a good name for the skrub function, but we can definitely postpone renaming it to a later PR.

skrub/_data_ops/_estimator.py
jeromedockes: yes that is correct. the goal is to get the nth item without materializing the whole sequence in a list. i'll add a comment

skrub/_data_ops/_estimator.py
jeromedockes: I agree

skrub/_data_ops/_skrub_namespace.py
jeromedockes: my thinking was that the only frequently-used splitting (at least the only scikit-learn one) where splits really have an order is the time series one, for others the ordering is arbitrary. note that this kind of splitting (use data already available at this time point to predict data that will come later) wider than strict forecasting setting where there is a strong dependency on time. for timeseries split, the first can have a small training set, and does not use most of the data which lies beyond the end of the first test set, so taking the last split is typically more interesting. moreover, it is much better to take the last one if we further subdivide the train part e.g. for hyperparameter tuning. so all in all I think last is a better default, and the split_index parameter lets you pick another one. I'll add a sentence to mention timeseries splits as you suggest

skrub/_data_ops/_skrub_namespace.py
jeromedockes: yes, scikit-learn train_test_split only returns X_train, X_test, it does not return the indices so we have no way of getting them

jeromedockes (commented):

skrub/_data_ops/_skrub_namespace.py
rcap107: yes it makes sense, thanks for the explanation

skrub/_data_ops/_skrub_namespace.py
```diff
@@ -2009,11 +2009,31 @@ def train_test_split(
...
+              :meth:`~DataOp.skb.mark_as_X`, that splitter is used.
+              In this case the only accepted kwarg in ``split_func_kwargs`` is
+              ``'split_index'``, which indicates which split to use. By default
+              ``'split_index'`` is -1, i.e. the last split is used.
+
+              Note: the default is the last one because for time series
+              splitter it typically is the one that uses the most data, which
+              is particularly important if the train part will be further
+              subdivided e.g. for hyperparameter search. (For other splitters
+              split ordering is usually arbitrary.)
```
```suggestion
              Note: the default split is the last one because when using a
              time series splitter it typically is the one that uses the most data; 
              this is particularly important if the train part will be further
              subdivided e.g. for hyperparameter search. (For other splitters
              split ordering is usually arbitrary.)
```

rcap107: small edit for clarity, this will break the formatting

rcap107 (approved):

rcap107 (approved):
Aside from that suggestion, looks good to me. Thanks a lot @jeromedockes

PR #2214, opened 2026-07-15
pixi allow using arrow channel for scikit-learn if missing from anaco…

jeromedockes:
fix the CI

for some reason scikit-learn is currently missing from anaconda nightlies channel. this allows to fetch it from subsequent channels


rcap107 (approved):
We need to remember to do the TODO and remove the index strategy once the wheels appear, but as long as the CI is running we can merge this

PR #2216, opened 2026-07-15
Allow DropSimilar with Polars without PyArrow

### LINKED ISSUE #2187: DropSimilar should not compute Pearson correlation
DropSimilar requires pyarrow to be installed when the input is a polars dataframe, because compute_associations uses pandas to compute associations. However, DropSimilar does not use the Pearson correlations (which is what actually requires pyarrow). `compute_associations` should have a parameter to compute either Cramer V, Pearson or both (both by default), and DropSimilar should skip Pearson, since it's not used. 

We may also want to use that option in the tablereport and compute only cramerv with a warning when we have a polars dataframe and no pyarrow.

ShreyanshGoyal:
Fixes #2187.

## Summary

- add a backward-compatible option to skip Pearson correlation in column_associations
- have DropSimilar explicitly skip its unused Pearson correlation computation
- preserve the existing exception when Pearson correlation is requested for Polars without PyArrow
- allow DropSimilar to process Polars dataframes when PyArrow is not installed
- add regression coverage and document the user-facing change in CHANGES.rst

## Rationale

DropSimilar discarded the pearson_corr column immediately after calling column_associations, but computing that unused column converted Polars dataframes to pandas and made PyArrow mandatory. DropSimilar now requests only the Cramers V association it uses. Direct column_associations calls retain their requested behavior and raise if Pearson correlation cannot be computed.

## Test plan

- [x] Run the affected test files (25 passed, 2 skipped)
- [x] Confirm DropSimilar works with Polars while PyArrow imports are blocked
- [x] Confirm column_associations raises when Pearson is requested while PyArrow imports are blocked
- [x] Check the patch for whitespace errors

### DIFF
--- a/skrub/_column_associations.py
+++ b/skrub/_column_associations.py
@@ -13,18 +13,23 @@ _N_BINS = 10
 _CATEGORICAL_THRESHOLD = 30
 
 
-def column_associations(df):
+def column_associations(df, *, compute_pearson=True):
     """Get measures of statistical associations between all pairs of columns.
 
-    Reported metrics include Cramer's V statistic and Pearson's Correlation
-    Coefficient. The result is returned as a dataframe that contains the column
-    name and idx for the left and right table and both associations; results
-    are sorted in descending order by Cramer's V association.
+    Reported metrics include Cramer's V statistic and, unless disabled,
+    Pearson's Correlation Coefficient. The result is returned as a dataframe
+    that contains the column name and idx for the left and right table and the
+    requested associations; results are sorted in descending order by Cramer's
+    V association.
 
     Parameters
     ----------
     df : dataframe
         The dataframe whose columns will be compared to each other.
+    compute_pearson : bool, default=True
+        Whether to compute Pearson's Correlation Coefficient. Setting this to
+        ``False`` avoids converting Polars dataframes to pandas and therefore
+        does not require PyArrow.
 
     Returns
     -------
@@ -36,7 +41,8 @@ def column_associations(df):
     The result is returned as a dataframe with columns:
 
     ``['left_column_name', 'left_column_idx', 'right_column_name',
-    'right_column_idx', 'cramer_v', 'pearson_corr']``
+    'right_column_idx', 'cramer_v', 'pearson_corr']``. When
+    ``compute_pearson=False``, the ``'pearson_corr'`` column is omitted.
 
     As the function is commutative, each pair of columns appears only once
     (either ``col_1``, ``col_2`` or ``col_2``, ``col_1`` but not both).
@@ -108,6 +114,8 @@ def column_associations(df):
         _cramer_v_matrix(df),
         df,
     )
+    if not compute_pearson:
+        return cramer_v_table
     pearson_c_table = _compute_pearson(df)
     on = ["left_column_name", "right_column_name"]
     return _join_utils.left_join(
--- a/skrub/_column_associations.py
+++ b/skrub/_column_associations.py
@@ -16,11 +16,10 @@ _CATEGORICAL_THRESHOLD = 30
 def column_associations(df, *, compute_pearson=True):
     """Get measures of statistical associations between all pairs of columns.
 
-    Reported metrics include Cramer's V statistic and, unless disabled,
-    Pearson's Correlation Coefficient. The result is returned as a dataframe
-    that contains the column name and idx for the left and right table and the
-    requested associations; results are sorted in descending order by Cramer's
-    V association.
+    Reported metrics include Cramer's V statistic and Pearson's Correlation
+    Coefficient. The result is returned as a dataframe that contains the column
+    name and idx for the left and right table and the requested associations;
+    results are sorted in descending order by Cramer's V association.
 
     Parameters
     ----------
@@ -29,7 +28,8 @@ def column_associations(df, *, compute_pearson=True):
     compute_pearson : bool, default=True
         Whether to compute Pearson's Correlation Coefficient. Setting this to
         ``False`` avoids converting Polars dataframes to pandas and therefore
-        does not require PyArrow.
+        does not require PyArrow. This parameter is set to ``False`` when the
+        dataframe is Polars and PyArrow is not available.
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


skrub/_column_associations.py
```diff
@@ -36,7 +41,8 @@ def column_associations(df):
     The result is returned as a dataframe with columns:
 
     ``['left_column_name', 'left_column_idx', 'right_column_name',
-    'right_column_idx', 'cramer_v', 'pearson_corr']``
+    'right_column_idx', 'cramer_v', 'pearson_corr']``. When
+    ``compute_pearson=False``, the ``'pearson_corr'`` column is omitted.
```
rcap107:
```suggestion
    'right_column_idx', 'cramer_v', 'pearson_corr']``. 
    
    When ``compute_pearson=False``, the ``'pearson_corr'`` column is omitted.
```

skrub/_column_associations.py
```diff
@@ -13,18 +13,23 @@
...
 
-def column_associations(df):
+def column_associations(df, *, compute_pearson=True):
     """Get measures of statistical associations between all pairs of columns.
 
-    Reported metrics include Cramer's V statistic and Pearson's Correlation
-    Coefficient. The result is returned as a dataframe that contains the column
-    name and idx for the left and right table and both associations; results
-    are sorted in descending order by Cramer's V association.
+    Reported metrics include Cramer's V statistic and, unless disabled,
```
rcap107:
```suggestion
    Reported metrics include Cramer's V statistic and
```

the fact that Pearson's correlation can be disabled is a detail so it should be mentioned below

skrub/_column_associations.py
```diff
@@ -13,18 +13,23 @@
...
+    V association.
 
     Parameters
     ----------
     df : dataframe
         The dataframe whose columns will be compared to each other.
+    compute_pearson : bool, default=True
+        Whether to compute Pearson's Correlation Coefficient. Setting this to
+        ``False`` avoids converting Polars dataframes to pandas and therefore
+        does not require PyArrow.
```
rcap107: I'm thinking that this should be set to "False" by default if pyarrow is not installed and the given dataframe is polars. 

Then, the docstring should mention that ("this parameter is set to false by default if the dataframe is polars and pyarrow isn't available")

skrub/_column_associations.py
```diff
@@ -108,6 +114,8 @@ def column_associations(df):
         _cramer_v_matrix(df),
         df,
     )
+    if not compute_pearson:
```
rcap107, line 115: as mentioned in the other comment, this should also check whether pyarrow is available, and set `compute_parson=False` if it isn't

rcap107: there should also be the relative test

rcap107:
Hi @ShreyanshGoyal, as a general comment before working on a PR you should ask in the relevant issue first. This is to make sure that there aren't already people working on the issue, and to get more detail from the maintainers on what the PR should do. 

In this case it wasn't a big issue, but please keep it in mind for future contributions.

rcap107 (commented):
Hi @ShreyanshGoyal, thanks for working on this. 

The PR is doing most of what we want, but I think it would be better to change the default behavior so that, if pyarrow isn't available, the parameter is changed to False in any case. This would let polars dataframes go through without failing.

So, the code and the tests should be updated accordingly.

Also, there should be an entry in the changelog since this is a user-facing change.

Thanks again!

ShreyanshGoyal:
Thanks for the review. I’ve updated the default path so `column_associations` automatically disables Pearson only for Polars when PyArrow is unavailable, switched `DropSimilar` back to that default path, added the direct regression test, incorporated the docstring suggestions, and added the changelog entry. The two affected test files now pass all 39 tests, and Ruff checks are clean.

skrub/_drop_similar.py
```diff
@@ -139,15 +139,6 @@ def fit_transform(self, X, y=None):
...
-        if sbd.is_polars(X):
-            try:
-                import pyarrow  # noqa F401
-            except ImportError:
-                raise ImportError(
-                    "DropSimilar requires the Pyarrow package to run on Polars"
-                    " dataframes."
-                )
-
         association_df = column_associations(X)
```
jeromedockes: here we should use compute_pearson=False because we never need it

skrub/_column_associations.py
jeromedockes: please see my other comment, if the function is not able to compute what was requested it should signal that with an exception.

what we can do (in another pr, not here) is to automatically choose compute_pearosn in the tablereport, but tthere would need to be a message in the report explaining why the correlations are not there

skrub/_drop_similar.py
ShreyanshGoyal: Updated in 8fe32961. DropSimilar now passes compute_pearson=False explicitly because it never uses the Pearson result.

skrub/_column_associations.py
ShreyanshGoyal: Updated in 8fe32961. I removed the automatic fallback. When Pearson is requested without PyArrow, column_associations now retains its exception behavior, with a direct regression test covering it.

skrub/_column_associations.py
```diff
@@ -13,18 +13,22 @@
...
+    results are sorted in descending order by Cramer's V association.
 
     Parameters
     ----------
     df : dataframe
         The dataframe whose columns will be compared to each other.
+    compute_pearson : bool, default=True
+        Whether to compute Pearson's Correlation Coefficient. Setting this to
+        ``False`` avoids converting Polars dataframes to pandas and therefore
+        does not require PyArrow.
```
jeromedockes:
```suggestion
    compute_pearson : bool, default=True
        Whether to compute Pearson's Correlation Coefficient. Currently, computing Pearson correlations for polars DataFrames requires PyArrow to be installed and internally converts the dataframe to pandas.
```

jeromedockes (approved):
looks good with a slight rewording of the docstring entry, thank you very much!

rcap107:
Thanks a lot for the PR @ShreyanshGoyal

PR #2218, opened 2026-07-16
MAINT - add a command for testing all docstrings

rcap107:
Adding a command that tests all docstrings without running tests.

This should make it easier to debug and fix doctests in the documentation and the docstrings when they're worked on: checks can be run locally and skip running the actual tests.


jeromedockes (approved):
very useful thanks!

PR #2224, opened 2026-07-18
Remove confusing contributor note from SingleColumnTransformer.fit docstring

### LINKED ISSUE #2128: docs: Remove puzzling sentence in fit description of encoders inheriting from Single_column_transformer

Image

The sentence "Subclasses should implement fit_transform and transform." is irrelevant at best, and surprising when reading the documentation from a user point of view.

We should either remove it completely, or put it as a code comment to keep it for contributors, but remove it from the front documentation for users. It's already said in the documentation of the class, so in my point of view we can remove this sentence.

[link to the code](https://github.com/skrub-data/skrub/blob/0e3c5f4/skrub/_single_column_transformer.py#L170)

jaysparkx:
Closes #2128.

The sentence `Subclasses should implement fit_transform and transform.` in `SingleColumnTransformer.fit` is directed at contributors, but appears in the user-facing documentation of `fit`, where it is confusing (as noted in the issue). The same information is already documented at the class level, so this removes it from the method docstring.

No behavior change.

### DIFF
--- a/skrub/_single_column_transformer.py
+++ b/skrub/_single_column_transformer.py
@@ -173,8 +173,6 @@ class SingleColumnTransformer(SkrubBaseEstimator):
         This default implementation simply calls ``fit_transform()`` and
         returns ``self``.
 
-        Subclasses should implement ``fit_transform`` and ``transform``.
-
         Parameters
         ----------
         column : a pandas or polars Series


jaysparkx:
Closing this — withdrawing the change for now. Apologies for the noise, and thanks for maintaining the project.

Issue #2114, opened 2026-06-02
Accept estimators that don't inherit from `sklearn.BaseEstimator` in `tabular_pipeline`


betatim:
### Problem Description

I would like to use `tabular_pipeline()` with a classifier from cuml.

```python
from skrub import tabular_pipeline
from cuml.ensemble import RandomForestRegressor

pipeline = tabular_pipeline(RandomForestRegressor())
# TypeError: tabular_pipeline expects a scikit-learn estimator,
# 'regressor', or 'classifier' as its first argument.
```

I'm not sure why the [tabular pipeline checks if the estimator inherits from `BaseEstimator`](https://github.com/skrub-data/skrub/blob/9eb9d69643e15ee90188825e0cb2579e7e08479f/skrub/_tabular_pipeline.py#L249). Maybe the reason is that there is no other good way to check if something is a scikit-learn compatible estimator? In Python we often do duck typing, but I think that means we don't find out till later if the thing the user passed is the right kind of duck (a scikit-learn compatible estimator).

### Feature Description

Make it possible to use `tabular_pipeline` with estimators that are scikit-learn compatible but not from scikit-learn. I've tried xgboost and lightgbm, they both pass the `isinstance` check because they inherit from `BaseEstimator`.

But they don't get the special treatment in

https://github.com/skrub-data/skrub/blob/9eb9d69643e15ee90188825e0cb2579e7e08479f/skrub/_tabular_pipeline.py#L255-L272

So maybe a good idea for how to handle both the `isinstance` check and the "is this a tree model?" is what I am looking for?

### Alternative Solutions

You can work around this with the following:

```python

pipeline = make_pipeline(
    TableVectorizer(),
    RandomForestRegressor(n_estimators=50, random_state=0),
)
```

But you miss out on the special treatment for tree based models/you have to know to do it yourself

jeromedockes:
thank you, @betatim !

about the isinstance check: there is no compelling reason, really. we can relax it a bit and check that the input is not a type and has a couple of things like `get_params` and `set_params`

about the special treatment for tree-based models, for the moment it is a simple hard-coded list of scikit-learn estimators that we know to be based on trees (and actually it is missing a few). I am not sure how we could decide more generally if the input model needs scaling, one-hot-encoding of categories and spline-encoding of datetimes, if you have ideas they are welcome. maybe we could have a parameter to override the current heuristic

betatim:
[...]

PR #2225, opened 2026-07-19
Fix Issue 2114: Accept estimators that don't inherit from sklearn.BaseEstimator in tabular_pipeline

ljdyer:
# Fix Issue 2114: Accept estimators that don't inherit from sklearn.BaseEstimator in tabular_pipeline

## Description

All suggestions in the issue comments made sense, so I implemented them as specified. There were three main changes:

1. Loosened the requirement for estimator to inherit from sklearn.BaseEstimator

- Per the suggestion in the issue comments, the new implementation is to require a 'scikit-learn compatible' estimator, which is defined as an object that has get_params and set_params attributes
- Check made explicit with `is_scikit_learn_compatible` variable
- `isinstance(estimator, type)` check carried out before `is_scikit_learn_compatible` check since the validity check no longer makes sense for types. Class name no longer included in error message, since it may be displayed even for invalid types. Still serves its purpose of helping the user to spot the mistake.
- Tested (in `test_sklearn_incompatible_learner_fails` and `test_sklearn_incompatible_learner_succeeds`) with dummy classes that define `get_params`/`set_params`.

2. Replaced HGBT and tree ensemble estimator checks with class name-based checks

- Per suggestion in issue comments, replaced specific `isinstance` checks with broader substring match checks to determine need to HGBT/tree ensemble treatment (`HistGradientBoosting`, `RandomForest`)
- This ensures compatibility with libraries that use same class names as scikit-learn (e.g. cuml)
- `XGB` and `LGBM` included in substrings for tree ensemble check, for compatibility with xgboost and lightgbm libraries
- Checks are made explicit (`_is_hgbt_estimator`, `_is_tree_ensemble_estimator`)
- Existing tests pass with no modifications. 
- Added tests for tree ensemble treatment for dummy classes with `RandomForest`/`XGB` in their names

3. Added parameters to allow the caller to request tree ensemble/HBGT treatment, overriding default heuristics

- Added `is_hgbt_estimator`/`is_tree_ensemble_estimator` parameters, per the suggestion in the issue comments
- Explained in docstrings that default heuristics still apply and will work in most cases, but that these parameters can be used for overrides
- Added test to verify that `is_tree_ensemble_estimator` results in tree ensemble treatment in a case where this would not be triggered by the default heuristic

Addresses #2114

## Checklist

- [ X ] I have read the contributing guidelines
- [ X ] I have added tests that verify the bug fix
- [ X ] I have added an entry to CHANGES.rst describing the fix
- [ X ] My code follows the code style of this project
- [ X ] I have checked my code and corrected any misspellings

## How Has This Been Tested?

All existing and new tests in test_tabular_pipeline.py
(+ entire test suite)

## AI Disclosure

- [ X ] This PR contains AI-generated code
    - [ X ] I have tested the code generated in my PR
    - [ X ] I have read and understood every line that has been generated by the AI agent
    - [ X ] I can explain what the AI-generated code does

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -44,9 +44,15 @@ Changes
   :user:`Jérôme Dockès <jeromedockes>`.
 
 - Made the following changes to :func:`tabular_pipeline`:
-  - Estimators are no longer required to inherit from :class:`sklearn.BaseEstimator`. Instead, scikit-learn compatibility check is based on presence of `get_params` and `set_params` attributes.
-  - Requirement for special treatment for tree ensemble/HGBT models is determined based on class name substring matching, rather than exact type matching.
-  - Parameters are provided to override default heuristics and enforce special treatment for tree ensemble/HGBT models.
+  - Estimators are no longer required to inherit from :class:`sklearn.BaseEstimator`.
+    Instead, scikit-learn compatibility check is based on presence of `get_params`
+    and `set_params` attributes.
+  - Requirement for special treatment for tree ensemble/HGBT models is determined
+    based on class name substring matching, rather than exact type matching.
+  - Parameters are provided to override default heuristics and enforce special
+    treatment for tree ensemble/HGBT models.
+
+  :pr:`2225` by :user:`Laurence Dyer <ljdyer>`. 
 
 Bugfixes
 --------
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -49,12 +49,10 @@ Changes
 
 - Made the following changes to :func:`tabular_pipeline`:
   - Estimators are no longer required to inherit from :class:`sklearn.BaseEstimator`.
-    Instead, scikit-learn compatibility check is based on presence of `get_params`
-    and `set_params` attributes.
+    Instead, scikit-learn compatibility check is based on presence of the methods:
+    `get_params`, `set_params`, `fit`, `predict`.
   - Requirement for special treatment for tree ensemble/HGBT models is determined
     based on class name substring matching, rather than exact type matching.
-  - Parameters are provided to override default heuristics and enforce special
-    treatment for tree ensemble/HGBT models.
 
   :pr:`2225` by :user:`Laurence Dyer <ljdyer>`. 
 
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -43,6 +43,10 @@ Changes
   :pr:`2222` by :user:`Ashwin V. Mohanan <ashwinvis>`, with guidance from
   :user:`Jérôme Dockès <jeromedockes>`.
 
+- Made the following changes to :func:`tabular_pipeline`:
+  - Estimators are no longer required to inherit from :class:`sklearn.BaseEstimator`. Instead, scikit-learn compatibility check is based on presence of `get_params` and `set_params` attributes.
+  - Requirement for special treatment for tree ensemble/HGBT models is determined based on class name substring matching, rather than exact type matching.
+  - Parameters are provided to override default heuristics and enforce special treatment for tree ensemble/HGBT models.
 
 Bugfixes
 --------
--- a/skrub/_tabular_pipeline.py
+++ b/skrub/_tabular_pipeline.py
@@ -1,5 +1,4 @@
 from sklearn import ensemble
-from sklearn.base import BaseEstimator
 from sklearn.impute import SimpleImputer
 from sklearn.pipeline import make_pipeline
 from sklearn.preprocessing import OrdinalEncoder
@@ -26,7 +25,7 @@ _TREE_ENSEMBLE_CLASSES = (
 def tabular_pipeline(estimator, *, n_jobs=None):
     """Get a simple machine-learning pipeline for tabular data.
 
-    Given either a scikit-learn estimator or one of the special-cased strings
+    Given either a scikit-learn compatible estimator or one of the special-cased strings
     ``'regressor'``, ``'regression'``, ``'classifier'``, ``'classification'``, this
     function creates a scikit-learn pipeline that extracts numeric features, imputes
     missing values and scales the data if necessary, then applies the estimator.
@@ -47,7 +46,9 @@ def tabular_pipeline(estimator, *, n_jobs=None):
 
     Parameters
     ----------
-    estimator : {"regressor", "regression", "classifier", "classification"} or sklearn.base.BaseEstimator
+    estimator : {"regressor", "regression", "classifier", "classification"} or scikit-learn
+        compatible estimator
+
         The estimator to use as the final step in the pipeline. Based on the type of
         estimator, the previous preprocessing steps and their respective parameters are
         chosen. The possible values are:
@@ -58,7 +59,8 @@ def tabular_pipeline(estimator, *, n_jobs=None):
         - ``'classifier'`` or ``'classification'``: a
           :obj:`~sklearn.ensemble.HistGradientBoostingClassifier` is used as the final
           step;
-        - a scikit-learn estimator: the provided estimator is used as the final step.
+        - a scikit-learn compatible estimator: the provided estimator is used as the final
+          step.
 
     n_jobs : int, default=None
         Number of jobs to run in parallel in the :obj:`TableVectorizer` step. ``None``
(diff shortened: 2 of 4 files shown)


skrub/_tabular_pipeline.py
```diff
@@ -58,13 +58,24 @@ def tabular_pipeline(estimator, *, n_jobs=None):
...
-        - a scikit-learn estimator: the provided estimator is used as the final step.
+        - a scikit-learn compatible estimator: the provided estimator is used as the final
+          step.
 
     n_jobs : int, default=None
         Number of jobs to run in parallel in the :obj:`TableVectorizer` step. ``None``
         means 1 unless in a joblib ``parallel_backend`` context. ``-1`` means using all
         processors.
 
+    is_tree_ensemble_estimator : bool, default=None
```
jeromedockes: sorry that wasn't clear from the issue discussion, but let's keep things simple for now and not add those extra parameters

skrub/_tabular_pipeline.py
```diff
@@ -240,31 +251,46 @@ def tabular_pipeline(estimator, *, n_jobs=None):
...
+    is_scikit_learn_compatible = hasattr(estimator, "get_params") and hasattr(
+        estimator, "set_params"
+    )
+    if not is_scikit_learn_compatible:
         raise TypeError(
-            "tabular_pipeline expects a scikit-learn estimator, 'regressor',"
-            " or 'classifier' as its first argument."
+            "tabular_pipeline expects a scikit-learn compatible estimator as its first"
+            " argument. The estimator object must have 'get_params' and 'set_params'"
+            " attributes."
```
jeromedockes: can we add 'fit' and 'predict' to the list? thanks!

skrub/_tabular_pipeline.py
```diff
@@ -285,7 +311,7 @@ def tabular_pipeline(estimator, *, n_jobs=None):
     if not is_estimator_from_tabicl:
         if not get_tags(estimator).input_tags.allow_nan:
```
jeromedockes: here we could be a bit more lenient on input estimators that don't implement the full scikit-learn api with something like

```
try:
    allow_nan = get_tags(estimator).input_tags.allow_nan
except AttributeError:
    allow_nan = False # assume we need imputation
if not allow_nan:
    steps.append(SimpleImputer(add_indicator=True))
```

skrub/tests/test_tabular_pipeline.py
```diff
@@ -14,6 +15,19 @@
 )
 
 
+def fake_get_tags(_):
```
jeromedockes: with the catching tag attribute error suggested below we can remove this

skrub/tests/test_tabular_pipeline.py
```diff
@@ -36,14 +50,32 @@ def test_bad_learner():
...
+    """Test that a TypeError is raised when the estimator does not have a
+    `set_params` attribute"""
+    sklearn_incompatible_learner = type("", (), {"get_params": 123})()
+    with pytest.raises(
+        TypeError, match=".*expects a scikit-learn compatible estimator"
+    ):
+        tabular_pipeline(sklearn_incompatible_learner)
+
+
+def test_sklearn_compatible_learner_succeeds(monkeypatch):
```
jeromedockes: can we create a real estimator instead,

```python
class Regressor:
    def fit(self, X, y=None): return self
    def predict(self, X): return np.zeros(X.shape[0])
    def get_params(self): return {}
    def set_params(self, **params): return self

then make a pipeline and check we can call fit and predict on toy data and get the expected prediction? thanks
```

skrub/tests/test_tabular_pipeline.py
```diff
@@ -66,6 +98,70 @@ def test_tree_learner():
     assert tv.datetime.periodic_encoding is None
 
 
+def test_tree_ensemble_treatment_for_any_random_forest(monkeypatch):
+    """Test that special treatment for tree ensemble models is applied when
+    substring 'RandomForest' appears in estimator class name"""
+    IAmARandomForestEstimator = type(
```
jeromedockes: in those instead of calling type we can subclass the dummy regressor defined earlier eg
```
class UserRandomForest(Regressor): pass
```

jeromedockes (commented):
thank you very much @ljdyer !! this looks great. 

- I think we can remove the parameters that override the kind of estimator for now, because we would need to discuss more what are the classes of estimator and the parameter names
- the tests can be simplified slightly

looks great otherwise, thanks :)

skrub/_tabular_pipeline.py
```diff
@@ -240,31 +251,46 @@ def tabular_pipeline(estimator, *, n_jobs=None):
...
-            f" argument. Pass an instance of {estimator.__name__} rather than the class"
-            " itself."
+            "tabular_pipeline expects a scikit-learn compatible estimator instance as"
+            " its first argument, but you have passed a type. Pass an instance of the"
+            " estimator rather than the class itself."
         )
-    if not isinstance(estimator, BaseEstimator):
+
+    is_scikit_learn_compatible = hasattr(estimator, "get_params") and hasattr(
+        estimator, "set_params"
```
Sanjays2402: This accepts non-callable attributes, so an object with `get_params = 1` and `set_params = 1` passes the guard but the returned pipeline raises `TypeError` as soon as `get_params()` or `clone()` touches it. Should these checks use `callable(getattr(..., None))` instead?

ljdyer:
Thanks @jeromedockes for the feedback! I'll be working on this again in the next few days. Cheers

skrub/_tabular_pipeline.py
ljdyer: No problem, will do.
Just want to point out that this makes the requirements a bit more stringent that previously with `BaseEstimator` inheritance. For example [AgglomerativeClustering](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.AgglomerativeClustering.html) inherits from `BaseEstimator` and has `get_params`/`set_params`/`fit` but does not have have `predict`. But perhaps we want to exclude it or are not concerned because it is neither a regressor or a classifier?
Another example that has the first three but does not have `predict` is [PCA](https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html), though this does not inherit from `BaseEstimator`.

skrub/_tabular_pipeline.py
ljdyer: Hi Sanjay, I agree that it's fine to check for `callable(` as well. Extending this to `fit` and `predict` too. Only caveat is what I wrote in my response to Jerome's comment above about estimators that don't implement `predict`.

skrub/_tabular_pipeline.py
ljdyer: Looks good - implementing more or less as you wrote

skrub/tests/test_tabular_pipeline.py
ljdyer: Great advice - makes the tests much simpler

skrub/tests/test_tabular_pipeline.py
ljdyer: I was able to add the `fit` test with a pandas DataFrame, but with `predict` we run into issues as `check_is_fitted` gets called. Could monkeypatch, but in the interest of test simplicity could we skip the test with toy data?

ljdyer:
Thank you @jeromedockes for the review. I have resubmitted so please review again when convenient.

jeromedockes:
thanks a lot for making the changes @ljdyer ! I will look at it again in the next few days :)

CHANGES.rst
```diff
@@ -100,6 +100,15 @@ Changes
 - Removed the parameter ``how`` of :meth:`DataOp.skb.apply`. :pr:`2281` by
   :user:`Eloi Massoulié <emassoulie>`.
 
+- Made the following changes to :func:`tabular_pipeline`:
+
+  - Estimators are no longer required to inherit from :class:`sklearn.BaseEstimator`.
+    Instead, scikit-learn compatibility check is based on presence of the methods:
+    `get_params`, `set_params`, `fit`, `predict`.
```
```suggestion
    ``get_params``, ``set_params``, ``fit``, ``predict``.
```

rcap107 (approved):
Looks good to me!

PR #2231, opened 2026-07-25
fix(ToDatetime): accept pandas columns of datetime.date objects

### LINKED ISSUE #2084: BUG - Pandas columns created with datetime.date are not recognized as "date"
Columns that are defined with `datetime.date` are treated as "objects" by pandas, and do not have datetime dtype so they are not encoded as dates by the DatetimeEncoder. 

```python
import pandas as pd
import polars as pl
from datetime import date, datetime

df_pd = pd.DataFrame({"date": [date(2002, 1,1)], "datetime": [datetime(2002, 1,1)]})
df_pl = pl.DataFrame({"date": [date(2002, 1,1)], "datetime": [datetime(2002, 1,1)]})

DatetimeEncoder().fit_transform(df_pd["date"]) # rejected because it's not a date
DatetimeEncoder().fit_transform(df_pl["date"]) # works as intended
```

This is somewhat related to #2035, but more annoying to deal with because the column needs to be converted to datetime somehow before it can actually be seen as a datetime 🙃

Sanjays2402:
## Description

Fixes: #2084

Pandas has no dedicated dtype for `datetime.date`, so a column built from date objects has the `object` dtype: it is neither a date column nor a string column, so `ToDatetime` rejected it and the `DatetimeEncoder`/`TableVectorizer` never saw it. The equivalent polars column has the `Date` dtype and was handled correctly, so the two backends behaved differently.

## Changes

- Cast pandas `object` columns that contain only `datetime.date` to Datetime before the dtype checks in `ToDatetime.fit_transform`. Columns that are not entirely made of dates, and columns that fail to convert, are returned unchanged, so string parsing and the existing rejections are unaffected.
- Add a non-regression test, and update `test_to_datetime_func`, which previously excluded the pandas `date-col` because of this bug.
- Changelog entry.

## Testing

- [x] Unit tests added/updated
- [ ] Integration tests passed
- [x] Manual checks performed: `new test fails on main with RejectColumn and passes with the fix (stash-verified); skrub/tests/test_to_datetime.py and test_datetime_encoder.py pass`

## Checklist

- [x] Code follows project style guidelines (linting passes).
- [x] Tests added/updated for changes.
- [x] All tests pass locally.
- [ ] Documentation updated (if needed).


skrub/_to_datetime.py
```diff
@@ -82,6 +83,28 @@ def _convert_time_zone_polars(col, time_zone):
...
+    if not (sbd.is_pandas(column) and sbd.is_object(column)):
+        return column
+    not_null = sbd.drop_nulls(column)
+    if sbd.shape(not_null)[0] == 0:
+        return column
+    if not all(type(value) is datetime.date for value in not_null.head(_SAMPLE_SIZE)):
+        return column
+    try:
+        return pd.to_datetime(column)
+    except Exception:
```
rcap107, line 104: Please change this exception to the specific exception that is raised by pd.to_datetime instead

skrub/_to_datetime.py
```diff
@@ -82,6 +83,28 @@ def _convert_time_zone_polars(col, time_zone):
...
+    Pandas has no dedicated dtype for ``datetime.date``, so such a column has
+    the ``object`` dtype and is neither a date nor a string column. Polars, on
+    the other hand, stores them as ``pl.Date``. Casting them here makes the
+    handling of the two dataframe libraries consistent. Columns that do not
+    contain only dates are returned unchanged.
+    """
+    if not (sbd.is_pandas(column) and sbd.is_object(column)):
+        return column
+    not_null = sbd.drop_nulls(column)
+    if sbd.shape(not_null)[0] == 0:
```
rcap107, line 98: this line isn't covered by tests and coverage is complaining about it

skrub/_to_datetime.py
```diff
@@ -82,6 +83,28 @@ def _convert_time_zone_polars(col, time_zone):
...
+        return column
+    if not all(type(value) is datetime.date for value in not_null.head(_SAMPLE_SIZE)):
+        return column
+    try:
+        return pd.to_datetime(column)
+    except Exception:
+        return column
```
rcap107, line 105: same for this line, it should be covered

rcap107 (commented):
Hi @Sanjays2402, thanks for this PR. 

It looks good so far, but coverage of the tests needs to be improved. Since the missing lines are in the _cast_date_objects method, you can add a test specifically for that function, rather than using ToDatetime. 

Other than that, I think this issue had some kind of hidden caveat that I currently do not remember, and I'd like to weigh in with some other maintainers that are currently away before merging.

Sanjays2402:
Thanks @rcap107 — both addressed. The bare except in _cast_date_objects is now except (ValueError, TypeError), which covers what pd.to_datetime actually raises (e.g. OutOfBoundsDatetime/DateParseError are ValueErrors; I confirmed OutOfBoundsDatetime on out-of-bounds dates with pandas 2.2). And there are now four focused tests on _cast_date_objects itself rather than through ToDatetime: conversion of date objects, the empty/all-null column path, the non-date passthrough, and the conversion-failure fallback (the three lines coverage was flagging).

skrub/tests/test_to_datetime.py
```diff
@@ -249,3 +246,62 @@ def test_specific_time_encoding():
...
+    out = to_dt.fit_transform(col)
+    assert sbd.is_any_date(out)
+    assert out[0] == pd.Timestamp("2002-01-01")
+    assert pd.isna(out[1])
+
+    transformed = to_dt.transform(pd.Series([date(2005, 5, 5)], name="when"))
+    assert transformed[0] == pd.Timestamp("2005-05-05")
+
+
+def test_cast_date_objects_converts_date_objects():
```
rcap107: This and the other tests should use the df_module fixture from conftest to make sure that the new function works properly with polars

skrub/tests/test_to_datetime.py
```diff
@@ -249,3 +246,62 @@ def test_specific_time_encoding():
...
+    """A pandas object column of datetime.date becomes Datetime."""
+    col = pd.Series([date(2002, 1, 1), None, date(2003, 2, 2)], name="when")
+    out = _cast_date_objects(col)
+    assert sbd.is_any_date(out)
+    assert out[0] == pd.Timestamp("2002-01-01")
+    assert pd.isna(out[1])
```
rcap107: these two lines may be unneded after all, and may cause trouble after switching to df_module, so I think they can be removed

skrub/tests/test_to_datetime.py
```diff
@@ -249,3 +246,62 @@ def test_specific_time_encoding():
...
+
+
+def test_cast_date_objects_empty_column_unchanged():
```
rcap107: same comment about df_module here and below

skrub/tests/test_to_datetime.py
```diff
@@ -249,3 +246,62 @@ def test_specific_time_encoding():
...
+    assert _cast_date_objects(col) is col
+
+
+def test_cast_date_objects_non_date_column_unchanged():
+    """Object columns that are not all dates are returned unchanged."""
+    col = pd.Series(["hello", "world"], name="when")
+    assert _cast_date_objects(col) is col
+
+
+def test_cast_date_objects_conversion_failure_unchanged(monkeypatch):
```
rcap107, line 289: this is pandas specific so no need for df_module

rcap107 (commented):
Hi @Sanjays2402, thanks for the fixes! 

The tests still need some work, but we're getting close to merging this

rcap107:
Something else to mention: we did a pretty big refactoring of the docs, so you should merge with upstream to avoid conflicts @Sanjays2402

skrub/tests/test_to_datetime.py
Sanjays2402: Done — the three _cast_date_objects tests now take the df_module fixture, so they run against pandas, pandas-nullable-dtypes and polars (a84c6558). I dropped the two pandas-specific value assertions you flagged below since they couldn't survive the switch; the remaining sbd.is_any_date check still verifies the conversion.

skrub/tests/test_to_datetime.py
Sanjays2402: Removed in a84c6558.

skrub/tests/test_to_datetime.py
Sanjays2402: Done — applied to all three _cast_date_objects tests in a84c6558.

skrub/tests/test_to_datetime.py
Sanjays2402: Agreed, left as-is.

Sanjays2402:
Merged upstream main into the branch

rcap107:
@Sanjays2402 this PR is stuck because of some conflicts, and can be merged once they're addressed. Could you do that? Thanks!

Sanjays2402:
Done, merged upstream main into the branch and resolved the conflicts. The only real one was in the test imports (UTC on your side vs date/timezone on mine); everything else merged cleanly and the _cast_date_objects fix is untouched. ToDatetime suite is green (89 passed). Ready for another look whenever you have a moment, thanks!

rcap107 (approved):
Looks good to me, thanks a lot for the fix @Sanjays2402

PR #2236, opened 2026-07-29
series of changes to datasets API documentation

lisaleemcb:
Updated the documentation for the datasets available to download.

Changes consist of:
- added examples for all dataset fetchers
- cleaned up some docstrings and made their structure more uniform across datasets
- changed the References section in the `fetch_electricity_forecasting` docstring to a See Also

### DIFF
--- a/skrub/datasets/_fetching.py
+++ b/skrub/datasets/_fetching.py
@@ -8,15 +8,15 @@ from ._utils import download_dataset, load_dataset_files, load_simple_dataset
 
 
 def fetch_employee_salaries(data_home=None, split="all"):
-    """Fetches the employee salaries dataset (regression), available at \
+    """
+    Fetches the employee salaries dataset (regression), available at \
         https://github.com/skrub-data/skrub-data-files
 
-    Description of the dataset:
-        Annual salary information including gross pay and overtime pay for all
-        active, permanent employees of Montgomery County, MD paid in calendar
-        year 2016. This dataset is a copy of https://www.openml.org/d/42125
-        where some features are dropped to avoid data leaking.
-        Size on disk: 1.3MB.
+    Annual salary information including gross pay and overtime pay for all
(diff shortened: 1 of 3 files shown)


emassoulie:
Looking good! It seems the tests fail because loading a dataset involves printing progress, which leads to unexpected outputs. For instance, the last test that does this expects no output but has `Downloading 'videogame_sales' from https://github.com/skrub-data/skrub-data-files/raw/refs/heads/main/videogame_sales.zip (attempt 1/3)` as one.

The two ways this can be avoided are either to actually put this expected output in the example or bypass API testing entirely for these docstrings. The former will make the docstring a bit unwieldy, and require maintenance if the format of this output ever changes, so I'd lean towards the latter (besides, I think these specific examples can do without strict verification).

rcap107:
> Looking good! It seems the tests fail because loading a dataset involves printing progress, which leads to unexpected outputs. For instance, the last test that does this expects no output but has `Downloading 'videogame_sales' from https://github.com/skrub-data/skrub-data-files/raw/refs/heads/main/videogame_sales.zip (attempt 1/3)` as one.
> 
> The two ways this can be avoided are either to actually put this expected output in the example or bypass API testing entirely for these docstrings. The former will make the docstring a bit unwieldy, and require maintenance if the format of this output ever changes, so I'd lean towards the latter (besides, I think these specific examples can do without strict verification).

Good catch @emassoulie. I agree that the latter solution is probably the better one. Doctests here are not too important because the test suite is explicitly testing the same things elsewhere.

rcap107:
At a second glance, I think the doctests for single tables that only print the shape can be removed: that information is already present in the docstring.

However, for datasets that include multiple tables I think it's useful to have the keys of the bunch so that users can easily figure out what datasets they are working with.

rcap107 (approved):
Looks good to me, thanks a lot @lisaleemcb!

Issue #2237, opened 2026-08-01
GapEncoder(init='k-means') crashes on columns containing nulls


Hrafz:
### Bug description

`GapEncoder(init="k-means")` raises an `AttributeError`/`ValueError` from deep
inside scikit-learn's vectorizers whenever the fitted column contains a
missing value, while `init="k-means++"` (the default) and `init="random"` fit
the exact same column without error.

### Steps/code to reproduce

```python
import pandas as pd
from skrub import GapEncoder

s = pd.Series(["alice", "bob", None, "alice", pd.NA, float("nan")], name="s")

GapEncoder(n_components=3, init="k-means++").fit_transform(s)  # OK
GapEncoder(n_components=3, init="random").fit_transform(s)     # OK
GapEncoder(n_components=3, init="k-means").fit_transform(s)    # crashes
```

### Expected results

`init="k-means"` handles the column's null values the same way the other two
`init` options do (nulls are treated like the empty string throughout the rest
of `GapEncoder`, see `skrub._utils.unique_strings` and `test_missing_values`
in `test_gap_encoder.py`).

### Actual results

```
AttributeError: 'NAType' object has no attribute 'lower'
```
(the `None`/`pd.NA` entries reach `HashingVectorizer`/`CountVectorizer` first), or

```
ValueError: np.nan is an invalid document, expected byte or unicode string.
```
(if only `np.nan` is present), both raised from scikit-learn's text
preprocessor inside `skrub._gap_encoder.get_kmeans_prototypes`. Which
exception you get depends on which null sentinel is in the column, not on
`hashing`.

Root cause: `GapEncoder._init_vars` builds the null-substituted, deduplicated
array `unq_X` via `unique_strings(X, is_null)` and correctly uses it to fit the
n-gram vectorizer, but then calls
`self._init_w(unq_V[lookup], X)` — passing the **raw** `X` (still containing
`None`/`NaN`) rather than the null-substituted values. `_init_w`'s
`init="k-means"` branch re-vectorizes this raw `X` from scratch inside
`get_kmeans_prototypes`, which is where the null value reaches scikit-learn's
preprocessor and crashes. The `"k-means++"` and `"random"` branches never
touch `X` directly (only the already-vectorized `V`), which is why they are
unaffected.

I have a one-line fix (pass `unq_X[lookup]` instead of `X`) plus a regression
test and would be happy to submit a PR.

### Versions

skrub 0.11.dev0 (current `main`, c08c674)

[...]

PR #2238, opened 2026-08-01
Fix GapEncoder(init="k-means") crash on columns containing nulls

Hrafz:
Addresses issue  
https://github.com/skrub-data/skrub/issues/2237

## Description

`GapEncoder(init="k-means")` crashes on any column containing a missing
value, while the two other `init` options fit the same column fine:

```python
import pandas as pd
from skrub import GapEncoder

s = pd.Series(["alice", "bob", None, "alice", pd.NA, float("nan")], name="s")

GapEncoder(n_components=3, init="k-means++").fit_transform(s)  # OK
GapEncoder(n_components=3, init="random").fit_transform(s)     # OK
GapEncoder(n_components=3, init="k-means").fit_transform(s)
# AttributeError: 'NAType' object has no attribute 'lower'
```

Expected: `init="k-means"` handles the column's nulls the same way the other
two `init` options do (nulls are treated as the empty string throughout
`GapEncoder`, see `skrub._utils.unique_strings` and the existing
`test_missing_values`).

Actual: it raises from scikit-learn's text preprocessor. The exception type
depends on the column's dtype rather than on which sentinel appears first:

| column | `init="k-means"` raises |
|---|---|
| object dtype containing `np.nan` | `ValueError: np.nan is an invalid document, expected byte or unicode string.` |
| object dtype containing `None` | `ValueError: np.nan is an invalid document, expected byte or unicode string.` |
| `string` dtype containing `pd.NA` | `AttributeError: 'NAType' object has no attribute 'lower'` |

`init="k-means++"` and `init="random"` fit all three without error.

Cause: `_init_vars` builds the null-substituted array `unq_X` via
`unique_strings(X, is_null)` and uses it everywhere else in the method, but
was calling `self._init_w(unq_V[lookup], X)` with the raw `X` (still
containing the original null). Only the `"k-means"` branch touches `X`
directly — it re-vectorizes it from scratch inside `get_kmeans_prototypes`,
which is where the null reaches the preprocessor. `"k-means++"` and
`"random"` only ever consume the already-vectorized `V`.

Fix: pass `unq_X[lookup]` instead of `X` — the same null-substituted,
per-sample reconstruction already used throughout the rest of the method.
When there are no nulls, `unq_X[lookup] == X`, so this is a no-op for the
common case.

## Checklist

## How Has This Been Tested?

`test_missing_values` is now parametrized over `init`, so it also exercises
`"k-means"` (previously only the default `"k-means++"`).

Before the fix (test change only, source untouched):
```
skrub/tests/test_gap_encoder.py::test_missing_values[pandas-numpy-dtypes-k-means] FAILED
skrub/tests/test_gap_encoder.py::test_missing_values[pandas-nullable-dtypes-k-means] FAILED
2 failed, 4 passed, 44 deselected
```

After the fix:
```
skrub/tests/test_gap_encoder.py -q
50 passed
```

The full `skrub/tests` suite passes with no failures and no change in the
skipped/xfailed/xpassed counts beyond the newly-passing parametrizations.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -53,6 +53,9 @@ Bugfixes
 - The parallel coordinate plot created by :meth:`ParamSearch.show_results` could
   have incorrect tick labels in some cases. This has been fixed in :pr:`2215` by
   :user:`Jérôme Dockès <jeromedockes>`.
+- :class:`GapEncoder` with ``init="k-means"`` raised an error when the input
+  column contained missing values. This has been fixed in :pr:`2238` by
+  :user:`Achraf Ez <Hrafz>`.
 
 Deprecations
 ------------
--- a/skrub/_gap_encoder.py
+++ b/skrub/_gap_encoder.py
@@ -269,7 +269,7 @@ class GapEncoder(TransformerMixin, SingleColumnTransformer):
                 )
         _, self.n_vocab = unq_V.shape
         # Init the topics W given the n-grams counts V
-        self.W_, self.A_, self.B_ = self._init_w(unq_V[lookup], X)
+        self.W_, self.A_, self.B_ = self._init_w(unq_V[lookup], unq_X[lookup])
         # Init the activations unq_H of each unique input string
         unq_H = _rescale_h(unq_V, np.ones((len(unq_X), self.n_components)))
         # Update self.H_dict_ with unique input strings and their activations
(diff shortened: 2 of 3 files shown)


rcap107:
Hi @Hrafz, thanks for the PR. We'll be looking at this and #2240 in the next few days (most maintainers were off in the past few weeks)

rcap107 (approved):
Looks good to me, thanks @Hrafz!

PR #2243, opened 2026-08-05
Support tuple unpacking of DataOps

e-strauss:
`a, b, c = data_op` now creates a node per target instead of raising an error, so a deferred function returning several values can be unpacked as usual.

Python calls `iter()` on the right-hand side from an `UNPACK_SEQUENCE` instruction whose argument is the number of targets, so `DataOp.__iter__` reads that count off the calling frame's bytecode and creates an `AsTuple` node (which converts the result to a tuple and checks its length) plus one `GetItem` per target. A wrong number of targets is reported eagerly when a preview value is available, and at evaluation time otherwise.

Only `UNPACK_SEQUENCE` is recognised, so anything else (`for` loops, `list(data_op)`, `f(*data_op)`, `a, *rest = data_op`) keeps raising the previous TypeError, as does any case where the bytecode cannot be inspected. Indexing into the result stays available as a portable alternative.

Example:
```python
test = skrub.var("test", [1, 2])

@skrub.deferred
def process_test_data(test):
   left = test[0]
   right = test[1]
   return left, right

# works now without raising
left, right = process_test_data(test)
```


CHANGES.rst
```diff
@@ -12,6 +12,10 @@ Ongoing development
 
 New Features
 ------------
+- It is now possible to unpack a :class:`DataOp` that evaluates to several
+  values, for example ``train, test = data_op``. Each target becomes a DataOp
+  that extracts one of the values.
+  :pr:`3` by :user:`Elias Strauss <e-strauss>`.
```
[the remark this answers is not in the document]
e-strauss: typo, fixed now

skrub/_data_ops/_data_ops.py
```diff
@@ -700,6 +731,13 @@ def __bool__(self):
         )
 
     def __iter__(self):
+        # Unpacking (`a, b = data_op`) is supported: we know how many values are
+        # expected, so we can create a node for each of them. Any other kind of
+        # iteration would need the length of the result, which is unknown until
+        # the DataOp is evaluated.
+        if (arity := _unpack_arity()) is not None:
+            values = unpack(self, arity)
+            return iter([values[i] for i in range(arity)])
```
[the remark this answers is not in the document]
e-strauss: `() = data_op` binds nothing so raising is its only possible effect, and when a preview is available we already raise Python's exact message — treating arity 0 as unsupported would replace that correct error with a misleading one to fix a statement nobody writes.

jeromedockes:
thanks a lot for opening this PR, @e-strauss !!

for context, the original discussion where @e-strauss proposed this approach happened on the skrub discord:

https://discord.com/channels/1220094555282477159/1532375520405229649

skrub/_data_ops/_data_ops.py
```diff
@@ -197,6 +198,36 @@ def _format_data_op_creation_stack():
...
+    if (getframe := getattr(sys, "_getframe", None)) is None:
+        return None
+    try:
+        # skip the frames of this function and of DataOp.__iter__
+        frame = getframe(2)
+        for instruction in dis.get_instructions(frame.f_code):
+            if instruction.offset == frame.f_lasti:
+                if instruction.opname == "UNPACK_SEQUENCE":
+                    return instruction.arg
+                return None
```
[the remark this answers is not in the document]
jeromedockes: - even for large functions this time would be negligible
- if the == is not met we are in an error case anyway so speed is not so important we are not in a fast path but about to raise an exception about a programming error in client code
- (I don't see in which situation the == would not be true for one of the instructions but I could easily be wrong about that part)

so this comment is noise / needless optimization

skrub/_data_ops/_data_ops.py
```diff
@@ -197,6 +198,28 @@ def _format_data_op_creation_stack():
     return traceback.format_list(stack)
 
 
+def _unpack_arity():
+    """Number of targets in the unpacking assignment being executed, if any.
+
+    Read from the caller's ``UNPACK_SEQUENCE`` instruction, or None if not found.
+    """
+    if (getframe := getattr(sys, "_getframe", None)) is None:
```
jeromedockes, line 206: I guess this could be thrown in the same "best effort / catch-all" bag as the rest but don't have a strong preference either way

skrub/_data_ops/_data_ops.py
```diff
@@ -197,6 +198,28 @@ def _format_data_op_creation_stack():
...
+        return None
+    try:
+        # skip the frames of this function and of DataOp.__iter__
+        frame = getframe(2)
+        for instruction in dis.get_instructions(frame.f_code):
+            if instruction.offset == frame.f_lasti:
+                if instruction.opname == "UNPACK_SEQUENCE":
```
jeromedockes, line 213: could you add a short comment saying that we intentionally don't handle UNPACK_EX? thanks!

skrub/_data_ops/_data_ops.py
```diff
@@ -1662,6 +1692,33 @@ def pretty_repr(self):
...
+        expected, got = e.expected_length, len(result)
+        if got != expected:
+            problem = "not enough" if got < expected else "too many"
+            raise ValueError(
+                f"{problem} values to unpack (expected {expected}, got {got})"
+            )
+        return result
+
+    def __repr__(self):
+        return f"<{self.__class__.__name__} {short_repr(self.iterable)}>"
```
jeromedockes, line 1714:
```suggestion

    def __repr__(self):
        return f"<{self.__class__.__name__}: {self.expected_length} items>"
```

I know I suggested the current version but on second thought showing the repr of the parent is a bit redundant and not really consistent with most other dataops, showing the number of items is more informative and consistent with eg skrub.concat . (we know that expected_length is an actual number because AsTuple is only created by iter, though even if it wasn't the consequence would not be dramatic)

skrub/_data_ops/_data_ops.py
```diff
@@ -700,6 +723,13 @@ def __bool__(self):
         )
 
     def __iter__(self):
+        # Unpacking (`a, b = data_op`) is supported: we know how many values are
+        # expected, so we can create a node for each of them. Any other kind of
+        # iteration would need the length of the result, which is unknown until
+        # the DataOp is evaluated.
+        if (arity := _unpack_arity()) is not None:
+            values = unpack(self, arity)
+            return iter([values[i] for i in range(arity)])
```
jeromedockes, line 732: (nitpick) out of curiosity why do you prefer `iter([...])` rather than returning a generator expression here?

CHANGES.rst
```diff
@@ -12,6 +12,10 @@ Ongoing development
 
 New Features
 ------------
+- It is now possible to unpack a :class:`DataOp` that evaluates to several
+  values, for example ``train, test = data_op``. Each target becomes a DataOp
+  that extracts one of the values.
+  :pr:`2243` by :user:`Elias Strauss <e-strauss>`.
```
jeromedockes, line 18:
```suggestion
- It is now possible to unpack a :class:`DataOp` that evaluates to an iterable
  (of known size), for example ``first, second = data_op``. Each target becomes
  a DataOp that extracts one of the items.
  :pr:`2243` by :user:`Elias Strauss <e-strauss>`.
```

jeromedockes (commented):
looks great apart from some nitpicks! this is an issue that had come up a couple of times so it is a really neat improvement, thanks a lot @e-strauss ! :)

skrub/_data_ops/_data_ops.py
e-strauss: yes, makes sense! had a similar thought when I saw the printed DAG

skrub/_data_ops/_data_ops.py
e-strauss: ye, I agree, this will make the code a bit more readable

rcap107:
Hi @e-strauss, thanks a lot for the time you spent on this PR!

We're planning another release soon-ish (not before a couple of weeks) and it'd be great if this PR could be part of it. Do you think you'll have time to work on this? If not, no problem at all and no pressure: it'll just go in the next release some more time down the line.

e-strauss:
Hi @rcap107 , I will get to back it this week :)

rcap107:
> Hi @rcap107 , I will get to back it this week :)

Amazing, thanks a lot!

rcap107:
Just a heads up that the docs have been moved from doc/ to skrub/_docs

skrub/_data_ops/_data_ops.py
e-strauss: true, no need to materialize the list here, i did not know that I can use generator expr here, these part of python, I am not so familiar ^^

jeromedockes (approved):
cool, this is a very nice addition! thank you very much @e-strauss  :)

e-strauss:
> cool, this is a very nice addition! thank you very much @e-strauss  :)

thanks for the assistance :)

PR #2244, opened 2026-08-05
Add CategoricalEncoder combining OneHotEncoder and TargetEncoder

### LINKED ISSUE #2065: Encoder that combines One-hot and target encoding
### Problem Description

A suggestion from @ogrisel : an encoder that combines a one-hot encoding of the most frequent categories with target-encoding to also capture some information for less-frequent categories.

currently this has to be done manually in a dataop or FeatureUnion:

```python
one_hot_encoded = X.skb.apply(
    OneHotEncoder(
        sparse_output=False,
        dtype=np.float32,
        handle_unknown="ignore",
        max_categories=5,
    )
)
target_encoded = X.skb.apply(TargetEncoder(random_state=0), y=y)
encoded = one_hot_encoded.skb.concat([target_encoded], axis=1)
```

<details>
<summary>full example</summary>

```python
import numpy as np
import skrub
import pandas as pd
from sklearn.preprocessing import OneHotEncoder, TargetEncoder

csv = skrub.var("csv", skrub.datasets.fetch_employee_salaries().path)
df = csv.skb.apply_func(pd.read_csv)
y = df["current_annual_salary"]
X = df[["department", "employee_position_title"]]


one_hot_encoded = X.skb.apply(
    OneHotEncoder(
        sparse_output=False,
        dtype=np.float32,
        handle_unknown="ignore",
        max_categories=5,
    )
)
target_encoded = X.skb.apply(TargetEncoder(random_state=0), y=y)
encoded = one_hot_encoded.skb.concat([target_encoded], axis=1)


encoded.skb.full_report()
```

</details>

Image

which is quite verbose, and sorts the outputs by encoder whereas we would like them to be sorted by input column. it would be much nicer to have something like

```python
encoded = X.skb.apply(skrub.CategoricalEncoder())
```

if using a scikit-learn pipeline instead of dataops it is also doable but quite inconvenient too, we have to use a scikit-learn FeatureUnion inside of a skrub ApplyToCols or chain 2 ApplyToCols and leverage `keep_original` and `rename_columns` to allow processing the columns twice.

### Feature Description

an encoder that combines one-hot for the most important categories and a fallback targetencoding representation for the rare ones. the inner encoders could be hardcoded or configurable:

```
CategoricalEncoder(frequent=OrdinalEncoder(...), rare=TargetEncoder(), max_frequent=10)
```

(all names TBD)

### Alternative Solutions

_No response_

### Additional Context

_No response_

faithlesstomas:
# New Feature Pull Request

## Description

This PR introduces `CategoricalEncoder`, a single-column transformer that combines `OneHotEncoder` (for encoding frequent categories into one-hot binary columns) and `TargetEncoder` (for encoding categories using the target variable `y`).

Fixes #2065

## Checklist

- [x] I have added tests that verify the feature works
- [x] I have added necessary documentation:
  - [x] Docstrings for new functions/classes
  - [x] API documentation if applicable
  - [ ] Example notebook or script demonstrating usage
- [x] I have added an entry to CHANGES.rst describing the feature

If the feature inherits from `SingleColumnTransformer`:
- [x] `get_feature_names_out` has been implemented

## How Has This Been Tested?

- Added unit tests in `skrub/tests/test_categorical_encoder.py` covering:
  - Transformation on both Pandas and Polars Series.
  - Validation when target `y` is missing (`y=None`).
  - Cloning and isolation when custom `one_hot_encoder` or `target_encoder` instances are provided.
  - Integration with `ApplyToCols`.
  - Scikit-learn estimator compatibility (`check_is_fitted`, `clone`).
- Ran local test suites (`pytest skrub/tests/test_categorical_encoder.py` and `pytest skrub/tests/test_single_column_transformer.py`).
- Verified code style and formatting with `ruff check`.

## Example Usage

```python
import pandas as pd
from skrub import CategoricalEncoder

s = pd.Series(["a", "b", "a", "c", "d", "e", "a", "b", "c", "d"], name="col")
y = pd.Series([1, 0, 1, 0, 1, 0, 1, 0, 1, 0])

enc = CategoricalEncoder(max_categories=3)
res = enc.fit_transform(s, y)
print(res)

### DIFF
--- a/skrub/_categorical_encoder.py
+++ b/skrub/_categorical_encoder.py
@@ -107,14 +107,20 @@ class CategoricalEncoder(TransformerMixin, SingleColumnTransformer):
         X_pandas = sbd.to_pandas(column).to_frame()
 
         if sbd.is_dataframe(y):
-            y_pandas = sbd.to_pandas(sbd.col_by_idx(y, 0))
-        elif sbd.is_column(y):
-            y_pandas = sbd.to_pandas(y)
+            y_col = sbd.col_by_idx(y, 0)
         else:
-            y_pandas = y
+            y_col = y
+
+        if sbd.is_column(y_col):
+            y_vec = sbd.to_numpy(y_col)
+        else:
+            y_vec = np.asarray(y_col)
+
+        if y_vec.ndim > 1:
+            y_vec = y_vec.ravel()
 
         ohe_res = self.one_hot_encoder_.fit_transform(X_pandas)
-        te_res = self.target_encoder_.fit_transform(X_pandas, y_pandas)
+        te_res = self.target_encoder_.fit_transform(X_pandas, y_vec)
 
         if hasattr(ohe_res, "toarray"):
             ohe_res = ohe_res.toarray()
--- a/skrub/_categorical_encoder.py
+++ b/skrub/_categorical_encoder.py
@@ -87,9 +87,7 @@ class CategoricalEncoder(TransformerMixin, SingleColumnTransformer):
             DataFrame containing one-hot and target-encoded features.
         """
         if y is None:
-            raise ValueError(
-                "Target y must be provided to fit CategoricalEncoder."
-            )
+            raise ValueError("Target y must be provided to fit CategoricalEncoder.")
 
         if self.one_hot_encoder is None:
             self.one_hot_encoder_ = OneHotEncoder(
@@ -123,9 +121,7 @@ class CategoricalEncoder(TransformerMixin, SingleColumnTransformer):
         if te_res.ndim == 1:
             te_res = te_res.reshape(-1, 1)
 
-        ohe_names = list(
-            self.one_hot_encoder_.get_feature_names_out([col_name])
-        )
+        ohe_names = list(self.one_hot_encoder_.get_feature_names_out([col_name]))
         te_names = list(self.target_encoder_.get_feature_names_out([col_name]))
 
         ohe_df = sbd.make_dataframe_like(column, dict(zip(ohe_names, ohe_res.T)))
@@ -151,9 +147,7 @@ class CategoricalEncoder(TransformerMixin, SingleColumnTransformer):
         res_df : Pandas or Polars DataFrame
             Transformed features.
         """
-        check_is_fitted(
-            self, ["one_hot_encoder_", "target_encoder_", "all_outputs_"]
-        )
+        check_is_fitted(self, ["one_hot_encoder_", "target_encoder_", "all_outputs_"])
 
         X_pandas = sbd.to_pandas(column).to_frame()
 
--- a/skrub/_categorical_encoder.py
+++ b/skrub/_categorical_encoder.py
@@ -119,6 +119,12 @@ class CategoricalEncoder(TransformerMixin, SingleColumnTransformer):
         if y_vec.ndim > 1:
             y_vec = y_vec.ravel()
 
+        if y_vec.dtype == object:
+            try:
+                y_vec = y_vec.astype(float)
+            except (ValueError, TypeError):
+                pass
+
         ohe_res = self.one_hot_encoder_.fit_transform(X_pandas)
         te_res = self.target_encoder_.fit_transform(X_pandas, y_vec)
 
--- a/skrub/_categorical_encoder.py
+++ b/skrub/_categorical_encoder.py
@@ -1,0 +1,184 @@
+"""
+Implementation of CategoricalEncoder combining OneHotEncoder and TargetEncoder.
+"""
+
+import numpy as np
+from sklearn.base import TransformerMixin, clone
+from sklearn.preprocessing import OneHotEncoder, TargetEncoder
+from sklearn.utils.validation import check_is_fitted
+
+from . import _dataframe as sbd
+from ._single_column_transformer import SingleColumnTransformer
+
+__all__ = ["CategoricalEncoder"]
+
+
+class CategoricalEncoder(TransformerMixin, SingleColumnTransformer):
+    """Encode a single categorical column combining OneHotEncoder and TargetEncoder.
+
+    This transformer applies a :class:`~sklearn.preprocessing.OneHotEncoder`
+    to encode frequent categories into binary one-hot columns, and a
+    :class:`~sklearn.preprocessing.TargetEncoder` to target-encode the column.
+
+    Parameters
+    ----------
+    max_categories : int or None, default=10
+        Maximum number of categories for the ``OneHotEncoder``. If there are more
+        categories, the remaining ones are grouped into an infrequent category.
+        Ignored if a custom ``one_hot_encoder`` is provided.
+
+    one_hot_encoder : OneHotEncoder instance or None, default=None
+        Custom ``OneHotEncoder`` instance to use. If ``None``, a default
+        ``OneHotEncoder(max_categories=max_categories, sparse_output=False,
+        handle_unknown="ignore")`` will be used.
+
+    target_encoder : TargetEncoder instance or None, default=None
+        Custom ``TargetEncoder`` instance to use. If ``None``, a default
+        ``TargetEncoder()`` will be used.
+
+    Attributes
+    ----------
+    one_hot_encoder_ : OneHotEncoder
+        The fitted ``OneHotEncoder`` instance.
+
+    target_encoder_ : TargetEncoder
+        The fitted ``TargetEncoder`` instance.
+
+    all_outputs_ : list of str
+        The list of feature names created by the transformer.
+
+    Examples
+    --------
+    >>> import pandas as pd
+    >>> from skrub._categorical_encoder import CategoricalEncoder
+    >>> s = pd.Series(["a", "b", "a", "c", "d", "e", "a", "b", "c", "d"], name="col")
+    >>> y = pd.Series([1, 0, 1, 0, 1, 0, 1, 0, 1, 0])
+    >>> enc = CategoricalEncoder(max_categories=3)
+    >>> enc.fit_transform(s, y)
+       col_a  col_d  col_infrequent_sklearn  col
+    0    1.0    0.0                     0.0  ...
+    1    0.0    0.0                     1.0  ...
+    """
+
+    def __init__(self, max_categories=10, one_hot_encoder=None, target_encoder=None):
+        self.max_categories = max_categories
+        self.one_hot_encoder = one_hot_encoder
+        self.target_encoder = target_encoder
+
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by jeromedockes** [comment]
thanks a lot @faithlesstomas !!

for context, @faithlesstomas started working on this as part of the EuroScipy 2026 sprint.

(just so you know a few maintainers including myself are currently on vacation so review may be a bit slower than usual)

**Comment by lisaleemcb** [comment]
Hi @faithlesstomas , nice to see you again. :) It looks like your test coverage is low (~75%). While you wait for some more comments, perhaps you could update your tests to expand coverage?

**Comment by lisaleemcb** [comment]
Hi @faithlesstomas! Thanks for the updated tests. A couple more comments:

-  I think for now we don't need to the option of passing custom encoders, so maybe remove them for now?
- After some discussion, we think that shortening the name to CatEncoder would make it a little easier to type, so could you go ahead and apply that name change?

Also, there's a merge conflict in the changelog that needs resolving. Thanks again for all your work on this!

File: skrub/_categorical_encoder.py [hunk 1]
```diff
@@ -0,0 +1,244 @@
...
+
+        y : Pandas or Polars Series, DataFrame, or array-like
+            Target values for target encoding.
+
+        Returns
+        -------
+        res_df : Pandas or Polars DataFrame
+            DataFrame containing one-hot and target-encoded features.
+        """
+        if y is None:
```
rcap107, line 99: this check can be removed by making y a required parameter

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+        else:
+            self.one_hot_encoder_ = clone(self.one_hot_encoder)
+
+        if self.target_encoder is None:
+            self.target_encoder_ = TargetEncoder()
+        else:
+            self.target_encoder_ = clone(self.target_encoder)
+
+        col_name = sbd.name(column) or "categorical_enc"
+        X_pandas = sbd.to_pandas(column).to_frame()
```
rcap107, line 117: this functionality should be implemented in both pandas and polars (see https://github.com/skrub-data/skrub/pull/2253 for more context on how dispatched functions work)

converting to pandas from polars requires pyarrow which isn't a required dependency, so it should be avoided unless there's an alternative

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+            raise ValueError(
+                "CategoricalEncoder expects y to be one-dimensional or a "
+                f"single-column dataframe; got an array with shape {y_vec.shape}."
+            )
+
+        if y_vec.dtype == object:
+            try:
+                y_vec = y_vec.astype(float)
+            except (ValueError, TypeError):
+                pass
```
rcap107: Rather than converting the dataframe/series to numpy array, I would instead flip the conversion: if the user provides a np array, convert it to a dataframe with the same type as what is used by X, then use ToFloat transformer (https://skrub-data.org/dev/reference/generated/skrub.ToFloat.html#skrub.ToFloat) to check for size and type. 

ToFloat raises a ValueError if it's passed a dataframe with more than 1 dimension, and a RejectColumn error if the column cannot be converted to float, and converts it to float otherwise. 

I think this would simplify the code, but maybe there is a good reason to not do that. WDYT @jeromedockes

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+            [
+                "one_hot_encoder_",
+                "target_encoder_",
+                "one_hot_outputs_",
+                "target_outputs_",
+                "all_outputs_",
+            ],
+        )
+
+        X_pandas = sbd.to_pandas(column).to_frame()
```
rcap107, line 190: Same comment about converting to pandas here

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+        te_res = self.target_encoder_.fit_transform(X_pandas, y_vec)
+
+        if hasattr(ohe_res, "toarray"):
+            ohe_res = ohe_res.toarray()
+
+        self.one_hot_outputs_ = list(
+            self.one_hot_encoder_.get_feature_names_out([col_name])
+        )
+        target_outputs = list(self.target_encoder_.get_feature_names_out([col_name]))
+        self.target_outputs_ = _make_target_names_unique(
```
rcap107, line 158: this function could be replaced by the function pick_column_names in skrub/_join_utils.py

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+        return self._make_output(column, ohe_res, te_res)
+
+    def _make_output(self, column, ohe_res, te_res):
+        """Build the output without coercing the encoders' individual dtypes."""
+        ohe_df = sbd.make_dataframe_like(
+            column, dict(zip(self.one_hot_outputs_, ohe_res.T))
+        )
+        ohe_df = sbd.copy_index(column, ohe_df)
+
+        te_df = sbd.make_dataframe_like(
```
rcap107, line 207: I think it would be better to rename the column that is generated here to something like "{col}_target_sklearn", because otherwise there would be a column with the same name as the input column

in some cases it may be useful to keep the original column (it's a parameter in ApplyToCols) and keeping the original name would be confusing

skrub/tests/test_categorical_encoder.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+def test_categorical_encoder_y_none():
+    s = pd.Series(["a", "b", "a"], name="col")
+    enc = CategoricalEncoder()
+    with pytest.raises(ValueError, match="Target y must be provided"):
+        enc.fit_transform(s, y=None)
+
+
+def test_categorical_encoder_custom_encoders():
```
rcap107, line 61: this test should use df_module

skrub/tests/test_categorical_encoder.py
```diff
@@ -0,0 +1,250 @@
...
+    ]
+    assert res.shape == (15, 5)
+    assert enc.target_encoder_.target_type_ == "multiclass"
+
+
+@pytest.mark.parametrize(
+    "y, expected_message",
+    [
+        (
+            pd.DataFrame({"first": [0, 1] * 5, "second": [1, 0] * 5}),
```
rcap107, line 100: this also needs df_module

skrub/tests/test_categorical_encoder.py
```diff
@@ -0,0 +1,250 @@
...
+            "exactly one column",
+        ),
+        (np.ones((10, 2)), "one-dimensional"),
+        (np.asarray(1), "one-dimensional"),
+    ],
+)
+def test_categorical_encoder_rejects_non_1d_target(y, expected_message):
```
rcap107: this test needs to be updated after replacing the logic with ToFloat

skrub/tests/test_categorical_encoder.py
```diff
@@ -0,0 +1,250 @@
...
+
+    fitted = enc.fit_transform(s, y)
+    transformed = enc.transform(s)
+
+    for name in enc.all_outputs_:
+        assert sbd.dtype(sbd.col(fitted, name)) == sbd.dtype(sbd.col(transformed, name))
+    assert sbd.to_numpy(sbd.col(fitted, "col_a")).dtype == np.float32
+
+
+def test_categorical_encoder_stable_names_on_collision(df_module):
```
rcap107, line 159: what are you testing here?

skrub/tests/test_categorical_encoder.py
```diff
@@ -0,0 +1,250 @@
...
+)
+def test_categorical_encoder_rejects_non_1d_target(y, expected_message):
+    s = pd.Series(["a", "b"] * 5, name="col")
+
+    with pytest.raises(ValueError, match=expected_message):
+        CategoricalEncoder().fit_transform(s, y)
+
+
+def test_categorical_encoder_2d_string_target_and_sparse_output():
+    s = pd.Series(["a", "b", "c"] * 5, name="col")
```
rcap107, line 115: this also needs df_module

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+    """
+
+    def __init__(
+        self,
+        max_categories=10,
+        one_hot_encoder=None,
+        target_encoder=None,
+    ):
+        self.max_categories = max_categories
+        self.one_hot_encoder = one_hot_encoder
```
rcap107, line 80: for the first version of this transformer, we thought it would be better to start with a more basic version that does not allow to use custom transformers, so the one_hot_encoder and target_encoder parameters should be removed

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+
+        if self.target_encoder is None:
+            self.target_encoder_ = TargetEncoder()
+        else:
+            self.target_encoder_ = clone(self.target_encoder)
+
+        col_name = sbd.name(column) or "categorical_enc"
+        X_pandas = sbd.to_pandas(column).to_frame()
+
+        if sbd.is_dataframe(y):
```
rcap107, line 119: I think this section would be more readable with a refactoring like

```python
def check_y(col):
	if isinstance(col, np.array):
		y_col = col
	else:
		# this raises an exception if col isn't a series or dataframe
		y_col = sbd.to_numpy(col)

	if y_col.ndim == 2 and y_col.shape[1] == 1:
		y_vec = y_col ...
	elif y_col.ndim != 1:
		raise ...

```

skrub/_categorical_encoder.py
```diff
@@ -0,0 +1,244 @@
...
+
+        if y_vec.ndim == 2 and y_vec.shape[1] == 1:
+            y_vec = y_vec[:, 0]
+        elif y_vec.ndim != 1:
+            raise ValueError(
+                "CategoricalEncoder expects y to be one-dimensional or a "
+                f"single-column dataframe; got an array with shape {y_vec.shape}."
+            )
+
+        if y_vec.dtype == object:
```
rcap107, line 142: I don't think this part is needed, the TargetEncoder should handle non-numeric targets on its own

rcap107 (commented):
Hi @faithlesstomas, thanks a lot for this PR! I'm sorry it's been taking so long to provide feedback

I'm not the original author of the issue, so there may be some further discussion after this review, but I still wanted to give some info to work on while we wait for another eye

In general, I think the PR looks good, though the main change is a simplification of the code so that it is not possible to provide a custom OHE or TE. That will be a feature that will be implemented further down the line.

Aside from that, there are a few comments on clarity or some code changes. 

In any case, thanks for your work.

PR #2248, opened 2026-08-18
FIX - avoiding a deprecation warning for categorical data in Polars 2.0

rcap107:
This PR fixes a failing test that was caused by a deprecation warning for polars 2.0, which was added in #2233 

cat.get_categories will be removed, so this PR is replacing the line with `unique`.


lisaleemcb (approved):
Looks good! @rcap107

rcap107:
Thanks @lisaleemcb !

PR #2252, opened 2026-08-24
feat: add `accept_numeric` kwarg to `ToCategorical`

### LINKED ISSUE #2219: Change the behavior of ToCategorical so that it no longer rejects numeric features
### Problem Description

Currently, ToCategorical will not convert numerical features to categorical, rejecting them with a RejectColumn exception. This is intended behavior because in the Cleaner and TableVectorizer numeric features should be "left alone". However, since ToCategorical is now a public transformer, I think it's unintended behavior from a user's point of view: if they want to convert an integer column to categorical (e.g., because it's an integer index that should not be treated as a number), they should be allowed to do so. 

### Feature Description

ToCategorical should either have a parameter "accept_ints=True", or accept any number by default, so that whatever is the content of the column, the output is the column converted to categorical.

I would prefer the second solution, because I think it aligns better with what a user would want to do. 

### Alternative Solutions

_No response_

### Additional Context

This would be a breaking change in multiple ways.
The TableVectorizer and Cleaner now expect numerical features to be rejected by ToCategorical, so that behavior would need to be altered and adapted to the new data. 

Additionally, this is a breaking change because the behavior of ToCategorical would effectively "switch" from rejecting numeric columns to treating them as categorical. For this reason we would have to have two deprecation cycles: one to add some kind of "legacy behavior" and one to remove it. 

No PRs should be opened until we decide exactly how we want to deal with all of this.

lisaleemcb:
The encoder `ToCategorical` now also accepts type `int` columns, in addition to `str` and categorical dtypes via setting the new kwarg `accept_numeric`  to `'int'`.  In addition, `float` columns will be accepted if `accept_numeric='all'`. 

As currently implemented, the new default is `accept_numeric='int'` which means that columns of strings, categoricals, and ints will be marked as categorical after performing `fit_transform`. **Note** this is different from the current default behaviour, which only accepts strings, and categoricals, and will raise an error otherwise. The current default can be recovered by setting `accept_numeric=None`.

The rationale for the new default is that integers are used in many datasets to represent categories. But this would make it a breaking change for direct users, so perhaps we should discuss.

However, note that this change in the default would not affect the behaviour of the `TableVectorizer`, as the numerical columns are currently not selected to be sent to the `ToCategorical` encoder. Thanks to @rcap107 for beaucoup help with understanding its behaviour.

PS Also changed the docstring to make it more concise.

### DIFF
--- a/skrub/_to_categorical.py
+++ b/skrub/_to_categorical.py
@@ -9,7 +9,7 @@ class ToCategorical(SingleColumnTransformer):
     Convert a string column to Categorical dtype.
 
     This transformer ensures that a given string or categorical column has
-    Categorical dtype. This is done to mark columns to be treated as categorical
+    Categorical dtype so that it is treated as categorical
     by downstream transformers and learners.
 
     Notes
@@ -26,9 +26,14 @@ class ToCategorical(SingleColumnTransformer):
     a polars column with dtype ``String``, is converted to a categorical
     column. Categorical columns are passed through.
 
+    If ``accept_numeric`` is set to ``"all"``, then both integer and float
+    columns are accepted and converted to categorical. If it is set to ``"int"``,
+    then only integer columns are accepted. The default value is ``"int"``.
+
     Any other type of column is rejected by raising a ``RejectColumn``
     exception. **Note:** the ``TableVectorizer`` only sends string or
-    categorical columns to its ``low_cardinality_transformer``. Therefore it is
+    categorical columns to its ``low_cardinality_transformer``, regardless
+    of the inputted value of ``accept_numeric``. Therefore it is
     always safe to use a ``ToCategorical`` instance as the
     ``low_cardinality_transformer``.
 
@@ -88,6 +93,16 @@ class ToCategorical(SingleColumnTransformer):
         ...
     skrub.core.RejectColumn: Column 'c' does not contain strings.
 
+    Unless ``accept_numeric`` is set to ``"int"`` or ``"all"``, in which case integer
+    columns are accepted in the former case, and both integer and float columns are
+    accepted in the latter:
+
+    >>> to_cat.fit_transform(pd.Series([1.1, 2.2], name='c'), accept_numeric="all")
+    0    1.1
+    1    2.2
+    Name: c, dtype: category
+    Categories (2, float64): [1.1, 2.2]
+
     ``object`` columns that do not contain only strings are also rejected:
 
     >>> s = pd.Series(['one', 1], name='c')
@@ -143,13 +158,19 @@ class ToCategorical(SingleColumnTransformer):
     True
     """
 
-    def fit_transform(self, column, y=None):
+    def fit_transform(self, column, accept_numeric="int", y=None):
         """Fit the encoder and transform a column.
 
         Parameters
         ----------
         column : pandas or polars Series
             The input to transform.
+
+        accept_numeric : str, default="int"
+            How to handle numeric columns. If "int", will convert integer
+            columns to categorical. If "all", both float and integer columns
+            will be accepted. If None, no numeric
+            columns will be accepted.
 
         y : None
             Ignored.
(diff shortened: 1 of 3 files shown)


skrub/_to_categorical.py
```diff
@@ -143,14 +158,20 @@ class ToCategorical(SingleColumnTransformer):
...
-    def fit_transform(self, column, y=None):
+    def fit_transform(self, column, accept_numeric="int", y=None):
         """Fit the encoder and transform a column.
 
         Parameters
         ----------
         column : pandas or polars Series
             The input to transform.
 
+        accept_numeric : str, default="int"
```
rcap107: This should be moved to the init of the class with the relative docstring because it's one of the requirements of scikit-learn transformers

```diff
diff --git a/skrub/_to_categorical.py b/skrub/_to_categorical.py
index de42d0683..1f17803fc 100644
--- a/skrub/_to_categorical.py
+++ b/skrub/_to_categorical.py
@@ -158,7 +158,10 @@ class ToCategorical(SingleColumnTransformer):
     True
     """
 
-    def fit_transform(self, column, accept_numeric="int", y=None):
+    def __init__(self, accept_numeric="int"):
+        self.accept_numeric = accept_numeric
+
+    def fit_transform(self, column, y=None):
         """Fit the encoder and transform a column.
 
         Parameters
@@ -166,12 +169,6 @@ class ToCategorical(SingleColumnTransformer):
         column : pandas or polars Series
             The input to transform.
 
-        accept_numeric : str, default="int"
-            How to handle numeric columns. If "int", will convert integer
-            columns to categorical. If "all", both float and integer columns
-            will be accepted. If None, no numeric
-            columns will be accepted.
-
         y : None
             Ignored.
 ```

File: skrub/_to_categorical.py [hunk 2]
```diff
@@ -163,9 +184,17 @@ def fit_transform(self, column, y=None):
 
         if sbd.is_categorical(column):
             return column
-        if not sbd.is_string(column):
-            raise RejectColumn(f"Column {sbd.name(column)!r} does not contain strings.")
-        return sbd.to_categorical(column)
+        elif sbd.is_string(column):
```
rcap107: This set of if-else conditions can be simplified (a bit) by combining the conditions that return sbd.to_categorical. It's not a big simplification, but it does save some duplication. 

```diff
@@ -182,19 +179,19 @@ class ToCategorical(SingleColumnTransformer):
         """
         self.all_outputs_ = [sbd.name(column)]
 
+
         if sbd.is_categorical(column):
             return column
-        elif sbd.is_string(column):
-            return sbd.to_categorical(column)
-        elif sbd.is_integer(column) and accept_numeric in ("int", "all"):
-            return sbd.to_categorical(column)
-        elif sbd.is_float(column) and accept_numeric == "all":
+        if (
+            sbd.is_string(column)
+            or (sbd.is_integer(column) and self.accept_numeric in ("int", "all"))
+            or (sbd.is_float(column) and self.accept_numeric == "all")
+        ):
             return sbd.to_categorical(column)
-        else:
-            raise RejectColumn(
-                f"Column {sbd.name(column)!r} does not contain strings "
-                "or numerical data (if accept_numeric is 'int' or 'all')."
-            )
+        raise RejectColumn(
+            f"Column {sbd.name(column)!r} does not contain strings "
+            "or numerical data (if accept_numeric is 'int' or 'all')."
+        )
 
     def transform(self, column):
         """Transform a column.
```

skrub/tests/test_to_categorical.py
```diff
@@ -13,9 +13,21 @@ def test_to_categorical(df_module):
     # categorial columns are accepted
     assert ToCategorical().fit_transform(out) is out
     assert ToCategorical().fit(out).transform(out) is out
-    # non-string, non-categorical columns are rejected
+    # default behaviour accepts integer and string
+    # columns, but not float
+    i = df_module.make_column("c", [1, 2, None])
```
rcap107, line 21: here the asserts should check for equality with an expected column, which you can get by doing

```python
expected = sbd.to_categorical(i)
df_module.assert_column_equal(transformed, expected)
```

that way we know both that the code ran correctly, and that the output is what we're expecting to see

skrub/tests/test_to_categorical.py
```diff
@@ -13,9 +13,21 @@ def test_to_categorical(df_module):
...
+    assert ToCategorical().fit_transform(i, accept_numeric="int")
+    assert ToCategorical().fit(i).transform(i, accept_numeric="int")
+    # unless accept_numeric is None, in which case
+    # only string and categorical columns are accepted
+    with pytest.raises(RejectColumn, match=".*does not contain strings or*"):
```
rcap107: expected failures should be wrapped in their own test, and the test should be parametrized for float and int cases

rcap107 (commented):
Thanks for the PR @lisaleemcb! There are still a few things that we need to clarify, but here is an early review with some comments

skrub/tests/test_to_categorical.py
lisaleemcb: @rcap107 thanks for all the suggestions! I learned a lot. I think I implemented most of your suggestions.

lisaleemcb:
Note this PR is still failing because of the test case where the integer column contains None/null values. This desired behaviour in this situation is still tbd, and should be discussed.

rcap107 (approved):
Looks good to me, thanks @lisaleemcb !

PR #2253, opened 2026-08-26
DOC - adding dev documentation about the dataframe API and df_module

rcap107:
This PR is adding documentation about the dispatched dataframe API and the df_module and testing framework. 

For the moment, they have not been added to the rendered documentation, so they're only available from inside the repository. We can decide to put them on the website if needed. 

I asked an agent to generate the bulk of the material, then I reviewed it to remove/rephrase some of the content. I'll leave this in draft mode to get a few more eyes on and trim it more if necessary.


rcap107:
Somewhat unrelated, we currently don't have a dedicated place that we use to keep developer-specific documentation, so we might want to rework a bit the material so that the new contributor guide and the release guide are in the same place as the pages added in this PR, though that can be in a separate PR.

doc/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+* It must handle real user data — arbitrary DataFrames that arrive from outside
+  skrub.
+
+
+``df_module``: testing library-agnostic code
+---------------------------------------------
+
+``df_module`` solves a different problem: when a test needs to construct
+inputs, call a function, and check the output, it must do so in a way that
+works for all three configurations.  ``df_module`` provides a uniform
```
lisaleemcb, line 82: configurations is sort of vague here. maybe something about the 3 supported dataframe types?

It specifies it further below but I think it could be moved up to make it clearer.

doc/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+  implementation, invoked automatically.
+
+In the **test fixture**:
+
+* ``df_module`` — generic, runs for every configuration.
+* ``pd_module`` / ``pl_module`` — fixed to one backend, used when the test
+  itself is backend-specific.
+
+The design principle is the same in both cases: write the general case once
+and isolate backend differences to dedicated, clearly labelled places.
```
lisaleemcb, line 208:
```suggestion
and isolate backend differences to dedicated, clearly labeled places.
```

doc/development/dataframe_api.rst
```diff
@@ -0,0 +1,358 @@
+.. _dev_dataframe_api:
+
+The dispatch-based dataframe API
+=================================
+
+skrub targets both pandas and polars as first-class backends.  Rather than
+scattering ``if pandas … else polars …`` branches throughout the codebase, all
+dataframe and column operations are funnelled through a thin dispatch layer that
```
lisaleemcb, line 8:
```suggestion
dataframe and column operations are funneled through a thin dispatch layer that
```

doc/development/testing.rst
```diff
@@ -0,0 +1,329 @@
...
+twice.
+
+
+Anatomy of ``df_module``
+-------------------------
+
+``df_module`` is defined in ``skrub/conftest.py`` and returns a
+:class:`types.SimpleNamespace` with a consistent set of attributes.  The
+attributes are designed to normalise the differences between libraries so test
+bodies need no ``if pandas / if polars`` branches (with few exceptions).
```
lisaleemcb, line 54:
```suggestion
bodies need no ``if pandas / if polars`` branches (with a few exceptions).
```

doc/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+| **When it runs**          | Production — at import time and      | Test time — under pytest only      |
+|                           | when skrub functions are called      |                                    |
++---------------------------+--------------------------------------+------------------------------------+
+| **What it abstracts**     | *How* to perform an operation        | *How* to construct inputs and      |
+|                           | (fill nulls, get shape, cast, …)     | assert outputs in a test           |
++---------------------------+--------------------------------------+------------------------------------+
+| **Who uses it**           | skrub transformers, encoders,        | Test functions                     |
+|                           | utility functions                    |                                    |
++---------------------------+--------------------------------------+------------------------------------+
+| **Mechanism**             | ``functools.singledispatch`` +        | pytest ``@fixture(params=…)``      |
```
emassoulie, line 36:
```suggestion
| **Mechanism**             | ``functools.singledispatch`` +       | pytest ``@fixture(params=…)``      |
```

doc/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+|                           | ``specialize`` decorator             |                                    |
++---------------------------+--------------------------------------+------------------------------------+
+| **Result of abstraction** | A single call site (``sbd.fill_nulls``) | A single test body runs 3 times |
```
emassoulie, line 39:
```suggestion
| **Result of abstraction** | A single call site                   | A single test body runs 3 times    |
|                           | (``sbd.fill_nulls``)                 |                                    |
```

doc/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+|                           | ``skrub/_dispatch.py``               |                                    |
++---------------------------+--------------------------------------+------------------------------------+
+| **When it runs**          | Production — at import time and      | Test time — under pytest only      |
+|                           | when skrub functions are called      |                                    |
++---------------------------+--------------------------------------+------------------------------------+
+| **What it abstracts**     | *How* to perform an operation        | *How* to construct inputs and      |
+|                           | (fill nulls, get shape, cast, …)     | assert outputs in a test           |
++---------------------------+--------------------------------------+------------------------------------+
+| **Who uses it**           | skrub transformers, encoders,        | Test functions                     |
+|                           | utility functions                    |                                    |
```
emassoulie, line 34: I'm simplifying a bit, but these three lines seem to agressively repeat that `df_module` is for testing and the API for all the rest. This could be consolidated into a single line ("when it runs" for instance?), and the extra information in "what it abstracts" can probably be reformulated or merged with the "mechanism" line.

doc/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
++-----------------------------------------------+-----------------------------------+
+| I am unsure whether a new operation belongs   | If other transformers would       |
+| in ``_common.py`` or should be local.         | benefit from it: ``_common.py``.  |
+|                                               | If it is specific to one class:   |
+|                                               | local ``@dispatch``.              |
++-----------------------------------------------+-----------------------------------+
+
+A concrete heuristic: if you are writing code that will run when a user calls
+``TableVectorizer().fit_transform(df)``, use the dataframe API.  If you are
+writing code that only runs under ``pytest``, use ``df_module``.
```
emassoulie, line 186: In a more extreme version of my earlier comment, this could be easily be boiled down to "use df_module iff you're writing a test" with little loss of information. Does this table need to be as extensive, or can we simply keep the "concrete heuristic" summary (or even remove the section entirely)? The main benefit this table brings, giving concrete examples of API/df_module usage and introducing pd/pl_module, could probably be factored elsewhere.

doc/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+
+
+Parallel structure
+------------------
+
+Despite their different purposes, the two layers do mirror each other in one
+respect: both have a "all backends" path and a "specific backend" escape hatch.
```
emassoulie, line 193: Small detail, but should the main takeaway from this section really be "here is one point on which both structures are similar"? A name like "selecting specific backends in each mechanism" (or a more elegant version) feels more to-the-point.

doc/development/testing.rst
```diff
@@ -0,0 +1,329 @@
...
+.. code-block:: python
+
+    def test_column_names(df_module):
+        df = df_module.make_dataframe({"a": [1, 2, 3], "b": [4, 5, 6]})
+        names = sbd.column_names(df)
+        assert "a" in names
+        assert "b" in names
+
+If multiple types are required for a given test, then ``example_dataframe`` can
+be used to avoid boilerplate:
```
emassoulie, line 216: I don't see the connection between this heading and the example below: is "multiple types required for a given test" a common hurdle in testing? And how does checking the presence of specific columns in the `example_dataframe` illustrate `df_module` overcoming that hurdle?

skrub/_docs/development/testing.rst
```diff
@@ -0,0 +1,329 @@
...
+.. code-block:: python
+
+    def test_column_names(df_module):
+        df = df_module.make_dataframe({"a": [1, 2, 3], "b": [4, 5, 6]})
+        names = sbd.column_names(df)
+        assert "a" in names
+        assert "b" in names
+
+If multiple types are required for a given test, then ``example_dataframe`` can
+be used to avoid boilerplate:
```
rcap107: The section is about defining example dataframes, which is something that can be done either by using make_dataframe or by using the example_dataframe

Having to test multiple dtypes is a fairly common occurrence, but for those tests it's usually more convenient to just define a specific dataframe with make_dataframe rather than relying on the example_dataframe

I agree that the example with example_dataframe isn't particularly meaningful, maybe we can just remove it.

doc/development/api_vs_testing.rst
rcap107: I removed one of the lines, but I still think that "when it runs" and "what it abstracts" are distinct enough to keep both

skrub/_docs/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
++-----------------------------------------------+-----------------------------------+
+| I am unsure whether a new operation belongs   | If other transformers would       |
+| in ``_common.py`` or should be local.         | benefit from it: ``_common.py``.  |
+|                                               | If it is specific to one class:   |
+|                                               | local ``@dispatch``.              |
++-----------------------------------------------+-----------------------------------+
+
+A concrete heuristic: if you are writing code that will run when a user calls
+``TableVectorizer().fit_transform(df)``, use the dataframe API.  If you are
+writing code that only runs under ``pytest``, use ``df_module``.
```
rcap107: I shortened a bit the table because some of the lines were a bit too specific, but aside from that I think it's a convenient checklist for contributors as most of the points are decisions that come up frequently. Having everything in one spot makes it more convenient.

skrub/_docs/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+
+
+Parallel structure
+------------------
+
+Despite their different purposes, the two layers do mirror each other in one
+respect: both have a "all backends" path and a "specific backend" escape hatch.
```
rcap107: Good point, honestly I think we can remove this entire section

skrub/_docs/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
+.. _dev_api_vs_testing:
+
+Two layers, one purpose: the dispatch API and ``df_module``
```
jeromedockes, line 3: Thanks @rcap107 ! I think this will be very helpful to onboard new contributors and there is good advice in here. 

My main comment is I haven't read everything but I think this doc is much too long and the useful parts might not get read because of that. also I think it answers questions that do not make too much sense, eg the parallel between the library's dispatch mechanism and the pytest fixture. 

there should not be any discussion of design decisions, nor any explanation of how the dispatch function works. we only need a very short summary of how it can be _used_ for new contributors. I would recommend 2 sections

- how to write code that works on both pandas and polars dataframes
  - using the functions in skrub._datatrame
  - for complex needs: define your own dispatched function
- how to write tests that run on pandas and polars dataframes using df_module

each no longer than a page, ideally shorter.

also there should not be a list of "available functions" which will get outdated and is already available just by looking at the module

jeromedockes: also there are many examples in skrub code itself, so we just need a few pointers to help contributors get started and have a general idea of what is happening, and for details and conventions they can look at existing code + ask on github

rcap107: Noted, I'll shorten those sections. @emassoulie gave similar suggestions too, I just need to implement them. 

I'll also add more references to places in the code that do certain things.

rcap107:
I ended up adding the pages to the documentation to see how they would render. 

I want to do a bunch more editing before merging this, also based on feedback.

skrub/_docs/development/api_vs_testing.rst
```diff
@@ -0,0 +1,208 @@
...
+| **What it abstracts**     | *How* to perform an operation        | *How* to construct inputs and      |
+|                           | (fill nulls, get shape, cast, …)     | assert outputs in a test           |
++---------------------------+--------------------------------------+------------------------------------+
+| **Who uses it**           | skrub transformers, encoders,        | Test functions                     |
+|                           | utility functions                    |                                    |
++---------------------------+--------------------------------------+------------------------------------+
+| **Mechanism**             | ``functools.singledispatch`` +        | pytest ``@fixture(params=…)``      |
+|                           | ``specialize`` decorator             |                                    |
++---------------------------+--------------------------------------+------------------------------------+
+| **Result of abstraction** | A single call site (``sbd.fill_nulls``) | A single test body runs 3 times |
```
rcap107: as it turns out, having the same column width is just cosmetic for the rendering, but sphinx does complain if the total length of the borders (`+---+` etc) is not consistent

skrub/_docs/development/api_vs_testing.rst
rcap107: removed

PR #2254, opened 2026-08-27
MAINT - Rename textencoder

rcap107:
This PR renames the TextEncoder to LLMEncoder. 

I renamed all instances of TextEncoder (with the exception of what's in the changelog), added a deprecation warning and the relative test.

Since I had to modify the content of the `_text_encoder.py` file, I thought it would be better to first merge this PR and then open a separate PR to rename the file, to keep the diff simple.


rcap107:
Closed because I messed up the git history

PR #2255, opened 2026-08-27
Renaming TextEncoder -> LLMEncoder

rcap107:
This PR renames the TextEncoder to LLMEncoder. 

I renamed all instances of TextEncoder (with the exception of what's in the changelog), added a deprecation warning and the relative test.

Since I had to modify the content of the `_text_encoder.py` file, I thought it would be better to first merge this PR and then open a separate PR to rename the file, to keep the diff simple.


skrub/_text_encoder.py
```diff
@@ -452,3 +452,25 @@ def get_feature_names_out(self, input_features=None):
...
+    def __init__(self, *args, **kwargs):
+        import warnings
+
+        warnings.warn(
+            "TextEncoder is deprecated and will be removed in a future release. "
+            "Use LLMEncoder instead.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
+        super().__init__(*args, **kwargs)
```
GaelVaroquaux: A better way of doing this is to use the deprecated decorator of scikit-learn: https://scikit-learn.org/stable/modules/generated/sklearn.utils.deprecated.html
Something like:

```suggestion
@deprecated("TextEncoder has been renamed to LLMEncoder and will be removed in version 0.12.")
class TextEncoder(LLMEncoder):
     pass
```
I believe that this is important to avoid the use of "*args, **kwargs", which don't play well with our detection of arguments for get_params/set_params

GaelVaroquaux (changes requested):
One small comment, but that may be of importance

GaelVaroquaux (approved):
Thanks! LGTM, merging

PR #2267, opened 2026-09-03
DOC - add proper `meta` tags in the API pages

rcap107:
This PR modifies how API pages are rendered by sphinx so that their HTML includes `meta` tags that contain the first line of the docstrings as description.

This is because search engines rely on meta tags to describe the content of a page as their search summary (see https://developers.google.com/search/docs/appearance/snippet). 

Without this PR, the description was 

```
content="Gallery examples: Getting Started with skrub Quick overview of DataOps Hands-On with Column Selection and Transformers SquashingScaler: Robust numerical preprocessing for neural networks Sessions i..."
```

which is not useful at all, and is indeed the same description for all pages.


rcap107:
closing because the history was messed up

PR #2269, opened 2026-09-04
DOC - adding a page about common mistakes

rcap107:
This is a work in progress PR that should address a problem we have been discussing for a long time, that is providing users with a page that explains common mistakes and pitfalls. 

At the moment, it's in the how-to's section, but I am not entirely convinced it's the best place. Maybe it should be at the end of the user guide? 

Also, I haven't properly reviewed the actual content of the solutions. This is more of a POC to see what the page would look like and where to put it.

### DIFF
--- a/doc/guides/common_errors.rst
+++ b/doc/guides/common_errors.rst
@@ -1,0 +1,270 @@
+.. currentmodule:: skrub
+
+.. _common_errors:
+
+Common errors and gotchas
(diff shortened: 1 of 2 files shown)


rcap107:
From IRL discussion. 

The page should stay in how-to (no need for a new tab), but the title of the page should be changed. The template is also fine.

PR #2274, opened 2026-09-07
MAINT - move the docs folder to the skrub folder

rcap107:
This will be a heavy PR. The idea is that we should move most of the docs folder (user guide, how-tos, examples) to the skrub/ folder.

The main advantage is that by doing this the docs will be bundled with the wheel, so that whenever the package is installed docs will be present in the same place as the code. This should be especially useful for agents that install skrub as this should make it easier for them to understand what the library is actually doing, and thus produce better code. 

The main problem is that means moving most (not all!) the doc files, which of course will lead to a bunch of conflicts and a huge diff to go through. 

My hope is to be able to get this done by the time we want to release 0.11, but I am not entirely sure we'll manage to get everything done by then.


jeromedockes (approved):
thanks @rcap107 !

Issue #2116, opened 2026-06-02
doc - correct redirection link


MarieSacksick:
### Describe the issue linked to the documentation

In [contributing guide](https://skrub-data.org/stable/CONTRIBUTING.html#setting-up-the-environment), the link to "install from source" redirects to installation page, but not to the tab "from source". We should correct that to avoid useless clic + misunderstanding.  

### Suggest a potential alternative/fix

I suggest correcting the link to https://skrub-data.org/stable/install.html#advanced-usage-for-contributors instead of https://skrub-data.org/stable/install.html#installing-from-source.

jeromedockes:
thanks @MarieSacksick . unfortunately at least for me on firefox the link doesn't actually open the right tab. anyway it goes much beyond just installation, so I would suggest making the from source tab much shorter (just showing `pip install -e .`), and move all the content & explanations about git etc directly into the contributing guide

MarieSacksick:
Arg. A bit less straightforward then, but agree with your solution, let's do it this way!

MarieSacksick:
And in that case it means that this issue should be merged with https://github.com/skrub-data/skrub/issues/2117. Otherwise it's highway to conflicts.  
Therefore closing the other one + copy/pasting it here:  

### Describe the issue linked to the documentation

In [this part of installation guide](https://skrub-data.org/stable/install.html#run-the-tests), it states 
> If you see some warnings like:
UserWarning: Only pandas and polars DataFrames are supported, but input is a Numpy array. Please convert Numpy arrays to DataFrames before passing them to skrub transformers. Converting to pandas DataFrame with columns ['0', '1', …].
  warnings.warn(
This is expected, and you may proceed with the next steps without worrying about them. However, no tests should fail at this point: if they do fail, then let us know.

This is not relevant anymore since the fix done [here](https://github.com/skrub-data/skrub/issues/1828). 

### Suggest a potential alternative/fix

Remove that part.

vedant1100:
I would like to claim this issue.

rcap107:
> I would like to claim this issue.

Please to, thanks!

PR #2276, opened 2026-09-08
Doc correct redirection link

vedant1100:
# Documentation fix pull request

## Description

Addresses #2116 

In [Contrubuting Guide](https://skrub-data.org/stable/CONTRIBUTING.html#setting-up-the-environment), "Install" link redirects to "From Source" by default.


rcap107:
Hi @vedant1100, with your proposed solution the installation instructions would always point at the tab about installing from source: that's not what we want, because most users are not interested in that. 

What we would rather have is a bigger rework of the content of the "from source" tab: at the moment, it's way longer than it needs to be, and most of the information there is only relevant for people that want to contribute to the package. 

The updated version of the "From source" page in the installation page should just explain how to clone from the github repository, and how to install the package in editable mode from within a virtual env (`pip install -e .`). 

Then, the rest of the text should be moved to the CONTRIBUTING.rst file, in the section named "Setting up the environment". 

That way, contributor information is together with the rest, and the installation instructions contain only what's required to run the package. 

Finally, a major rework of the documentation was merged in the main repository, so I think that in the interest of avoiding conflicts it may be easier to close this PR and open a new one to start from a clean slate.

Does all this make sense to you?

vedant1100:
Ok. I will try to incorporate new changes and start a new PR.

vedant1100:
"Deep learning dependencies" are present in every page. Should I keep it in "from source" page at the bottom or move it to CONTRIBUTING.rst file?

rcap107:
> "Deep learning dependencies" are present in every page. Should I keep it in "from source" page at the bottom or move it to CONTRIBUTING.rst file?

Those should stay where they are because they are important for the regular installation.

PR #2277, opened 2026-09-08
DOC - expanding the docstring of data ops

rcap107:
Small change to the docstring to make the intent of data ops more evident, and adding crossref to the docs.


GaelVaroquaux (approved):
LGTM, merging

PR #2278, opened 2026-09-09
MAINT - adding plausible analytics

rcap107:
This PR adds a Plausible script to pages to track analytics on the skrub website. 

I added the same line that's used in scikit-learn's configuration to conf.py: https://github.com/scikit-learn/scikit-learn/blob/2cc7893a2097be891426ca6a4d9ea2e14cf82b9c/doc/conf.py#L233

Building the docs locally shows the script embedded in the head of all pages in the documentation, but I don't know if merging this in main will be enough to trigger the tracking, or if we need to to a full release to rebuild the entire website. 

Paging @ogrisel and @lesteve to know if this is all that's done on scikit-learn's side

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -252,6 +252,10 @@ html_theme_options = {
     #         "name": "PyData",
     #     },
     # ],
+    "analytics": {
+        "plausible_analytics_domain": "skrub-data.org",
+        "plausible_analytics_url": "https://views.scientific-python.org/js/script.js",
+    },
     "header_links_before_dropdown": 5,
     "icon_links": [
         {


jeromedockes (approved):
thanks, this will be very interesting!

jeromedockes:
if needed we can push a commit to the 0.10.X branch with the small doc change without needing to actually do a release (new wheel on pypi + version change)

rcap107:
> if needed we can push a commit to the 0.10.X branch with the small doc change without needing to actually do a release (new wheel on pypi + version change)

I think it's worth doing

lesteve:
I guess you know this but you can view the stats here: https://views.scientific-python.org/skrub-data.org

This seems to be working AFAICT :tada:.

rcap107:
> I guess you know this but you can view the stats here: [views.scientific-python.org/skrub-data.org](https://views.scientific-python.org/skrub-data.org)
> 
> This seems to be working AFAICT 🎉.

It is working! Only on the dev website though

@jeromedockes we should push this to the X branch then

jeromedockes:
thanks @lesteve !

Issue #2273, opened 2026-09-07
MAINT - bump min version of python and some packages for 0.11


rcap107:
With release 0.11 we should bump the minimum requirements of a few packages

- The min python version should be increased to 3.10
- The min version of scikit-learn should be increased to 1.5.2
- The min version of matplotlib can be increased from 3.6 to 3.9
- Scipy can be bumped from 1.9 to 1.14	
- Joblib could go from 1.4.0 to 1.4.2
- Pandas 2.0 has now been out for more than 2 years (but I understand that we won't be able to drop 1.5.3 forever more)
- Polars could go to 1.7 or 1.8. The main PR that is still blocked by a polars release is #1871, which needs 1.29 (released in December 2024)

The first two are the main things that need to be updated, for the others I just checked the release history to see which versions are around 2 years old and listed them there. Before dropping support we should look up which versions of each package are the most used and decide accordingly (hello pandas 1.5.3, installed on millions of legacy machines)

rcap107:
In general, no need to bump dependencies unless doing so makes our job easier. This issue can still be kept open as a reminder of the versions that can be bumped.

rcap107:
Closing this, as mentioned we will be bumping the versions as needed.

PR #2280, opened 2026-09-10
MAINT - bumping minimum reqs for release 0.11

rcap107:
Increasing the min reqs for python to 3.11 and sklearn to 1.5.2

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,6 @@ classifiers = [
   "Environment :: Console",
   "Intended Audience :: Science/Research",
   "Operating System :: OS Independent",
-  "Programming Language :: Python :: 3.10",
   "Programming Language :: Python :: 3.11",
   "Programming Language :: Python :: 3.12",
   "Programming Language :: Python :: 3.13",
@@ -33,7 +32,7 @@ requires-python = ">=3.10"
 dependencies = [
   "numpy>=1.23.5",
   "pandas>=1.5.3",
-  "scikit-learn>=1.4.2",
+  "scikit-learn>=1.5.2",
   "scipy>=1.9.3",
   "jinja2>=3.1.2",
   "matplotlib>=3.6.1",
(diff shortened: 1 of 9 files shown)


jeromedockes:
thanks @rcap107 ! 

once we drop python 3.10 we should simplify by removing code that handles the case where exceptions don't have the `add_note` method. the corresponding tests can be made a bit more specific, too. it is probably better to do that in a separate PR

rcap107:
We may want to deprecate a few more dependencies according to #2273, but those are not as critical as sklearn and python

pyproject.toml
jeromedockes: should this change to 3.11?

rcap107: yes probably, missed that one

skrub/_data_ops/tests/test_data_ops.py
```diff
@@ -191,10 +191,7 @@ def test_becomes_default():
     # default
     assert c.skb.eval({"b": 20}) == 21
     # Whereas b is not
-    if sys.version_info < (3, 11):
-        err_t, err_msg = RuntimeError, "Evaluation of node <Var 'b'> failed"
-    else:
-        err_t, err_msg = KeyError, "No value has been provided for 'b'"
+    err_t, err_msg = KeyError, "No value has been provided for 'b'"
```
jeromedockes: we don't need to put them in variables anymore we can just use pytest.raises(KeyError, ...)

skrub/_data_ops/tests/test_errors.py
```diff
@@ -99,10 +99,7 @@ def test_unpacking_wrong_number_of_targets_at_runtime():
     # without a value for 'a' the length is only known when the plan runs
     a = skrub.var("a")
     first, _second = a
-    if sys.version_info < (3, 11):
-        err_t, err_msg = RuntimeError, "Evaluation of node <AsTuple"
-    else:
-        err_t, err_msg = ValueError, "too many values to unpack"
+    err_t, err_msg = ValueError, "too many values to unpack"
```
jeromedockes: same

jeromedockes (approved):

Issue #2220, opened 2026-07-16
META - Remove deprecated features


rcap107:
This issue should be kept open and updated as we deprecate and remove new features. The idea is that we can keep track of which features have been deprecated and which can be removed because the deprecation cycle is over. 

As of July 2026, we can remove any deprecated feature after a full release (so if something was deprecated in skrub 0.9, it can be removed in 0.11).

## Latest version
0.10.1

## Currently deprecated, to be removed
- [ ] SKRUB_DATA_DIRECTORY in `_config.py` 
- [ ] `max_plot_cols` and `max_association_columns`
- [ ] `drop_if_unique` parameter in `DropUninformative` and `Cleaner`
- [ ] the `splitter` parameter of `.skb.train_test_split` has been renamed to `split_func`

m4nn2609-dot:
i would like to take it on please

rcap107:
Hi @m4nn2609-dot, you may take this up but please only remove 
- Parameters numeric_dtype and drop_if_unique in the Cleaner
- Parameters max_association_columns and max_plot_columns in TableReport

m4nn2609-dot:
sure, @rcap107 i will keep that in mind.

rcap107:
Thank you!

PR #2281, opened 2026-09-10
Removed deprecated `how` parameter from `skb.apply`

emassoulie:
Addresses part of https://github.com/skrub-data/skrub/issues/2220

Every reference to the `how` parameter has been removed, and every piece of logic depending on its value adjusted.


jeromedockes (approved):
LGTM, thanks!

PR #2284, opened 2026-09-14
Rearranging sections in the homepage

emassoulie:
The first two "example" rows of the homepage have been switched, as have their respective columns so the top-left of the page remains a piece of text.


rcap107 (approved):
Looks good to me, thanks @emassoulie !

PR #2285, opened 2026-09-15
FEAT - Optimize session encoder to improve performance

rcap107:
This PR refactors the logic used by the SessionEncoder to detect sessions to vectorize it rather than using a python for loop.

This solution improves the runtime for pandas by ~2x at least, and polars by at least 15x depending on the dataset.

### DIFF
--- a/.github/workflows/check_stub_files_diff.yaml
+++ b/.github/workflows/check_stub_files_diff.yaml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: prefix-dev/setup-pixi@v0.10.2
         with:
-          pixi-version: v0.68.0
+          pixi-version: v0.80.0
           frozen: true
 
       - name: Check stub file for `_data_ops.py` is up-to-date
--- a/.github/workflows/run-code-format-checks.yaml
+++ b/.github/workflows/run-code-format-checks.yaml
@@ -17,7 +17,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: prefix-dev/setup-pixi@v0.10.2
         with:
-          pixi-version: v0.68.0
+          pixi-version: v0.80.0
           frozen: true
 
       - name: Run tests
--- a/.github/workflows/test-javascript.yml
+++ b/.github/workflows/test-javascript.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: prefix-dev/setup-pixi@v0.10.2
         with:
-          pixi-version: v0.68.0
+          pixi-version: v0.80.0
           environments: ci-py314-latest-optional-deps
           # we can freeze the environment and manually bump the dependencies to the
           # latest version time to time.
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -16,8 +16,8 @@ jobs:
       matrix:
         os: [windows-latest, ubuntu-latest, macos-latest]
         environment:
-          - ci-py310-min-deps
-          - ci-py310-min-optional-deps
+          - ci-py311-min-deps
+          - ci-py311-min-optional-deps
           - ci-py311-transformers
           - ci-py314-latest-deps
           - ci-py314-latest-optional-deps
@@ -27,7 +27,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: prefix-dev/setup-pixi@v0.10.2
         with:
-          pixi-version: v0.68.0
+          pixi-version: v0.80.0
           environments: ${{ matrix.environment }}
           # we can freeze the environment and manually bump the dependencies to the
           # latest version time to time.
@@ -63,7 +63,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: prefix-dev/setup-pixi@v0.10.2
         with:
-          pixi-version: v0.68.0
+          pixi-version: v0.80.0
           environments: ci-nightly-deps
           # we can freeze the environment and manually bump the dependencies to the
           # latest version time to time.
--- a/.github/workflows/update_pixi_lock_files.yml
+++ b/.github/workflows/update_pixi_lock_files.yml
@@ -24,7 +24,7 @@ jobs:
       - uses: actions/checkout@v7
       - uses: prefix-dev/setup-pixi@v0.10.2
         with:
-          pixi-version: v0.68.0
+          pixi-version: v0.80.0
           run-install: false
 
       - name: Remove the current lock file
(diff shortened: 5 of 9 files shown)


rcap107:
I have no idea why coverage is hallucinating that the code isn't being covered

skrub/_session_encoder.py
```diff
@@ -51,54 +51,44 @@ def _get_session_column_pandas(
...
-        X_selected.groupby(grouper) if len(split_by_columns) > 0 else [("", X_selected)]
+    X_selected = X_with_order[selected + [row_order_col]]
+
+    # Only do sessionization on rows that have no nulls in the selected columns
+    # Columns with nulls get -1 in the session id
+    mask_null = X_selected[selected].isnull().any(axis=1)
+    X_has_nulls = X_selected.loc[mask_null].assign(**{session_id_column: -1})
+    X_clean = X_selected.loc[~mask_null]
+
+    sort_cols = (
```
jeromedockes, line 66: this is the same thing as `selected` above. (also selected could be renamed `selected_cols`)

skrub/_session_encoder.py
```diff
@@ -51,54 +51,44 @@ def _get_session_column_pandas(
...
-    grouper = split_by_columns[0] if len(split_by_columns) == 1 else split_by_columns
-    groups = (
-        X_selected.groupby(grouper) if len(split_by_columns) > 0 else [("", X_selected)]
+    X_selected = X_with_order[selected + [row_order_col]]
+
+    # Only do sessionization on rows that have no nulls in the selected columns
+    # Columns with nulls get -1 in the session id
+    mask_null = X_selected[selected].isnull().any(axis=1)
+    X_has_nulls = X_selected.loc[mask_null].assign(**{session_id_column: -1})
+    X_clean = X_selected.loc[~mask_null]
```
jeromedockes, line 64: X_non_null or something like that?

skrub/_session_encoder.py
```diff
@@ -51,54 +51,44 @@ def _get_session_column_pandas(
...
-    X_selected = s.select(X_with_order, selected + [row_order_col])
-    X_has_nulls = X_selected.loc[X_selected[selected].isnull().any(axis=1)]
-    # Assigning a session ID of -1 to rows with nulls in timestamp or group_by columns
-    # -1 rather than None because adding nulls to a pandas column of integers will
-    # convert it to float
-    X_has_nulls = X_has_nulls.assign(**{session_id_column: -1})
-
-    X_selected = X_selected.dropna(subset=selected)
-
-    # needed to avoid a warning with min deps
```
jeromedockes, line 69: we should keep that comment

skrub/_session_encoder.py
```diff
@@ -51,54 +51,44 @@ def _get_session_column_pandas(
...
+    if split_by_columns:
+        grouper = (
+            split_by_columns[0] if len(split_by_columns) == 1 else split_by_columns
         )
-        groups_with_session_ids.append(group_df_sorted)
-    X_with_session_id = pd.concat(groups_with_session_ids + [X_has_nulls], axis=0)
-
-    # Reordering rows back to the original order and selecting session id
-    return X_with_session_id.sort_values(by=row_order_col)[session_id_column]
+        diffs = X_sorted.groupby(grouper)[timestamp_column].diff().dt.total_seconds()
```
jeromedockes, line 77: can you add a comment that pandas groupby preserves row order within each group

skrub/_session_encoder.py
```diff
@@ -51,54 +51,44 @@ def _get_session_column_pandas(
...
+    else:
+        diffs = X_sorted[timestamp_column].diff().dt.total_seconds()
+
+    # boundary is a bool, True if a value is na (diff with previous is invalid
```
jeromedockes, line 81: all the comments until the end of the function repeat exactly the code so can be removed I think

skrub/_session_encoder.py
jeromedockes: actually we don't need an extra variable for this it can be merged into the assignment to X_sorted

skrub/_session_encoder.py
jeromedockes, line 104: not introduced by this PR but maybe we should use polars selector here instead?

jeromedockes, line 108: that comment seems out of sync with the line below. and here also we could use a more explicit name than `selected`

skrub/_session_encoder.py
```diff
@@ -119,46 +109,46 @@ def _get_session_column_polars(
     selected = split_by_columns + [timestamp_column]
 
     # Find rows with nulls in timestamp or group_by columns and assign them a session
-    # ID of -1
-    # -1 rather than None for consistency with pandas implementation
+    # ID of -1 rather than None for consistency with pandas implementation
```
jeromedockes, line 112: could we move the polars import into this function? so we don't need to catch an import error

skrub/_session_encoder.py
```diff
@@ -119,46 +109,46 @@ def _get_session_column_polars(
...
 
-    X_selected = X_selected.drop_nulls(subset=selected)
-
-    groups = (
-        X_selected.group_by(split_by_columns, maintain_order=True)
-        if len(split_by_columns) > 0
-        else [("", X_selected)]
+    # Work only on the columns that have no nulls in the timestamp or split_by columns
+    X_clean = X_selected.drop_nulls(subset=selected)
+    X_sorted = (
```
jeromedockes, line 119: here also, can we skip the temp variable X_clean and do sort(by=selected)

skrub/_session_encoder.py
```diff
@@ -119,46 +109,46 @@ def _get_session_column_polars(
...
         )
-        groups_with_session_ids.append(group_df_sorted)
-    X_with_session_id = pl.concat(groups_with_session_ids + [X_has_nulls])
-
-    # Reordering rows back to the original order and selecting only the session id
-    return X_with_session_id.sort(by=row_order_col)[session_id_column]
+    else:
+        # Otherwise, it's just over the timestamp
+        diff_expr = pl.col(timestamp_column).diff().dt.total_seconds()
+    _diff_column = f"_diff_skrub_{random_string()}"
```
jeromedockes, line 133: nit: `_diff_column -> diff_column`

skrub/_session_encoder.py
```diff
@@ -119,46 +109,46 @@ def _get_session_column_polars(
...
+    X_sorted = X_sorted.with_columns(diff_expr.alias(_diff_column))
+    # Identify session boundaries based on the time differences, null values
+    # are added when there isn't a value to diff against (i.e., this is the
+    # first record in a group)
+    boundary = (
+        pl.col(_diff_column).is_null()
+        | pl.col(_diff_column).is_nan()
```
jeromedockes, line 140: how do both null and nan occur here? does it depend on the polars version?

jeromedockes (commented):
thanks @rcap107 ! looks good besides a few nitpicks

skrub/_session_encoder.py
rcap107: I don't remember the reason, I think it was because of the polars version, but removing is_nan doesn't fail with min optional deps so 🤷 I'll remove it

PR #2286, opened 2026-09-15
clarified docstring of tabular_pipeline

lisaleemcb:
quick fix to clarify that, when given a input pipeline, `tabular_pipeline` does not return the same pipeline, but instead a new pipeline with the original pipeline steps


jeromedockes (approved):
LGTM, thanks a lot @lisaleemcb !

PR #2289, opened 2026-09-17
removed all order_by references 

lisaleemcb:
the `order_by` parameter is deprecated and is now removed.


rcap107 (approved):
Looks good to me! I think we should add an entry in the changelog to mention this was removed after all. Once that's done, we can merge the PR.

rcap107:
We're good! Thanks @lisaleemcb

PR #2291, opened 2026-09-17
Memoize read-only DataOp traversals on shared DAGs

### LINKED ISSUE #2288: Extend the `_Evaluator` result cache to the other DataOps traversals
### Describe the bug

`_Evaluator.handle_data_op` returns the cached `impl.results[self.mode]` before
recursing (`_evaluation.py:330`), so evaluation visits each node once and is linear in
the number of nodes.

No other `_DataOpTraversal` subclass has an equivalent. They walk the graph as a
**tree** rather than a **DAG**: a node reached by *k* distinct edges is re-traversed
*k* times, recursively. A plan whose steps each reuse the previous result *k* times
therefore costs `O(k^depth)` per traversal, for a graph with only `O(depth)` nodes.

This matters because `checked_data_op_constructor` runs four of these traversals on
*every* DataOp constructed, and `evaluate()` runs another via `nodes()` /
`_cache_pruner`. Node reuse is ordinary skrub (one table feeding several branches that
are later combined), and `.skb.if_else()` produces it structurally at every level,
since condition, "then" and "else" all reference the same upstream node.

### Steps/Code to Reproduce

```python
import time

import skrub
from skrub._data_ops._evaluation import nodes


@skrub.deferred
def f(a, b):
    return a + b


for n in range(8, 16):
    t0 = time.perf_counter()
    v = skrub.var("v", 1.0)
    for _ in range(n):
        v = f(v, v)          # each step reuses the previous result twice
    print(f"  depth={n:>2}  {len(nodes(v)):>3} nodes  {time.perf_counter() - t0:>8.3f}s")
```

Swapping the body for `v = (v > 1).skb.if_else(halve(v), v)` (fan-out 3 instead of 2)
gives the same curve with base 3.

### Expected Results

Time roughly linear in the number of nodes — these graphs have 9–37 nodes, so
milliseconds.

### Actual Results

Time multiplies by exactly the fan-out per level of depth.

```
fan-out 2, f(v, v)                  fan-out 3, if_else
  depth= 8    9 nodes    0.031s       depth= 6   19 nodes    0.087s
  depth=11   12 nodes    0.240s       depth= 8   25 nodes    0.758s
  depth=13   14 nodes    0.950s       depth=10   31 nodes    7.052s
  depth=15   16 nodes    3.715s       depth=12   37 nodes   64.974s
```

16 nodes in 3.7 s; 37 nodes in over a minute.

## Which traversals lack the cache

Counting `handle_data_op` calls while building the `if_else` chain — `_Evaluator` is
flat at 7 visits per level, everything else triples:

| depth | `_Evaluator` | `_FindConflicts` | `_ChoiceGraph` and others |
|------:|-------------:|-----------------:|--------------------------:|
| 5     | 35           | 1,500            | 13,812                    |
| 7     | 49           | 13,645           | 125,550                   |

| pass | class | triggered from |
|---|---|---|
| `find_conflicts` | `_FindConflicts` | every node construction (`_data_ops.py:425`) |
| `_find_dataframe` → `find_arg` | `_FindArg` | same |
| `check_choices_before_Xy` → `choice_graph` | `_ChoiceGraph` (largest) | same |
| ↳ `find_X` / `find_y` | `_FindNode` ×2 | `_evaluation.py:919` |
| `nodes()`, `_cache_pruner` | `_Graph` | every `evaluate(..., clear=True)` |

`_ChoiceGraph`, `_ChoiceEvaluator`, `_Printer` and `_Cloner` inherit the un-cached base
`handle_data_op` (`_evaluation.py:240`); `_FindConflicts`, `_FindArg`, `_FindNode` and
`_Graph` override it but call `super()` unconditionally.

`_Cloner` is a near miss: `compute_result` checks `self._replace`
(`_evaluation.py:628`), so the clone it returns is a properly shared DAG — but the
check runs *after* the children are visited, so the traversal is still exponential
(0.005 s → 0.220 s from depth 8 to 14). This affects `make_learner()` and
`set_params()`.

Note also that `skrub.set_config(eager_data_ops=False)` does not help: `check_data_op`
is decorated `allow_skipping=False`, so `eval()` pays the full cost once anyway
(0.005 s to build, 150 s to evaluate, at depth 12).

### Suggested fix

These passes are read-only, so visiting a shared node twice cannot change their result
— the same reasoning that makes the `_Evaluator` cache sound. A `seen` set of node ids
in `_DataOpTraversal.run()`, used by the read-only subclasses, should cover it;
`_Graph` and `_Cloner` need their per-edge bookkeeping kept, but can hoist their
existing identity check to *before* the recursion instead of after.

Monkey-patching that in gives:

```
depth   build     eval   nodes
   12  0.025s   0.001s      37     (was ~65 s build)
  200  4.582s   0.015s     601
 1000    ----   0.077s    3001
```

Evaluation becomes linear. Build drops to `O(n^2)`, and the remainder is inherent to a
per-traversal cache: `impl.results` lives on the *node* and outlives the traversal, so
previewing a new node is O(1), whereas a `seen` set lives on the traversal object and
dies with it, so each of the O(n) constructions re-walks the whole graph. Removing that
too would mean giving the checks node-level subtree summaries (names, uuids, X/y marks,
choices) merged from children rather than recomputed — a larger change, and probably
not worth it, since real plans are small enough that quadratic is irrelevant. The
exponential is the part that bites.

### Versions
```
Python 3.14.5 macOS-26.6.2-arm64-arm-64bit-Mach-O
skrub 0.11.dev0 (main, c08c674)
sklearn 1.9.0 | pandas 3.0.5 | numpy 2.5.1
```

00200200:
# Bug Fix Pull Request

## Description

Read-only DataOp walks treated shared graphs as trees. A plan like `f(v, v)` went exponential with depth. `_Evaluator` already cached; the other passes did not.

I memoize visited nodes on the read-only traversals, hoist the identity check in `_Graph` / `_Cloner` so they still record every edge / return a shared clone, and key `_ChoiceGraph` by `(node, current outcome)`.

Fixes #2288

## Checklist

## How Has This Been Tested?

`pytest skrub/_data_ops/tests/test_evaluation.py` (25 passed). A depth-16 `f(v, v)` now builds in tens of milliseconds.

## AI Disclosure

- [x] This PR contains AI-generated code
    - [x] I have tested the code generated in my PR
    - [x] I have read and understood every line that has been generated by the AI agent
    - [x] I can explain what the AI-generated code does


jeromedockes:
thank you for opening the PR @00200200  but next time ask to be assigned to an issue before starting to work on it. this is a core part of the library and the issue will be handled by one of the core devs or the contributor who opened the issue, who already knows this part of the code well

00200200:
Got it — I'll ask to be assigned before starting on issues like this next time. Thanks for the note.

jeromedockes:
> Got it — I'll ask to be assigned before starting on issues like this next time. Thanks for the note.

Thanks for understanding! :)

